PR #2, opened 2019-12-13
Add Skull Notifier Plugin

Trevor159:


Trevor159:
lgtm

Trevor159:
I should probably note this is not tested but probably works.

PR #15, opened 2019-12-20
Added withes-bronzemanmode

robinwithes:
This plugin is meant to be a close bronzemanmode replication with some nice qol features like faster unlock messages depending on how big the queue is and a few commands that are toggleable to aid you in using the plugin.

Bronzemanmode itself is a custom gamemode similiar to ironman but once you receive a item it "unlocks" making you able to buy it in the GE.

### DIFF
--- a/plugins/withes-bronzemanmode
+++ b/plugins/withes-bronzemanmode
@@ -1,0 +1,2 @@
+repository=https://github.com/robinwithes/withes-bronzemanmode.git
+commit=1fe81adcc30e429e9cea04f1c78ef1918bc5f2ff


Nightfirecat:
Code lgtm, though it would be nice to see this plugin meaningfully differentiate itself from [dekvall/bronzeman-mode](https://github.com/dekvall/bronzeman-mode)

robinwithes:
> Code lgtm, though it would be nice to see this plugin meaningfully differentiate itself from [dekvall/bronzeman-mode](https://github.com/dekvall/bronzeman-mode)

This is the same concept but with more features & it was tested for almost half a year. Adam said in discord that it should be fine to pr this as long as i just slap my name on it to differentiate it which i did.

PR #28, opened 2019-12-22
region-locker: Change repo

slaytostay:
Change the region locker repo to take it out of dekvall's hands
Adds plugin icon

### DIFF
--- a/plugins/region-locker
+++ b/plugins/region-locker
@@ -1,2 +1,2 @@
-repository=https://github.com/slaytostay/plugin-hub.git
+repository=https://github.com/slaytostay/region-locker.git
 commit=549cefd8e3f7dafdf21adc5c449271053165e5eb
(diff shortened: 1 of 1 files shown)


Adam-:
WIthout consent of the original plugin owner, we don't allow just forcibly taking over other users plugins. Maybe @dekvall is fine with this? If not, it should just be added as a second plugin (see how eg. there are two bronzeman plugins currently).

Also your plugin description appears to be advertising another client "ChunkLite"?

dekvall:
He's the original plugin author and i asked him if he wanted to maintain it, so this is ok with me.

Adam-:
Okay. Can you update the plugin description?

slaytostay:
Yeah maybe I should have clarified that, devkall did ask me to replace his repo.
ChunkLite was just RuneLite with this plugin added and some changes to the client for shading and other chunk features, I will not be maintaining the client any longer because of the external plugin. But since original users of the client know the name I thought it would be appropriate as a tag.

Adam-:
The tag is probably fine, I mean this: https://github.com/slaytostay/region-locker/blob/region-locker/src/main/java/com/regionlocker/RegionLockerPlugin.java#L67

slaytostay:
Oh of course, I forgot that that was still there. On it.

PR #30, opened 2019-12-23
Add loot logger plugin

TheStonedTurtle:
Adds a plugin that logs Loot Tracker data locally and adds a new side-panel UI for viewing it.

Depends on the LootReceived event being added from runelite/runelite#10484


TheStonedTurtle:
Now that RL is on v1.6.1 and the LootReceived event exists this should be good. Travis is passing.

PR #45, opened 2020-01-01
create custom left click dropping

JZomerlei:
A common request I see is people wanting to not shift click drop, this plugin allows a typed list like

```raw trout,raw salmon``` or ```iron ore```

 to have those things left click drop.

### DIFF
--- a/plugins/custom-leftclick-dropper
+++ b/plugins/custom-leftclick-dropper
@@ -1,2 +1,0 @@
-repository=https://github.com/JZomerlei/red-external-plugins.git
-commit=5c4890fc006895144f4f2a14ac5f0ea5fe77ea1c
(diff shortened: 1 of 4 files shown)


JZomerlei:
I fixed the displayname and support links. I moved to my own repo that is not a fork of the example/template, but still is using branches for each external plugin, I don't want to have a bunch of repos on my github account when a commit hash to a branch is sufficient. 

I renamed the plugin in the hub to 'zom-leftclick-dropper' 

I introduced an isEquipment() to prevent use and drop being swapped on equipment so that shift key interactions works on those items if they typed in the item name into the list. 

I was dumb and overwrote history when I added isEquipment() but there is at least a history here and in my commit where I mention it was added... Just looks silly to have initial commit & a message about a change. CBA splitting the commits at this point but will if I have to.

PR #48, opened 2020-01-02
Add Hiscore killcount plugin

melkypie:
A rendered down version of the Hiscore plugin that looks up players killcounts.
Looking at the wiki this would not make it into the client as an official plugin because of the possibility of spoofing the data. But with user taking full responsibility with external plugins, I hope it can be merged as an external plugin.

I think it would be really useful for people doing raids and trying to see if they are bringing along someone inexperienced or with decent amount of experience.

### DIFF
--- a/plugins/hiscorekc
+++ b/plugins/hiscorekc
@@ -1,0 +1,2 @@
+repository=https://github.com/melkypie/Hiscorekc.git
+commit=8af6bfc9e02bb1f2a2305f4158019ace3a3f4b65


Trevor159:
I am pretty sure we are waiting on jagex to fix the hiscores for cerb and callisto before we add this to the main client.

melkypie:
Well if so, this could be a temporary solution until they fix it. After then just remove this from the external plugins.

Adam-:
I think this should go into the main client instead.

melkypie:
I could integrate this into the HiScores plugin, did not do it because on the wiki it says that adding a killcount lookup would be straight up rejected due to spoofing.

Adam-:
This isn't spoofable though, it comes from the Jagex hiscores.

melkypie:
Well that is why I was confused about how it is spoofable but I didn't want to take any chances so I just made it an external plugin

melkypie:
Closing this as boss kilcount was added to main plugin

PR #59, opened 2020-01-11
add bats-locator plugin

chestnut1693:
This plugin will help locate bats at the thieving room inside Chambers of Xeric. See [original post](https://github.com/runelite/runelite/pull/10540).


chestnut1693:
I would prefer some opinions before this makes it through.

Trevor159:
> I would prefer some opinions before this makes it through.

Opinions on what? We normally do not code review external plugins other than making sure they are not malicious. Like I said on your PR to the main repo you can come in the discord and ask if you have any specific questions. This LGTM otherwise.

chestnut1693:
I said I'm leaving it as it is but i want to add some minor changes.

Adam-:
We can merge this now if you want, or do you prefer if we waited?

PR #63, opened 2020-01-12
Add bank value history tracker

AdrianLeeElder:
Adds a new panel in the right-hand menu for tracking bank value over time. 

Current features:
- Automatic, periodic tracking of the current bank value (can be disabled, so only manual entries are added
- Data is tracked based on bank tab numbers, so you can filter data by bank tabs. 
- Graph displaying current bank value over time
- Simple time selection ranges (Today, 24 hours, 6 months) or an advanced time selection using a date/time picker.
- Ability to open in a new window

More information/images located on the repo [readme](https://github.com/AdrianLeeElder/runelite-bank-history-plugin).

### DIFF
--- a/plugins/bank-value-tracker
+++ b/plugins/bank-value-tracker
@@ -1,2 +1,2 @@
 repository=https://github.com/AdrianLeeElder/runelite-bank-history-plugin.git
-commit=fa9c494b2ce0c2b9e785b18d3ec2bf17156d7dad
+commit=ec29eddc8236f03c65b7f85846f73c09fd78b3ea
--- a/plugins/bank-value-tracker
+++ b/plugins/bank-value-tracker
@@ -1,2 +1,2 @@
 repository=https://github.com/AdrianLeeElder/runelite-bank-history-plugin.git
-commit=ec29eddc8236f03c65b7f85846f73c09fd78b3ea
+commit=6221eca8471aa2c280f249362b68c89f431cb1ce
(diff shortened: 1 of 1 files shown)


loldudester:
You need to set up your runelite-plugin.properties file. Especially the `plugins=` line.

AdrianLeeElder:
Is that build failure something I can fix on my end?

abextm:
Please add the [gradle witness plugin](https://github.com/signalapp/gradle-witness) to your build to verify the shasums of the 2 non-runelite dependencies you added.

AdrianLeeElder:
This is my first time using the `gradle-witness` plugin. It worked locally using gradle `4.10.2`, but when trying locally using the version of travis (5.1.1) it complains about:

` Could not find method dependencyVerification() for arguments`

The readme for the plugin said to put `dependencyVerification` directly below dependencies, but it doesn't recognize it there.

abextm:
So apparently the witness plugin doesn't support gradle 5, and doesn't actually work correctly. So now we use gradle nightly, follow the steps here <https://github.com/runelite/plugin-hub#third-party-dependencies>

abextm:
Do not store date in `~/.runelite/cache` that you expect to persist.

abextm:
`~/.runelite/plugins` is also not the right place; Files in it get removed if they are not the jar for a currently running plugin. You should put your stuff in `~/.runelite/bank-value-tracker`

abextm:
You put the wrong commit sha in here

PR #74, opened 2020-01-24
Add Essence Running Plugin

AaronPoon:
Change the default shift option that is displayed when hovering over objects associated with essence running

### DIFF
--- a/plugins/essence-running
+++ b/plugins/essence-running
@@ -1,2 +1,2 @@
 repository=https://github.com/AaronPoon/essence-running.git
-commit=d8545c2abc537ca6a5b23f66da5e11d535c47429
+commit=6c220645fc2a93c5f5d7cb9a4937cb9b4d44f470
--- a/plugins/essence-running
+++ b/plugins/essence-running
@@ -1,0 +1,2 @@
+repository=https://github.com/AaronPoon/essence-running.git
+commit=6dea2f2c219ab0dd3dbdb5d0c66230946d55846e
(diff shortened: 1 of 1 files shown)


Nightfirecat:
Does this swap menu entries of teleport jewellery only in the inventory? I didn't see anything in the plugin code which would block that swapping from happening on the equipment tab.

AaronPoon:
Yes it only does it in inventory. The user can configure what they want shift+click to be directly in the UI. The options available are an enum in: EssenceRunningItemShiftClick.java
Essentially only the options available while the items are in your inventory.

Swapping options while equipped are different (ex. Glory would should Edgeville, Draynor VIllage, etc).

AaronPoon:
I did some more testing since your comment got me thinking for the case of Crafting Cape where in both Inventory & Equipped, it is called "Teleport". I did a print and found out that, the format for equipped menuEntries on a equipped item is different than an item in the inventory:

Equipped
option: examine, identifier: 10, type: 1007
option: teleport, identifier: 3, type: 57
option: boost, identifier: 2, type: 57
option: remove, identifier: 1, type: 57

Inventory
option: use, identifier: 9781, type: 38
option: teleport, identifier: 9781, type: 35
option: wear, identifier: 9781, type: 34
option: drop, identifier: 9781, type: 37
option: examine, identifier: 9781, type: 1005

My code relies on the fact that the default entry for an item in Inventory is always the one before the "examine", hence my swapPrevious() code. However, for an Equipped item, it seems the examine will always be the first entry. Thus, swapPrevious will always do nothing since the entry before "examine" is nothing.

abextm:
You should add an explicit check to block that anyway

AaronPoon:
I can add the priority check for the type since it looks like the same logic as swapBankOp. Unless you know of some other way to check if an item is in inventory vs equipped.

AaronPoon:
Updated so rather than checking
`option.equals("examine")`, it now checks
`menuEntry.getType() == MenuAction.EXAMINE_ITEM.getId()`
which guarantees the MenuAction is from within the inventory.

PR #89, opened 2020-02-05
Update essence-running

AaronPoon:
Introduced new EssenceRunningUtils.java to hold all helper methods that shouldn't belong in EssenceRunningPlugin.java

Introduced two new features:
Session Statistics
- Displays statistics such as Pure essence/Binding necklace traded per runner and Fire runes crafted

Prevent Fire runes:
- Forces menu to open when you click the Fire Altar if you would accidentally craft Fire runes


AaronPoon:
Pushed another change to set 'highlightTradeSent' to disabled by default.

PR #97, opened 2020-02-14
Mlm alternate overlays

BuzzfeedBlue:
Adds an overlay to motherlode mine paydirt veins as usual, but uses the minimap mining site icon instead of the mining skilling icon. Purely an aesthetic change.


Nightfirecat:
So if I'm understanding this correctly, this is exactly the same as the existing MLM plugin, except it changes the icon from the larger mining icon (as would be seen on the skills tab) to the world map "Mining site" icon? I can't help but feel this should either be (a) a personal plugin because it's something so specific that you will be the only user of it, or (b) a change added to the released MLM plugin to allow a configurable icon if other players may want this.

BuzzfeedBlue:
Even if I were the only one to use the plugin, is the hub not the place for secondary non essential plugins? If its something you think that no one will use, why would it be in the main runelite motherlode plugin rather than tucked away in the hub, where only people who are looking for additional ways to customize the client may be looking?

PR #105, opened 2020-02-17
Add equipment-inspector plugin

botanicvelious:
Allows you to see the equipment that someone has equipped.

### DIFF
--- a/plugins/equipment-inspector
+++ b/plugins/equipment-inspector
@@ -1,2 +1,2 @@
 repository=https://github.com/botanicvelious/Equipment-Inspector.git
-commit=70b3654d4acb430620d5727ae35ee272b1e22fd9
+commit=d7b7ac4863d23e9744c258f2b8ac9a30e1190b31
--- a/plugins/equipment-inspector
+++ b/plugins/equipment-inspector
@@ -1,2 +1,2 @@
 repository=https://github.com/botanicvelious/Equipment-Inspector.git
-commit=d0deb21e4ecbe3d554c71be49f18d5b270c0f3a6
+commit=70b3654d4acb430620d5727ae35ee272b1e22fd9
--- a/plugins/equipment-inspector
+++ b/plugins/equipment-inspector
@@ -1,2 +1,2 @@
 repository=https://github.com/botanicvelious/Equipment-Inspector.git
-commit=8e1f4adf5d56896209259f2e0bc3d45ccaaba7e6
+commit=65bc068020427719adbcee3654062f58b9b23e6d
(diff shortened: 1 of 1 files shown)


botanicvelious:
note: there is no features like how much items are worth etc, just the names and icons

botanicvelious:
updated to not show charges on equipped items and always show max charges

dekvall:
Even without price information, I believe that a plugin like this can be abused in PvP scenarios. It would still be significantly easier to determine potential risk with this plugin rather than by visual inspection. I think that if it were to be added, it should be disabled in PvP altogether.
Furthermore, the implementation is mostly a copy of the unlicensed code in https://github.com/runelite/runelite/pull/3733. Though, maybe that's fine since it was PRed with the intention of adding it to the main client.

botanicvelious:
Updated it to not work in any pvp area

dekvall:
I guess it's fine then, please add a link to your repository in the `support` field of `runelite-plugin.properties`

@sean-duffy As the original author, are you ok with this?

sean-duffy:
Fine by me, will look forward to using it in the plugin hub! 👍

Nightfirecat:
For what it's worth, I don't think this needs a non-PVP limiter. Generally speaking, anyone who is familiar with PVP in almost any capacity will be able to identify items (and therefore risk) without any aid, so this wouldn't impact them whatsoever.

botanicvelious:
> 
> 
> For what it's worth, I don't think this needs a non-PVP limiter. Generally speaking, anyone who is familiar with PVP in almost any capacity will be able to identify items (and therefore risk) without any aid, so this wouldn't impact them whatsoever.

While I personally agree especially since it normalizes all charges etc on items I can see the argument, so fair enough.

@sean-duffy thanks :) only downside is that to normalize charges etc it no longer shows the trim version name etc... tho it does show the base item so you can at least figure it out

@dekvall done

botanicvelious:
> 
> 
> Like I mentioned before, I don't think the wilderness check is valuable since it doesn't really impact players who know what kind of valuables to look for, and it can be worked around by any of the following:
> 
>     * Perform the lookup on a player while standing outside of the wilderness
> 
>     * Perform the lookup on a player while in a PVP world (afaict there's no check for this)
> 
> 
> If someone was unaware enough to not be able to recognize gear but still wanted to use this to identify targets to potentially lure into PVP, this could still be used in those cases. I'll reiterate that my opinion is that there are basically 0 players for whom that'll be the case where this plugin would aid them in doing so.
> 
> Regardless, LGTM

If it will be approved without the wilderness check I will remove it, I tend to agree.

abextm:
imo this is fine without a wildy check; you still need to know the value of an item, and if you know that you probably know what it looks like.

botanicvelious:
update to remove the pvp check! 👍

PR #106, opened 2020-02-17
add fullscreen plugin

dekvall:
Might or might not work with multiple monitors. I don't know.


abextm:
I think this will break the tray icon on linux btw

PR #107, opened 2020-02-18
Update clan-chat-warnings

MarbleTurtle:
Added a cooldown to notifications (Big thanks to the guy hopping for Kraken)


MarbleTurtle:
I have a serious problem of making too many features but least I'm out of ideas and don't want to add any more to this.

PR #114, opened 2020-02-22
add teleport-logger

leejt:
Just a small utility for wiki users to systematically get coordinates/time for all of the teleport/transportation options in the game.


leejt:
Decided to keep this separate from other data gathering plugins for now, but will likely merge into a singular one in the future. It now correctly removes the button on shutdown.

PR #112, opened 2020-02-21
Create Better-Polybar-Integration

TrontheTechie:
Adding more output messages to a polybar module from runelite

PR #116, opened 2020-02-24
Update polybar-integration (v1.2)

SirArchLinux:
Adds more animations and idle-only mode


abextm:
Does this include everything from #112?

PR #128, opened 2020-02-29
Create Ground Markers With Labels plugin

Stevester118:
Original Ground Marker plugin with additional labeling functionality so that the user knows what each ground marker is for


abextm:
Don't share a config group with a core plugin

Stevester118:
i fixed it, sorry gonna resubmit again

PR #142, opened 2020-03-05
Crab Stun Timers

AnkouOSRS:
Don't merge until version 1.6.9 is released as this plugin relies on a change to the GraphicChanged event which is part of that version. It will not work properly without that change.

This plugin displays a timer over crabs in Chambers of Xeric to show how long they will remain stunned.


AnkouOSRS:
Pushed a change to add a check to ensure the actor in the event is a crab to avoid timers being created while fighting Vasa.

PR #152, opened 2020-03-10
zulrah-helper: remove nav button from toolbar on shutDown

while-loop:
Correctly remove navigation button from toolbar on `shutDown()`.

also official v1.0.0 release


while-loop:
Does the plugin-hub support using the `tag` as the commit? instead of a commit hash?

Nightfirecat:
Tags do not uniquely identify source, so no they are not supported as it would be a security issue if we did.

PR #153, opened 2020-03-12
PvP Performance Tracker

Matsyir:
Same as discussed in [Runelite PR 9861](https://github.com/runelite/runelite/pull/9861). My README has a lot of details, there's also a lot of dicussion on the old PR. Basic idea: tracks number of successful off-pray hits during 1v1 PvP fights. For example, if your opponent is praying melee, you must hit him with magic or ranged for a successful off-pray hit. Displayed live on an overlay and fight history saved in a panel alongside total kills/deaths.

Haven't really made changes since then, only some very minor things. I also named it more generally around PvP rather than focused on LMS.

Closes runelite/runelite#8269

### DIFF
--- a/plugins/pvp-performance-tracker
+++ b/plugins/pvp-performance-tracker
@@ -1,0 +1,2 @@
+repository=https://github.com/Matsyir/pvp-performance-tracker.git
+commit=0ea2204746cecef9709a2301e696b7a797b97f1d


Nightfirecat:
I've updated this PR to close runelite/runelite#8296 as I don't imagine we would be taking up any of these features in the main client and this would address the needs in that issue nicely.

PR #157, opened 2020-03-14
PvM Anti Drag

jcwhisman:
Anti Drag plugin that does not require shift. 

Anti-Drag is disabled when a player is in any kind of interaction with another player.


jcwhisman:
@Nightfirecat just curious, not trying to rush. This is my first plugin-hub submission, and am not familiar with the process. What is the typical wait time for PR approval and merging?

Nightfirecat:
Based on discussion, we're not likely to accept this as a plugin-hub plugin. Either we'll make changes to the main client plugin in this direction, or we wouldn't accept this and leave the main client version as-is. (see runelite/runelite#11036)

jcwhisman:
Totally fine @Nightfirecat. That PR in the main client is mine as well. However it is taking the spec orb varbit approach and overall is a much better solution anyways than the one in this repo.

PR #178, opened 2020-03-25
Pvp performance tracker: Add new statistic, various improvements

Matsyir:
This update was in collaboration with Mazhar, @maz_rs on twitter, @voiderman1337 on github. He implemented the core code of the newly tracked statistic: Deserved damage. This checks your gear as well as the opponent's, calculates accuracy, max hit, and gets an estimated hit based on that data. This is added to a total counter, to help see if you performed better than your opponent. This really adds to the off-pray hit tracker, since it will heavily penalize you for doing things like barraging in d'hide, or getting ranged/melee'd in mystic too often (even if you are better with the prays, this can severely reduce your overall performance). The off-pray hits alone wouldn't detect cases like this. This doesn't involve hitsplats in any way, it really just calculates your 'average' damage you should've gotten based on you and your opponent's gear.

The deserved damage also isn't perfect, there is a lot of hardcoding involved, but it should work in the vast majority of cases. In the end, any inaccuracies the plugin has will be passed on to both fighters, but they are rare.

There are many other various improvements. Fights are now saved across restarts using a simple json encoded string, with a config limit. Default limit (1000 fights) should result in about 500kb storage. Storage isn't a problem, but at ridiculously high numbers (10k+) the ram usage is significant (but useable). I doubt you could get to that amount of fights realistically though. Also fixed some odd behaviour where following or trading others would show the overlay right away. This shouldn't happen anymore until actually in a fight. The UI code was optimized a lot, saving re-used elements and updating their text rather than re-creating the components on every update. Shouldn't be too noticeable, but why not be as robust as possible.

README is very detailed if you would like more info.


Nightfirecat:
This looks like it needs rebasing. This PR should be changing the existing `pvp-performance-plugin` file's commit, not trying to re-add the file.

Matsyir:
I tried to fix it, not sure if I did that right. I don't understand how it behaved that way since I'm just changing the line in one file, which should be the only difference between my branch & main repo.

PR #192, opened 2020-04-03
Reclaiming Profiles plugin

Spedwards:
The Profiles plugin was authored by me in its entirety and uploaded by Gudi with permission from me after I lost interest in the game (#53).

I am reclaiming this plugin and pointing it to my repo as I would like to make improvements to the plugin.

### DIFF
--- a/plugins/profiles-panel
+++ b/plugins/profiles-panel
@@ -1,2 +1,2 @@
 repository=https://github.com/Spedwards/runelite-plugins.git
-commit=451acdc2f002502a58b0fe5c491c678a052902a3
+commit=73e8ab742ab2f4e3900472d2e44ea715204cb0b6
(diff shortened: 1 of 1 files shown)


Adam-:
I've asked Gudi about this and he says it is fine.

Nightfirecat:
It would be really helpful if you could rebase your repo on mlgudi/profiles@e251564 so we can see a diff between your repo's tip and what we're currently building for this plugin.

Nightfirecat:
This change adds password storage to this plugin which might be dangerous. We might want to wait on #159 to add that functionality as users should be warned if plaintext passwords are being stored as a result of using this plugin.

Spedwards:
> This change adds password storage to this plugin which might be dangerous. We might want to wait on #159 to add that functionality as users should be warned if plaintext passwords are being stored as a result of using this plugin.

It will be stored on their local system and it is completely optional, but I had the same thought.

Spedwards:
Changed the commit to use as one before password support was implemented.

PR #206, opened 2020-04-10
map-waypoints

iipom:
Plugin that enables waypoints on the world map, similar to Pre-EOC. Once a waypoint is set with a double-click on the world map (or shift-click), an overlay will show you the direction, and the number of tiles you are away from that waypoint. A waypoint is removed by double-clicking or shift-clicking on it.


loldudester:
1) You don't need to close and open new PRs to fix them
2) You need to point to the main plugin file in runelite-plugin.properties. In your case the plugins line should be `com.iipom.mapwaypoint.MapWaypointPlugin`

PR #213, opened 2020-04-15
Map waypoints

iipom:
This is probably the last update I will do for this plugin. It does everything I wanted it to now.


nlruk:
@iipom Awesome plugin!!! Can you please make the yellow arrow more aesthetically pleasing and can the blue focus waypoint on the world map be a different shape so we can easily differentiate from our current location

PR #215, opened 2020-04-15
Create player-tile

1Defence:
Highlights the local players server-tile, different from the local tile available within the player-indicators  plugin which is delayed by animations.


1Defence:
Animation delays, moving diagonals especially cause your character to visually misrepresent your real location in game; thus this highlights your characters actual location on a tile by tile basis, useful for pathing and timing actions.
vGY20EsTlg

1Defence:
The plugin no longer works in PVP worlds to prevent aiding lures, added framework for region-based removal if necessary in the future.

abextm:
btw you can use `@Alpha` on `Color` config values so you don't need to support opacity manually. Also we don't care if it works in pvp or not, and your check only checks pvp worlds, not pvp zones.

PR #222, opened 2020-04-17
Update NPC Idle Timer (v1.1)

vonpawn:
Adds secondary color to overlay timer when time is below configured threshold


Horselord1:
Not sure how to get in contact with you vonpawn but I was hoping I could ask if you can change the timer to ticks instead of seconds because I used the seconds timer and it just looks weird since its still only updating every tick and counting down weirdly

PR #233, opened 2020-04-21
Created clan roster helper plugin

simbleau:
Created plugin file for clan roster purifier plugin


simbleau:
The clan roster purifier plugin takes a truthful source of a clan extract and informs the clan owner what actions they must take to match the truthful source. The clan roster source file is a file which contains an array of objects which each have a 'rsn' value and a 'rank' value. This file acts as the truthful roster, and this plugin tells you how to match that file.

More information: https://github.com/simbleau/clan-roster-purifier

PR #249, opened 2020-04-28
Update: Profiles (v1.2.0)

Spedwards:
Adds password support. Password support is completely optional and use is at end user discretion.
Password support requires an encryption key to use and is not saved anywhere.

Depends on #159

### DIFF
--- a/plugins/profiles-panel
+++ b/plugins/profiles-panel
@@ -1,2 +1,3 @@
 repository=https://github.com/Spedwards/runelite-plugins.git
-commit=c2dd59dcdf324df01c54aeba4cce4577c645f642
+commit=169f26f0d116f8dc56bc2bec0ef7d68ee20e46dd
+warning=This plugin optionally can save your password, encrypted using a key of your choice, to a file on your computer.


Nightfirecat:
This code looks fine from what I can tell. I've updated the post to reflect that this depends on us adding support for warnings.

plugins/profiles-panel
```diff
@@ -1,2 +1,3 @@
 repository=https://github.com/Spedwards/runelite-plugins.git
-commit=c2dd59dcdf324df01c54aeba4cce4577c645f642
\ No newline at end of file
+commit=169f26f0d116f8dc56bc2bec0ef7d68ee20e46dd
+warning=Password support is completely optional and use is at your discretion
```
Nightfirecat, line 3:
```suggestion
warning=This plugin can save your password, encrypted using an external key, to a file on your computer.
```

Nightfirecat:
At this time, we're not comfortable handling passwords in any capacity even as is implemented here. For the time being, we recommend users make use of password managers for that, possibly leveraging a plugin which can interface directly with that [such as raiyni's bitwarden plugin](https://github.com/raiyni/bitwarden-plugin).

PR #253, opened 2020-04-29
add text-to-speech plugin

Trevor159:
pretty terrible but it works.

The licensing should be fine I think


Trevor159:
Closing forever probably.

PR #281, opened 2020-05-08
Create theatre-of-blood-stats

HSJ-OSRS:
Provides a more detailed break down of the rooms in theatre of blood by showing time splits and damage dealt


Hydrox6:
Your `runelite-plugin.properties`'s `plugins` variable must include the package.

`plugins=hsj.external.theatreofbloodstats.TheatreOfBloodStatsPlugin`

PR #300, opened 2020-05-15
Update: Group Iron Man (v1.1.1)

Spedwards:
Adds Group Iron Man icon in v1.1.0 and fixes a couple bugs in v1.1.1.


mehmetk95:
Hi,
are you able to implement that you cant pick items from other players into the group ironman plugin? At least have it as a option u can tick?

arrafrost:
I believe there is currently no way to determine between ground items dropped by yourself or other players, or by monsters killed by yourself or by others. As such it seems unlikely to me that this feature would be able to be added.

mehmetk95:
ah i see, it is the only thing remaining to get the real group ironman feeling, so would be very nicee to have it :/

PR #306, opened 2020-05-17
Add resource-packs plugin

melkypie:
Doesn't have any proper resource packs yet, releasing this mostly for people to be able to create some.


melkypie:
Made the readme clearer and moved chatbox.png from /tab to /chatbox and renamed it to background.png

melkypie:
Updated to include some more sprites that can be replaced and added the first proper resource pack.

PR #308, opened 2020-05-18
Add custom-emoji to plugin-hub/plugins.

coizioc:
Custom Emoji is a plugin that replaces Discord-style emoji (e.g. `:lol:`) in chat messages with user-defined images.


Nightfirecat:
We've declined other plugins which would have added discord-style triggers for emoji in the past since we did not want to pollute other user's chat if they did not have the plugin active, or were playing on some other client.

coizioc:
Alright, thanks for your time.

PR #313, opened 2020-05-22
Inventory-Setups-1.6

dillydill123:
Inventory Setups 1.6:

Adds a way to view setups in game using the "Show worn items" menu.

Adds Spellbook to the setup so you can make sure you are on the correct setup

Adds a help button that links to the read me (https://github.com/dillydill123/inventory-setups#inventory-setups) which now contains pictures and gifs to help new users. This button can be hidden in the config (learning from jagex's mistakes).

### DIFF
--- a/plugins/inventory-setups
+++ b/plugins/inventory-setups
@@ -1,2 +1,2 @@
 repository=https://github.com/dillydill123/inventory-setups.git
-commit=3102b4d1622885bad907077dee03b63a6b2184a4
+commit=7372f1bfd1f715ab8638fa6db563ec6e6048c971


Nightfirecat:
FYI `assert` will cause client hanging if it evaluates to `false`.

dillydill123:
Are asserts enabled in releases? I thought they are only active during debug. This is how I’ve always used them (in C++ for example).

I think the asserts are fine and won’t get triggered. But if you have a strong opinion on removing them I don’t have a problem either.

abextm:
asserts are only enabled if you run with `-ea`

Nightfirecat:
I believe they're only enabled for players running dev builds, but that will likely affect some nonzero number of players. See #292 for reference

dillydill123:
It sounds like it won’t affect many players if one does trigger and I like having them in there for debug purposes. If you still think they should be removed it’s not a problem though.

Nightfirecat:
I'll merge it as-is, but I worry with things like these that it will cause more traffic to `#support` if any of the assertions are buggy or overly strict.

PR #314, opened 2020-05-23
Update Flipping Utilities (v1.3.1)

Belieal:
Adds new multi-client support so that the plugin is able to properly communicate and store historical trade data between multiple simultaneous client sessions.

Adds GE tracking links that allows the user to right-click an item on the plugin panel to take them to the item's respective official OSRS GE or Platinum Tokens page. More links may be added in the future as long as the websites are respectable (No ads for private servers, cheats or RWT etc.) and their owners give us permission to link to their sites.

Descriptions of smaller additions and bug fixes can be seen on the [repo's changelist](https://github.com/Belieal/flipping-utilities#ge-tracking-links). :)

### DIFF
--- a/plugins/flipping-utilities
+++ b/plugins/flipping-utilities
@@ -1,3 +1,3 @@
 repository=https://github.com/Belieal/flipping-utilities.git
-commit=7407fab8e07365cf0229875e1dfdb32954e41909
+commit=33eb1db336a8692b9712b3f3db8370370b2f5af4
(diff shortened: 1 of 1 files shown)


Adam-:
This probably requires a warning message for the `platinumtokens.com` access. Did we standardize one yet?

abextm:
Since this just `LinkBrowser`s you, imo its fine for this to not have a warning if the menu reads `Open in PlatinumTokens.com`

Adam-:
It looks like it does read that. I suppose its fine then.

PR #319, opened 2020-05-26
Add Magic Secateurs plugin

Ryanbarker0:
The plugin adds a notification when the magic secateurs are in inventory and not equipped. Acts as a reminder to equip them before harvesting crops on farm runs to maximize yield and XP.

### DIFF
--- a/plugins/magic-secateurs
+++ b/plugins/magic-secateurs
@@ -1,2 +1,2 @@
 repository=https://github.com/Ryanbarker0/magic-secateurs.git
-commit=d3c78882262eaf7c03c94537508e9e6e8563a06d
+commit=36816c80988fb06d01b3b4e53e80981e6629d163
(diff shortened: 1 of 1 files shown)


abextm:
its highly recommended to not set `enabledByDefault = false` on external plugins; it confuses people because they install the plugin and it does nothing by defaultt

raiyni:
Fwiw, you can use the inventory changed event to see if secateurs are in the inventory instead of checking every render.

basicDamen:
You can also utilize ItemId to grab the item’s id instead of hardcoding the value.

Nightfirecat:
I'll echo what the others have said as well. We don't mind much the quality of the code in general, but removing `enabledByDefault = false` [and adding a LICENSE to the repo](https://github.com/runelite/plugin-hub#licensing-your-repository) should be done before we merge it.

PR #320, opened 2020-05-27
Updated chompy hunter to include config and notifications.

ejedev:
Updated the plugin to allow users to toggle a notification for when a chompy spawns.


Nightfirecat:
Why has the author changed?

ejedev:
@Nightfirecat I changed it to my OSRS username. I can change it back to my github name if that would work better

Nightfirecat:
No that's fine, just making sure that was an intentional change.

PR #322, opened 2020-05-27
Added Wiki Bank Tag Integration Plugin

MitchBarnett:
This Plugin allows BankTags to be applied to items using the categories on the http://osrs.wiki

Usage example chat command "::bt Potions" to tag all potions.


Nightfirecat:
The `author` in runelite-plugin.properties is shown in the plugin hub panel. I would recommend abbreviating it to your display name and adding alternate contact methods to your README as needed.

MitchBarnett:
@Nightfirecat Thanks for that I wasn't sure where `author ` was used, I have changed that and also added the repository URL as the support link.

PR #330, opened 2020-05-29
resource-packs: add multiple resource packs and more textures

melkypie:
Adds a bunch of new textures and adds the ability to switch between multiple resource packs


Zalmix:
Melkypie, how can I get in contact with you? I would want a menu entry swapper coded, but I didn't get it working for me

PR #353, opened 2020-06-05
Create wiki-search-shortcuts

IronCory:
Currently just a configurable hotkey to open the wiki search interface in runelite


IronCory:
I'm unsure what caused this error or what it even means :(

abextm:
you told it the plugin class is `com.wikishortcuts.WikiShortcutsPlugin` but you didn't provide that class

IronCory:
thanks for you help abex, i forgot to update the plugin class after i changed the plugin name

abextm:
It would probably be fine for this to be an option in the core wiki plugin if you want.

IronCory:
the hotkey being added to the core plugin works for me! I made it as an external plugin only so that I could still have it if it were to be rejected from the core plugin and so I was less likely to break the core wiki plugin.

PR #360, opened 2020-06-06
Create ninja-kick

ErinTales:
Allows you to set a blacklist of players to be automatically kicked on joining a clan chat. Also allows you to type a command to kick a player who has not yet joined.

Developed to help out a clan running short on ignore list space :)


dekvall:
In my opinion, auto-kicking players seems a bit sketchy.

abextm:
We have decided that this is close to, if not, breaking the rules, and we do not want this.

PR #361, opened 2020-06-06
Create hide widgets

PresNL:
This plugin enables you to hide all UI widgets in Oldschool Runescape using a hotkey. The default hotkey for this is CTRL + H.

This plugin only works in resizeable mode and does nothing in fixed mode. 

This was originally requested in this issue: [https://github.com/runelite/runelite/issues/10803](https://github.com/runelite/runelite/issues/10803)

demo


Hydrox6:
Usually it's a bad idea to have the plugin disabled by default, but I'm not sure for this case

PresNL:
that was still on for before it was a hotkey toggle I'll update the plugin to be enabled by default

PresNL:
I have added the license and made sure the plugin is now enabled by default

PR #362, opened 2020-06-06
Ninja kick

ErinTales:
Two functions:

1. Allows you to use a chat command to kick a player from a clan chat. This is useful for a large/popular cc with an overflowing ignore list who wants to be able to preemptively refresh the kick on a user when the key rank/owner is not on to ban them.

2. Contains a blacklist which notifies you whenever a user on it enters the clan, allowing you to easily kick them. Removed the functionality to have this auto-kick them, as that's why this plugin was rejected before.

### DIFF
--- a/plugins/ninja-kick
+++ b/plugins/ninja-kick
@@ -1,2 +1,2 @@
 repository=https://github.com/ErinTales/ninja-kick.git
-commit=d057dc60f37725837f02c50ceb1efee48e553536
+commit=609269b2b3130cc516b556b9eee9ee682549cd29
--- a/plugins/ninja-kick
+++ b/plugins/ninja-kick
@@ -1,0 +1,2 @@
+repository=https://github.com/ErinTales/ninja-kick.git
+commit=00c5b928a9eecfeb2b861a7e85ad243e60d2d047
--- a/plugins/ninja-kick
+++ b/plugins/ninja-kick
@@ -1,2 +1,2 @@
 repository=https://github.com/ErinTales/ninja-kick.git
-commit=c3c09b0896ab1999c536a2becf616df9a95265ce
+commit=77897eb95371e86e99cb49743bbf67c4aff606f4
--- a/plugins/ninja-kick
+++ b/plugins/ninja-kick
@@ -1,2 +1,2 @@
 repository=https://github.com/ErinTales/ninja-kick.git
-commit=609269b2b3130cc516b556b9eee9ee682549cd29
+commit=c3c09b0896ab1999c536a2becf616df9a95265ce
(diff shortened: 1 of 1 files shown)


Adam-:
This is allowing anyone to type `!kick <rsn>` and it will force a kick from your client, which seems pretty bad, and is probably outside of the "non-malicious" guarantee we try to keep.

ErinTales:
I guess I don't see how it's malicious? If a cc moderator wants to abuse their power, this isn't going to let them do it much easier. It's just an alternate way to do what you can already do.
Am I misunderstanding what you mean? It's not like you can kick random people without the ability to already do so, or anything like that.

Adam-:
1) random unsuspecting user installs plugin
2) everyone in user's CC now can kick each other via !kick

ErinTales:
Aah, I see what you mean. Whoops, that's kind of an obvious oversight.

If I make it so you have to be the one doing it, would that be ok?

Adam-:
That's probably okay. FYI there is a event for client commands eg `::kick` you can use too.

ErinTales:
Dang it I committed the old one...

ErinTales:
Wait no I didn't never mind. Should be good now.

MarbleTurtle:
Coming from another large cc (and I made clan warnings and had the same ideas that you are doing) having a high ranking member able to go scorched earth without any evidence of them being the cause can lead to issues, the ability to kick a player not in cc is a grey area as its something that is normally not possible but I'm unsure if Jagex cares/detects this I personally decided against doing it after asking my cc for input, and unfortunate the Ninja kick feature is not going to be possible after the next rl update due to the kick messages having the name feature being reverted after a message from Jagex on it being abusable.

ErinTales:
You can already kick a player not in the CC, if you've seen them type in another clan chat, or if it's been any amount of time and you somehow still the message loaded in your chatbox (i.e. you log out, and come back 5 hours later, and the message is still there).

When you say it won't be possible after the next update, do you mean that the plugin is going to break & won't be fixable? I understand why the feature was removed, but this isn't actually sending the player's name as a chat message, anything typing out the players name that you're kicking here is a clientside message.

MarbleTurtle:
I meant the ninja kick itself with it not saying anything in chat, that relies on a script that is being changed back to not be editable

ErinTales:
Aah, I see what you're saying. So the entire thing will be pointless in less than a week anyway.
I'm not too familiar with RL, one of our clan leaders approached me several weeks back, asking if this was possible, and I didn't think it was until I saw the update this week. I didn't realize that it was actually impossible when he asked, and that it's going to be impossible again very soon.

PR #363, opened 2020-06-07
Update rune.cafe plugin

jmhertlein:
Adding trade reporting, better logging, and now doesn't spam the backend if an API key isn't loaded yet.


plugins/rune-cafe-cashflow
```diff
@@ -1,3 +1,3 @@
 repository=https://github.com/rune-cafe/rc-plugin-cashflow.git
-commit=f9d2df237df697ad5298116ed0ec3ff684f10392
-warning=This plugin submits your username, GE history, and IP address to a 3rd party website not controlled or verified by the RuneLite Developers.
+commit=1348743249bdadc90d710d379301f83e752eac9a
+warning=This plugin submits your username, GE history, GE trades and IP address to a 3rd party website not controlled or verified by the RuneLite Developers.
```
Spedwards, line 3: GE trades is implied in GE history.

PR #369, opened 2020-06-10
SuppliesTracker: Remove itemcomp from swing thread

Dyldozer:
Internal changes to remove itemcomp call from swing
Added ToB and vorkath death cost
Moved thrownaxe to ammo


JZomerlei:
Similar to how you added the cost of death for TOB can you add the cost of dying at Vorkath? Cost is 100,000. 


Like the plugin a lot.

PR #372, opened 2020-06-11
Add click-remapping plugin

NielsHeltner:
A plugin that allows binding a key to perform a left-click.


Hydrox6:
We've decided that keyboard to mouse (or vice versa) rebindings are not something we want in the client in any capacity.

NielsHeltner:
Fair enough. What is the reasoning behind that?

Hydrox6:
We had a discussion about it, and have decided that mouse <-> keyboard remapping was just too sketchy, and something we didn't want at all.

PR #395, opened 2020-06-16
update resource-packs and clan-chat-country-flags

melkypie:
Only merge this after the runelite update


abextm:
Merged as 013af5efe51a22f512f9c5761301e70140ab44d5

PR #404, opened 2020-06-18
add RuneWatch plugin

while-loop:
Adds a plugin to label and add investigate options to players.

More details and pics [here](https://github.com/while-loop/runelite-plugins/blob/runewatch/README.md)

| NOTE: I've been in contact with Will (Will osrs#0480) from RuneWatch and he gave me the okay to use their public API & use the RuneWatch logo. Along with some feature requests (such as trade screenshots). |
| --- |

### DIFF
--- a/plugins/runewatch
+++ b/plugins/runewatch
@@ -1,2 +1,3 @@
 repository=https://github.com/while-loop/runelite-plugins.git
 commit=7cc220e93c011d1641d344f656a0e8a996fc87f1
+warning=This plugin submits your IP address to a 3rd party website not controlled or verified by the RuneLite Developers.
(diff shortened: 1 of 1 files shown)


Nightfirecat:
This plugin should have a warning stating that it will send user IPs to an external service, similar to [the warning found in this plugin](https://github.com/runelite/plugin-hub/blob/626be7a/plugins/rune-cafe-cashflow)

while-loop:
Added warning message.

abextm:
lgtm, waiting on an ok from Will to merge this

RuneWatch:
> lgtm, waiting on an ok from Will to merge this

Thanks for holding it for me, I'm happy for this to go, as is. I'm happy with the 30 minute sync delay. However I do have another endpoint: `https://www.runewatch.com/api/cases/{rsn}` if you would like latest data on a case (could be useful for '**investigate**' lookups).

Please note that in the event no cases are found, the response will look like:

`[
    {
        "error" => "unable to get cases"
    }
]`

while-loop:
Sounds good! I'll implement username lookup in the next release. I will need to do a bit of refactoring to handle the async responses.

PR #417, opened 2020-06-22
add CoX external plugin

Blackberry0Pie:
Adds a fork of the Chambers of Xeric plugin with additional features.


Trevor159:
I will say what I said in discord here, I think copying the whole raids plugin is a bad idea and implementing raidscout/raidreset events in the main client would be better. Those are all you need for this I think.

Nightfirecat:
See runelite/runelite#11973. Once merged, this plugin can be greatly simplified

Blackberry0Pie:
Updated to use RaidScouted event.

PR #419, opened 2020-06-24
resource-packs: update to 1.1

melkypie:
add resource pack hub functionality


melkypie:
Removed the dependency and made some optimizations

melkypie:
Fixed a styling bug

PR #421, opened 2020-06-25
Create runewatchcheck

claythearc:
I created a plugin that submits a HTTP Query to http://www.runewatch.com/api/cases/{player_name} everytime a player joins a friends chat. It will then call client.AddChatMessage() and list the results. This will hopefully prevent people from getting scammed as much =).


abextm:
You need to get this okayed by the RuneWatch staff

Trevor159:
I am going to close this as it has been almost two months without a response from the owner. If you want this reopened just let us know.

PR #422, opened 2020-06-26
Create level30wilderness

yrrah2:
My plugin is a very simple plugin which does the following:

Between Level 30 Wilderness and Level 31 Wilderness there is now a blue line when the plugin is enabled

There is currently no config or settings to change, and when the plugin is disabled the blue line is removed

The purpose is to see where you can activate certain teleports from


yrrah2:
unsure what i did wrong as the plugin works on my end

PR #423, opened 2020-06-26
Create level30wilderness

yrrah2:
My plugin is a very simple plugin which does the following:

Between Level 30 Wilderness and Level 31 Wilderness there is now a blue line when the plugin is enabled

There is currently no config or settings to change, and when the plugin is disabled the blue line is removed

The purpose is to see where you can activate certain teleports from

### DIFF
--- a/plugins/level30wilderness
+++ b/plugins/level30wilderness
@@ -1,0 +1,2 @@
+repository=https://github.com/yrrah2/level30wilderness.git
+commit=2ac296a53b694d431305d0fd04ee3f1f8498ece5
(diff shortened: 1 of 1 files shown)


Nightfirecat:
I'm of the opinion this plugin would be fine. I think that, unlike multi-combat indicators, this offers little "competitive" advantage in the wilderness. I'd like to hear other contributors' opinions on this though.

while-loop:
@Nightfirecat, I see the "advantage" you gain with this, but it's no different than marking all wilderness lvl 30 tiles.

But going with that logic, I think this wouldn't be allowed based on Jagex's message about 3rd party clients.
> Anything that automatically indicates where to stand, or not to stand. This applies to only automatic indicators, and not tiles which have been **_manually marked_**.

Nightfirecat:
Given that particular bit of guidance, and concerns from other admins/contributors, we will not be adding this plugin at this time.

PR #446, opened 2020-07-06
update optimal-quest-guide

cesoun:
A bug fix for users using stretched mode and expand to screen display settings. This would cause the plugin not to update when logging in for users with this setting. The plugin now checks for the minimap widget group instead. This also causes the plugin to update right away after login instead of after pressing `Click Here to Play` which is nice.

Thank you to Dahl for finding and reporting this bug.

Also a spelling mistake was fixed for the side panel. `Optimal QuestInfo Guide` -> `Optimal Quest Guide`


Nightfirecat:
This can't be merged as-is, you should rebase your `plugin-hub` repository's branch on master then push the commit which updates your plugin's hash.

PR #497, opened 2020-07-26
Create underwater-agility

R438:
Plugin with helpful features for underwater agility and thieving. Includes a chest timer, tear counter, tear rate, and highlighting of currents, puffer fish and chests.


Nightfirecat:
Your runelite-plugin.properties file is not set up correctly. It should Have all of its properties updated, particularly `plugins`, which should point to your plugin's package path. In this case, that would be `com.swim.SwimPlugin`

PR #509, opened 2020-07-30
the-gauntlet

dutta64:
An all-in-one plugin for The Gauntlet.

* Custom resource tracking
* Resource image and outline overlays
* NPC outline overlays
* Starting-room utilities outline overlays
* Timer


Cyborger1:
You code doesn't seem to compile because you've use `Set.of`, which was only added in Java SDK 1.9. Try using `ImmutableSet.of` from `com.google.common.collect` instead, such as seen in [ChatMessageManager.java](https://github.com/runelite/runelite/blob/37e025ed799f15739e0029dff5a3bbac2394c2d7/runelite-client/src/main/java/net/runelite/client/chat/ChatMessageManager.java), for example.

PR #524, opened 2020-08-02
Add the no social messages plugin

hex-agon:
A simple plugin that removes the 'player has logged in' and 'player has logged out' messages from ever appearing.


hex-agon:
Exists in the chat filter plugin

PR #537, opened 2020-08-10
Add accounts-login plugin

wesley-221:
Add a simple plugin that allows you to save an account and password to a local file. When you click on the tile it simply pastes the username and password to the login screen. 

Somewhat similar to the "Profiles" plugin, but since the that plugin hasn't been updated in a while and doesn't support password saving in the current version, I decided to make my own plugin.

### DIFF
--- a/plugins/accounts-login
+++ b/plugins/accounts-login
@@ -1,0 +1,2 @@
+repository=https://github.com/wesley-221/runelite-accounts-login.git
+commit=ef9b037d771fb3891f1abbf6a934a5c452602292


loldudester:
The profiles plugin's 1.2.0 update PR (#249), was rejected for password handling, with this reasoning from Nightfirecat:
> At this time, we're not comfortable handling passwords in any capacity even as is implemented here. For the time being, we recommend users make use of password managers for that, possibly leveraging a plugin which can interface directly with that such as raiyni's bitwarden plugin.

PR #540, opened 2020-08-11
Add ToB time infobox plugin

winterdaze:
Adds an infobox with the room's time after it is completed.
Resets when you enter the next raid (manually reset by toggling the plugin).

img

Would close https://github.com/runelite/runelite/issues/11876


winterdaze:
Created the license. Should I update the hub commit hash or is it okay to keep it as it is?

PR #549, opened 2020-08-17
Update collection-log v1.1

evansloan:
Load total items collected on plugin startup/player login


melkypie:
You should remove the `enabledByDefault = false` as it might confuse people when they install the plugin as to why it is not working. The same reason as was said in #319.

PR #570, opened 2020-08-26
not-empty: add plugin

dekvall:
Give users the ability to hide annoying right-click options on items


Nightfirecat:
Based on Discord conversation it sounds like we want this to not allow customizing the hidden entries allowed.

PR #573, opened 2020-08-27
Inferno Wave Splits plugin

usa-usa-usa-usa:
Displays a chat message with inferno elapsed time for the various inferno wave milestones.

To avoid re-inventing the wheel. The plugin is dependent on @winterdaze 's recent addition of the tzhaar timers on the base clients Timers plugin. This is stated within the support documentation for the plugin.

All this plugin does is parses the Wave: # chat message to determine if it is a split wave, reads the text on the infobox implemented in the Timers plugin, then displays a chat message with the elapsed time.

### DIFF
--- a/plugins/inferno-wave-splits
+++ b/plugins/inferno-wave-splits
@@ -1,0 +1,2 @@
+repository=https://github.com/usa-usa-usa-usa/inferno-wave-splits.git
+commit=5e0265350ff0d7420440965cf0b666f2cf4ebcf5


abextm:
btw, <https://github.com/usa-usa-usa-usa/inferno-wave-splits/blob/5e0265350ff0d7420440965cf0b666f2cf4ebcf5/src/main/java/com/infernosplittimer/InfernoSplitTimerPlugin.java#L108> is probably not reliable because the item manager makes no guarantees about reference comparison

usa-usa-usa-usa:
> btw, https://github.com/usa-usa-usa-usa/inferno-wave-splits/blob/5e0265350ff0d7420440965cf0b666f2cf4ebcf5/src/main/java/com/infernosplittimer/InfernoSplitTimerPlugin.java#L108 is probably not reliable because the item manager makes no guarantees about reference comparison

This was me dabbling with the runelite API for pretty much the first time ever after some discussion with the guys in the inferno discord, plus I am not a Java guy. So I get its kinda wonky. I was initially trying to do this as an extension of the core timers plugin, instead of adding an entirely separate timer myself.

If I stay that route, could you point me in a better direction for better direction for retrieving the elapsed time from the core timers plugin? Or is the only real reliable way here to have this plugin also do the timing?

PR #575, opened 2020-08-27
NMZ Optimal Points plugin 1.0.1

bloogeyz:
Updated runelite version in gradle.build to update required library


abextm:
Why did you make this pr? You shouldn't need to update the runelite version in your build.gradle except for when you run it from the test.

bloogeyz:
> 
> 
> Why did you make this pr? You shouldn't need to update the runelite version in your build.gradle except for when you run it from the test.

Seems I'm on the wrong track, plugin was reported not working in the game but is working locally so I had assumed it was a library issue but based on what you've said it shouldn't be.

abextm:
It likely doesn't work because you are not distributing `nmzbosses.csv`

bloogeyz:
> 
> 
> It likely doesn't work because you are not distributing `nmzbosses.csv`

Thank you Abex I should've caught that, have updated the branch and the commit to match

PR #577, opened 2020-08-29
Create item-comparison

intervenience:
Allows users to compare items stats in the sidebar, instead of having to load the wiki in a browser or visit the GE in game.
It queries the oldschool.runescape.wiki API for the search parameters input by the user and then will either: 
a) return an error if it does not exist
b) accept the page redirect by the wiki for searches that are similarly named (ie Firecape redirects to Fire_cape) and then
c) load the contents of the page found by the wiki API.

The contents are scanned for the item stat bonuses, which are then displayed in their appropriate text field next to icons to indicate what kind of stat bonus they are.

image

### DIFF
--- a/plugins/item-comparison
+++ b/plugins/item-comparison
@@ -1,0 +1,2 @@
+repository=https://github.com/intervenience/Runelite-Item-Comparison.git
+commit=0e73e552776427f1069baa1d46aba3ff268783d8
--- a/plugins/item-comparison
+++ b/plugins/item-comparison
@@ -1,2 +1,2 @@
 repository=https://github.com/intervenience/Runelite-Item-Comparison.git
-commit=0e73e552776427f1069baa1d46aba3ff268783d8
+commit=226b5cb94aba03c6080ee61dd906c18a3b802648


Nightfirecat (changes requested):
`plugins=com.intervenience.comparison` must be `plugins=com.intervenience.comparison.ComparisonPlugin` in your runelite-plugin.properties file.

Nightfirecat:
This looks like it should function fine, but you can definitely improve the performance of this plugin by leaning on `ItemManager.getItemStats()` and getting itemIDs via scanning item names (using `itemManager.getItemComposition().getName()` while scanning the client's item list) instead of doing a new wiki request each search--we already fetch item stats from our wiki scraping in ItemManager so that data is already present in the client.

If you want to implement that, I recommend [seeing how this is accomplished in ChatboxItemSearch](https://github.com/runelite/runelite/blob/4668140df145a683bb9a919f08dc0f28efa8384c/runelite-client/src/main/java/net/runelite/client/game/chatbox/ChatboxItemSearch.java#L295-L328).

intervenience:
Yeah at the time of writing the plugin I didn't realise the item data was already in the client. And I'm not sure if the item manager has something like this, but with the wiki if I was to search 'ags' for instance it'll return a redirect to 'Armadyl godsword' (the downside being more lag for a result), but I'll look into that some more. Thanks Nightfirecat

abextm:
Since this has outstanding required changes and hasn't been updated for a while, I am closing it as stale. You can reopen it once you make the required changes.

PR #580, opened 2020-08-30
Update Raid Tracker: added ui and config

CasvM:
datatracker config

### DIFF
--- a/plugins/raid-data-tracker
+++ b/plugins/raid-data-tracker
@@ -1,2 +1,2 @@
 repository=https://github.com/CasvM/raid-tracker.git
-commit=3e5dafc22a41ad7f7898a54188825b5fbdcc15ec
+commit=959d696982bd530fe3af9f43f0b2046af5299216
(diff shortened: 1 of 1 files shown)


abextm:
when writing to the .runelite you need to chose a name that is easily identifiable as your plugin; `loots` is not specific enough. Preferably use `raid-data-tracker`.

CasvM:
I could do that, I figured I could re-use the folder system from thestonedturtle's loot logger.
You think I should change to raid-data-tracker?

abextm:
unless you need to share data with that other plugin (which should also use a better name) you need to change the name to something identifiable.

CasvM:
Just changed the main directory from loots to raid-data-tracker, thanks for the quick replies!

WorstHerp:
@CasvM Hey, is it possible to add loots you've logged using other sources to your log inside the raid data tracker?

CasvM:
while it is possible, it can't be done automatically. The data tracker tracks the loots by adding a line to the raid_tracker_data.log whenever you finish a raid, there's no way of determining the variables if it's not done from within the raid. 

You can add the variables manually though, simply copy one of the lines in raid_tracker_data.log and put the correct values in there.

PR #581, opened 2020-09-01
Add pathfinder plugin

Milo-Piazza:
Pathfinder is similar to the built-in tile indicator plugin, but it computes and highlights the path the player's character will take to the hovered tile. It is mainly meant for Zalcano to help navigate the red symbols. For now it assumes there are no obstructions between the player and the tile, but this could be changed later.

I couldn't determine whether or not this breaks Jagex's third party client rules; to me it's in a bit of a grey zone. Sorry in advance if it is against the rules. I also couldn't find any plugins implementing this functionality already.

### DIFF
--- a/plugins/pathfinder
+++ b/plugins/pathfinder
@@ -1,0 +1,2 @@
+repository=https://github.com/Milo-Piazza/pathfinder.git
+commit=36577c233135e3c3ff3fbc4f4a18cbe587c5549b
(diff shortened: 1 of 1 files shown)


Nightfirecat (changes requested):
You must [license your repository.](https://github.com/runelite/plugin-hub#licensing-your-repository)

melkypie:
As I can see you added the license file but you should update the commit hash to point to the commit which has the license file included.

PR #583, opened 2020-09-03
Create music-replacer

alowaniak:
Gives the ability to replace music tracks with music of your own choosing (as requested in https://github.com/runelite/runelite/issues/3390).
See the [readme](https://github.com/alowaniak/music-replacer/blob/master/README.md) for more details.

### DIFF
--- a/plugins/music-replacer
+++ b/plugins/music-replacer
@@ -1,0 +1,2 @@
+repository=https://github.com/alowaniak/music-replacer.git
+commit=08a95c37b79fefdf52c1594991c94fc7e5841954


alowaniak:
Locally my build works fine, but the travis build fails because it can't find any of the JavaFX dependencies.
I added an implementation 'org.openjfx:javafx-media:14' dependency but it seems these jars are empty and the actual content exists in platform specific jars (of which I'm already not sure how I would even go about to add as dependency since it's a classifier and not a seperate Artifact).
I'm suspecting my local build works because I've got JavaFX bundled in my (oracle) jdk1.8, and it isn't bundled in the OpenJDK on the travis build (and obviously also won't be on the jdk 11 build).
~~What I also don't understand is that earlier I added the same dependency in the RuneLite maven pom (I changed the music plugin to add my functionality) and I could see in my IntelliJ dependencies that it added the javafx-media:win:14 while I didn't do anything additional. So I don't understand how in the RuneLite build it added those but for the Gradle it doesn't .~~
I'm guessing the different platform classifiers is something built-in in maven but not in Gradle (which is why the javafx-gradle-plugin exists).

I'm not sure what I need to do to fix this.
I've found https://github.com/openjfx/javafx-gradle-plugin which seems like it could help a bit, however that plugin is built with Java 11 and also needs at least 11 to run (which I assume won't happen with the java 8 build on travis).
I'm also not sure how much it could help, because afaict it will pull in the correct platform dependency for the platform it's building on.
However I assume that people using the plugin will need the JavaFX build for the platform they are running RuneLite on.
(Unless perhaps for the MediaPlayer no platform-specific things are hit? Or I just pull in all patform jars in the plugin somehow?)

So I'm a bit lost.
And have a hard time finding any other media player libraries for java.
I might actually try to package `ffmpeg` into the plugin to convert the files to `.wav`.🤷‍♂️
If anyone has a better idea please do tell.

PR #598, opened 2020-09-08
Create fishing-notifier

Bram91:


Nightfirecat:
Just to understand the use case here, is this just to have a separate notifier plugin with instant notifications instead of using the idle notifier plugin with a very short idle time?

Broooklyn:
That is correct - Think Abex's Idle Notifier but for Fishing instead of Mining

Bram91:
Yup that's correct

Nightfirecat:
You must [license your repository](https://github.com/runelite/plugin-hub#licensing-your-repository).

Bram91:
Oops 🙃  fixed now.

PR #600, opened 2020-09-09
Flipping-Utilities Bugfix

zumaad:
Sneaky bug resulting in an error bc because we were setting who an offer was made by incorrectly


zumaad:
wow thanks a ton for the swift merge @Nightfirecat

PR #607, opened 2020-09-11
Create thenightmare

Falsifyed:


Hydrox6:
There should only be 2 lines in the plugin file; see every other plugin file for examples

Also, the second feature in your description is banned by the [Jagex Statement](https://secure.runescape.com/m=news/another-message-about-unofficial-clients?oldschool=1)

Hydrox6:
Your plugins repo also needs to be public

Falsifyed:
I removed the repo after reading the Jagex Statement, thanks for the info!
________________________________

PR #616, opened 2020-09-17
SuppliesTracker: Persistence update

Dyldozer:
Tracks supplies used between accounts and sessions

-loads previous supplies on log in


Nightfirecat:
You should use a more specific directory for saving this data. I recommend the same name as your plugin: `supplies-tracker`

Dyldozer:
> You should use a more specific directory for saving this data. I recommend the same name as your plugin: `supplies-tracker`

fixed

PR #620, opened 2020-09-20
Flipping-Utilities: Flipping toolbar

zumaad:
**toolbar related:**
added flipping panel toolbar (ability to see favorited items, sort by roi, sort by profit, sort by time)

added ge favorite search

added ability to favorite items

**misc:**
added a config option for whether a user wanted to see items in the flipping tab in verbose view or not

fixed some of the stat panels sorting methods

added some logging statements to see how long rebuilds took


Adam-:
tangential, but we are getting a lot of reports of this plugin crashing clients on login. Can you look into fixing that? I've been recommending to people for a week now to just remove this plugin.

zumaad:
Yes, several people have contacted me about it :(. I've been looking into it, it has to do with large trade files being accumulated over time and the plugin trying to rebuild all the panels on login which can cause a spikes in CPU which cause crashes for some people. I've been looking at ways to optimize it. I think i just won't show all the panels, i'll implement some sort of pagination so only 10 panels are ever visible at a time. That should improve performance a ton.

plugins/flipping-utilities
```diff
@@ -1,2 +1,3 @@
 repository=https://github.com/Belieal/flipping-utilities.git
-commit=02837a0e71dab2752a4a86f7cf29b62533d1727e
+commit=be7b37ad3197c74f06c0b99851a9496f82cfcbfa
+
```
Nightfirecat, line 3: This line is causing the build failure--remove it.

Nightfirecat:
What is the source of the added images?

zumaad:
My bad, i added attribution to the icon creators in the Icon Atrtibutions section of the readme

PR #621, opened 2020-09-22
Update raid data tracker to v1.3.1

CasvM:
### v1.3.1
- Fixed a bug caused by not importing the StringEscapeUtils dependency in build.gradle (not tested).
- Improved ui performance by not pausing the swing thread, but instead updating the panel when the variable is loaded.


abextm:
This probably won't work on release because you are using a different version of commons-text than the client

PR #633, opened 2020-09-28
Disable crowdsourcing plugin

Adam-:


Nightfirecat:
Could we not just remove this instead?

melkypie:
I am pretty sure wiki wants to keep the plugin but needs time to remove the existing functionality so it doesn't collide with the core plugin. I think they mentioned that they want to use it for stuff that needs to get data fast and doesn't require the amount of data the core plugin can provide.

PR #657, opened 2020-10-05
Add simple license check

Nightfirecat:
This commit adds a build-time check for a LICENSE file with the BSD 2-Clause "Simplified" license header.

Unfortunately, GitHub's license API does not allow for checking the license type of a repo at a given commit, so we cannot use that here.

### DIFF
--- a/build_plugin.sh
+++ b/build_plugin.sh
@@ -86,6 +86,7 @@ SIGNING_KEY="" REPO_CREDS="" gradle \
 
 [ -s "$BUILDDIR/plugin.jar" ]
 [ -s "$BUILDDIR/plugin.manifest" ]
+[ -f "$BUILDDIR/LICENSE" ]
 
 cat "$BUILDDIR/plugin.manifest"
 


Nightfirecat:
For what it's worth, while this passes Travis right now, this would cause build failures due to numerous plugins which don't have licenses, and a few which have the wrong license:

```bash
$ for file in *; do                                                             
  if ! [ -f "$file/LICENSE" ]; then                                             
    echo "No license file found for '$file'"                                    
  else                                                                          
    if ! grep -q 'BSD 2-Clause License' "$file/LICENSE"; then                   
      echo "Possibly invalid license found for '$file'"                         
    fi                                                                          
  fi                                                                            
done                                                                            
No license file found for 'abexs-idle-notifier'                                 
Possibly invalid license found for 'advanced-notifications'                     
No license file found for 'bank-tab-names'                                      
No license file found for 'data-export'                                         
No license file found for 'deathindicator'                                      
No license file found for 'detached-camera'                                     
No license file found for 'discord-loot-logger'                                 
No license file found for 'emoji-palette'                                       
No license file found for 'emojiscape'                                          
No license file found for 'essence-running'                                     
No license file found for 'ess-pouch'                                           
No license file found for 'fake-iron-icon'
Possibly invalid license found for 'fixed-mode-hide-chat'
No license file found for 'global-consciousness'
No license file found for 'gpu-experimental'
No license file found for 'green-screen'
No license file found for 'health-notifier'
No license file found for 'http-server'
No license file found for 'inventory-summary'
No license file found for 'master-farmer'
No license file found for 'monkey-metrics'
No license file found for 'nature-rune-chest'
No license file found for 'profiles-panel'
No license file found for 'raid-points-overlay'
No license file found for 'region-locker'
No license file found for 'runite-rocks'
No license file found for 'tob-damage-counter'
No license file found for 'tob-healthbars'
No license file found for 'tray-notifications'
No license file found for 'zom-dense-essence'
No license file found for 'zom-leftclick-dropper'
No license file found for 'zom-nmz-util'
No license file found for 'zoom'
```

abextm:
I explicitly didn't make the script generate a license because that seems like it wouldn't legally enforceable.

I don't particularly care if the license is BSD-2, as long as it is permissive enough for us to distribute the resulting binary on it's own, so maybe just a check that the file exists is fine.

Maybe we can check the commit time and only enforce this for plugins updated after a certain date so we don't need to update ~20 plugins. Something like `[ "$(git log -1 --format=%ct "$PLUGINFILE")" -gt 1601945234 ]` maybe

Nightfirecat:
Since people are submitting plugins to us in the first place anyway, I figure something to generate a license in the first place would be fine and would help cut down on "Please add a license" type reviews. I can definitely make the license check more permissive. From <https://choosealicense.com/licenses/>, it looks like Apache, MIT, Boost Software, and Unlicense don't require distributing source. That said, Apache and MIT seem to indicate we'd need to distribute licenses even with binaries so even those seem incompatible.

As for licensing other repos, I see this as a long-term effort that we shouldn't pass up. I don't mind waiting to get this PR merged while we reach out to those plugin authors to license/change licenses as needed.

Nightfirecat:
I've dropped the template change and changed the license check to be a simple file check. From my previous comment, I feel we should spend the time to make sure all hub plugins are licensed so I'd prefer not adding leniency for older submissions.

Nightfirecat:
Abex has a commit for the new build script which will accomplish this.

PR #681, opened 2020-10-14
Add manifest for Dorgesh-Kaan lights plugin

andmcadams:
Plugin that tells you the locations of broken lamps in Dorgesh-Kaan based on the associated varbit. As you move around the city, the overlay will update with a list of known broken lights and the number of unknown ones (if any). These will be removed as you fix them. Note that the varbit only changes when you move between the two map squares and planes of the city, so you will need to move around quite a bit to discover all the lights.


dekvall:
Please [license](https://github.com/runelite/plugin-hub#licensing-your-repository) your repository.

PR #682, opened 2020-10-15
Add Share Links plugin

kashmoneygt:
### Intro
Hi all,
This is my first time contributing to RuneLite, so I'm very excited to present my plugin.
I created a plugin that lets users share links like Spotify music with others on OSRS. 

Right now, the plugin only works with Spotify tracks but I'm planning to add much more functionality ([see Roadmap](https://github.com/kashmoneygt/share-media#roadmap)) if this plugin gets approved.

### How it works
The idea:
- Both the sharer and sharee need to have the plugin installed.
- The sharer needs to copy the Spotify song link to their clipboard and type "!share".
- The sharee needs to do nothing.
- There is a Share Links Panel (similar to News Feed Panel) that maintains a list of shared links. It will get updated when links are shared.
- Both sharer and sharee can click on these links to open them. These links are not persisted to disk or RuneLite account -- it is on the roadmap.

See this gif

### Spotify APIs
At first, I was talking to Spotify APIs using client credentials. However, there is no way to provide my client secret to this plugin without adding it as plain text to the repo. Therefore, I'm opting to use Authorization Code Flow with PKCE ([see Spotify Auth docs](https://developer.spotify.com/documentation/general/guides/authorization-guide/)).

This means that when the user does "!share" with a Spotify link for the first time, a browser window will open and the user will be asked to authenticate with Spotify. I do cache the Spotify access token between RuneLite sessions, so the user will never get asked to authenticate more than once an hour. It's pretty annoying still, but there's no way for me to get around this.

Please let me know what you guys think. Open to any and all feedback. :)


loldudester:
This is essentially chat macro-ing, which isn't great.

kashmoneygt:
I wanted to explain the flow a bit more cause my code can be confusing (my b).

1. `updateChatMessageWithClipboardLink()` called for local player
2. `onChatMessage()` called for all users and looks for the message `!share <link>`

`updateChatMessageWithClipboardLink()` is activated when the local player types `!share`. It is activated before `onChatMessage()` because `onChatMessage()` has a priority of -1. It looks for a valid link in local player's clipboard and updates `!share` (hasn't been sent to others yet) to `!share <link>`. If it doesn't find a valid link in clipboard, it doesn't update the chat message.

Then, `onChatMessage()` (for all users including local player) looks for `!share <link>` simply adds the link as an item to their panel. Calls Spotify APIs here to get metadata about link.

So the message that is sent is the same for all users. Recipients see the message that the local player sees. The link in the message is from local player, not a link from recipient's clipboard.

---

A few things to note:

I specifically made sure that no new messages are created. Only the original message is updated. This structure is extremely similar to the structure in Chat Commands Plugin. That plugin registers async listeners for commands like "!total" -> local player types "!total" -> all users first see "!total" -> the message in chatbox is updated to the actual value of the total for all users. The only difference is that mine is synchronous because the message needs to contain the link before it's sent because `onChatMessage()` is looking for a message containing the link.

I also made sure that the links in the updated message can't be malicious. Since user clipboard content must adhere to Spotify regex pattern, nothing besides "spotify:track:<track_id>" and "https://open.spotify.com/track/<track_id>" will work. If the clipboard content isn't valid, the message will just go through as `!share` instead of `!share <link>`.
I can create a list of lots of edge cases and run them through the regex if you'd like to see something more concrete.

raiyni:
As @loldudester mentioned in their edit, this plugin won't work people because you are only reading the local clipboard. It seems to work for you because you have two clients open on the same pc, reading the same clipboard.

kashmoneygt:
Good point -- let me do some testing with clients on different PCs and get back to you.

abextm:
Since this has outstanding required changes and hasn't been updated for a while, I am closing it as stale. You can reopen it once you make the required changes.

PR #685, opened 2020-10-19
Add Bank Memory plugin

Lazyfaith:
Features include:

- Remember users banks across all acounts (all accessible at any time)
- Automatic updating as their bank contents changes
- Saving a snapshot of a bank
- Comparing 2 bank saves to see the item differences
- Quick search of bank contents

### DIFF
--- a/plugins/bank-memory
+++ b/plugins/bank-memory
@@ -1,2 +1,2 @@
 repository=https://github.com/Lazyfaith/runelite-bank-memory-plugin.git
-commit=eb57232651d2686ff33e23b764471481d29849e0
+commit=0b981e57520df2e9ce2a1502c11f058a913180c0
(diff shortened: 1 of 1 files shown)


abextm:
You are adding guice to your deps which is already included from runelite-client. This will break if the versions are not in sync.

Lazyfaith:
Thanks for looking over this!

I've removed the Guice dependency, though I did keep the guice-testLib dependecy which I think was the right thing to do as RuneLite declares it in the "test" scope so doesn't provide it transitively. I've not much experience meddling with Gradle builds though so please let me know if that was wrong of me.

Nightfirecat:
Since you are only deviating from the default build.gradle in test dependencies, it's not needed to pin your dependencies.

PR #692, opened 2020-10-21
Add ability to build with github actions

melkypie:
Once merged this needs secrets (REPO_CREDS, REPO_ROOT, SIGNING_KEY) setup for it to work (tell me if I missed any).
This is ran only manually under Actions->plugin-hub builder->Run Workflow.

Unfortunately due to github actions cache not supporting workflow_dispatch, there is no way to keep a gradle cache.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,0 +1,23 @@
+name: plugin-hub builder
+
+on:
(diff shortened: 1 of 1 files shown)


abextm:
Added in df657e20f58f6f47492ba6594356df3fb320be53

PR #698, opened 2020-10-24
Add extended-api plugin

manmademagic:
This is a slightly more advanced version of  [Adam-'s HTTP Server Plugin](https://github.com/Adam-/runelite-plugins/tree/http-plugin)

It allows for changing ports, as well as exposing some extra data.

### DIFF
--- a/plugins/extended-api
+++ b/plugins/extended-api
@@ -1,0 +1,2 @@
+repository=https://github.com/manmademagic/Extended-API.git
+commit=bf453c450e440612c3a10d6d5cc1d0ac024741ba


Nightfirecat:
Is there any reason this isn't an update to that plugin instead of a completely separate plugin?

manmademagic:
> 
> 
> Is there any reason this isn't an update to that plugin instead of a completely separate plugin?

No, not really, I was just following the wiki instructions for submitting.
I'm not super versed with doing pull requests, so my apologies if this was the wrong way.

What would be the best way to handle updating the current one?
Just changing [plugin-hub/plugins/http-server](https://github.com/runelite/plugin-hub/blob/master/plugins/http-server) to use my repo's details?

Nightfirecat:
I would recommend forking adam's repository, making the changes you want to contribute, then opening a PR against his repository, which he can then accept and update his plugin's record on this repository to use that updated code. There's not necessarily any problems with having multiple plugins which have similar functions, but if this is literally just the same as another one with a few enhancements, it would overall be better to then keep it to that one plugin.

manmademagic:
Yep, that's a fair call

Feel free to reject this one then, and I'll go through and fork through Adam's repo

PR #704, opened 2020-10-26
Added Toxic NPCs plugin

ejedev:
This plugin makes NPC's flame you if you die to them. 

implements runelite/runelite#12683


Nightfirecat (commented):
I would recommend removing `enabledByDefault = false` from your plugin descriptor since that is a confusing experience for users to install a new plugin and not have it active right away.
Why is `loadWhenOutdated = true` set?

Finally, I should mention that some contributors/maintainers have argued we may not want this feature in RuneLite at all, due to the feel-bad nature of the plugin. I'm personally of the opinion that this would be fine as implemented here though.

ejedev:
Closed due to feedback from development channel. Plugin will not be developed further by myself.

PR #706, opened 2020-10-26
Update: Profiles (v1.1.5)

kmarius:
Post UsernameChanged events to make the plugin work with the farming timers and Bitwarden plugin. @Spedwards can take back the plugin when he finds time for development.


Nightfirecat (changes requested):
As mentioned on Discord, we don't allow plugin transfers without the original author's express permission.

abextm:
Since this plugin deals with credentials I don't want to transfer it without @Spedwards permission. As he hasn't given that for over a month I am closing this as stale.

PR #707, opened 2020-10-27
Add Shortest Path plugin

Runemoro:
Simple plugin that draws the shortest path to a chosen destination on the map. Here's what it looks like: https://streamable.com/ln5q6c

### DIFF
--- a/plugins/shortest-path
+++ b/plugins/shortest-path
@@ -1,2 +1,2 @@
 repository=https://github.com/Runemoro/shortest-path.git
-commit=1a6e5183d09e2b375ac45c26206af18828b6fecf
+commit=7debf49c53b8c922cc1dd46a0b23ec3956db18f4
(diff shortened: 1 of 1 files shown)


Nightfirecat (changes requested):
Please add a support URL to your `runelite-plugin.properties` file. Most plugin authors refer users to the GitHub page for their plugin&ndash;in this case, that would be <https://github.com/Runemoro/shortest-path>.
Where are the `collision-map` and `transports.txt` resource files sourced?

Runemoro:
`collision-map` is generated by me using RuneLite's cache API.

`transports.txt` is from [the RS wiki data](https://docs.google.com/spreadsheets/d/1lqFDK8Yhpx4KmaublYh50v8a_ucPfBUiBeQN2E_C7_Y/edit), licensed under CC BY-NC-SA 3.0 (same as rest of the wiki content). I added attribution in the readme and the top of that file.

PR #733, opened 2020-11-03
Spawn Marker plugin

TalesRK:
This plugin marks the tile where creatures have spawned, with the creature's name on it.
It can be useful for hunting, combat, slayer, etc.

### DIFF
--- a/plugins/spawn-marker
+++ b/plugins/spawn-marker
@@ -1,0 +1,2 @@
+repository=https://github.com/TalesRK/spawn_marker.git
+commit=706bbc49841f1fff657d53fa0be10eb8a97f20a8


abextm:
This seems pretty similar to the respawn timer functionality provided by NPC Indicators?

TalesRK:
I was not aware of this feature in RuneLite and looking at it I can see that it is similar, for sure.
There is a difference that the tile is not marked on the map when you hunt or kill something.
If someone wants to track all of a creature's spawns, for example, a red chimp, it will need to manually mark the tiles where the spawns are displayed for 3 seconds while doing the activity they really want to do.
With this plugin, it will be marked automatically and for as long as they wish.

PR #735, opened 2020-11-03
Refactor Tob Drop Chance plugin

winterdaze:


Hydrox6 (approved):
Code LGTM, not 100% sure about moving repo and packages like that so I'm going to wait for probably Abex to confirm.

PR #738, opened 2020-11-04
Add the osleague.tools exporter plugin

spudjb:
This plugin allows users to export their completed leagues tasks both in CSV format, and a format compatible with https://www.osleague.tools which is a very handy website to track progression!


spudjb:
I talked with the developers over at osleague.tools and there's another plugin in development with better integration with the website. So I'm closing my PR in favur of that upcoming one.

PR #755, opened 2020-11-09
add dont-eat-it

Hydrox6:
A simple plugin that swaps *every* item's left click to Use, as long as they have one.


Hydrox6:
forgot a logo

PR #772, opened 2020-11-11
Update Damage Counter to v1.3

0anth:
- Included kill time in overlay, rather than mouseover
- Fixed bug with Barrows brothers


0anth:
- Updated to v1.3
- Included config option to always hide overlay

PR #780, opened 2020-11-11
Update Corp FFA07 plugin

Box-Of-Hats:
## New Features
- Allow tagging of players
- Take screenshots of crashers
- Check gear on spawn instead of just when a player attacks corp
- Show the player's wielding a non-spec weapon if a player has 0/1 specced
- Regroup config options to be more user friendy
- Remove "always on" config option as it was not needed

## Bug fixes
- Remove teled players from the player count
- Fix duplicate player names in player list

### DIFF
--- a/plugins/corp-ffa07
+++ b/plugins/corp-ffa07
@@ -1,2 +1,2 @@
 repository=https://github.com/Box-Of-Hats/runelite-plugin-corp-ffa.git
-commit=615c7e8d2c2cfd4bcc89b2124b4f6e6debeb1cb0
+commit=dc538decc6f71d6b6bcc48ea7d86c4ea8214f8fa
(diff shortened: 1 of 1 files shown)


abextm:
Generally we disallow player tagging, though its probably fine if you make it limited strictly to the Corp cave. As it is this works anywhere as long as you have visited the cave once

Box-Of-Hats:
Whenever the player changes location, there is a check for if they are in the corp cave which makes the plugin active/disabled. If  `IsActive` is false (i.e not in corp cave) then player tagging and the other features aren't enabled. Because of this logic and in my testing, this should already be limited to corp cave.

If this is still against the rules then I'm happy to remove the tagging feature :)

PR #783, opened 2020-11-12
Update better-renderer to v1.4

Runemoro:
Adds support for Linux, macOS, and 32-bit Windows.

The native inside lwjgl3-awt-0.1.8-SNAPSHOT.jar is the same one as https://github.com/LWJGLX/lwjgl3-awt/blob/main/native/macosx/liblwjgl3awt.dylib (using a custom build until they merge this PR).


Runemoro:
Switched to jitpack rather than jar to make verifying the dependency easier.

abextm:
jitpack makes it difficult to verify and will probably make your plugin randomly stop building so we disallow it

Runemoro:
Ok, I reverted it back to including the jar.

PR #803, opened 2020-11-18
Add hydrate reminder plugin

jmakhack:
Hydrate Reminder

A simple plugin that sends reminders to take hydration breaks on a configurable time interval.

Reminder notifications can be set to be sent via chat notification, computer notification, or both.

<hr>

Never forget to drink water while exploring the world of Old School RuneScape ever again! 

ﾟ+｡:.(\*･ω･)o旦　旦o(･ω･\*).:｡+ﾟ


Hydrox6:
You could have just pushed another commit to the other PR if you wanted to update it while it's still under consideration

PR #814, opened 2020-11-25
Update Os league plugin: Remove hardcoded task count

chaiinchomp:
Game update today duplicated a task (nice one jagex) which broke the plugin since it depends on a hardcoded task count. Removed that to just rely on checking the task filters state instead.


tylerthardy:
I approved this merge. @chaiinchomp is a contributor on the Runelite plugin.

PR #816, opened 2020-11-26
Create alch-items

ajsprojects:
Plugin that shows you the best items to high alch


ajsprojects:
Any idea what I can do to reduce the output jar file size?

Adam-:
It's probably due to this `htmlunit` you've added, how large is it?

ajsprojects:
> It's probably due to this `htmlunit` dep you've added, how large is it?

Only 2MB according to maven repo. If it is causing the issue is there any workout aside from not using it?

Adam-:
The total size of it also includes its transitive dependencies, so it could be large. If you are using it to read the wiki, maybe look at using the wikis api instead?

PR #823, opened 2020-11-30
Create alchitems

ajsprojects:
Updated PR for alch items plugin - Plugin that shows you the best items to high alch!

I removed item trade volume for now, will consider it as a future update once I decide the best route to take


abextm:
The wiki asks that all requests to it have a ua set. Also you can get all of this information from the ItemManager so you don't need to request anything in the first place.

ajsprojects:
> The wiki asks that all requests to it have a ua set. Also you can get all of this information from the ItemManager so you don't need to request anything in the first place.

I can add user agent

You can get some of the information but what about trade limit for example? Also there would be no way of knowing which items you should query through item manager without querying them all and checking profit surely?

ajsprojects:
Updated with user agent

Also with a future update I'm hoping to add item trade volume information in so i'd need a way of doing that anyway
At the moment it only takes a few seconds to fetch and parse the data so by the time runelite has fully loaded the plugin is pretty much ready

raiyni:
> > The wiki asks that all requests to it have a ua set. Also you can get all of this information from the ItemManager so you don't need to request anything in the first place.
> 
> I can add user agent
> 
> You can get some of the information but what about trade limit for example? Also there would be no way of knowing which items you should query through item manager without querying them all and checking profit surely?

We scrape trade limits from the Wiki

abextm:
also scraping html from the wiki is incredibly brittle; there are much better ways to scrape what you want that is simpler to implement, faster to run, is lower load for the wiki, more reliable, and doesn't require pulling in an HTML parser

ajsprojects:
Where can I get trade limit and trade volume data from runelite? Also there would be no innate way to know which items to even query for this data so the wiki is an easy way to get a decent list of items. 
I've looked at using the media wiki api but its not exactly user friendly and would require parsing alot more data than my current solution. I'm open to hear how you would implement this in a better way

abextm:
You can get the ge limit from `ItemManager::getItemStats`. To get a list you can just iterate every item that is `tradeable` (which actually marks if it is GE-able). If you need help operating the wiki's api you can always ask for help in [their discord](http://discord.gg/runescapewiki)

abextm:
That UA is obviously wrong. Please don't submit semi malicious code to here again.

ajsprojects:
> That UA is obviously wrong. Please don't submit semi malicious code to here again.

Considering I only added it at your request, assuming it was intentionally malicious would be a bit cynical. As a reviewer for an open source voluntary project I would have hoped for a bit more open-ness and guidance for best practice. Feel free to close the PR :)

abextm:
The point of the user agent is to identify the agent (your software) to the web server for the server operators reference. Using a browser's UA for an automated tool does the exact opposite; it actively obfuscates what your tool is. Doing this is basically only useful if you are abusing a service and don't want the owner of said service to be able to stop you easily. Automated tools like this typically use a user agent in the form `AlchItems/1.0 (+https://github.com/ajsprojects/AlchItemsPlugin)`, which gives a clear name and a easy way to contact you.

ajsprojects:
> The point of the user agent is to identify the agent (your software) to the web server for the server operators reference. Using a browser's UA for an automated tool does the exact opposite; it actively obfuscates what your tool is. Doing this is basically only useful if you are abusing a service and don't want the owner of said service to be able to stop you easily. Automated tools like this typically use a user agent in the form `AlchItems/1.0 (+https://github.com/ajsprojects/AlchItemsPlugin)`, which gives a clear name and a easy way to contact you.

So if I refactor the user agent are you happy to approve it? Or do you still think it needs reworking?

I don't see any easy way to get volume traded of items nor a list of items to check if they are profitable without scanning thousands of items through item manager

Adam-:
> 
> 
> > The point of the user agent is to identify the agent (your software) to the web server for the server operators reference. Using a browser's UA for an automated tool does the exact opposite; it actively obfuscates what your tool is. Doing this is basically only useful if you are abusing a service and don't want the owner of said service to be able to stop you easily. Automated tools like this typically use a user agent in the form `AlchItems/1.0 (+https://github.com/ajsprojects/AlchItemsPlugin)`, which gives a clear name and a easy way to contact you.
> 
> So if I refactor the user agent are you happy to approve it? Or do you still think it needs reworking?
> 
> I don't see any easy way to get volume traded of items nor a list of items to check if they are profitable without scanning thousands of items through item manager

Yes. We just don't want to make the wiki sysadmin's lives harder since RuneLite has a lot of users and can generate large amounts of traffic, and it is rude to try and masquerade it as normal web browser traffic. With enough traffic, they (the wiki) may even force you over to use their API or so, but we don't want to aid you in eg. trying to bypass them protecting themselves.

ajsprojects:
> > > The point of the user agent is to identify the agent (your software) to the web server for the server operators reference. Using a browser's UA for an automated tool does the exact opposite; it actively obfuscates what your tool is. Doing this is basically only useful if you are abusing a service and don't want the owner of said service to be able to stop you easily. Automated tools like this typically use a user agent in the form `AlchItems/1.0 (+https://github.com/ajsprojects/AlchItemsPlugin)`, which gives a clear name and a easy way to contact you.
> > 
> > 
> > So if I refactor the user agent are you happy to approve it? Or do you still think it needs reworking?
> > I don't see any easy way to get volume traded of items nor a list of items to check if they are profitable without scanning thousands of items through item manager
> 
> Yes. We just don't want to make the wiki sysadmin's lives harder since RuneLite has a lot of users and can generate large amounts of traffic, and it is rude to try and masquerade it as normal web browser traffic. With enough traffic, they (the wiki) may even force you over to use their API or so, but we don't want to aid you in eg. trying to bypass them protecting themselves.

Yes as in I need to do more refactoring aside from fixing the user agent?

Adam-:
I don't think you need to do anything besides fixing the ua. Maybe abex will see something else.

ajsprojects:
Updated UA

Not sure if I messed up pulling updates into my runelite plugins branch so let me know if i've made a mistake

abextm:
Your history on this branch is all messed up and updating 5 different plugins that you don't own. You will need to force push this branch with a fixed history.

Since you are pulling in jsoup you MUST [pin your dependencies](https://github.com/runelite/plugin-hub#third-party-dependencies)

I would like it, though I don't strictly require it, if you fixed it to not hit the wiki like this; it is very fragile and the alchemy page is the slowest page on the wiki. Everything you currently get from it is available in the client from the ItemManager like I said. Scanning the whole item list on the client is orders of magnitude faster than doing any HTTP request. Also you may want to look into https://meta.weirdgloop.org/w/Forum:PSA:_Grand_Exchange_changes as may contain a solution to your volume problem soon.

ajsprojects:
Ill close the PR for now because something went wrong with pulling updates and ill test out item manager 
Cheers

PR #824, opened 2020-11-30
Add easyparty plugin

wesley-221:
A plugin that allows you to create a party using the PartyService by sharing a party UUID, without having to add people on Discord. 

On a later date I will try to implement inviting through Runescape itself so you can simply enter a command in chat and it'll generate an invite for people to accept.


abextm:
This needs to remove the nav button on `shutDown`. Otherwise this is fine.

PR #835, opened 2020-12-04
Add Custom Settings Easter Eggs plugin

Cyborger1:
Allows players to set up custom easter eggs for the reworked settings interface.

egg_example


raiyni (approved):
FWIW you can use guava splitters to trim results and omit empty results

Cyborger1:
I'm looking into using Guava and checking VARC instead of getting out the Widget text.

Cyborger1:
I changed the separator from `=` to `=>` so make sure to hit reset in the config of fix them up yourself if you were testing this on your end.

PR #838, opened 2020-12-05
Create custom-item-hovers

geel9:
Add custom-item-hovers

This plugin essentially just reads json files from a user-specified directory, which contain text to display on hovering over specified items.

For example, this hover:

https://reel.geel.tf/sopaszqsb88ekdfh22c6.png

Can be produced by this json:

```
{
  "items_regex": ["MAGIC_(SEED|SEEDLING|SEEDLING_W|SAPLING)"],
  "hovers": [
    [
    	"<col=3498db>Farming:</col>", 
    	"-Level: 75", 
    	"-Payment: 25x Coconut", 
    	"-Check XP: 13,768"
    ]
  ]
}

### DIFF
--- a/plugins/custom-item-hovers
+++ b/plugins/custom-item-hovers
@@ -1,0 +1,2 @@
+repository=https://github.com/geel9/runelite-custom-item-hovers.git
+commit=b73a62a54b82c3e5658063a376409d65c67d2a7b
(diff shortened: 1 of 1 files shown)


abextm:
This should not use the user's home directory if the config has an empty string. Generally plugins should only read/write to a plugin-named folder in the `.runelite` directory (`RuneLite.RUNELITE_DIR`)

geel9:
I'm not sure where it used the home directory, but I agree that a user-specified directory is a bad idea regardless.

I've changed it so it creates and exclusively uses a `customitemhovers` directory in the `.runelite` directory.

PR #840, opened 2020-12-06
Update Flipping-Utilities

zumaad:
Adding third party dependencies for excel exporter


zumaad:
closing cause the excel dependencies are too large :(

zumaad:
Is there anyway to check what is contributing to plugin size? Perhaps I'm adding something unnecessary in there that I can prune.

PR #841, opened 2020-12-06
Add shared implings plugin

peterpanneke3:
Adds the SharedImplings plugin.

The plugin allows to automaticaly share the location of scouted dragon & lucky implings with other users.

Readme available at https://github.com/peterpanneke3/sharedImplings/blob/master/README.md

### DIFF
--- a/plugins/shared-implings
+++ b/plugins/shared-implings
@@ -1,0 +1,2 @@
+repository=https://github.com/peterpanneke3/sharedImplings.git
+commit=2e43934d9a9849872c7490cd3cf4d7d755d588be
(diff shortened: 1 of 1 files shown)


abextm:
I would appreciate it if you used OkHttp's websocket client so I don't have to verify this dep

peterpanneke3:
> OkHttp's websocket client



> I would appreciate it if you used OkHttp's websocket client so I don't have to verify this dep

Fair point! I'll get on with it

abextm:
Thanks that makes it much easier.

You can use `@Slf4j` instead of using the LoggerFactory manually, and [this](https://github.com/peterpanneke3/sharedImplings/blob/760eed8ee9d0b38fbf77a35f9498486579f6c163/src/main/java/be/sharedimplings/SharedImplingsPlugin.java#L192-L193) can just be `client.getCachedNPCs()[imp.getNpc().getIndex()] != null`, or better yet it could just be a Set<NPC> that you only add/remove from in onNpc(De)?spawned

PR #844, opened 2020-12-06
Emote Clue Items v1.0.3

larsvansoest:
Emote Clue Items is a simple RuneLite plugin which highlights items that are required for emote clue steps. 

Maintaining free bank space can be quite cumbersome, especially when you are not sure which items should be kept for future clue scrolls. With this plugin, throwing away items may be a bit less stressful, as this plugin aims to highlight all emote clue items.


abextm:
you didn't put the plugin descriptor in the plugins folder

larsvansoest:
I am currently working on fixing the build fail. This is my first plugin, and unfortunately I forgot to fill in the properties file.

larsvansoest:
@abextm thanks for your response. As of now, I have successfully built the pull request with my plugin in its plugins folder.

abextm:
[ImageIO usages must be wrapped with ImageUtil](https://github.com/larsvansoest/emote-clue-items/blob/19efb29600d2e8df4d662fcc9c8fea22cc8a6b70/src/main/java/com/larsvansoest/runelite/clueitems/overlay/icons/EmoteClueIconProvider.java#L36)
you want to use `ImageUtil.getResourceStreamFromClass`

larsvansoest:
@abextm thanks for your quick response. Conforming to your instructions, I have wrapped ImageIO using util. In the future, is there somewhere I can find a list of these coding requirements? If so, I may be able to prevent these issues in any future release.

abextm:
We don't have a proper list. The ImageIO thing is really the only instance of "you must use api X instead of Y", because using ImageIO without the synchronization that ImageUtil provides can crash the client.

PR #851, opened 2020-12-08
Update Damage Counter to v1.5.0

0anth:
Counter now automatically resets when engaging combat with a different boss.
Counter will no longer automatically reset when teleporting/running away from boss.
Counter will still automatically reset when boss dies.
Extra config option added to allow custom NPCs to be tracked by name.


abextm:
you changed the capitalization of your config group, so it will reset everyone's config for this plugin to the default

0anth:
Whoops. Reverted config group name change, thanks.

PR #853, opened 2020-12-09
Sharedimplingsv03

peterpanneke3:
version 0.3:
-Add NINJA and MAGPIE imps
-Add configuration options for puro-puro


peterpanneke3:
Add direction to the impling text in overlay if the impling is fairly close and in the current world

PR #871, opened 2020-12-16
world-location: API change fix & config simplification

XrioBtw:
Fix for upcoming 1.6.36 API changes.
Needs `hold-for-release`.
Also contains some config changes.


abextm:
@XrioBtw This fails against the current 1.6.36 snapshot

PR #879, opened 2020-12-21
updated ban-list

fangthe1st:
removed onWidgetHiddenChanged


abextm:
`onScriptPostFired` is a hot event, you must always check the scriptid before doing other work with it

PR #881, opened 2020-12-21
example-plugin: Remove slf4j-simple, use logback-classic from rl client

llamositopia:
The RL client jar already contains an slf4j implementation, logback-classic. The existing example-plugin template inserts a test depedency on slf4j-simple, and removes logback-classic as a transitive dependency of runelite-client.

This has the unfortunate side-effect of causing a ClassNotFoundException when using the --debug flag, since runelite-client expects a logback-classic Logger instance. 

This PR updates the project template to instead use the included logback-classic, removes slf4j-simple, and provides a logback-test.xml which only takes effect when testing the plugin, and is not included in the packaged jar. From this, we no longer have the crash when using --debug.


Adam- (approved):

templateplugin/src/test/resources/logback-test.xml
```diff
@@ -0,0 +1,11 @@
+<configuration>
+    <appender name="STDOUT" class="ch.qos.logback.core.ConsoleAppender">
+        <encoder>
+            <pattern>%d{HH:mm:ss.SSS} [%thread] %-5level %logger{36} - %msg%n</pattern>
```
abextm, line 4: This should mirror exactly what the client has. I would be fine with just not including this file to keep the template as clean as possible.

Adam-: I think the idea was to prevent the tests from writing to the .runelite logs dir right? If so this file would be required.

abextm: Yes, but in practice there are < 5 plugins with tests so I'm fine with not worrying about this.

llamositopia: New commit added that removes it. I'll add something to the wiki about this if it's merged.

PR #883, opened 2020-12-23
Create do-not-kick

MarbleTurtle:
Removes kick option from clan chat on configurable option


abextm:
I think it won't break anything, but normally we config groups are lowercase alphanumeric, so `@ConfigGroup("donotkick")`

PR #885, opened 2020-12-23
Add logging instructions regarding changes in #881

llamositopia:
Related to #881, added instructions to the README on how to create alternative logging setups, and requirements to do so.

### DIFF
--- a/README.md
+++ b/README.md
@@ -77,6 +77,29 @@ There are two methods to create an external plugin, you can either:
 
  8. When you have your plugin working. Commit your changes and push them to your repository.
 
+### Changing the default logging behaviour
+
+New plugin hub projects will default to using the same logging configuration as the standard RuneLite client.
+RuneLite uses logback-classic, with a configuration to log to to ~/.runelite/logs, using the INFO logging level.
+To enable DEBUG level logging, you can use the --debug CLI flag, as described in [Using the client developer tools](https://github.com/runelite/runelite/wiki/Using-the-client-developer-tools).
+
+If you prefer to use a different logging configuration, you can create your own logback-test.xml within your project's src/test/resources folder.
+This file *must* be a test resource, so that it is not included in your plugin distribution.
+As an example, the following logback-test.xml file would log only to console, with the DEBUG level:
+```xml
+<configuration>
+    <appender name="STDOUT" class="ch.qos.logback.core.ConsoleAppender">
+        <encoder>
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -77,6 +77,29 @@ There are two methods to create an external plugin, you can either:
 
  8. When you have your plugin working. Commit your changes and push them to your repository.
 
+### Changing the default logging behaviour
+
+New plugin hub projects will default to using the same logging configuration as the standard RuneLite client.
+RuneLite uses logback-classic, with a configuration to log to to ~/.runelite/logs, using the INFO logging level.
+To enable DEBUG level logging, you can use the --debug CLI flag, as described in [Using the client developer tools](https://github.com/runelite/runelite/wiki/Using-the-client-developer-tools).
+
+If you prefer to use a different logging configuration, you can create your own logback-test.xml within your project's src/test/resources folder.
```
abextm, line 86: I really don't see why anyone would want to do this. IMO it really isn't relevant to this document at all. It certainly shouldn't be included in the middle of the setup flow

llamositopia: Alr, I mentioned that I could add this documentation in #881 and didn't get any pushback. Considering the previous default was to use a completely different logger, I felt that it would be useful to know what the change was for new plugins. It's also right at the end of the setup instructions anyway.

Feel free to close if you think it's not useful.

abextm: Its between the repo creation and the licensing step of setup. If anything it should just be a link to the wiki at the bottom of the document

llamositopia: I added it to this README since this is the only page on the wiki regarding plugin hub development, and this information is only relevant to plugin hub development. If you like, I can move it to the bottom of the file, or we can make a whole new page on the wiki for it.

abextm: the entire point of this is that you can use `--debug` just like if you were not working on the a hub plugin

llamositopia: Right, but that information isn't shown or linked anywhere in the plugin hub documentation. The crux of this information is just letting new plugin hub developers know they can use --debug in this scenario, and where to read more information the developer tools that runelite-client provides, by linking to https://github.com/runelite/runelite/wiki/Using-the-client-developer-tools. 

The logback-test,xml is described as optional, and I feel could be useful for devs that don't want their hub dev logging to mix up with normal execution runelite logs, but we can remove that if you think it's too much info.

llamositopia:
Spoke on Discord. Expanding the plugin hub section of the wiki, and adding this as a page.

https://github.com/runelite/runelite/wiki/External-Plugin-Development-Logging

PR #887, opened 2020-12-24
Update Inventory-value-overlay to latest commit

wikiworm:
New feature to allow the inventory value calculation to ignore items.
Added defensive code.


abextm:
Changing your config group will cause all of your users' configs to get reset. Are you sure you want that?

wikiworm:
> 
> 
> Changing your config group will cause all of your users' configs to get reset. Are you sure you want that?

Yes -- I think that's desired with the new config improvements.

PR #888, opened 2020-12-24
Add combo-runes-only v1.0

llamositopia:
This plugin removes the Craft-rune option from the fire altar, so that users don't accidentally make fire runes instead of lava/steam/whatever


llamositopia:
I messed this up, will recreate shortly

PR #889, opened 2020-12-24
Add combo-runes-only v1.0

llamositopia:
This plugin removes the Craft-rune option from the fire altar, so that users don't accidentally make fire runes instead of lava/steam/whatever


abextm:
essence-running already does this btw

llamositopia:
I tried Essence Running and it wasn't for me. It's not an exact-match feature wise, since it opens the right click menu on click, instead of just ignoring it, and I didn't want to bring in all the other unwanted pieces of Essence Running.

llamositopia:
Originally I also wasn't planning to submit this since it's really small, but another user in the discord mentioned they would like to have it as well.

PR #900, opened 2020-12-30
Banked XP plugin PR

seelrr:
This plugin calculates the total amount of XP in your bank and displays it in an overlay (instead of a side panel). By being able to see your total banked XP across 9 skills (in one view) with just a click of a button, it makes consuming this data quick and easy. Also, because this plugin uses an overlay, it allows streamers to display this neat information to viewers.

### DIFF
--- a/plugins/bank-xp-value
+++ b/plugins/bank-xp-value
@@ -1,2 +1,2 @@
 repository=https://github.com/seelrr/bank-xp-value.git
-commit=8eac8fd9b8007c1f3f13126fd86bbfc451b2e925
+commit=ed19835d634e91582a007ced8689f955be1f6896
--- a/plugins/banked-xp
+++ b/plugins/banked-xp
@@ -1,2 +1,2 @@
 repository=https://github.com/seelrr/banked-xp.git
-commit=b7f4327d7231060d49f0128856f8f5f4acf21d7a
+commit=7c2db23b2789597a6ffda26d88460b00d6a5a7d3
--- a/plugins/banked-xp
+++ b/plugins/banked-xp
@@ -1,0 +1,2 @@
+repository=https://github.com/seelrr/banked-xp.git
+commit=b7f4327d7231060d49f0128856f8f5f4acf21d7a
--- a/plugins/banked-xp
+++ b/plugins/banked-xp
@@ -1,2 +1,2 @@
-repository=https://github.com/seelrr/banked-xp.git
-commit=7c2db23b2789597a6ffda26d88460b00d6a5a7d3
+repository=https://github.com/seelrr/bank-xp-value.git
+commit=8eac8fd9b8007c1f3f13126fd86bbfc451b2e925


raiyni:
FWIW this already exists as a plugin with a panel.

seelrr:
> FWIW this already exists as a plugin with a panel.

The reason why I created this is because I found the panel implementation to be tedious by having to flip through every option to see the totals. My version spits it all out in one window, along with a total across the skills. While the panel implementation is great, I think mine can appeal to those who appreciate quick data, and being able to see it all at once. Additionally, an overlay implementation may also appeal to those who prefer having more on-screen, such as streamers.

psikoi:
Surely it would be better for everyone if you contributed to the original plugin and added the overlay as a config, right?

seelrr:
> Surely it would be better for everyone if you contributed to the original plugin and added the overlay as a config, right?

Good observation! I thought about that myself, but decided not to, for multiple reasons:

For starters, the original states: "In order for the UI to register Inventory changes you must select a new skill via the dropdown". That said, I can't take out a bunch of items, and immediately see how it affects the totals. This was the correct design choice considering the panel would be rendering non-stop if it wasn't optimized; however, because my implementation is only active for a short period of time (only in the bank when toggled), I can allow for constant rendering. That means the results can dynamically change, which is useful. Because of the original's stated design principles, I wouldn't want to implement something that conflicts with them. #1.

#2, an overlay with his would require configuring values across multiple skills before getting results. For example, molten glass defaults to a beer glass. I can't support the efficiency I'm aiming for if I have to set multiple values before getting the results. My implementation uses a fixed XP rate for the items (most popular/best use for each item), which allows me to avoid that process -- a little, but significant, time saver. 

#3 An item in the original plugin will be applied to multiple skills, which means you can't have a number resembling the total amount of XP in your bank. That may not seem significant, but I think it would be cool for players to know exactly how much XP they have in their bank. My current version only supports 1 to 1 (i.e. logs only to firemaking), but I plan on making config options that allow you to pick and choose.

Essentially, if I wanted to implement theses concepts in the original, it would require a lot "ugly" changes that I doubt would even be approved (conflicting design principles, a bunch of config options, etc). From the way I see it, these are two different plugins with two different ideas -- his more geared towards the grinding player, while mine is more for the frequent bank inspecting.

seelrr:
> > Surely it would be better for everyone if you contributed to the original plugin and added the overlay as a config, right?
> 
> Good observation! I thought about that myself, but decided not to, for multiple reasons:
> 
> For starters, the original states: "In order for the UI to register Inventory changes you must select a new skill via the dropdown". That said, I can't take out a bunch of items, and immediately see how it affects the totals. This was the correct design choice considering the panel would be rendering non-stop if it wasn't optimized; however, because my implementation is only active for a short period of time (only in the bank when toggled), I can allow for constant rendering. That means the results can dynamically change, which is useful. Because of the original's stated design principles, I wouldn't want to implement something that conflicts with them. #1.
> 
> #2, an overlay with his would require configuring values across multiple skills before getting results. For example, molten glass defaults to a beer glass. I can't support the efficiency I'm aiming for if I have to set multiple values before getting the results. My implementation uses a fixed XP rate for the items (most popular/best use for each item), which allows me to avoid that process -- a little, but significant, time saver.
> 
> #3 An item in the original plugin will be applied to multiple skills, which means you can't have a number resembling the total amount of XP in your bank. That may not seem significant, but I think it would be cool for players to know exactly how much XP they have in their bank. My current version only supports 1 to 1 (i.e. logs only to firemaking), but I plan on making config options that allow you to pick and choose.
> 
> Essentially, if I wanted to implement theses concepts in the original, it would require a lot "ugly" changes that I doubt would even be approved (conflicting design principles, a bunch of config options, etc). From the way I see it, these are two different plugins with two different ideas -- his more geared towards the grinding player, while mine is more for the frequent bank inspecting.

TL;DR - the logic of the original is limited to dealing with one skill at a time. Its options and preferences allow for more personalized use, but at the cost of displaying multiple skills/convenience (correct me if I'm wrong). It would require a completely foreign logic/calculation algorithm and data set to implement it as an additional feature to the original. Enough for me to think it's deserving of it's own plugin.

I think there's a huge "market" of interest that isn't/can't be satisfied with the original. I can see people using mine while staring at their bank, pulling it up real quick right before closing it, comparing total banked XP amounts with friends, or seeing the effects of certain items as the totals dynamically change. It's just interesting data that people wouldn't bother with if it required a second longer.

abextm:
You need to choose a name that differentiates this from the existing Banked Experience plugin more.

seelrr:
Okay! How does "Bank XP Value" sound? This name gives me the idea of adding XP tooltips to items in the bank, which could distinguish them ever further.

seelrr:
> You need to choose a name that differentiates this from the existing Banked Experience plugin more.

Okay! How does "Bank XP Value" sound? This name gives me the idea of adding XP tooltips to items in the bank, which could distinguish them ever further.

Mrnice98:
if i am understanding this correctly, you wont be able to choose what method you want to plugin to calculates the XP. It just assumes you are going to be using it in the most "efficient way" ?

abextm:
`Bank XP Value` seems fine to me

seelrr:
> if i am understanding this correctly, you wont be able to choose what method you want the plugin to calculate the XP. It just assumes you are going to be using it in the most "efficient way" ?

Precisely. I implemented it this way for three reasons: 

#1 The number of multi-use items is very small. Out of a 3B bank loaded with random items, I used the original Banked Experience plugin to determine how many configurable items I had, and there were literally only 2: molten glass and gold bars. That said, I think giving the option to choose the method of using an item adds so much complexity and reduces ease of use for very minimal benefit. 

#2 Giving the option to choose how to use an item defeats the purpose of "Total Banked XP". The value is meaningless/doesn't work if it can be changed outside of items. Like, imagine if the Bank Value plugin allowed you to configure items to previous prices -- it would make your total bank value illegitimate. I want the values listed in my plugin to be fixed and comparable amongst players. In other words, my plugin is objective, while the original Banked Experience is subjective.

#3 It's 2020, who doesn't use the meta?

abextm:
[this](https://github.com/seelrr/bank-xp-value/blob/8eac8fd9b8007c1f3f13126fd86bbfc451b2e925/src/main/java/com/bankxpvalue/ItemDataCache.java#L61) won't work since its trying to read a file off disk, not out of resources. You should have [something like this](https://github.com/runelite/runelite/blob/431e09588bd505147f03e2c88786deed9834ff84/runelite-client/src/main/java/net/runelite/client/game/ItemVariationMapping.java#L56-L57). The file you are loading needs to be in `resources`, not `java` or it will not be included in the jar. Also gson can populate directly into fields, you don't need to do it manually.

Also you should rename the file you put in this repo to your plugin's new name.

seelrr:
> [this](https://github.com/seelrr/bank-xp-value/blob/8eac8fd9b8007c1f3f13126fd86bbfc451b2e925/src/main/java/com/bankxpvalue/ItemDataCache.java#L61) won't work since its trying to read a file off disk, not out of resources. You should have [something like this](https://github.com/runelite/runelite/blob/431e09588bd505147f03e2c88786deed9834ff84/runelite-client/src/main/java/net/runelite/client/game/ItemVariationMapping.java#L56-L57). The file you are loading needs to be in `resources`, not `java` or it will not be included in the jar. Also gson can populate directly into fields, you don't need to do it manually.
> 
> Also you should rename the file you put in this repo to your plugin's new name.

Couldn't figure out how to use gson to populate directly to the fields for the life of me. I don't think it's possible without changing the entire json file. If you have any pointers, I'll be more than happy to change it, but I left the manual implementation for now.

abextm:
```
class ItemDataContainer {
    List<ItemData> items;
}

ItemDataContainer data = gson.fromJson(..., ItemDataContainer.class);
```
should work

PR #917, opened 2021-01-03
Updated to v1.1

Elrol:
Removed Restrictions
Added Customization
Fixed Bugs

### DIFF
--- a/plugins/bone-running
+++ b/plugins/bone-running
@@ -1,2 +1,2 @@
 repository=https://github.com/Elrol/Run4LessPlugin.git
-commit=c945882835a8fc552008a0af0acc8dffc87b14e6
+commit=a7ff721555d36159c8ee8f1783cd30095050e087
--- a/plugins/bone-running
+++ b/plugins/bone-running
@@ -1,2 +1,2 @@
 repository=https://github.com/Elrol/Run4LessPlugin.git
-commit=a7ff721555d36159c8ee8f1783cd30095050e087
+commit=5c246f621889c8278594e6ef61119cb319d7bae9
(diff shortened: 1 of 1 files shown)


pokefreak:
we are still waiting on this to be checked elrol?

abextm:
Don't set an incorrect user agent https://github.com/Elrol/Run4LessPlugin/compare/dcadca1ba598b7e186c8981faf7607902dd42a22...Elrol:05ab652adf6d2744f703bccf65e9760b7d1ef8a3#diff-57ce4e164218995deb522b01e7a7e9256db73fcb597d2a6b3aa3632a0c8c75e5R37
(also you should really use okhttp, its much better than java's stdlib's http stuff)
ImageIO use must be in a `synchronized(ImageIO.class)` https://github.com/Elrol/Run4LessPlugin/compare/dcadca1ba598b7e186c8981faf7607902dd42a22...Elrol:05ab652adf6d2744f703bccf65e9760b7d1ef8a3#diff-b6c4ed828c64c3c7fe3236aa13c400b3867af0f53b476dd3f25cec0ed9d0846aR45

Nightfirecat (approved):

abextm:
You should still set a user agent, just not one that isn't yours. Again okhttp does this for you; you should use it.

You still haven't synchronized your ImageIO access. Again you should use okhttp for downloading that image too. Passing urls to things that handle them transparently is almost always wrong.

abextm:
Okhttp is already included in the client. manually pulling in a version will cause it to conflict and break

Elrol:
Good to know, wasnt aware it was included. Give me just a moment

Elrol:
Well then, looks like there are some issues that were not showing up in my ide. I will fix those really fast.

PR #927, opened 2021-01-06
Update OSRS Streamers to version 1.4

rhoiyds:
- Adding colours to right hand click menu depending on streamers live/not live status (toggleable config)
- Using config manager as a cache for the streamers config file and limiting retrieval from GitHub to once every 24 hours
- Adding basic tests
- Updating version number

### DIFF
--- a/plugins/osrs-streamers
+++ b/plugins/osrs-streamers
@@ -1,2 +1,2 @@
 repository=https://github.com/rhoiyds/osrs-streamers.git
-commit=0566d53f5dc1d1a33c8fcebd8e914b305db1da29
+commit=6657d6c23cbbd42f15b075cf465347e4027eaedc
--- a/plugins/osrs-streamers
+++ b/plugins/osrs-streamers
@@ -1,2 +1,2 @@
 repository=https://github.com/rhoiyds/osrs-streamers.git
-commit=1841f6aba571db177abf5f175be9343c1dec4399
+commit=0566d53f5dc1d1a33c8fcebd8e914b305db1da29


abextm:
You really shouldn't put this much data in the config manager. It has a limit of (iirc) 200k, and you are putting in about 190k. You can use RuneLiteAPI's OkHttp Client and it will handle caching for you automagically.

rhoiyds:
@abextm Damn, I wish I had known this before haha. I saw that InventorySetups was using a similar amount of data so I assumed it would be OK if I did too. I mainly just wanted to limit every client uneccessarily fetching the config file from GitHub multiple times a day.
If you can assure me that using RuneLiteAPI's OkHttp Client will achieve this, then I'll swap out my system for that quick-smart.
In terms of Git/GitHub workflow would you like me to close this PR, make these changes in a new commit and then create a new PR from scratch when I'm done?

abextm:
Using the OkHttp client will make it respect the caching headers that github puts on it, which is probably fine. I think github will use Last-Modified-Since or an etag, so it won't actually redownload anything. You can close this or not, I don't really care. If you do close it you can push a new commit to this branch when you have it fixed and reopen it.

rhoiyds:
Beautiful. I'll use this. Sounds like a much better plan, thanks for the careful review. 
I'll make the changes in a few hours and have a crack at pushing a commit to this branch instead of creating a new PR.

rhoiyds:
Changes made as per discussed, Abex. 
Hope this solution is OK with you.
As always, any problems, just let me know 👍

PR #936, opened 2021-01-07
region-locker: Update to RuneLite 1.6.36.1

slaytostay:
Currently the region-locker plugin is not appearing in the plugin hub, hoping updating will fix that.

Update: Added the LICENSE file


loldudester:
You're missing a license file. A new requirement since this plugin was last updated.

slaytostay:
> 
> 
> You're missing a license file. A new requirement since this plugin was last updated.

Ah that makes sense, thanks.

PR #937, opened 2021-01-07
Create gecalc

cman8396:
A Runelite plugin that add the ability to use math/s to set price and quantity in the Grans Exchange window, and more...


abextm:
This won't work on some clients since the javascript interpreter isn't always available. Also I don't really like this from a security standpoint anyway.

PR #939, opened 2021-01-10
Create nexus-map

Antipixel:
Enables you to navigate the Nexus portal menu in the player owned house as an interactive map


abextm:
You could probably halve your loc with some `@Getter`/`@Data`

PR #946, opened 2021-01-12
Update OSRS Streamers to v1.4.3

rhoiyds:
When logging in, prompt users with a misconfigured token to properly configure their token with a chat command ::gettoken

### DIFF
--- a/plugins/osrs-streamers
+++ b/plugins/osrs-streamers
@@ -1,2 +1,2 @@
 repository=https://github.com/rhoiyds/osrs-streamers.git
-commit=b74be12c2d8b28861afe8b85d4250de7b69d7c1a
+commit=5cbab5e9083a4413c6bee77285e5efc5fccd92b2


Nightfirecat:
What would the user flow look like if this opened the authorization link directly from the client--would that work?

rhoiyds:
With the OAuth process, upon authorizing through twitch, Twitch will send a request containing the token in it's headers to the URL specified in the Twitch Application (atm it's set to the URL of the Github page hosting the token generator)
AFAIK if I was to set this to localhost, potentially allowing the auth request to start from the client, there would need to be some kind of webserver on the client machine listening for web traffic to receive the aforementioned request from Twitch - which is the limiting factor in this case. I don't think RuneLite can currently support this with it's dependencies,and  I wouldn't want to cause a mess trying to implement this with my limited knowledge on the topic.
Take everything I say with a grain of salt, I'm really not the best at auth flow.
Althought if someone here truly thinks it's possible, I'm willing to give it a crack.

Nightfirecat (commented):
Code change looks fine. Only concern would be that this directs to a website we have no control of its content, but I don't think there's really a way to phish users for anything besides the information provided via Twitch's oauth flow?

rhoiyds:
The website is simply a GH Page hosted on the [OSRS Streamers plugin repository](https://github.com/rhoiyds/osrs-streamers/blob/master/_layouts/default.html), all code for the site is visible if you're worried about phishing. Nothing too malicious can happen here I don't think.

PR #951, opened 2021-01-13
Add gear-inspector plugin 

rivalosrs:
I created a plugin for inspecting a player's gear as my first plugin. It wasn't till after that I saw the Equipment Inspector plugin already exists. [Adam said it was okay though.](https://discord.com/channels/301497432909414422/419891709883973642/798951653696339968)


abextm:
I would really like to avoid having 2 plugins that do the exact same thing unless there is a reason the existing one is unsuitable for your purpose.

Nightfirecat:
Closing due to inactivity.

PR #964, opened 2021-01-16
Makes code easier to read and changes name

Karambtwo:
Changed big if statements to a loop to check against a list of items as well as changing CoX Special Loot Hider to CoX Censor


Nightfirecat:
This PR needs to be rebased

PR #981, opened 2021-01-20
Add Fight Analysis/Merge, Fight Log Details, Discord Msg Copy

Matsyir:
Fairly big update, but not much for average users. Tested in LMS as always. Details/images below

**Fight Analysis/Merge:** This was discussed a bit in discord#development, in a nutshell, it's a way to combine Fight data from two different clients to get more detailed stats, since some data is only available client-side. Manual copy/paste merge, no networking stuff. *Big thanks to dekvall for leading me in the right direction for "matching" logs from two different clients using the client tick.

The code/"algorithm" to do this is probably a bit more complicated than it needs to be, but I was worried about reliability when doing it in a more direct way. This feature is mostly for tournament players, for extra verification/authenticity. Since you need your opponent's data, it also requires a certain level of trust between participants, although it's not straightforward to edit the data without fully understanding the code, especially when merging as you'd have to edit attack logs 1-by-1. Plus, you would be able to see where the fight was edited using the logs/detailed logs, if you were able to notice the difference at least.
Fight analysis panel part 1
===>
Fight analysis panel part 2

**Fight Log Details:** Displays in-depth details about every single attack log shown in the Fight Log panel. Accessed by clicking a row of the Fight Log panel/table. (The image shown is the results of a merged fight, so regular fights contain a little bit less data)
image

**Discord Msg Copy:** simply copy the fight data as a discord-formatted message, because why send screenshots for text data?
image
(hmm.. ironic)

Reformatted the way the "Magic Hit Luck" stats is displayed, now shows `successfulHits / totalMageAtkCount (luck%)`, where luck% is: 0% = expected luck, >0% = lucky, <0% = unlucky. Previously was a less intuitive `successfulHits / deservedHits `

Also moved my classes to organize them a bit more, split into "MVC" although not fully conforming to that design/framework.

Likely a few unrelated minor improvements/fixes.

The fight analysis/merge was suggested to me by the youtuber Pan1c, who said he'll use this for tournaments, and apparently many others in the 'nh bridding' community would as well. He's given me many design suggestions, as well as asking others for more feedback, so I think the feature is in a good spot right now.

### DIFF
--- a/plugins/pvp-performance-tracker
+++ b/plugins/pvp-performance-tracker
@@ -1,2 +1,2 @@
 repository=https://github.com/Matsyir/pvp-performance-tracker.git
-commit=0affe718324dcd5f1636143227ff2e9a46083cbc
+commit=ac3c3b91ff46659cb7ead395185f7cab64d544d9


Nightfirecat:
Does this run without tripping assertions with `-ea`? It looks like there's a couple places this is calling client methods off client thread

Matsyir:
> Does this run without tripping assertions with `-ea`?

No. Fails on many spriteManager.getSprite calls when I had dev tools/-ea on. Tried to fix but was unable. Never had issues without -ea though. How do I fix this? I looked around a few plugins that display item sprites on JLabels and didn't notice what they do differently that was specific to retrieving the sprite.

Matsyir:
Fixed the `-ea` assertions, did a few refactors, and re-tested in LMS with `-ea` & devtools on the whole time this time, making sure to use every UI element. Working well now. Added sprite cache, since there's a small number of those that are re-used, but not item cache, since there could potentially be a large amount of items saved which aren't often re-used, so I don't think it would benefit as much (especially a potential problem for people who leave the client open for many days without ever restarting). Thanks again for helping fix the assertions!

PR #984, opened 2021-01-20
Added Set Timer Plugin

InfernoStats:
This is a TzKal-Zuk set timer plugin that is completely manual and does not use any game events. It is a replica of the commonly used timer here: https://bistools.github.io/timers.html


Nightfirecat:
This needs to remove the nav button on shutdown. LGTM otherwise

nbulischeck:
> This needs to remove the nav button on shutdown. LGTM otherwise

Thanks for catching that. I'll push a new commit

PR #1000, opened 2021-01-21
add wom flags

dekvall:
image


Nightfirecat:
What's the source of these flag icons? Are they licensed in a way that allows them to be used like this--without attribution or redistribution of their license?

Nightfirecat:
[Per their license](https://github.com/twitter/twemoji/blob/master/LICENSE), that license must be explicitly copied to the repo in some way such that it would be distributed with copies of the plugin, so somewhere in `src/main/resources/`

PR #1008, opened 2021-01-23
Add fashionscape plugin

equirs:
edit: working on some bug fixes and changing the entry points so that I don't have to mess with the right click menu options on the equipment UI

### DIFF
--- a/plugins/fashionscape
+++ b/plugins/fashionscape
@@ -1,2 +1,2 @@
 repository=https://github.com/equirs/fashionscape-plugin.git
-commit=a4f48f43f93adca229356034927ec3b8634d40ef
+commit=8a67b84ba6450a996fe39e1123aeb4f7f6159db2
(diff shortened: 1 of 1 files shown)


equirs:
Well, since https://github.com/runelite/plugin-hub/pull/1009 was merged and it has significant overlap with this, I think it's probably better to just close this, yeah?

Nightfirecat:
As was mentioned in Discord, [Jagex's statement specifically lists equipped item menu modifications as a prohibited behavior](https://secure.runescape.com/m=news/another-message-about-unofficial-clients?oldschool=1):

> Modifies the menu options on worn/equipped items

I don't see any reason this could not coexist with the transmog plugin, but if you feel you'd rather close it, that is also fine.

equirs:
Alright, I changed a bunch of stuff. No more modifying menu entries. I think the UI is better suited for a side panel anyway; there's a lot more room there than the game's native widget stuff.

Nightfirecat:
Does this plugin run without tripping assertions with `-ea`? It seems like it is calling some ItemManager methods off the client thread, which could cause errors.

equirs:
Oh lol, I wish I had realized that sooner. Went through all the plugin's functions and I don't think it's calling any of these methods off the client thread now.

Nightfirecat:
FYI ConfigChanged often runs on EDT, not on client thread: https://github.com/equirs/fashionscape-plugin/blob/45f8181ae3e3fdf141738d1489561020e6ece569/src/main/java/eq/uirs/fashionscape/FashionscapePlugin.java#L140

PR #1016, opened 2021-01-24
New - Chat Translator plugin

simbleau:
This plugin translates chat box input for the user with their own Google Cloud API account.
More information: https://github.com/simbleau/chat-translator#chat-translator


simbleau:
It looks like it's working finally, but Travis is denying it because the JAR is 22MB (max is 10). Is this something we can overlook? The size is due by the Google Cloud Translate library. I would understand why a chat translation library might be large.

simbleau:
I am going to re-do this PR with an okhttp REST API wrapper instead of using the Google client library, because it's huge.

PR #1017, opened 2021-01-25
Add new osrs wiki crowdsourcing plugin

pwatts6060:
A plugin to trial new features to the osrs wiki crowdsourcing e.g. mining crowdsourcing.


abextm:
This should be PRed to the existing plugin

PR #1019, opened 2021-01-25
Add Tears of Guthix Indicator

LlemonDuck:
This plugin adds a small quest icon to the skill that will receive your Tears of Guthix exp reward. Useful to see if you need to train some skills to ensure that ToG goes into the skill you want.


Nightfirecat:
Your runelite-plugin.properties file should have a support URL, it can be a link to your repo or your issue tracker.

PR #1024, opened 2021-01-27
Add "Hotkeyable Menu Swaps" plugin to plugin hub

geheur:
This plugin adds some menu entry swaps that can be activated by holding customizable hotkeys instead of just shift (e.g. hold "T" to swap "withdraw-10" but hold "F" to swap "withdraw-5" in the bank). Currently affects bank withdraw/deposit, spirit tree/fairy ring, and occult altar.


dekvall:
There are no changes in this PR

geheur:
My mistake, it should be correct now.

PR #1032, opened 2021-01-28
Updated plugin to display timer in an infobox

InfernoStats:
It was requested that the plugin have an optional infobox be displayed for users who didn't want to look at the panel as frequently.

Old commits were squashed due to some history issues w/ using both Intellij and WSL. You can probably find the old version in the reflog to diff if need be.


abextm:
Your reflog is only kept locally. Destroying your history like this means this has to be reviewed as a new plugin, rather than a update, which is quite annoying for us.

PR #1040, opened 2021-01-30
Add Ground Object Hider plugin

Leo288:
This plugin intends to improve performance on low-end devices, as well as removing visual distractions from the game.

See: https://github.com/Leo288/object-hider/blob/master/README.md


Leo288:
This won't pass CI until 1.6.39 is released, as it uses a newer API for object hiding.

PR #1059, opened 2021-02-03
Add TaskJingle plugin

maximonster:
Created a plugin that plays a short jingle when completing a slayer task

### DIFF
--- a/plugins/TaskJingle
+++ b/plugins/TaskJingle
@@ -1,0 +1,2 @@
+repository=https://github.com/maximonster/TaskJingle.git
+commit=a997c12aef577e7853f1590e58e4a6691cd9a2f4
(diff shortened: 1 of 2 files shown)


abextm:
This is just the example plugin

maximonster:
> This is just the example plugin

Sorry I'm new to IntelliJ and didn't realise it wasn't pushed immediatly.

Nightfirecat:
This will not work for other users as your `/libs` directory is not included in the build. You should [follow the directions in our README on adding and pinning third-party dependencies](https://github.com/runelite/plugin-hub#third-party-dependencies).

abextm:
Its on the build classpath so it will be included. You also don't need to pin it by hash since its checked into the repo.

abextm:
However, you aren't actually loading the mp3 that comes with your plugin. Also any files you have in .runelite should be named similar to your plugin, so `task-jingle.mp3` or `task-jingle/TaskDone.mp3`

abextm:
You still aren't reading the mp3 that comes with your plugin. You are looking for a file in .runelite instead, so this won't actually work for users unless they put a file there their self. Also you don't need to add a `verification-metadata.xml` since your deps are pinned by committing them to your repo.

PR #1067, opened 2021-02-07
Add multicolor-highlights

sgfost:
NPC highlighting plugin with (currently) very simple functionality, but allows for multiple (up to 5) colors for different NPCs.

### DIFF
--- a/plugins/multicolor-highlights
+++ b/plugins/multicolor-highlights
@@ -1,0 +1,2 @@
+repository=https://github.com/sgfost/multicolor-highlights.git
+commit=0ad4513c87b28e755ed1717a660bfab93232c439


oeg91:
Does this plugin works for tob nylocas room when the spiders/crabs does switch styles? i mean, when the Nylocas Ischyros switch to Nylocas Toxobolos, the color change aswell?

Hydrox6:
Considering that would be against the rules, no. That would also make it ineligible to be in the client.

oeg91:
I see, had no idea, thanks for let me know.

Nightfirecat:
@sgfost Apparently you've deleted or privated your repo, is that intentional?

sgfost:
Intentional, yes. Thoughtful, no. Forgot how this was set up. Unsure if it even worked and no one seemed interested in forking it. What do you typically do with abandoned plugins?

Nightfirecat:
We can disable it in the hub so it isn't included in builds

stormvansoldt:
@sgfost I was actually looking into forking this plugin to add some more functionality to it, would it be possible to make the source available again? I'd really appreciate it :)

stormvansoldt:
Just wanted to give a shout to @sgfost for making the source available. Already forked it and have a few updates in the works.

PR #1070, opened 2021-02-07
Update fashionscape to v1.1.1

equirs:
Fixes a rather annoying bug where the player's body parts would disappear when turning the plugin off. Turns out I was always debugging with items equipped in-game and never noticed, heh.


equirs:
Think this fixes it now. It was a threading issue turns out

PR #1072, opened 2021-02-07
Adding plug-in OSRSKillboard

Osrskillboard:
Plugin to log PvP kills to an online killboard.

### DIFF
--- a/plugins/osrskillboard
+++ b/plugins/osrskillboard
@@ -1,3 +1,3 @@
 repository=https://github.com/Osrskillboard/OSRSKillboard-plugin.git
-commit=13d8103dd69fdc70052dd5a5637ef402e7db7aa1
+commit=b2af3851e1725709f4bfc3b3986e653506f89afa
 warning=This plugin submits your username, Gear, PvP Kills, PvP Loot and IP address to a 3rd party website not controlled or verified by the RuneLite Developers.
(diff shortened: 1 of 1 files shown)


melkypie:
This needs a warning added as it's sending data to `https://api.osrskillboard.com/`. You can use [this](https://github.com/runelite/plugin-hub/blob/master/plugins/rune-cafe-cashflow#L3) as an example

Osrskillboard:
Updating license to BSD 2-Clause "Simplified" Licens.

abextm:
This won't work for people not logged into a runelite account, which is probably not what you want.

Osrskillboard:
Hey @abextm, - could you expand on this? 

The only thing that doesn't seem to work in test is getting the prices of the loot. However, the same happens with the LootTracker add-in and that works in the live client when I'm not logged in.

abextm:
[This code](https://github.com/Osrskillboard/OSRSKillboard-plugin/blob/b2af3851e1725709f4bfc3b3986e653506f89afa/src/main/java/com/osrskillboard/OsrsKillboardPlugin.java#L101-L109) runs when you login/logout of your RuneLite account, so your client will be nulled out if you aren't logged into RuneLite. Since you client has no dependence on the rl account logged into, this seems pointless.

Osrskillboard:
@abextm thanks for the info! Cleared that up. I still and getting to know the codebases and... Java.

abextm:
You still also have [this](https://github.com/Osrskillboard/OSRSKillboard-plugin/commit/578b0590eebd93a9e54f8eb21f4fd282ea15b524#diff-8919b879a0fa56bcaed9bd474f70c05b5be02c15b86279a298000b3fa69dfb3fR118-R131) and a no-op invokeLater callback

Osrskillboard:
Thanks for the reviews and the pointers. I appreciate your patience.

Also cleared up the unused ClientThread and SessionManager.

PR #1089, opened 2021-02-12
Create AfterDark-PvM-Loot-Tracker

NovaGTX:
This plugin stores specific PvM drops in the AfterDark drop database. This plugin requires a User ID and Auth Token and is only designed to be used for AfterDark PvM competitions.


dekvall:
For future reference there is [LootReceived](https://static.runelite.net/api/runelite-client/net/runelite/client/plugins/loottracker/LootReceived.html) which might be of interest to you, and you also need to [pin your dependencies](https://github.com/runelite/plugin-hub/#third-party-dependencies).

NovaGTX:
I've pinned the dependencies, should be good to go?

Nightfirecat (approved):
Deps and code look fine to me. sha256 sums match those of maven central

PR #1127, opened 2021-02-20
Create drop-simulator

mxp190009:
This drop simulator plugin allows for the simulation of any number of trials of most NPCs with a drop table. Uses the [osrsbox-api](https://api.osrsbox.com/index.html) to gather an NPC's drop data. Drops are then simulated and displayed in the plugin panel. Hovering over a drop will display the name, quantity, and value of the item stack.


Nightfirecat:
FYI you can push updates to your branch instead of closing and opening new PRs

PR #1136, opened 2021-02-23
Camera Points Plugin

NullNameException:
This plugin allows you to save and load camera viewpoints (pitch/yaw/angle).
I took some inspiration form the Screen Markers and other plugins UI to make it more intuitive to use.

### DIFF
--- a/plugins/camera-points
+++ b/plugins/camera-points
@@ -1,2 +1,2 @@
-repository=https://github.com/NullNameException/camera-points
+repository=https://github.com/NullNameException/camera-points.git
 commit=f321622c90f72bd51ce317761905bb7b710b79a7
--- a/plugins/camera-points
+++ b/plugins/camera-points
@@ -1,2 +1,2 @@
 repository=https://github.com/NullNameException/camera-points.git
-commit=f321622c90f72bd51ce317761905bb7b710b79a7
+commit=0fa0a58d5253b8d9fc5b415d55cd4dffbc825d4d
--- a/plugins/camera-points
+++ b/plugins/camera-points
@@ -1,0 +1,2 @@
+repository=https://github.com/NullNameException/camera-points
+commit=f321622c90f72bd51ce317761905bb7b710b79a7


abextm:
This is a rejected feature. Also you don't need to create a new pr every time your build fails.

NullNameException:
Does that mean setting a fixed camera angle is not allowed at all?

Also I'm sorry, I didn't know there was a way to know the build is going to fail before creating the pr.

Hydrox6:
> I didn't know there was a way to know the build is going to fail before creating the pr.

You can just push more commits to the branch and the PR will update, btw

abextm:
> Does that mean setting a fixed camera angle is not allowed at all?

Correct.

NullNameException:
> > I didn't know there was a way to know the build is going to fail before creating the pr.
> 
> You can just push more commits to the branch and the PR will update, btw

Thank you, I will do that in the future


> > Does that mean setting a fixed camera angle is not allowed at all?
> 
> Correct.

Would it be okay to provide a fixed set of some camera angles to load instead of being able to save and load custom ones?
Since both of these features ("North/South/East/West on compass" and "Ctrl to reset zoom") already exist as plugins.

abextm:
Just adjusting the zoom is fine, I don't want pitch/yaw adjustment more granular than just cardinal directions.

NullNameException:
I understand. So if I changed the plugin to only save/load the zoom value it would be fine?

abextm:
Yes. You can push the updated commit to this branch and re-open the pr if you do so.

NullNameException:
Would it also be okay to set one of the cardinal directions (calling the same script as the compass) along with the zoom?
And if yes would I need to modify the script myself if the user does not use the compass plugin or will the modified version still be loaded?

NullNameException:
The plugin now only allows to load cardinal directions (using the compass client script) along with the zoom. If that's still too much I can change it to only the zoom being adjustable.
I hope this works now.

PR #1148, opened 2021-02-25
Update NMZ Optional Points Plugin

bloogeyz:


Hydrox6:
You don't have to open a new PR if you want to add even more while the existing one is open; just push another commit to the PR branch and it'll all update

bloogeyz:
> 
> 
> You don't have to open a new PR if you want to add even more while the existing one is open; just push another commit to the PR branch and it'll all update

MB I was actually worried that I'd need to spend more time fixing my plugin so I closed the branch pre-empitvely, should've just moved it back to a draft.

Hydrox6:
You could always close the PR, then reopen it and push the changes to it once you're done

PR #1155, opened 2021-02-27
Create bot-detector

Ferrariic:
Client side (This Plugin): 

The Bot Detector Plugin gathers player names from within OSRS in order to aid in the training and evaluation of a bot-detecting neural network. This plugin pulls names from in-game upon a PlayerSpawned event, adds those player names to a temporary text file, removes duplicates from the text file by parsing into another file, and uploads that final text file to a one-way website which stores those names for use in the neural network. The upload of player names is controlled by the account user - by turning off the plugin. Currently - after sending a name batch to the listening server it is necessary to restart Runelite if another batch of collected names would like to be sent. This will be changed in future updates, along with making the plugin more user friendly and interactive for the player. 

----

Server side: Once names are sent by the user, these player names are sent through the OSRS Hiscores by a python script, and stats are pulled. These stats are then evaluated by a neural network and grouping algorithm which organizes player data from the Hiscores into distinct groups - and evaluates them against current and known bots in order to detect botting or malicious behaviors. Suspicious accounts are then forwarded to tipoff@jagex for final evaluation and consideration.

### DIFF
--- a/plugins/bot-detector
+++ b/plugins/bot-detector
@@ -1,2 +1,3 @@
 repository=https://github.com/Ferrariic/bot-detector.git
 commit=2f7a166171e1e92bcc82b73663f544cb779698ae
+warning=This plugin submits the names of encountered Players, and your IP address to a 3rd party website not controlled or verified by the RuneLite Developers.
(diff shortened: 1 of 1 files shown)


raiyni:
This needs a warning similar to other plugins that post out to a 3rd party server.

ex

https://github.com/runelite/plugin-hub/blob/master/plugins/discord-rare-drop-notificater

Ferrariic:
> This needs a warning similar to other plugins that post out to a 3rd party server.
> 
> ex
> 
> https://github.com/runelite/plugin-hub/blob/master/plugins/discord-rare-drop-notificater

Added a warning, thank you for notifying me!

PR #1156, opened 2021-02-27
Update Drop Simulator to v1.1

mxp190009:
- Fixed manually changing the number of trials to update properly
- Fixed instances of individual item popup menu info lingering upon right click simulating more drops


Hydrox6:
Just want to mention: I think you wanted to updated the version in `build.gradle` to 1.1 but forgot

PR #1160, opened 2021-03-01
ProximityScape - Discord

warnerblue:
Proximity chat for OSRS using Discord.


Nightfirecat:
You should remove `enabledByDefault = false` from your plugin--it is a common source of confusion for users.

warnerblue:
Removed

> You should remove `enabledByDefault = false` from your plugin--it is a common source of confusion for users.

Removed

PR #1172, opened 2021-03-02
Update fashionscape to 1.1.3

equirs:
1.1.3 expanded the swap capabilities to kits, which required reworking the entire swap logic (sorry for big diff)
also added a menu option to "copy-outfit" from other players, so now you can instantly become a clone of someone else

### DIFF
--- a/plugins/fashionscape
+++ b/plugins/fashionscape
@@ -1,2 +1,2 @@
 repository=https://github.com/equirs/fashionscape-plugin.git
-commit=9a074c1a7e75e8abf9291940f42bd6789a9c0045
+commit=93bfe021fccbb07c0b93a6f5a8e94e89a36adad1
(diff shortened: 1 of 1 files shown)


abextm:
We don't allow reflection into client fields. We can add a getter for this if you want.

equirs:
> We don't allow reflection into client fields. We can add a getter for this if you want.

That is unfortunate. I would appreciate if a getter were added, if it's not too much trouble

equirs:
I've removed the egregious reflection-reliant code and the features that depend on it (i.e., changing colors) and have changed versions back to 1.1.3 since both 3 and 4 were never released. There is some dead code as a result of this, but I hope to bring it back at some point if the RuneLite api will allow it.

Nightfirecat:
Which of this PR and #1347 should we be reviewing for merging?

equirs:
> Which of this PR and #1347 should we be reviewing for merging?

Sorry if that's confusing, I would like this reviewed and merged first. The other PR was put up to prevent this one from growing any larger, and that one won't build until the next release anyway.

PR #1178, opened 2021-03-03
Update optimal-quest-guide

cesoun:
Bump to resolve fps pinning.


abextm:
this commit does nothing

PR #1186, opened 2021-03-07
Update raid data tracker to v1.4.2

CasvM:
### v1.4.2
- Added dey0 time tracking to the raid_tracker_data.log file, toggleable in the settings. Adding these times to the ui are planned for the next version.
- Increased the amount of date filters in the filter panel: added a filter for 12 hours, 3 days and 3 months.
- Deleted verification-metadata as it was not necessary and caused the plugin not to build as of v1.7.0 runelite.
- Fixed raid times > 1 hour not counting the extra hour.
- Fixed a bug where the filters > 7 people wouldn't work; all the kills were shown instead.
- Fixed a bug where times wouldn't be tracked if precise timing was enabled. The times are now rounded in the tracker instead.


Nightfirecat:
@Cyborger1 Do you want this plugin to be transferred?

PR #1196, opened 2021-03-09
Add clan-event-attendance

JoRouss:
A plugin used to generate a list of clan members attending an event.

It is designed so any clan member can take attendance at an event and post attendance to their clan's social media.

It tracks the event duration and for every clan member, the time spent at the event and how late they joined the event.

At the end of the event, you can copy/paste the attendance list to your clan's social media, like Discord.


JoRouss:
Battle tested it tonight during two huge clan events. Found an issue when ppl hop into the event or log in directly on the spot.
I will fix this and un-draft the pr.

Nightfirecat (changes requested):
[Do not use `ImageIO.read()`](https://github.com/JoRouss/runelite-ClanEventAttendance/blob/5087b34fa1acac2571a2df8206c9f1e4a592f20e/src/main/java/com/ClanEventAttendance/ClanEventAttendancePlugin.java#L98). Use [`ImageUtil.loadImageResource()](http://static.runelite.net/api/runelite-client/net/runelite/client/util/ImageUtil.html#loadImageResource(java.lang.Class,java.lang.String)) instead.

PR #1211, opened 2021-03-13
Update Elysium Events plugin to use Google Sheets API

cmsu224:
I've added a new plugin panel that contains clan-related information such as boss of the week, skill of the week, clan events, hall of hame, and general announcements.  The data is pulled from a public spreadsheet (which is what most clans use to keep track of their players).  I am only using the API to read data from spreadsheets, it is not writing anything else out.


Broooklyn:
This contains extraneous changes. The only file that should be changed in this PR is elysiumevents-plugin

cmsu224:
Removed all extraneous changes from the PR

cmsu224:
Fixed build errors

package/verification-template/build.gradle
```diff
@@ -55,6 +55,12 @@ dependencies {
 	thirdParty("com.diffplug.spotless:spotless-plugin-gradle:5.2.0") {
 		because "ring-of-recoil-notifier"
 	}
+	thirdParty("com.google.apis:google-api-services-sheets:v4-rev581-1.25.0") {
+		because "elysiumevents-plugin"
+	}
+	thirdParty("com.google.guava:guava:20.0") {
```
Adam-, line 61: I don't think it will be possible to override the guava version at runtime, since we control it as it is a direct dep of runelite-client. You should probably just exclude the dep from whatever is pulling it in.

cmsu224:
Excluded guava 20.0 and other unused external libraries from dep
image

Adam-:
You can add the same exclusions to the verification metadata dep, which will prevent it from pinning those deps which you are no longer including.

cmsu224:
Updated the metadata dep with the same exclusions.

PR #1217, opened 2021-03-15
create Force Recolor plugin

Alexsuperfly:
Force recolors all game messages that contain specified text.

For example writing `divine` in the config will strip and replace the red color from the messages in chat telling you they are about to wear off.  
Or `exchange` to change that awful green color that can barely be read with the transparent chatbox.


Alexsuperfly:
closing for the moment cuz im thinking of expanding the features of the plugin

PR #1224, opened 2021-03-17
Update pvp-performance-tracker to v1.4.3

Matsyir:
Fixed a layout bug that occurred when changing the state of the "config warning", which aims to help users acknowledge the existence of the config. Small bugfix, but it's a confusing bug that likely happens to most new users, and is only fixed by restarting the plugin. Used GridLayout in a dumb way + wasn't using constants, and the row count got messed up. Now using constants so this won't happen again. I probably shouldn't be using GridLayout in this case, but at this point it works and I'm afraid to change it. May do so in the next update where I plan to do a lot of UI changes.
image


Matsyir:
Added another minor change to clarify that warning label & its related tooltips.

PR #1225, opened 2021-03-17
Update Clan event attendance to V1.3

JoRouss:
Display real-time attendance

I changed the JTextArea for a JLabel and formatted everything in HTML to add a color code for active/inactive members.

Added a couple of settings.

You now need to press the "Copy to clipboard" button to export the generated text.


JoRouss:
Hmmm something is wrong with my branch, will close and redo...

PR #1228, opened 2021-03-17
Update Drop Simulator to v1.3

mxp190009:
Clue scroll trials can now be simulated


mxp190009:
switching from sqlite to json

PR #1230, opened 2021-03-17
Update Bank Tag Layouts to v1.1

geheur:
* Adds importing/exporting of layout-ed tag tabs.
* Adds layout placeholders to indicate when an item is in the layout but not in your bank. You can right click these to remove them from the layout.
* Adds "Enable layout by default" config option. You can still disable the layout for individual tabs even when this option is enabled.
* Prints tutorial message when you drag an item on a tab and you don't have any layout-ed tag tabs yet.
* Prints warning message when you might have unintentionally reordered your actual bank instead of a layout.
* Bug fixes.

### DIFF
--- a/plugins/bank-tag-layouts
+++ b/plugins/bank-tag-layouts
@@ -1,2 +1,2 @@
 repository=https://github.com/geheur/bank-tag-custom-layouts.git
-commit=fa2ad3ccf6c38c6596575725efb7fad862a955dc
+commit=907274e15ba27b30e50349ed5626da3451790e86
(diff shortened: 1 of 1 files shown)


abextm:
we disallow reflection in most cases for hub plugins

geheur:
Does this mean it was a mistake that 1.0 was accepted? It also had some reflection in it.

abextm:
Yes, I wouldn't have accepted that

geheur:
Updated this to remove the reflective stuff. Hope it's ok now.

PR #1239, opened 2021-03-19
Update inferno-splits-logger

Magnusrn:
Added logging for splits on death. Fixed a problem with textfiles failing to write within the dir


Hydrox6:
You need to rebase this branch onto `upstream/master` for it to work

PR #1253, opened 2021-03-21
Add blue-prox-chat

ThatGamerBlue:


abextm:
We don't allow non maven central repositories in plugin hub plugins. Also no native code unless you are using official opus binaries. Also you are pulling in a lot of reflection plugins which I generally also disallow.

ThatGamerBlue:
> We don't allow non maven central repositories in plugin hub plugins. Also no native code unless you are using official opus binaries. Also you are pulling in a lot of reflection plugins which I generally also disallow.

Any easy way to get the versions of kryo and reflectasm into an allowed repo? Src is at https://github.com/thatgamerblue/kryo and https://github.com/thatgamerblue/reflectasm on the branches named as the versions, modifications are to remove most of the reflection and all the sun.misc.Unsafe usages

abextm:
you can put it in central or you can just check in a binary jar to your repo, but I won't allow reflectasm since its too hard to verify runtime code gen.

ThatGamerBlue:
> you can put it in central or you can just check in a binary jar to your repo, but I won't allow reflectasm since its too hard to verify runtime code gen.

I'll see what I can do about removing reflectasm in that case, mind if I DM you on discord any other questions to not flood here and the dev channel in rl disc?

abextm:
sure.

ThatGamerBlue:
Should be good now I hope :)

abextm:
kryo still has too much reflection in it for me. You are just using it as DataInputStream basically, so it should be pretty easily to just use that.

ThatGamerBlue:
> kryo still has too much reflection in it for me. You are just using it as DataInputStream basically, so it should be pretty easily to just use that.

moved kryo/kryonet into the source tree and deleted a whole bunch of it, i believe there's no reflection left outside of the regular rl client stuff, though im not 100% sure

abextm:
Don't do this https://github.com/ThatGamerBlue/osrs-prox-chat/blob/a336be2ae9a10b5e2d1bf2b9ba1dec723e2eb49b/src/main/java/com/thatgamerblue/osrs/proxchat/client/ProxChatClientPlugin.java#L34

Also this comment is wrong, Client tick is fired ever 20ms, which is probably not what you want https://github.com/ThatGamerBlue/osrs-prox-chat/blob/a336be2ae9a10b5e2d1bf2b9ba1dec723e2eb49b/src/main/java/com/thatgamerblue/osrs/proxchat/client/ProxChatClientPlugin.java#L199

otherwise a336be2ae9a10b5e2d1bf2b9ba1dec723e2eb49b lgtm

ThatGamerBlue:
restored the commented plugin name, and updated the comment to say 20ms, was a remnant from a less granular position system

PR #1260, opened 2021-03-23
Flipping Utilities

zumaad:
added wiki integration


zumaad:
rip, need to find a substitute for Pair, as its not available

PR #1265, opened 2021-03-23
InfernalFC

Rickboy26:
Clan plugin working with API data from websites.

Will be extended to support multiple clans (As long as they have an API)

Currently available are :

Member data lookup
->Username
->Rank
->Points
->Alts

CM Meta lookup
-> Raids data

Gear requirements
-> Images of required gear for the clan.


Nightfirecat (changes requested):
You should not be deleting build.gradle, you should be [adding your needed dependencies to it, then updating the verification metadata file by running `../gradlew --write-verification-metadata sha256`](https://github.com/runelite/plugin-hub#third-party-dependencies)

Rickboy26:
Done.

Nightfirecat (changes requested):
[Your runelite-plugin.properties file does not actually end up building the plugin.](https://github.com/Rickboy26/Infernal/blob/ad0f86977c6972e31dcaf43bebc022f56e2ad4cf/runelite-plugin.properties#L6) That line should read `plugins=com.clanevents.ClanEventsPlugin`

This plugin makes requests to external websites in several places ([1](https://github.com/Rickboy26/Infernal/blob/ad0f86977c6972e31dcaf43bebc022f56e2ad4cf/src/main/java/com/clanevents/panels/Lookup.java#L73), [2](https://github.com/Rickboy26/Infernal/blob/ad0f86977c6972e31dcaf43bebc022f56e2ad4cf/src/main/java/com/clanevents/ClanEventsPanel.java#L136)); you must add a `warning` to your plugin-hub entry, such as `warning=This plugin submits your IP address to a 3rd party server not controlled or verified by the RuneLite Developers.` [(example)](https://github.com/runelite/plugin-hub/blob/master/plugins/proximityscape-discord)

Instead of [requiring gson in your build dependencies](https://github.com/Rickboy26/Infernal/blob/ad0f86977c6972e31dcaf43bebc022f56e2ad4cf/build.gradle#L26), you should be able to use RuneLite's gson instance.

Because you're [depending on `google-api-services-sheets`](https://github.com/Rickboy26/Infernal/blob/ad0f86977c6972e31dcaf43bebc022f56e2ad4cf/build.gradle#L16), you should [add your plugin to the `because` field of plugin-hub's build.gradle.

[**Do not fake a user agent.**](https://github.com/Rickboy26/Infernal/blob/ad0f86977c6972e31dcaf43bebc022f56e2ad4cf/src/main/java/com/clanevents/panels/Ranks.java#L73) Use a custom agent, such as `RuneLite/InfernalFC plugin`

Nightfirecat:
Also, you should change the plugin's class package so that it does not conflict with [the existing Clan Events Plugin](https://github.com/cmsu224/clan-events/blob/master/src/main/java/com/clanevents/ClanEventsPlugin.java). Otherwise, this would cause both plugins to be turned on or off when using the plugin enable/disable switches.

Rickboy26:
Changed plugin build file.
Added warning
Changed Agent
Didn't change gson as been told, and used.

Also renamed to InfernalFCPlugin

PR #1293, opened 2021-03-29
Introduce irc-plugin

ryanwohara:
Integrating IRC into the OSRS chat box.

Short clip to demonstrate it: https://streamable.com/iiwtfc

Features:
1 channel to chat with
Notices
Private messages
IRC services: /ns /cs /bs /hs

Known issues:
Only one nick is supported. If that nick is in use your connection will fail.
Incompatible with Twitch usage

### DIFF
--- a/plugins/irc
+++ b/plugins/irc
@@ -1,0 +1,2 @@
+repository=https://github.com/ryanwohara/irc-plugin.git
+commit=b80b9c226917ccdfacc7d38fee11b9a9352f8da6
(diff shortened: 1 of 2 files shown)


abextm:
All of your licenses say Adam, not you

ryanwohara:
@Adam- I fixed the license to have your name as requested and now it's being rejected because of that. Please confirm which license I should proceed with.

PR #1304, opened 2021-04-01
create stack-value

RealSheeva:
Initial commit for submitting Stack Value plugin. This plugin highlights the value of stacks by allowing you to set value breakpoints and their respective colors. This plugin is a refactored, updated version of the ItemRarity plugin by Zamatica, so there is some shared code.

### DIFF
--- a/plugins/stack-value
+++ b/plugins/stack-value
@@ -1,0 +1,2 @@
+repository=https://github.com/RealSheeva/stack-value.git
+commit=2e39a66828a1671cdc17d719b7affc02c4ce658a
--- a/plugins/stack-value
+++ b/plugins/stack-value
@@ -1,2 +1,2 @@
 repository=https://github.com/RealSheeva/stack-value.git
-commit=2e39a66828a1671cdc17d719b7affc02c4ce658a
+commit=94b57c960ecd56c3d88b4589c7b924d9091ad442


RealSheeva:
This plugin shows the value of Item Stacks when enabled. Allows users to set their own breakpoints and colors. Calculates based off GE PRICE ONLY currently. This was due to very abnormal value calculations on certain items. This could be enabled / disabled in future versions, but I would -not- recommend this, as the anomalous behavior is more confusing than helpful with high-alch values enabled for item-stacks.

abextm:
How is this different from the bank heatmap plugin?

RealSheeva:
@abextm This is much different. I made it because the heatmap wasn't at all what I was looking for.

This plugin lets you set colors, set value thresholds, uses a completely different visual overlay, and isn't restricted to the bank.

Also - in my recent testing, the heatmap doesn't even work. It colors everything > 100k blue, and everything else no color.

raiyni:
image

Heatmap 100% works. Maybe you had it on for your entire back tab and you couldn't see your most valuable stack. I would gladly accept contributions to my plugin to have value thresholds or ignored items.

RealSheeva:
Okay, then I have misunderstood the purpose of the heatmap. If it is simply to highlight your *highest* value, than its purpose is completely different than mine altogether.

The purpose of my plugin is to color item stacks based on their value. The heatmap plugin isn't useful for me in this way, because it gives no indication of the actual value of items, simply relative value based on visible items.

If there is additional work to be done to get this plugin on the hub, please let me know.

RealSheeva:
FYI, all I did was change the functionality of https://github.com/Zamatica/ItemRarity and order the menu entries so that they are actually in order. I'm pretty Zamatica just wrote this, then forgot about it.

abextm:
Then why not submit this as a pull to Item Rarity instead of a separate plugin?

RealSheeva:
Because the end-goal of the two plugins are completely different. The ItemRarity plugin is meant to show the rarity of individual items. It has no care about the value of a stack at all. The purpose of this is to, at a glance, see exact price ranges for entire stacks of items.

RealSheeva:
It didn't seem appropriate at all in this situation, since I changed the fundamental purpose of the plugin, only re-using code to make my job easier.

PR #1311, opened 2021-04-04
Create cmAutoSplitter

SkyBouncer:
Plugin for automating splits for the Live Splits program for Cox Challenge Mode.


SkyBouncer:
Changing file name to all lowercaps

PR #1313, opened 2021-04-04
Cm auto splitter

SkyBouncer:
Plugin for automating splits in the LiveSplit program for cox challenge mode.


Nightfirecat:
We don't want plugins to send keypresses with `Robot`.

SkyBouncer:
Do you have a preferred alternative, or is keypresses a no go?

Nightfirecat:
Other plugins which interact with external plugins will either invoke commands ([eg. bitwarden plugin](https://github.com/runelite/plugin-hub/blob/master/plugins/bitwarden-plugin#L3)) or load an SDK to interface with some external software. I'd recommend looking into alternatives such as that. I don't see there being situations where simulating keypresses will be acceptable.

Hannah-GBS:
LiveSplit has a [server component](https://github.com/LiveSplit/LiveSplit.Server) you could use. Not sure about other split programs.

SkyBouncer:
Ok, it's updated.

Nightfirecat:
Your LICENSE must give attribution to De0 as you are using some of his code. This looks good otherwise.

SkyBouncer:
I'm not that familiar with open source licenses, but hopefully it's good now

PR #1320, opened 2021-04-07
Update weapon-charges

ZemyCode:
Added Thresholds, accidentally broke stuff with my commit history, this should fix it.


Adam- (changes requested):
The original commit no longer exists, however when this happens the bot is flagging it as a plugin change, and the compare link the bot generates causes github to show a made up diff. So this is effectively a new plugin. Is there something we can do about this? @abextm

PR #1330, opened 2021-04-10
Add interactive chat plugin

richardcane:
v1.0 adds the option for users to include clickable wiki links to their chat messages.

### DIFF
--- a/plugins/interactive-chat
+++ b/plugins/interactive-chat
@@ -1,0 +1,2 @@
+repository=https://github.com/richardcane/interactive-chat.git
+commit=6cd0319d57ef53bce90c911ce55290dbcb256e14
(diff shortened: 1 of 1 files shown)


Nightfirecat:
The created `hitboxWidget` should be explicitly deleted on plugin shutdown. (see [`Widget#deleteAllChildren()`](http://static.runelite.net/api/runelite-api/net/runelite/api/widgets/Widget.html#deleteAllChildren()))

richardcane:
The hitbox parent widget is unfortunately `WidgetInfo.CHATBOX_MESSAGE_LINES`, meaning `deleteAllChildren` puts the chatbox in a state where I'd need to reinitialise it. Instead, I borrowed an approach I found in the Wiki plugin. I've tested it and it has the intended effect, please let me know if it doesn't satisfy the requirement.

https://github.com/runelite/runelite/blob/master/runelite-client/src/main/java/net/runelite/client/plugins/wiki/WikiPlugin.java#L127

PR #1338, opened 2021-04-16
Create Spring-Garden

Jonesa5:
This plugin highlights the tiles that are used to walk through the Sorceress's Garden/Spring either green when it is safe to walk to them or red when it is not safe. It uses the Spring Elemental NPCs' positions to determine if it is safe to move to the next tile. The README.md file has more details on what it does.
[README.md](https://github.com/runelite/plugin-hub/files/6322867/README.md)


geheur:
Just in case you didn't know, there is a plugin for summer garden already: https://github.com/LlemonDuck/one-click-summer-garden

PR #1339, opened 2021-04-16
Create spring-garden

Jonesa5:
This plugin marks and colors the tiles that are safe or not safe to walk to while doing the Sorceress's Garden/Spring. The tiles are green when it is safe to move to them, and they are red when it is not safe to move to them. The plugin uses the Spring Elementals' locations' to determine if it is safe to move or not.

### DIFF
--- a/plugins/spring-garden
+++ b/plugins/spring-garden
@@ -1,2 +1,2 @@
 repository=https://github.com/Jonesa5/Spring-Garden-Plugin.git
-commit=f3285139f925d559c7d3419d1bdcebfcf1b10219
+commit=aa96e9e24a069bb5d61ad1bf18edc88b7c278144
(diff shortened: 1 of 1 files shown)


geheur:
Don't create new pull requests when you need to make a change, just update the commit hash, commit, and push. Github will detect the new commit.

Jonesa5:
It says that the build has failed do to missing package "net.runelite.client.plugins.springgarden". I'm not sure why this is coming up. Do you have any suggestions?

geheur:
You imported the wrong thing. I'm guessing you installed a snapshot version of runelite that had your plugin built-in, and then when you moved it to a plugin hub panel and copy-pasted it over you had the old import statements which imported files from your fork of runelite.

I would change the runelite version in build.gradle to 1.7.5 instead of 1.7.5-snapshot.

melkypie:
Your import statements lead to net.runelite.client.plugins.springgarden which is a non-existend plugin. You should instead change them to import your plugin-hub classes under package com.majesty373.springgarden. This works locally due to you having compiled a SNAPSHOT version of the client and using that when building your hub plugin.

Jonesa5:
Thank you for all the information! Im pretty sure it worked this time

PR #1371, opened 2021-04-25
Update quest-helper to v2.1.0

Zoinkwiz:
- Adds Karamja Achievement Diary
- Adds sections to sidebar to better organise helpers
- Adds a generic helper for locating Mage Arena II bosses after that miniquest


abextm:
added as #1381

PR #1376, opened 2021-04-26
Create Player-List-Indicators

Coltonbranker:
Based on Player Indicators. Allows the user to add players names to lists. These lists are then used to highlight the name of those players in different colours over their heads, on the minimap and in right click menu boxes.

### DIFF
--- a/plugins/Player-List-Indicators
+++ b/plugins/Player-List-Indicators
@@ -1,0 +1,2 @@
+repository=https://github.com/Coltonbranker/Player-List-Indicators.git
+commit=44463eed81f48102677ad3562e514eb4a610ba90


Adam-:
I believe the plugin file names must be lowercase, so try renaming it to eg `plugins/player-list-indicators`

Hydrox6:
This is [Rejected](https://github.com/runelite/runelite/wiki/Rejected-or-Rolled-Back-Features):
> Highlighting "Offline" Friends (and generic player highlighting): Not included to respect players with their messaging mode set to Private and to avoid harassing players (see runelite/runelite#951)

Coltonbranker:
@Adam-  sorry I'm kind of new to all this, how would I go about editing it to lowercase without removing this pull request and making another.

@Hydrox6 the plugin that this is based off of, Player Indicators which is a main plugin (not via pluginhub) has an option to highlight all other players, this would just be highlighting specific ones instead of every one, I don't believe thats any different than the original

Hydrox6:
> I don't believe thats any different than the original

It's very different. Highlighting everyone doesn't allow you to easily pick off specific targets for ragging/harassment. It's why the rejected line exists.

Coltonbranker:
alright thank you for the clarification

PR #1389, opened 2021-05-02
Another-bronzeman-mode: Fixed collection log and added settings for moving unlocks and allowing trading.

CodePanter:
- Corrected using a fixed value for the collection log category size.
- Added a setting to either have the bronzeman unlocks on the 'index' page, or moved to the 'Other' tab.
- Added a setting for allowing and disallowing trading with other users.


CodePanter:
I updated this pull request because I realized my code still included a bunch of imports I added while testing. I removed those, as they weren't necessary.

PR #1390, opened 2021-05-02
AntiSpam plugin

CameronWhiteCS:
Update: Didn't know this plugin already existed. Closed.


LlemonDuck:
What are the benefits of using this plugin over the built-in ChatFilter plugin?

PR #1391, opened 2021-05-03
Update corp-ffa07 - Fix chat message bug

Box-Of-Hats:
When checking if a player with 0/1 specs got kill credit, it was not correctly working for players with spaces in their names.

The regex has been adjusted to accommodate for this


Adam-:
FYI most of the time they use non-breaking spaces in player names, so I think this might not work

Box-Of-Hats:
@Adam- Thanks for the heads up. Are you happy to put this through in the meantime for testing purposes or would you rather I figure it out privately first?

Nightfirecat:
For what it's worth, you're also missing `-` in the player name regex.

Box-Of-Hats:
That simple fix turned out to be more fiddly than expected. Thanks both for the help with player names! 

The PR should be good to go now

PR #1397, opened 2021-05-04
Add DPS Calculator

LlemonDuck:
v1, I'm ready for bug reports

### DIFF
--- a/plugins/dps-calculator
+++ b/plugins/dps-calculator
@@ -1,0 +1,2 @@
+repository=https://github.com/LlemonDuck/dps-calculator.git
+commit=60a77623630ea53b7c8d964bd0f6ce5ca44bcdc0


LlemonDuck:
Squashed initial development history

Nightfirecat:
Have you run this plugin with assertions enabled? I see a number of cases where `client` methods are called off-thread which would likely trip assertions, such as: [1](https://github.com/LlemonDuck/dps-calculator/blob/4750e70b378a23b8be0365062d9d448563b331e7/src/main/java/com/duckblade/osrs/dpscalc/ui/equip/EquipmentPanel.java#L191), [2](https://github.com/LlemonDuck/dps-calculator/blob/4750e70b378a23b8be0365062d9d448563b331e7/src/main/java/com/duckblade/osrs/dpscalc/ui/skills/SkillsPanel.java#L71),

LlemonDuck:
Good call, I've updated the plugin with `ClientThread` usage.

PR #1404, opened 2021-05-07
Create roof-toggle

ChrisScott9456:
Allows use of a keybind for toggling roofs on and off without needing to open the in-game settings menu.

It simply has a default keybind "R" (configurable) that runs the setremoveroofs script that toggle roofs on and off.


plugins/roof-toggle
```diff
@@ -0,0 +1,2 @@
+repository=https://github.com/ChrisScott9456/runelite-rooftoggle
```
Nightfirecat, line 1:
```suggestion
repository=https://github.com/ChrisScott9456/runelite-rooftoggle.git
```
LGTM otherwise.

PR #1412, opened 2021-05-08
Update resource-packs to 1.3.2

melkypie:
Update the arceuus spellbook sprites and add some new sprites that jagex has added.


Shadax-OSB:
> Update the arceuus spellbook sprites and add some new sprites that jagex has added.

Just want to say a huge thank you for keeping on updating the pack whenever Jagex make changes to the client! Now I'm off to customise the new additions to keep my pack in line with the previous version.. going to be fun!

PR #1420, opened 2021-05-10
Update interactive chat to v1.1

richardcane:
Adds support for split chat messages and fixed game client layout

Fixes:

https://github.com/richardcane/interactive-chat/issues/2
https://github.com/richardcane/interactive-chat/issues/6


Nightfirecat:
It looks like your mouse listener is never unregistered if the plugin is disabled, please fix that--register and unregister it in your plugin on startup and shutdown instead of your overlay on instantiation, for example.

Does [this line](https://github.com/richardcane/interactive-chat/compare/8c15806c878f3285ddda85c3de6f1cd0e25cc8cf...richardcane:37e3bf9dee2d14c4e965d5afab3295d979554956#diff-303f78893197a591d936b06da1ec968ba20e18588d99c3c31d9816c9839baf43R60) trigger an assertion with `-ea`? I'm not sure what thread `MouseEvent` is fired on.

richardcane:
> It looks like your mouse listener is never unregistered if the plugin is disabled, please fix that--register and unregister it in your plugin on startup and shutdown instead of your overlay on instantiation, for example.
> 
> Does [this line](https://github.com/richardcane/interactive-chat/compare/8c15806c878f3285ddda85c3de6f1cd0e25cc8cf...richardcane:37e3bf9dee2d14c4e965d5afab3295d979554956#diff-303f78893197a591d936b06da1ec968ba20e18588d99c3c31d9816c9839baf43R60) trigger an assertion with `-ea`? I'm not sure what thread `MouseEvent` is fired on.

As mentioned [in discord](https://discord.com/channels/301497432909414422/419891709883973642/842683397251923978), will fix the registration issue.

[This line](https://github.com/richardcane/interactive-chat/compare/8c15806c878f3285ddda85c3de6f1cd0e25cc8cf...richardcane:37e3bf9dee2d14c4e965d5afab3295d979554956#diff-303f78893197a591d936b06da1ec968ba20e18588d99c3c31d9816c9839baf43R60) does not trigger any assertion with `-ea` but I will look into forcing it onto the client thread for clear intent if nothing else.

PR #1423, opened 2021-05-13
Add group bronzeman plugin

NicholasDenaro:
This plugin iterates on the bronzeman gamemode by adding a way to sync items across a group of accounts. It also adds some quality of life features like marking ground items which are valid for the bronzeman to pick up (which includes PVP drops, naturally spawning items, and projectiles). The group sync is done via Google Sheets and requires at least single user in each group to supply client credentials to members in their group (each user can provide their own).

### DIFF
--- a/package/verification-template/build.gradle
+++ b/package/verification-template/build.gradle
@@ -59,9 +59,20 @@ dependencies {
 		because "chat-logger"
 	}
 	thirdParty("com.google.apis:google-api-services-sheets:v4-rev581-1.25.0") {
-		because "elysiumevents-plugin"
+		because "elysiumevents-plugin, group-bronzeman-mode"
 		exclude group: 'com.google.guava'
-		exclude group: 'com.google.oauth-client'
+        	exclude group: 'com.google.ouath-client'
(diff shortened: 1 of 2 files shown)


Nightfirecat:
Does this plugin run without error with `-ea`? I see you call `client.getUsername()` off the client thread here: [`d26568ff4fe14ca3b99442cca2815289daf318ea..735b924e27abaefe846ce53a258f787cb51db429?short_path=7f23a1e`#diff-3ac9c9e7e6](https://github.com/NicholasDenaro/another-bronzeman-mode-group/compare/d26568ff4fe14ca3b99442cca2815289daf318ea..735b924e27abaefe846ce53a258f787cb51db429?short_path=7f23a1e#diff-3ac9c9e7e63f86ba4fc5579b614b9aafc51151d637d3a1e249ff71ecf36495e3R107) and in a couple other places in GoogleSheetSyncer.

NicholasDenaro:
Talking with base plugin owner, possibly going to merge there instead of creating yet another bronzeman plugin.

PR #1431, opened 2021-05-17
Added Inferno Stats Plugin

InfernoStats:
This plugin collects a lot of interesting information around inferno runs. Most useful are the wave-by-wave breakdown in the panel and the special attack counter that would help players know if bringing a SGS/Eldritch staff is worth it or not.

Panel that displays a wave-by-wave breakdown:

unknown

Within the plugin panel are clickable JPanels that take you to a [line of sight tool](https://infernostats.github.io/inferno.html?melee=[[16,17]]&bat=[[22,5],[23,12]]) with the monsters pre-placed on the map which is very helpful for sharing spawns with others when seeking advice if you have a bad wave.

### DIFF
--- a/plugins/inferno-stats
+++ b/plugins/inferno-stats
@@ -1,0 +1,2 @@
+repository=https://github.com/InfernoStats/InfernoStats.git
+commit=8217c1cc63044503b2ae1ae49a997c82e860d64b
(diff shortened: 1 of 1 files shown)


Hydrox6:
Don't close this and make another PR to try and fix the issue, just add a new commit and push it.

InfernoStats:
Sure thing

Nightfirecat:
The nav button must explicitly be removed from the client toolbar on plugin shutdown. It's currently only added depending on your location and config on startup, however if you were to start and stop the plugin while still in the inferno or change your config after starting the plugin, it would not be subsequently removed.

InfernoStats:
@Nightfirecat Thanks for the feedback! I've gone ahead and added `clientToolbar.removeNavigation(navButton);` in the shutdown function.

PR #1444, opened 2021-05-20
create herblorerecipes

skiclimbcode:
This is a small plugin that will show a tooltip overlay on herblore ingredients, showing which potions that item is a primary or secondary ingredient for, with configuration options to toggle 1st/2nd ingredients as well as herblore level requirements.
This version does not support barbarian mixes.
For performance reasons, a tooltip cache gets created as users hover over herblore ingredients and gets cleared with certain configuration changes.

Please let me know if any changes need to be made!


Hydrox6:
Your `runelite-plugin.properties` hasn't got its `plugins` field set right. It needs to point to your Plugin file, and it seems you have the package wrong. Just make the change in the plugin's repo, add a new commit with the update commit hash to this PR, then push it.

PR #1448, opened 2021-05-20
Add DiscordNotifier Plugin

jkufro:
This was created in collaboration with @nodeak

Repo: https://github.com/jkufro/DiscordNotifier/


jkufro:
Updated commit hash is for removing the dependency on Twilio - we were using it previously but no longer need it.

PR #1460, opened 2021-05-25
fake-iron-icon: v1.0.3

ThatGamerBlue:
No idea why that build failed /shrug


Nightfirecat:
Not sure what the first run failure was about... I triggered a re-run, and now it's failing due to missing LICENSE file. (plugins without them are required to add them with any updates)

PR #1462, opened 2021-05-26
Added support for tooltips on seeds used in Herblore

skiclimbcode:
This was a feature request made for the plugin.
This adds tooltips to seeds used in herblore - screenshot added to README along w/ relevant configuration option.


Nightfirecat:
FYI if you change your config `keyName`s, it will reset the config for your users.

skiclimbcode:
> 
> 
> FYI if you change your config `keyName`s, it will reset the config for your users.

Thanks for the heads up! I'll revert the `keyName`s.

PR #1480, opened 2021-05-31
Slash-Swapper Plugin added to the plugin hub

nganzalo:
A Plugin that allows users to send messages in their clan chat with '/' instead of '//' and send messages in their friends chat with '//' instead of '/'.

Closes runelite/runelite#13651.

### DIFF
--- a/plugins/slash-swapper
+++ b/plugins/slash-swapper
@@ -1,0 +1,2 @@
+repository=https://github.com/nganzalo/Slash-Swapper.git
+commit=538185792bb9db694b5389bf4dfaced7f0f4c40d
--- a/plugins/slash-swapper
+++ b/plugins/slash-swapper
@@ -1,2 +1,2 @@
 repository=https://github.com/nganzalo/Slash-Swapper.git
-commit=538185792bb9db694b5389bf4dfaced7f0f4c40d
+commit=9e2a7a7e395e37bcbfe42496dd9d3fb35f22370e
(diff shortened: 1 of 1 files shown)


nganzalo:
There was an error in the runelite.plugins.properties file that caused the build to fail earlier. I fixed that now. Sorry about that!

VinceS22:
Hello, so I was taking a peek at the / issue because it is bothering me and I saw this code block. Might be worth integrating it into chat channels? Still new contributing to this project but it seems that integrating what you have in the default block of this switch statement might be the way to go.

https://github.com/runelite/runelite/blob/657132b39b1374d66f47a9c98f08f43f0be2e61d/runelite-client/src/main/java/net/runelite/client/plugins/chatchannel/ChatChannelPlugin.java#L626

nganzalo:
> Hello, so I was taking a peek at the / issue because it is bothering me and I saw this code block. I would think it would be worth integrating it into the chat channels plugin and making an option. Still new contributing to this project but it seems that integrating what you have in the default block of this switch statement might be the way to go.
> 
> https://github.com/runelite/runelite/blob/657132b39b1374d66f47a9c98f08f43f0be2e61d/runelite-client/src/main/java/net/runelite/client/plugins/chatchannel/ChatChannelPlugin.java#L626

I am familiar with that plugin, When I initially made this pull request a few people in the Runelite discord also recommended that to add it to that plugin. As of now I made this pull request because I felt like the Chat Channels plugin already has a lot going on and thought that it might be better if this were its own simple plugin. However, if the Runelite devs think that this would be better in the chat channel plugin I can work on adding that with a config option.

PR #1483, opened 2021-06-01
Update ItemRarity 1.2

Zamatica:
1. Added version to 1.2

2. Rarities are in their own sections in the config

3. Removed "GE" from the config menu to reduce confusion.

4. Added Stack Counts as an option for value calculations


Nightfirecat:
FYI, if you change your config's `keyName`s, it will reset the saved config for your users. LGTM aside from that

Zamatica:
Ahhh, I'll avoid changing them in the future. At the time I was experimenting with layout order. But thankfully position exist. 

I'm not sure why it's showing that with this request though. Those changes should already have been merged in from a few days ago.

PR #1512, opened 2021-06-09
Added Barbarian Assault pb Tracking

SkylerPIlot:
This plugin serves as a carbon copy of the barbarian assault plugin found ingame along with just adding the extra functionality of saving the best run time to the user's config file, which can then be called up using the !pb commands

### DIFF
--- a/plugins/barbarian-assault-pbs
+++ b/plugins/barbarian-assault-pbs
@@ -1,2 +1,2 @@
 repository=https://github.com/SkylerPIlot/Barbarians-Assault-Personal-Bests.git
-commit=87c2a58cfc7395c651490669b262a71fc80f3b11
+commit=3390567b7be0913ab55ea39b9432d2001ffea52d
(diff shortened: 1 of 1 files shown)


Nightfirecat:
The configurations this plugin sets should be identifiable by the same config keys it uses for user configs, ie. `BaPb`

SkylerPIlot:
The main goal with saving it under "personalbest" was the ability to automatically be integrated into the !pb command so someone could just do !pb without me having to write my own handling for chat commands. I can easily integrate my own chat command handling though and will update the pr when its implemented.

SkylerPIlot:
This newest commits saves the configurations this plugin makes under the BaPB key
This newest update also adds a new chat command to all who download the plugin. 

!ba

which can be used with !ba Barbarian Assault
!ba Ba healer
and so forth for all roles that will allow pbs to be shown to all who install and turn on the plugin.
I plan to update the readme with all changes and how to run the plugin.

LlemonDuck:
Wouldn't this plugin be better as just the chat command and pb detection only? Why do you need to duplicate the existing core plugin?

SkylerPIlot:
I duplicated the plugin functionality so I could just have them activate it and deactivate the main one. Could just as easily be gutted down to just saving the personal best and the chat command.

SkylerPIlot:
There we go, now the plugin only adds the !ba command as well as saving into the user's config file

SkylerPIlot:
That was weird the bot couldn't seem to find my license when doing the workflow even though the commit in question had a license file, for now, I merged the branch into master along with updating my build.gradle so the workflows should now work as expected

PR #1513, opened 2021-06-09
Added clan notifications plugin

itzwolly:
* The plugin utilizes the new clan chat and sends a message to the clan chat channel when a user levels up a skill.
* Future updates include: boss drops and various other relevant notifications.

### DIFF
--- a/plugins/clan-notifications
+++ b/plugins/clan-notifications
@@ -1,0 +1,2 @@
+repository=https://github.com/itzwolly/clannotifications.git
+commit=c57273be0e7a0e97137fca7ac6a4d38a2982c696


LlemonDuck:
This call <https://github.com/itzwolly/clannotifications/blob/c57273be0e7a0e97137fca7ac6a4d38a2982c696/src/main/java/com/clannotifications/ClanNotificationsPlugin.java#L102> will only send to the local player's chatbox, not to the whole clan.

Implementing something that *will* send to all players in the clan would be macroing, and not permitted by Jagex.

itzwolly:
> 
> 
> This call https://github.com/itzwolly/clannotifications/blob/c57273be0e7a0e97137fca7ac6a4d38a2982c696/src/main/java/com/clannotifications/ClanNotificationsPlugin.java#L102 will only send to the local player's chatbox, not to the whole clan.
> 
> Implementing something that _will_ send to all players in the clan would ~be macroing, and not permitted by Jagex~ require a more complex setup, like using the web service to transmit the information between clients.

Ahh alright, thanks for letting me know. I will look into it.

Cyborger1:
Zoinkwiz's Broadcaster (https://github.com/Zoinkwiz/broadcaster) seems to do what you want to do already, but for Friend Chats.

You could probably either modify his plugin for your purpose or contribute to his plugin to add functionality with the new Clan Chats.

itzwolly:
> 
> 
> Zoinkwiz's Broadcaster (https://github.com/Zoinkwiz/broadcaster) seems to do what you want to do already, ~but for Friend Chats~.
> 
> ~You could probably either modify his plugin for your purpose or contribute to his plugin to add functionality with the new Clan Chats.~
> 
> **EDIT:** Actually his plugin doesn't care about the chat type and allows you to configure private broadcast groups, so I think simply using that plugin with your clan members will do what you need, but you could contribute adding different broadcast options.

ohh, I wasn't aware that a plugin like that already existed, I guess I didn't look good enough before starting. Eventhough the Abbly solution does a great job of making it work, I'm not a big fan of the limited amount of traffic for a free user (and using a 3rd party service). 

I will try to look for a better solution, but in the meantime this request could be removed. Once I figure out how I want to do it, I'll create a new request.

PR #1557, opened 2021-06-17
Add tob-hm-timer

Loze-Put:
Creates an overlay to display several times during the Theatre of Blood: Hard Mode. Current features are:

- Time to beat
- Time elapsed
- Time remaining
- Boss splits

An example:
image

### DIFF
--- a/plugins/tob-hm-timer
+++ b/plugins/tob-hm-timer
@@ -1,0 +1,2 @@
+repository=https://github.com/Loze-Put/tob-hm-timer.git
+commit=fb9c6510b3dc3ec7ee1845072a9e5dbdf061b6f7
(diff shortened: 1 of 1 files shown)


raiyni:
What about working with @HSJ-OSRS  to add an overlay mode to theatre-of-blood-stats?

Loze-Put:
> What about working with @HSJ-OSRS to add an overlay mode to theatre-of-blood-stats?

I have taken a look at different plugins such as theatre-of-blood-stats and tob-infoboxes. While I wouldn't be against merging with one of the existing plugins, I do think they target a different use case. Both of these plugins focus heavily on the time taken per boss and show the amount of time spent at a certain boss (stage). This plugin focusses on the overall HM timer and is intended to give the user an indication of how they are progressing towards that goal.

PR #1569, opened 2021-06-19
Create spacebar-checker

jcbelanger:
Spacebar checks are a fun chat game that people play when chat hits a lul. This plugin counts consecutive spacebar messages in chats to easily count spacebar check highscores.


raiyni (changes requested):
plugins needs to be set to `com.github.jcbelanger.spacebar.SpacebarCheckerPlugin`

PR #1572, opened 2021-06-20
Update quest-helper to v2.2.0

Zoinkwiz:
- Adds Falador, Kandarin and Fremmy Diaries
- Fixes issues with Karamja Diary not progressing
- Fixes to various Zeah quest helpers which broke with the new quest


raiyni (approved):
LGTM, long read.

PR #1577, opened 2021-06-22
Add TobMistakeTracker v0.1

QuestingPet:
Tracks mistakes made by players throughout the Theatre of Blood.

This new plugin will track mistakes for you and your teammates in the Theatre of Blood. It will also track mistakes for other raiders while you're spectating.

By default, when detecting a mistake, all players with this plugin will receive a public message of the mistake, a chat overhead above the player who made the mistake, and the mistake will be added to the Tob Mistake Tracker side-panel.

Currently, only the first couple of rooms (and deaths throughout the raid) are finished and being submitted in this version. I want to get this out the door in its current state so players can start to use it, and release more features very shortly. Refer to the [README](https://github.com/QuestingPet/TobMistakeTracker/blob/master/README.md) for more information and screenshots.

---
Regarding why I think this plugin is allowed: I don't feel as if it's doing anything against the spirit of the game. All mistakes are shown to the player *after* they've already made them, and are actively being punished by the game already. I made sure to never give the player information too early so that they could use it for decision making (like showing where/when certain mistakes could/would happen). 

For example, when you die, well, you're already dead anyway, and you can't do anything about that. When you get stunned by a bloat hand, the mistake doesn't notify you until *after* it's already too late to do anything about it. For maiden blood, you're already actively losing hp, prayer, and healing the boss, and I make sure not to alert you until the tick where that's relayed to you already, not any sooner.

If for some reason any of this is deemed against the spirit of the game, I'm more than happy to have a discussion about what we can do to fix it, including adding additional delays on when mistakes are revealed to the player, if we deem necessary.

NB: I know this is a big PR, and I really appreciate your time taking a look and considering it. I've spent countless hours the past week implementing all this, and I'm really passionate about it. I look forward to the discussion!

### DIFF
--- a/plugins/tob-mistake-tracker
+++ b/plugins/tob-mistake-tracker
@@ -1,0 +1,2 @@
+repository=https://github.com/QuestingPet/TobMistakeTracker.git
+commit=703ae3253a395bbb4cb94d8376b3a2c25d50603a
--- a/plugins/tob-mistake-tracker
+++ b/plugins/tob-mistake-tracker
@@ -1,2 +1,2 @@
 repository=https://github.com/QuestingPet/TobMistakeTracker.git
-commit=703ae3253a395bbb4cb94d8376b3a2c25d50603a
+commit=e1196be5b02ca95e6f135f061c8c604b4fe0c6bc


LlemonDuck:
Code structure change recommendations:

* Method is never used, but should be
https://github.com/QuestingPet/TobMistakeTracker/blob/703ae3253a395bbb4cb94d8376b3a2c25d50603a/src/main/java/com/tobmistaketracker/TobRaider.java#L43-L45

* Should be enum
https://github.com/QuestingPet/TobMistakeTracker/blob/8dbd1440626176f58f402800b5b9d5f697be94f3/src/main/java/com/tobmistaketracker/TobMistakeTrackerPlugin.java#L68-L70

* Can be moved out of `startUp`
https://github.com/QuestingPet/TobMistakeTracker/blob/8dbd1440626176f58f402800b5b9d5f697be94f3/src/main/java/com/tobmistaketracker/TobMistakeTrackerPlugin.java#L137

* Too deeply nested
https://github.com/QuestingPet/TobMistakeTracker/blob/8dbd1440626176f58f402800b5b9d5f697be94f3/src/main/java/com/tobmistaketracker/TobMistakeTrackerPlugin.java#L327-L341

* Never used?
https://github.com/QuestingPet/TobMistakeTracker/blob/6e87a19eba104f7c6cc7b0f15da01df2bcf9ee9b/src/main/java/com/tobmistaketracker/detector/BloatMistakeDetector.java#L42

* Doesn't need to extend `BaseTobMistakeDetector`, and likely shouldn't if it exclusively delegates, and doesn't detect mistakes on its own
https://github.com/QuestingPet/TobMistakeTracker/blob/d155f609420259398eebcc75c72a828c7945dabc/src/main/java/com/tobmistaketracker/detector/MistakeDetectorManager.java#L26

PR #1594, opened 2021-06-26
Create partyreadycheck

scainburger:
Display an alert and play a sound to check ready status of your party. Currently only works for ToB parties.


raiyni:
Do not set `loadWhenOutdated` and `enabledByDefault`

scainburger:
Done

scainburger:
My bad, that @Extension annotation was left over from OpenOSRS development. Thanks for your patience :)

scainburger:
My bad, that annotation was left over from development within another client. Thanks for your patience :)

PR #1598, opened 2021-06-27
Update quest-helper to v2.2.2

Zoinkwiz:
- Adds better handling of duplicate items in quest bank tab
- Many small fixes to existing helpers


Zoinkwiz:
Closing for a moment as found an issue with the world map causing a crash, will re-open once resolved.

PR #1605, opened 2021-06-29
update tob-mistake-tracker to v1.0

QuestingPet:
# Rev 1

This update sees the introduction of Verzik P2 mistake detection, which includes being bounced, bombed, and walking on acid (HM).

Also in this update is the transition from a debug plugin config to using developerMode for debug/developing. This is especially useful for separating mistake files on disk, so that my production/live account does not interfere with my dev build.

I have tested this on both my dev build and the live build and confirmed that no debug overlays show in the live build.

---
# Rev 2

Also added Verzik P3 mistake detection, which includes standing on a web tile and purpling. This will mark the 1.0 release for the plugin.

Note that the detection (and by proxy messaging) for the web mistake is intentionally delayed until *after* the web despawns, whether by being attacked by a player or by expiring. This is so to avoid the scenario where teams use this communication to "cheat" by providing pivotal information about this mechanic during the raid.

Additionally, it would be nice to get the animations/graphics for verzik p3 unmasked so that I could add the tank meleeing the team as a mistake easily.

---
# Rev 3

I removed the debug overlays and a lot of the mentions of `developerMode` as per @Hydrox6 's request. I decided to keep it in for the reading/writing of the mistakes, as I think it's important to me or any other developer to have a separate mistakes file for development/release.


QuestingPet:
This is now updating the plugin to v1.0 with the completion of verzik detection -- updated the description with more info.

Hydrox6 (changes requested):
Waiting on dev code to be removed, author aware of this request

QuestingPet:
Updated to remove debug overlays and most of the `developerMode` references that I feel are not pertinent to the release, addressing @Hydrox6 's comments.

PR #1638, opened 2021-07-07
Update classes-of-gielinor for RuneLite 1.7.15

Kraelll:
Updated plugin for RuneLite 1.7.15


Hydrox6:
You don't need to do this. That field is only for local testing

Also, clean your branch before you make a new PR, it will make your life a lot easier

PR #1643, opened 2021-07-08
Update pvp-performance-tracker to v1.4.4

Matsyir:
Minor updates:
- Limit amount of previous fights rendered in the Fight History Panel (while keeping their stats saved and included in the Total Stats), to avoid excessive UI Creation/ram usage. This is configurable under the "Max Rendered Fights" config, default to 200.
- Added dps/deserved damage support for crystal armor boosts on crystal bow and fbow.


Matsyir:
I just noticed that I only applied the first rendering fix while multiple fights are added to the panel, not when only one fight is added. Going to fix quick but it's pretty minor if anyone happened to be about to merge.

Matsyir:
Ok it's good to go now, re-tested in LMS with the extra fixes.

PR #1645, opened 2021-07-08
Update tob-mistake-tracker to v2.1

QuestingPet:
Changes:
* Add config for toggling mistake overhead text
  * Disabling overhead text would make it so that mistakes no longer appear to be said by players when detected. All other mistake detection/tracking implementation would stay the same (including viewing mistakes in the panel).
* Fix bug where plugin thinks a player dies who didn't actually die
  * This change makes use of onAnimationChanged with AnimationID.DEATH instead of the previously used onActorDeath. This is because onActorDeath would sometimes get triggered even if the server determines that the player is not actually dead (like regen-ing hp on the same tick that you would go to 0).

### DIFF
--- a/plugins/tob-mistake-tracker
+++ b/plugins/tob-mistake-tracker
@@ -1,2 +1,2 @@
 repository=https://github.com/QuestingPet/TobMistakeTracker.git
-commit=ab7d8470f9aefa4f50fc5e664e108724222015d3
+commit=b9838b82fc1edc3a821898829bd99d9f2a8cefaf


QuestingPet:
Turns out that changing the death detection logic leads to a regression, as now it takes longer (a few extra ticks) to detect if the player is dead, and so they can still accumulate mistakes in that time. This is because mistake detection for most bosses/rooms is directly tied to whether or not the player is deemed "dead" (according to the plugin). 

For example, if a player is standing on maiden blood, previously it would stop counting maiden blood as mistakes as soon as they "died" (onActorDeath). However, when switching to onAnimationChanged, since the death animation actually doesn't trigger until around 2 ticks *after* the player has actually died, extra erroneous maiden blood mistakes would be counted for the player, up until the animation started.

There are a few things I can do to fix this:
* Tie "mistakes" to actual hitsplats instead of just assuming that they're happening based on player position and death status
  * This won't work for mistakes that don't necessarily have hitsplats though, or if the hitsplat timing doesn't necessarily line up
* Use a more robust method to determine if we should continue counting mistakes for a player, which does not need to be the same as if the player is "dead" according to the plugin.

Overall I'll have to rethink this and re-test this a bit, and will come back to it another time. For now, I will continue using the old detection which includes the rare edge case described in the description above. I'll submit a new PR with just the config change.

PR #1659, opened 2021-07-12
Update hydrate-reminder

jmakhack:
Are you down bad and feeling extremely thirsty?
Fear not my fellow adventurers for hydrate reminder v1.1.0 is here!

hydrate_water_dance


Nightfirecat:
Your build.gradle changes aren't building because you've changed all the versions to ones which we don't have permitted in our dependency verification. I recommend you reverse those changes.

PR #1672, opened 2021-07-15
Update Runekit plugin

ImToxicCode:
Feature added: Notify raid members when other raid members have disconnected.
Raid members must have plugin installed to get notifications.


plugins/runekit
```diff
@@ -1,3 +1,3 @@
 repository=https://github.com/ImToxicCode/runekit.git
-commit=3ae3e07e2bbe0381166c3508c96bd09373f364ac
+commit=7955515ae39b8db43ade5dae2a9ff2c7280a9a75
 warning=This plugin communicates the stage of crops and birdhouses to a 3rd party server not controlled or verified by the RuneLite Developers.
```
Nightfirecat, line 3: This should be updated. Perhaps you should make it more general so it doesn't need to be changed each time a feature is added?

ImToxicCode: I updated the file.

plugins/runekit
```diff
@@ -1,3 +1,3 @@
 repository=https://github.com/ImToxicCode/runekit.git
-commit=3ae3e07e2bbe0381166c3508c96bd09373f364ac
-warning=This plugin communicates the stage of crops and birdhouses to a 3rd party server not controlled or verified by the RuneLite Developers.
+commit=caedbaafef4b1b00fe36f1b123fb06dd26572b42
+warning=This plugin sends data to a 3rd party server not controlled or verified by the RuneLite Developers.
```
Nightfirecat, line 3: Maybe this can be a little more specific, eg.
```suggestion
warning=This plugin sends player data to a 3rd party server not controlled or verified by the RuneLite Developers.
```

PR #1686, opened 2021-07-20
Add C Engineer: Completed plugin

m0bilebtw:
This plugin announces, with custom sound clips I personally requested from C Engineer, when the player reaches a new level, or completes a quest, achievement diary, or a new collection log slot. It also announces player death using the sound clip from his PvP HCIM dying. There is also an optional chat message for these same events.

This is simply a 'for fun' plugin!

When requesting the sound clips, I made it clear that they will be used for this exact plugin, and he gave both permission and the files themselves for this to go ahead over a discord text chat. He also gave permission for me to clip his youtube profile picture/icon for the plugin icon, as I have done.

### DIFF
--- a/plugins/c-engineer-completed
+++ b/plugins/c-engineer-completed
@@ -1,2 +1,2 @@
 repository=https://github.com/m0bilebtw/c-engineer-completed.git
-commit=7ab0801ebd6c969c38345b787f7d61444922aabb
+commit=1cfd686777e8d8ecb5db5c557f16271af53b0dcc
--- a/plugins/c-engineer-completed
+++ b/plugins/c-engineer-completed
@@ -1,0 +1,2 @@
+repository=https://github.com/m0bilebtw/c-engineer-completed.git
+commit=7ab0801ebd6c969c38345b787f7d61444922aabb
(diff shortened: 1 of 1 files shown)


m0bilebtw:
Screenshot from 2021-07-20 05-55-47
Screenshot from 2021-07-20 05-56-20
Screenshot from 2021-07-20 05-56-37

Attaching screenshots of the conversation I had with C Engineer and Pept. Pept is a high rank in the C Engineer clan chat and friends chat, which I am a slightly lower rank in, who helped me get in direct touch with C. If further proof is needed please let me know!

Nightfirecat:
[Are the various streams opened in SoundEngine ever closed?](https://github.com/m0bilebtw/c-engineer-completed/blob/1cfd686777e8d8ecb5db5c557f16271af53b0dcc/src/main/java/com/github/m0bilebtw/SoundEngine.java) Where possible, they should be explicitly closed or opened in a try-with-resources block so the memory can be freed.

m0bilebtw:
> [Are the various streams opened in SoundEngine ever closed?](https://github.com/m0bilebtw/c-engineer-completed/blob/1cfd686777e8d8ecb5db5c557f16271af53b0dcc/src/main/java/com/github/m0bilebtw/SoundEngine.java) Where possible, they should be explicitly closed or opened in a try-with-resources block so the memory can be freed.

Ah, no. I'll get on that.

PR #1695, opened 2021-07-21
Create lingering-click-tooltips

ryanxie0:
This is a new plugin to generate lingering tooltips/toasts for left click actions.


raiyni:
Your plugin name in the properties file needs to be properly cased

ryanxie0:
Just updated the properties file

raiyni:
You need to push the new commit hash in the plugin file here

PR #1706, opened 2021-07-25
Update npc-dialog-log to v1.3

neilrush:
Updated icon
Added config panel with options to toggle all features
Added options to show dialog overhead
Fixed chat messages having right click options


neilrush:
Are there any issues I can resolve?

raiyni:
I think your shutdown should check if the person is logged in?

neilrush:
Shutdown now checks if the user is logged in. Also, the `lastMessageTickTime` list is now cleared if the user logs out/hops etc.

PR #1713, opened 2021-07-26
Update runekit

ImToxicCode:
Updated logging.. Version might have been incorrect? Works fine in debug mode but not in production.


raiyni:
Those log statements won't be ran for users anyway.

ImToxicCode:
> Those log statements won't be ran for users anyway.

I know, I wanted to see if updating the commit string would update it properly.

raiyni:
This update wouldn't make a difference for that since nothing is being changed.

ImToxicCode:
> This update wouldn't make a difference for that since nothing is being changed.

Never mind, I just tried it again and its working now. That so odd.

Thank you

PR #1719, opened 2021-07-27
Add combat-indicators

dekuScrub731:
Show highlight of Player and NPCs when in combat

### DIFF
--- a/plugins/combat-indicators
+++ b/plugins/combat-indicators
@@ -1,2 +1,2 @@
-repository=https://github.com/dekuScrub731/combat-indicators
+repository=https://github.com/dekuScrub731/combat-indicators.git
 commit=fe029069d4ed70396a2d1c9c17ae800b18440568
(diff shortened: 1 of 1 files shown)


LeikvollE:
In violation of "Indicates which player an NPC is focused on"? Otherwise this could be merged with the interaction highlighter?

dekuScrub731:
@LeikvollE - Yeah, I did consider that text before I started with this plugin. I ended up interpreting the spirit of that guidance as not allowing an NPC to show highlighting to various players (including players other than local). If the consensus is that this violates Jagex guidance, I can rework to only show what the local player is targeting.

I also saw the Interaction Highlighter plugin PR yesterday right before I submitted this pull request. I agree, it makes sense to group similar highlighting functionality together, so I'm ok with it being merged.

LeikvollE:
@dekuScrub731 I don't know if you have read the discord discussion about the interaction highlighter, but it will be included in core and Adam has added functionality somewhat similar to this plugin.

dekuScrub731:
Closing this PR as most of "player initiated" actions are covered under interaction highlighter changes. NPC initiated highlighting may be a gray area in terms of Jagex permitted actions.

PR #1723, opened 2021-07-28
add motivation-plugin

sqlskid:
This plugin will send some motivating messages after death in your chat.


Hydrox6 (changes requested):
You should not have `loadWhenOutdated = true` in your plugin descriptor

PR #1732, opened 2021-08-01
Add Music Mask plugin

lequietriot:
Allows the user to play higher quality MIDI music over the Old School RuneScape game. This plugin supports SoundFont (.sf2) files and features an upgraded RuneScape sound engine for playback.

### DIFF
--- a/plugins/music-mask
+++ b/plugins/music-mask
@@ -1,2 +1,2 @@
 repository=https://github.com/lequietriot/music-mask.git
-commit=234945ca252f4eea77d1d872e9dccadb7642fc71
+commit=4bcb9d65f42c2826152aee8359efe002167245fd
--- a/plugins/music-mask
+++ b/plugins/music-mask
@@ -1,2 +1,2 @@
 repository=https://github.com/lequietriot/music-mask.git
-commit=5466b23892f9c88fb332e5a1757e846d24d0f482
+commit=db3c132d137746d5e89407244d611ed6243eac40
--- a/plugins/music-mask
+++ b/plugins/music-mask
@@ -1,0 +1,2 @@
+repository=https://github.com/lequietriot/music-mask.git
+commit=31fb78f96a9328a7baba95171b7d292b87933567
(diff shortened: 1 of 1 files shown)


Nightfirecat:
Is [the source in `com.musicmask.musicsystem`](https://github.com/lequietriot/music-mask/tree/234945ca252f4eea77d1d872e9dccadb7642fc71/src/main/java/com/musicmask/musicsystem) your own, or is it from some other package? It does not have any attribution currently, but looks like library code.

lequietriot:
> Is [the source in `com.musicmask.musicsystem`](https://github.com/lequietriot/music-mask/tree/234945ca252f4eea77d1d872e9dccadb7642fc71/src/main/java/com/musicmask/musicsystem) your own, or is it from some other package? It does not have any attribution currently, but looks like library code.

Most of that code comes from https://github.com/lequietriot/RuneScape-MIDI-Player-Suite (Another package) which is based on the OSRS sound engine. Some of the imports are from a package com.sun.media.sound to support SoundFont 2 files.

Nightfirecat:
I guess more specifically, is that all code that you wrote and own, and have license to publish? Is it _based on_, or _ripped from_ the osrs sound engine? (we likely can't publish the latter, but the former should be fine)

lequietriot:
> I guess more specifically, is that all code that you wrote and own, and have license to publish? Is it _based on_, or _ripped from_ the osrs sound engine? (we likely can't publish the latter, but the former should be fine)

Most of the code in the musicsystem source is based on the osrs sound engine. I could probably create an alternate sound engine code if needed for publishing, although it would still be based on the old runescape way of handling midi music.

Nightfirecat:
FYI all the code in `com.sun.media.sound` is licensed under GPL2, which requires work containing or derived from it also be licensed under GPL2; [your license is currently the BSD-2 Clause "Simplified" license](https://github.com/lequietriot/music-mask/blob/db3c132d137746d5e89407244d611ed6243eac40/LICENSE).

PR #1738, opened 2021-08-03
Create costume-room-highlighter

michaelperino:
Plugin highlights all items that can be put into a house costume room as defined on the wiki. Only highlights in bank, color and enabled storage locations changeable in settings.


plugins/costume-room-highlighter
```diff
@@ -0,0 +1,2 @@
+repository=https://github.com/michaelperino/CostumeRoomChecker
```
Cyborger1, line 1: The repo clone url needs to end in .git, this this:
```suggestion
repository=https://github.com/michaelperino/CostumeRoomChecker.git
```

PR #1746, opened 2021-08-05
Create monster-examine

leejt:
Very similar to item-interactions except it reports monster examines.


leejt:
After further consideration, this is going to be rolled into crowdsourcing with a flag in #1752

PR #1747, opened 2021-08-05
Native Mac Notifications

fatfingers23:
This implements native Mac notifications via how IntelliJ implements them. 

If this is tested locally it will show the notification came from Java, but should say Runelite in production since the code is run from the Runelite bundle.


Adam-:
We already have a client dependency on jna, which is transitive via the client, so it should not be something which you have to add a dependency for.

fatfingers23:
Did not realize it was already a dependency. I switched over to it and this should be good to go now. Thank you!

Adam-:
Additionally it would be good to make your plugin not break massively whenever installed on a non osx machine.

fatfingers23:
Added a check to see if the it is Mac OS if not it turns off the plugin.  Client no longer crashes on windows when a notification is delivered with the plugin installed.

PR #1748, opened 2021-08-06
Create thrall-helper

PortAGuy:
This plugin displays a red box when the player does not have a thrall active


raiyni:
Can this be integrated into `advanced-notifications`

PortAGuy:
While it probably could be, I am very new to github and programming in general and after a quick look I wouldn't be able to make the changes there

PR #1751, opened 2021-08-07
Create respawns

leejt:
This plugin is for a wiki project to add respawn times for items and scenery to article infoboxes. This is meant to augment the main crowdsourcing.


leejt:
After further consideration, this is going to be rolled into crowdsourcing with a flag in #1752

PR #1771, opened 2021-08-13
Create rl-weather (Weather effects for OSRS)

bogstandard:
Barebones Weather effects for osrs, in keeping with the look and feel of the game. Crunchy optional audio effects and osrs art-style friendly looking rain, snow and lightning.

Happy to make amends to the codebase if needed, any advice is very welcome.

### DIFF
--- a/plugins/rl-weather
+++ b/plugins/rl-weather
@@ -1,0 +1,2 @@
+repository=https://github.com/bogstandard/rl-weather.git
+commit=1f6bf0ce1c8e8ecac7be5f4a0daed3aa9335649c


Nightfirecat:
FYI, we do have a `@Units` annotation which can be added to configs to make it more clear what they represent for configs like this: https://github.com/bogstandard/rl-weather/blob/1f6bf0ce1c8e8ecac7be5f4a0daed3aa9335649c/src/main/java/com/rlweather/RlweatherConfig.java#L39. Might be useful for a later update.

PR #1798, opened 2021-08-22
Intial commit for Life Saver

keywok:
Howdy! The plugin is a re-worked notifier based off the Idle Notifier that is within RuneLite already. The key difference is this version forces through a notification even if you are intereacting with the client constantly. Main uses being thieving or playing a HC and not wanting the notification to go through.

image


keywok:
Sorry for the additonal commits I only realised after I had created the request that there was already a plugin with the name I had randomly picked.

PR #1809, opened 2021-08-26
clan hall checker

Gmoley:
checks what clan members are in sight upon using !Clanmembers in public chat. and prints them in chat 
this is feature is only available when in the clan hall


Nightfirecat:
Your repo is either private or deleted, please look into that.

PR #1816, opened 2021-08-28
added gauntlet-minimap

VicSegers:
My first plugin, feedback is welcome!

The plugin shows the resource nodes of the gauntlet on the minimap. Has the option to hide/display certain resource nodes.

### DIFF
--- a/plugins/gauntlet-minimap
+++ b/plugins/gauntlet-minimap
@@ -1,2 +1,2 @@
 repository=https://github.com/VicSegers/gauntlet-minimap.git
-commit=d4c274393ffc4b5b9c7a7c760a21a5e3c4cf6e56
+commit=ef13ee67325d5b8134ee4f49d1d466bc048b7573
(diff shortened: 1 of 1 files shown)


IllyaMyshakov:
This could work well as an addon to the existing phub plugin "The Gauntlet" by dutta64. Though tbf with uncapped render distance, the dutta64 plugin functionally addresses the same need as this plugin.

You can have a look at this for reference: https://runelite.net/plugin-hub/show/the-gauntlet

VicSegers:
> This could work well as an addon to the existing phub plugin "The Gauntlet" by dutta64. Though tbf with uncapped render distance, the dutta64 plugin functionally addresses the same need as this plugin.
> 
> You can have a look at this for reference: https://runelite.net/plugin-hub/show/the-gauntlet

That's true, but dutta64 said "Only problem is that I don't have membership/I'm not actively playing osrs, so I can't test it." that's why I didn't add to this plugin.

PR #1826, opened 2021-09-02
Runecrafting LClick Only

Airacan:
Plugin that changes most default actions (Left Clicks) related to Runecrafting so that the player no longer needs to R-Click and navigate through tons of menus. Helps the wrist with long Runecrafting sessions.

Example: Default action of Essence Pouches is normally 'Fill', this has been replaced with 'Empty'. Whereas the Fill option is set to be the default action while in Bank. This is especially useful for places like the Ourania Altar, where acidentally clicking off the bank costs an extra 20+ Runes.


VicSegers:
This is already possible with a Shift-left click when using the menu entry swapper.

Nightfirecat:

PR #1827, opened 2021-09-02
Add clan-member-list-sort plugin

jodelahithit:
This plugin adds a sort button to the clan member list.


Adam-:
The 2nd plugin change doesn't diff because you are both changing commit and repo at once, can you change just the repo?

jodelahithit:
There are no changes to the skilling-notifications plugin. I’m not sure why it’s showing up in this pull request. 
Both plugins are in the same repo but on different branches.

Adam-:
If the commit is changing then there are by definition changes 
image

jodelahithit:
My apologies, I've realized my mistake and will open a new pull request once I've fixed my fork of plugin-hub.

PR #1828, opened 2021-09-03
Add clan-member-list-sort plugin

jodelahithit:
This plugin adds a sort button to the clan member list.


Nightfirecat:
[`initWidgets()` should be called on the client thread.](https://github.com/jodelahithit/runelite-plugins/blob/12c43b716374d01e20151466103ca9aca9d99746/src/main/java/eu/jodelahithit/clanmemberlistsort/ClanMemberListSortPlugin.java#L55) `startUp()` and `shutDown()` are both invoked on the EDT.

jodelahithit:
Thank you for your feedback. I've fixed the issue in [this](https://github.com/jodelahithit/runelite-plugins/commit/bcd24a28171f44b94a004579df6e597aefd8fbb9) commit.

PR #1833, opened 2021-09-05
Add toxic-npcs plugin

ValerianClerc:
[This plugin](https://github.com/ValerianClerc/toxic-npcs) is intended to make the RuneScape playing experience optionally more immersive by roasting players for their mistakes. It's funny when you get hit for a 73 or you die, and a boss roasts you.

There's a lot of room for expanding on this plugin, such as having bosses roast you for safeing, missing prayers, having other players "roast" you, etc. Currently we have support for the following features:
- get roasted on death
- get roasted when boss hits you for a large number (customizable per boss)
- users can customize the frequency of the roasts

Picture attached for reference:

image

Thank you :)


Hydrox6:
See #704

PR #1840, opened 2021-09-07
Clan member list sort

jodelahithit:
-Added reverse sort option
-Added icon
-Improved sorting based on rank
-Updated readme


raiyni:
You can add methods to config interfaces that act as setters do you don't need to do `configManager.setConfiguration(CONFIG_GROUP, "activeSortType", type)` btw

PR #1843, opened 2021-09-08
feat: wilderness-level-indicators added

arnaudperalta:
New plugin intended to highlight players in the wilderness within the same combat bracket.


raiyni:
Goes against `Offers additional information about other players, with the purpose of scouting PvP targets`

PR #1851, opened 2021-09-10
Updated gauntlet-minimap to v1.1.1

VicSegers:
Cleaning up some code and adding additional tags, no features were changed.


Nightfirecat:
If there's no behavior change, you don't need to release it fyi

VicSegers:
Ok my bad!

PR #1856, opened 2021-09-11
Add old-control-run plugin

LordOfOtakus:
This plugin undoes the change to Control-Click running and reverts it to its old behavior, which was the Control-Click will result in the player always running, regardless of current run state.


Adam- (changes requested):
keylisteners are required to be unregistered at shutdown

LordOfOtakus (commented):
New commit fixing issue

PR #1858, opened 2021-09-13
update coffin-counter to v1.1

Hydrox6:
Adds infoboxes to show the contents of the coffin, so the inventory doesn't have to stay open all the time


Hydrox6:
forgot something, oops

PR #1865, opened 2021-09-14
Two step zoom

SkyBouncer:
Plugin that easily allows you to swap between two zoom levels from the keyboard.
No longer have to fiddle with the slider to get it just right if you sometimes want to zoom in, and sometimes want to play on your "normal" level of zoom.


Nightfirecat:
You should probably [remove `enabledByDefault = false`](https://github.com/SkyBouncer/TwoStepZoom/blob/39b6b6e875d6232f58c1a536353f57bd68d07161/src/main/java/sky/twoStepZoom/TwoStepZoomPlugin.java#L19) as it can be confusing to users to install a plugin which is not enabled when installed.

PR #1871, opened 2021-09-16
Update inferno-stats plugin to latest

Supalosa:
Added a 'idle tick' tracker to the plugin so players can analyse how many ticks they weren't attacking something during a run. Does not actively notify players or tell them what to do so I don't think this breaks any rules.

### DIFF
--- a/plugins/inferno-stats
+++ b/plugins/inferno-stats
@@ -1,2 +1,2 @@
 repository=https://github.com/InfernoStats/InfernoStats.git
-commit=a36d75e16b7e679246c3e9d00a22cd8fdbb1dc06
+commit=fcdebffcc360a8c2acfa460d6de17896480b414e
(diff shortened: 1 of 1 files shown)


Nightfirecat:
Please don't log debug-type information at info level every tick: [`a36d75e16b7e679246c3e9d00a22cd8fdbb1dc06..fcdebffcc360a8c2acfa460d6de17896480b414e`#diff-d55ed8f7f5](https://github.com/InfernoStats/InfernoStats/compare/a36d75e16b7e679246c3e9d00a22cd8fdbb1dc06..fcdebffcc360a8c2acfa460d6de17896480b414e#diff-d55ed8f7f56094a220b2ee35b6f57ff646f3ea70f249d847dc3f412a4618a8baR94). I strongly recommend using `log.debug()` instead wherever possible.

Supalosa:
Good catch, Fixed.

PR #1875, opened 2021-09-18
Update Barbarian Assault PBS to version 1.2.0

SkylerPIlot:
Adds better support and tracking for different team composition and tracks roles more efficiently


Nightfirecat:
Please use `log.debug` instead of `log.info` for your debug logging. Info logs are logged to users' logs, which pollutes them when we need to analyze them in the support channel, for instance.

SkylerPIlot:
I did think about getting rid of the log.info before you said something, but now its all fixed :).

PR #1883, opened 2021-09-21
Created Bank Withdraw Swapper

Sholto:
Plugin that allows you to change the left click withdraw amount for different items.
For example can configure Fire runes to left click Withdraw-All and Abyssal whip to left click Withdraw-1.
For things like Blast furnace, you want to withdraw all your Coal, however you only want to withdraw 1 Stamina pot dose.

Only allows the withdraw amounts that already exist in the bank. Does **not** allow custom amounts to be set.


Nightfirecat:
After some discussion, we decided we do not want to add this plugin as it would make banking for certain activities too trivial.

PR #1904, opened 2021-09-25
FlippingUtilities

zumaad:
quick summary:
- Moved the plugin to our org 
- Added slots syncing feature which submits current slots (if users opt in) to our api.
- Added a warning for the feature above ^

### DIFF
--- a/plugins/flipping-utilities
+++ b/plugins/flipping-utilities
@@ -1,3 +1,3 @@
 repository=https://github.com/Flipping-Utilities/rl-plugin.git
-commit=b419f7dd4601c4c2adb62fd818c990e2f9c03390
+commit=40583d4a673a8d114b082769e922827d1590b638
 warning=If you've explicitly opted in to our syncing feature by logging in to our api, your GE data will be synced with our api. If you do not opt in, all of your data remains stored locally and the plugin still remains functional.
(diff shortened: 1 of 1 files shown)


AntonyGarand:
The RL github app doesn't reference the right place as we switched our repo home
You can view the changes from the new repo instead:
https://github.com/Flipping-Utilities/rl-plugin/compare/09b17807a785285512be2d87ef286c74ead963ae...7043841d6da9270cb3baa79221e2958aed727fa6

From `09b17807a785285512be2d87ef286c74ead963ae` to `7043841d6da9270cb3baa79221e2958aed727fa6`

plugins/flipping-utilities
```diff
@@ -1,2 +1,3 @@
-repository=https://github.com/Belieal/flipping-utilities.git
-commit=09b17807a785285512be2d87ef286c74ead963ae
\ No newline at end of file
+repository=https://github.com/Flipping-Utilities/rl-plugin.git
+commit=40583d4a673a8d114b082769e922827d1590b638
+warning=If you've explicitly opted in to our syncing feature by logging in to our api, your GE data will be synced with our api. If you do not opt in, all of your data remains stored locally and the plugin still remains functional.
```
Nightfirecat, line 3: Keep in mind that this message should be written from the perspective of RuneLite. With that in mind, this is not an acceptable warning message format--you should look to other examples in the hub, such as `This plugin submits Grand Exchange transactions and your IP address to a 3rd party website not controlled or verified by the RuneLite Developers`

AntonyGarand: Is this new text fine?
If not, we're open to suggestions: Trying to show that it is optional, while keeping the warning information relevant if they do log in.

> If you opt in to do so, this plugin will submit Grand Exchange transactions and your IP address to a 3rd party website not controlled or verified by the RuneLite Developers. If you do not opt in, this plugin will store your data only locally and will still remain functional.

zumaad:
Closing to split the PR into two separate ones

PR #1158, opened 2021-02-27
Added new plugin: twitch-emotes

dylanblokhuis:
Twitch allows users to express themselves with "twitch" emotes in chat. This plugin brings this functionality to RuneLite. Most users also install browser extensions from third party platforms like [BetterTwitchTV](https://betterttv.com/) and [FrankerFaceZ](https://www.frankerfacez.com/). This plugin communicates with a service which caches the integration with these platforms. You can find the repository here: https://github.com/dylanblokhuis/twitch-emotes-warehouse.git

PR #43, opened 2019-12-29
Create twitch-emotes

loldudester:
Adds Twitch Global Emotes to chat, similar to the Emojis plugin.

PR #1946, opened 2021-10-06
Add twitch-emotes plugin

FraserKillip:
This plugin works in an identical way to the default Emoji plugin. It will replace a few sets of characters with images commonly used on twitch


Nightfirecat:
Does this suffer the same issues as #43 and #1158?

FraserKillip:
> Does this suffer the same issues as #43 and #1158?

Oh yes I completely didn't think about that. I will look into alternatives 👍

PR #1953, opened 2021-10-07
Update input-recolor

Ririshi:
A re-write of large parts of the plugin, as well as a display name change and some bug-fixes. @MarbleTurtle invited me to be a contributor for the Color-Coordinator repository, since he is currently rather inactive. That's why this PR is coming from me, and my plugin hub fork, and not from him.

### DIFF
--- a/plugins/input-recolor
+++ b/plugins/input-recolor
@@ -1,2 +1,2 @@
 repository=https://github.com/MarbleTurtle/Color-Coordinator.git
-commit=8f1b5b3d74ca6bfe69fce2e85e7b8ddfa48a8383
+commit=813a35df61fbc948a60c0cc51cfc9ea40383bafe


Ririshi:
I'm clearly new to this plugin hub PR thing. Looks like the shadowJar plugin blocking the build script, so I'll remove that and add a proper version bump, then add another commit to my plugin-hub fork.

Nightfirecat:
@MarbleTurtle can you confirm this?

Ririshi:
Will do! Can I ask why that's a problem? Other plugin also seemed to have it. Makes me wonder if there is a specific list of requirements that I can check somewhere, for any future plugins / updates I create.

Alexsuperfly:
log.info will spam the logs with useless information, so when we have to look through them in the support channel it becomes quite annoying

PR #1965, opened 2021-10-10
Add new plugin for Tithe Farm

TicTac7x:
Tithe farm plugin that:
* shows progress of the plants that only you have planted with game-tick precision
* highlights farming patches on mouse hover
* highlights seeds and watering cans in the inventory
* shows overlay of available/total amount of water in the inventory


Nightfirecat:
`TitheOverlayPoints#shutDown()` is calling client methods off client thread. (shutDown is run on EDT) You should fix this by using `clientThread#invokeLater()`.

PR #1967, opened 2021-10-10
Add ToB Gear Checker plugin

ArtsicleOfficial:
A plugin that automatically double-checks everything for your ToB setup.
In-use
Configuration


Nightfirecat:
["gearchecker" is probably a bit too generic for a config name](https://github.com/ArtsicleOfficial/tob-gear-checker/blob/d56cba4b6c6fa401a3d24467647f1632ce2ab961/src/main/java/com/gearchecker/GearCheckerConfig.java#L8), can you make it "tobgearchecker" to match the plugin name?

PR #1976, opened 2021-10-11
Tempoross Harpoonfish Notifier Plugin

Matt-J-H:
A plugin that notifies you when a double Harpoonfish spot appears while doing Tempoross


Hydrox6:
Instead of having an entire plugin for something this narrow, it should be added to the existing [Tempoross plugin](https://runelite.net/plugin-hub/show/tempoross)

Matt-J-H:
That would be ideal for sure but the original author doesn't seem to be active anymore and I'm not sure how to contact him other than creating an issue on the repo. Any thoughts?

Matt-J-H:
Done, now we wait and hope, lol

PR #1981, opened 2021-10-14
Create friend-live-locations

TiboDeMunck:
This plugin allows for 'live location' sharing between players, based on account type you will see your friends/trusted acquaintances on the world map. It is possible to only receive data without sending it when going into the wilderness.

This plugin uses an external API that has to be setup for each group of friends (see README).


Nightfirecat:
[Your plugin's config group should not be "example".](https://github.com/TiboDeMunck/runelite-live-friend-locations-plugin/blob/9136920cacf6e671f596d8104eec7bdc36d235f8/src/main/java/com/livefriendlocations/LiveFriendLocationsPluginConfiguration.java#L7)

PR #1994, opened 2021-10-18
Create discord-webhook-notifier

Kipx:
A Discord Webhook Notifying plugin that will be able to notify a channel or server about certain events using a webhook. Currently, the plugin only supports quest notifications, but will be expanded upon to include drops, boss kills, skilling, and whatever else may be neat. Started a GIM with my friends and was inspired to make this when I saw that there wasn't already a Questing notifier.


WilliamWinterDev:
Have also done the same here; https://github.com/runelite/plugin-hub/pull/1982 but merged a lot of the discord webhooks into one

PR #2035, opened 2021-10-30
Instant damage calculator

geeckon:
Fix for non-grouped xp drops


mkdrs:
Hello, the plugin does not seem to work inside LMS specifically.  

Here is an example with character inside Nightmare Zone - showing accurate hits: 
[https://gyazo.com/09f94d8503d1af84d02df4ff8f30d75e](url)
09f94d8503d1af84d02df4ff8f30d75e-png


Here is an example with the same account/settings inside Last Man Standing - showing inaccurate hits: 
[https://gyazo.com/042bf1655853beec94f61501e675dffb](url)
760fb7e781e3076562369cb9ea040567
042bf1655853beec94f61501e675dffb-png

PR #2036, opened 2021-10-30
Create runelogger

Joopjr:
Added new plugin. This plugin collects data from the player (for example level ups, duel results, quest completions) and stores and display them on runelogger.com.


Nightfirecat:
You must add a `warning` to your plugin since it makes calls to external services.

Joopjr:
I've added the warning to this commit.

PR #2045, opened 2021-11-01
Bank Quantities plugin config

CubeLuke:
This plugin will add a new tooltip to items in player inventory to show the current quantity amount that they have in their bank. Useful for if you get a drop and want to know how much of that item you already have banked.  

Example Screenshots:


raiyni:
You need to submit the commit hash from your actual plugin, not the plugin-hub branch.

raiyni:
Your plugin class should be `com.cubeluke.BankQuantitiesPlugin`

CubeLuke:
My mistake on the incorrect has and the url being wrong, it was late so I thought I had put the correct one. Updated the plugin as well for the package name to be the full `com.cubeluke.BankQuantitiesPlugin` as mentioned above.

Nightfirecat:

PR #2051, opened 2021-11-02
update phillip-hue-integration

nganzalo:
- Added feature to set smart lights back to a default color after the plugin triggers.

- Removed bug where drops over 10k would not trigger plugin.

### DIFF
--- a/plugins/phillip-hue-integration
+++ b/plugins/phillip-hue-integration
@@ -1,2 +1,2 @@
 repository=https://github.com/nganzalo/Hue_Implementation.git
-commit=91c7973d3fb36dedc4c41714c13bd719b8e46df5
+commit=c043da885ba051b38dbcf5c7ea36194aee7afd95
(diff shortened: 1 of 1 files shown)


Hydrox6 (changes requested):
For the test command, you should really be using a colon command like `::testlights`, as they never leave the client like normal chat messages will.

See the following example on how the Clue Scroll plugin implements the `::clue` command (you don't need a dev mode check btw):
https://github.com/runelite/runelite/blob/master/runelite-client/src/main/java/net/runelite/client/plugins/cluescrolls/ClueScrollPlugin.java#L648

nganzalo:
> For the test command, you should really be using a colon command like `::testlights`, as they never leave the client like normal chat messages will.
> 
> See the following example on how the Clue Scroll plugin implements the `::clue` command (you don't need a dev mode check btw): https://github.com/runelite/runelite/blob/master/runelite-client/src/main/java/net/runelite/client/plugins/cluescrolls/ClueScrollPlugin.java#L648

This change has been implemented and commit updated to support this change.

PR #2074, opened 2021-11-07
Add Narration plugin

adamk33n3r:
What started out as a project to have the game read quest dialog to me, turned into more of an accessibility thing. This adds a bunch of ways that the game can read to you. There are still plenty more additions that can be made, I'm sure, but a few of the big ones are:
- Dialog
- Objects/NPCs/Items/UI you click on
- Hotkeyed narration of what you're hovering over (including items in the right click menu)
- Hotkeyed quantity of hovered item


abextm:
freetts does network io, disk io, reflection and external process execution in ways that are not trivial for me to guarantee that they are sane or safe. As such it cannot be used on the plugin hub.

adamk33n3r:
> freetts does network io, disk io, reflection and external process execution in ways that are not trivial for me to guarantee that they are sane or safe. As such it cannot be used on the plugin hub.

Oh interesting, I didn't realize it did network stuff, or disk for that matter. I guess it might just have that possibility and I was just not using it. I can try to look for another library. How did you discover that FreeTTS had these things? Do you know if MaryTTS has these issues? Or do you think all local TTS engines will have those issues?

abextm:
> How did you discover that FreeTTS had these things?

I look at the jar file

> Do you know if MaryTTS has these issues?

No clue. I know it is pretty sizable though. There is a previous attempt at tts using it.

adamk33n3r:
I guess now might be a good time to start a discussion about including a tts api in core runelite. Would you be able to hide those security concerns behind the api so that plugins wouldn't be able to perform them? I could see adding at least a simple engine (like FreeTTS) to core being really helpful for different plugins to be able to add that functionality without having to all include a separate ~8mb or w/e in their build (and also hiding the security issues if possible). It would also open up the potential to add narration accessibility to the core runelite client in the future (menu navigation etc).

abextm:
its far to niche to be in core, and shuffling the problem to core does not make it go away

adamk33n3r:
Gotcha, that's unfortunate. I guess I'll explore different options in the meantime. Thank you.

PR #2081, opened 2021-11-12
Add Qolting plugin

ArtsicleOfficial:
The Qolting (QoL + Alting) plugin is a specialized overlay for alting vyrewatches in darkmeyer.

Settings

As seen with the Ear Blasters, it allows the user to set custom noises for three scenarios: Your blood shard loot, someone else's blood shard loot, or any onyx bolt tips.

It tells the user when they have auto-retaliate off or don't have their vyre outfit on using a shouty white square.
TOO AFK!

It allows the user to get special flashy notifications for blood shards using a shouty white square.
Shard notification

It gives the user a flashing prayer indicator along the top bar of their screen (configurable). As can be seen here, there is also an overlay for the account's RSN. The colors of this prayer indicator are completely customizable.
Prayer notification

It tracks profit gained by summing the value of items picked up over the session (which turns green after 10,000,000).
Profit gained

It lists all items of a certain threshold of value.
Threshold list

It gives a moveable overlay for the number of items left in the inventory.
Overlay

Finally, it tells the player when their inventory is full and that they should "go bank" with a shouty white square.
Go bank

### DIFF
--- a/plugins/qolting
+++ b/plugins/qolting
@@ -1,2 +1,2 @@
 repository=https://github.com/ArtsicleOfficial/Qolting.git
-commit=dfad513d7aff0297c09e7436304e7dbdd08af09e
+commit=395d394bb3c760a81c1327c0d0599e3643f81ce4
--- a/plugins/qolting
+++ b/plugins/qolting
@@ -1,2 +1,2 @@
 repository=https://github.com/ArtsicleOfficial/Qolting.git
-commit=395d394bb3c760a81c1327c0d0599e3643f81ce4
+commit=742c0efb5240f30e74cb0c985921cf0d68b46e0a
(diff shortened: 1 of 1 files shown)


Nightfirecat:
[This isn't great.](https://github.com/ArtsicleOfficial/Qolting/blob/742c0efb5240f30e74cb0c985921cf0d68b46e0a/src/main/java/com/qolting/QoltingPlugin.java#L243-L255) You should [reference how audio clip loading works in the main client](https://github.com/runelite/runelite/blob/73b81701f8ffb7e79fa9f3d7247ae3c9986568c0/runelite-client/src/main/java/net/runelite/client/Notifier.java#L478-L506), primarily focusing on gracefully handling failures to avoid memory leaks.

ArtsicleOfficial:
> [This isn't great.](https://github.com/ArtsicleOfficial/Qolting/blob/742c0efb5240f30e74cb0c985921cf0d68b46e0a/src/main/java/com/qolting/QoltingPlugin.java#L243-L255) You should [reference how audio clip loading works in the main client](https://github.com/runelite/runelite/blob/73b81701f8ffb7e79fa9f3d7247ae3c9986568c0/runelite-client/src/main/java/net/runelite/client/Notifier.java#L478-L506), primarily focusing on gracefully handling failures to avoid memory leaks.

It should hopefully work well now!

PR #2086, opened 2021-11-13
Add plugin: Oof-Sound

MakingStan:
Makes an oof sound when you die or take damage.

Overlay(4K HD ULTRA ;))
afbeelding

2 Whole different oofs!
afbeelding


Nightfirecat:
The Minecraft "oof" sound is fine to use (though you need to [abide by the Creative Commons 3.0 license](https://creativecommons.org/licenses/by/3.0/) which you're currently not doing), but I don't believe the Roblox sound is copyright free/royalty free. Can you provide more clear information on that?

MakingStan:
Hey, sorry for the late response. The ReadMe should be updated (i'm now following the rules of the https://creativecommons.org/licenses/by/3.0/ license ). Sadly i had to remove the roblox sound. Because there is no way i could use that with proper rights (looked into that).

Nightfirecat:
You're still including the roblox sound in the resources directory--that must be removed, not simply left unused.

MakingStan:
Hey I’m sorry I removed it local but ofc GitHub doesn’t remove it if it’s still their. My apologies. Should be all set in a second.

Nightfirecat:
[This code will leak memory if errors are encountered.](https://github.com/MakingStan/Oof-Sound/blob/2fb642480a353be059b2158332bd4f53fc7d5901/src/main/java/com/code/OefPlugin.java#L98-L121) You should take reference from how clips are loaded in the Notifier in core for how you should do that.

Also, I still don't see any attribution for the minecraft oof sound. It is licensed under CC, so you must provide attribution in order to use it.

MakingStan:
Hey, i think i fixed the memory leak. 

I still wouldn't know why this:
Schermafbeelding 2021-11-29 om 21 28 30
wouldn't be enough for that license.

Nightfirecat:
This is still missing some code present in Notifier, namely the `clip.close();` code path. Reassigning `clip` without closing the previous handle may leak memory also.

MakingStan:
This:
Schermafbeelding 2021-11-30 om 16 16 16

should fix it.

Nightfirecat:
I think you forgot to push new code. I'm not sure what the clip position check in that image is meant to do, it looks like that could still lead to clips leaking memory. (If the condition is false, the code would go on to behave the same way it does now, unless there are other changes not shown)

MakingStan:
Hash should be updated. If the condition if false the clip isn't even open so I would not be able to close it then. (Sorry for the      inconvenience)

MakingStan:
Are my changes fine? Don't want to sound unpatient but we should probaly make an end with this PR haha.

PR #2104, opened 2021-11-17
Update effective-level to v1.1

XrioBtw:
Adds [invisible non-combat boosts](https://osrs.wiki/Temporary_skill_boost) such as celestial ring/signet (+4 mining), crystal saw (+3 construction) and skilling guilds (+7 fishing, mining, woodcutting).


Nightfirecat:
It looks like you have done this change wrong, unless you meant to create an additional config and plugin file at the root of your project.

XrioBtw:
Thank you for pointing that out. This has now been fixed and tested

PR #2106, opened 2021-11-19
Update gauntlet-minimap to v1.2.2

VicSegers:
Fixed a bug that caused resetting the collected resources when changing rooms.


VicSegers:
Fixed another bug that caused the Demiboss icons to not display.

PR #2111, opened 2021-11-23
Gachi Flip Flops

hectim:
Play a random Gachi sound when an achievement is reached. This includes levelling up, completing a diary, combat achievement, quests, dying


Nightfirecat:
Do you have the rights to use/distribute all these sounds? I know they're well-known memes at this point, but we need to be permitted to distribute them.

Nightfirecat:

PR #2112, opened 2021-11-23
Create ferocitylootlogger

gord5500:


gord5500:
We are doing a bingo event and using the plugin Adam made for sending web hooks to Discord. This is just an addition in that we can allow the user to enter a list of items to also send if they don't meet the GP threshold

gord5500:
Changed the plugins property in runelite-plugin.properties to 'com.ferocitylootlogger.FerocityLoggerPlugin' which should be the correct path. Thanks Nightfirecat for your help!

Nightfirecat:
[Your config group](https://github.com/gord5500/loot-logger-extended/blob/71f63f827861ff8168ffdb1a5dcad54a62ec0d1f/src/main/java/com/ferocitylootlogger/FerocityLoggerConfig.java#L10) should be unique to your plugin so it does not conflict with others.

Nightfirecat:

PR #2120, opened 2021-11-26
Added a Trouble Brewing plugin

Biffo89:
During a game of Trouble Brewing, the plugin displays how many pieces of eight the player is eligible for, from both the player's contribution (max. 100 points) and the team's success (2 per bottle of rum).

### DIFF
--- a/plugins/trouble-brewing
+++ b/plugins/trouble-brewing
@@ -1,0 +1,2 @@
+repository=https://github.com/Biffo89/Trouble-Brewing.git
+commit=8487a46bbde35a47e6d13543d3943f864f85c775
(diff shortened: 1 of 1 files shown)


Nightfirecat:
[You register an overlay in `startUp()` here](https://github.com/Biffo89/Trouble-Brewing/blob/8487a46bbde35a47e6d13543d3943f864f85c775/src/main/java/com/biffo89/troublebrewing/TroubleBrewingPlugin.java#L44), but don't ever remove it. It must be removed in `shutDown()`.

Biffo89:
Sweet, I've made that changed now.

PR #2132, opened 2021-11-29
add prifddinas plugin

DavidVentura:
Removes some clutter from prif


Nightfirecat:
I think [this line may be incorrect,](https://github.com/DavidVentura/runelite-plugins/blob/d4471fead262c0256e3594e36dd9012e1a07bea2/src/main/java/dev/davidv/prifddinas/PrifddinasPlugin.java#L33) as that object is named "Flax". LGTM otherwise

DavidVentura:
I've updated the plugin removing that line -- I wanted to remove some ground decoration and got the object ID instead. In the end I removed the code to remove the ground decorations (rocks, grass, etc) as the perf didn't change at all and it has a significant visual impact

PR #2139, opened 2021-12-01
new plugin - maple XP bar

SirHitech:
Maple XP Bar adds a skill experience progress bar over the chat box for a chosen skill


Adam-:
You need to add your plugin to the "plugins" folder, like the others.

PR #2159, opened 2021-12-08
Barb Assault Personal Bests Update to 1.3.0

SkylerPIlot:
Adds saving role-specific pbs to !ba r by tracking the role in the 0.XX part of the double
also add.csv tracking of the roles of everyone on the team and the overall time of the run in a .csv file


abextm:
any files you write should be in .runelite with a name that matches your plugin

PR #2164, opened 2021-12-10
Update Inventory Setups to use new MenuEntry API

dillydill123:
Uses new MenuEntry API


dillydill123:
I refactored to use the onClick callback and fixed-up my commits so it's just one commit on this PR.

PR #2172, opened 2021-12-11
Update sprite-markers to work with API changes

ConradMS:
Update to work with changes to MenuEntry


Adam-:
You need to rebase this to only be editing your plugin and not flipping utils.

PR #2209, opened 2021-12-18
Update rl-weather

bogstandard:
Mark Sound Effects as Beta and trial a fix to get them working cross-platform.

### DIFF
--- a/plugins/rl-weather
+++ b/plugins/rl-weather
@@ -1,2 +1,2 @@
 repository=https://github.com/bogstandard/rl-weather.git
-commit=29643ce14591d1e4465e03e041196d13b1f7f1bb
+commit=7cc366374c60177a98b0a00386d33a098d72c9a9


bogstandard:
Thanks for merging, apologies for reopening but I'm wondering if anyone can help.

This patch was an attempted fix to get the sounds playing. It hasn't worked. 
I've got the sounds in the Resources folder similar to other sound-playing plugins & fetch them as a resource stream like those too. But they only play when in development test mode when building from IntelliJ IDEA - not when installed via the Plugin Hub.

Is it possible I'm missing a crucial build step to get these sound resources to work in the production version?

Obviously I don't want to waste your time by making PR's for patches that don't work!

Adam-:
Your resource loading code looks okay to me other than it is leaking the input stream returned by `getResourceAsStream`. But that wouldn't cause it not play.

PR #2210, opened 2021-12-19
Notification when you are close to level

Vanillj:
It was a request here: [https://github.com/runelite/runelite/issues/14442](https://github.com/runelite/runelite/issues/14442)

This plugin will help users from missing when they reach a level and help prevent leveling by accident.

The plugin checks if the XP difference is less than some XP threshold the user can set (default to 5k). There is also a delay between the notification to prevent it from spamming each xp drop (defaults to 5 min).


plugins/Xp-To-Level-Notification
```diff
@@ -0,0 +1,2 @@
+repository=https://github.com/Vanillj/XpToLevelNotification
```
Nightfirecat, line 1: The repository URL is malformed
```suggestion
repository=https://github.com/Vanillj/XpToLevelNotification.git
```

Nightfirecat:
Additionally, plugin file names must be lowercase alphanumeric + dashes. try: "xp-to-level-notification"

plugins/Xp-To-Level-Notification
Vanillj: I appreciate the correction.

Vanillj:
Overlooked that, it should work now.

PR #2216, opened 2021-12-20
Add Spotify plugin

Enovale:
Please let me know if I misunderstood something or this plugin breaks a rule.


LlemonDuck:
Your commit reference points to a non-existent commit. Did you forget to push in your plugin repo?

Enovale:
Oh, so it is. I thought I made it public quite a while ago and never noticed. Fixed.

Enovale:
I was reading the other PRs here and noticed a comment about needing to be compatible with JDK 8 and I'm feeling stupid. I've made a branch that should be compatible, let me know if I should merge it.

Enovale:
My god the dependency isn't compatible with Java 8... Okay well I might need to rework this

Adam-:
If at all possible, try to minimize the deps you are adding (reuse existing deps we have or find libs which have fewer deps) - since we have to audit all of it its rather time consuming on our side.

Also we had some discussion before about allowing plugins to mark themselves as requiring 11 instead of 8, and then making it so if you do try to install the plugin and you cannot it will prompt you to update. Though we don't have any work on it currently if someone wanted to do that I wouldn't be opposed to it.

Enovale:
Im gonna look into forking the official API library and backporting it to JDK 8. I may also try and implement my own library referencing another Java application ive seen that does this, but this is a lot of work that I'm not sure I want to put in.

Enovale:
Hopefully this should fix it, I noticed the Java 9 breaking change was done semi recently and the older version still works so I've downgraded the project to use that version...

Nightfirecat:

PR #2224, opened 2021-12-22
Add Hidden Drop Party plugin

Krazune:
Hides drops. Useful for streamers that want to make drop parties on stream but don't want to show where the good drops were placed.

Demo


Krazune:
A lot has changed in the plugin, mostly the code structure, but it should be in a stable state now, and ready to be added to the client.

PR #2233, opened 2021-12-24
update dps-calculator to v1.4.11

LlemonDuck:
Adds scythe multi-hit dps calculation


LlemonDuck:
Updated to also include revenant ether wildy weapon bonuses

PR #2248, opened 2021-12-31
Create bank-equipment-stat-filter

adamgiles1:
Allows you to use a panel to view equipment in your bank filtered by slot/stat, so you can find out what your best gear for certain scenarios is quickly.


adamgiles1:
I believe I've fixed the issue that caused the build to fail. Let me know if there's a way to test the build process on my own, so I don't have to blindly make changes and hope it fixes the issue.

Nightfirecat:
[The nav button you add on plugin startup](https://github.com/adamgiles1/bank-equipment-stat-filter/blob/0c8c8b5f665d63be89c0dfa8f655f4123503e907/src/main/java/bankequipmentstatfilter/BankEquipmentStatFilterPlugin.java#L96) is never removed. Please ensure it is removed on plugin shutdown.
LGTM otherwise

adamgiles1:
I've updated the plugin to remove the nav button on shutdown.

PR #2253, opened 2022-01-02
Add Loot Lookup Plugin

donth77:
Add plugin to quickly reference the monster drop tables from OSRS Wiki in the right sidebar panel

[README](https://github.com/donth77/loot-lookup-plugin/blob/master/README.md)

Features
- "Lookup Drops" menu option
- Search bar for monster names
- Collapsible sections for each drop table
- Collapse all / Expand all button
- Display rarity as percentage
- Button to open the OSRS Wiki page for the monster in default web browser
- List and Grid views
- Links to wiki pages for items in List view
- Customize colors for rarity and prices

### DIFF
--- a/package/verification-template/build.gradle
+++ b/package/verification-template/build.gradle
@@ -41,7 +41,7 @@ dependencies {
 		because "discord-rare-drop-notifier"
 	}
 	thirdParty("org.jsoup:jsoup:1.13.1") {
-		because "discord-rare-drop-notifier, drop-simulator, fire-beats, loot-table"
+		because "discord-rare-drop-notifier, drop-simulator, fire-beats, loot-table, loot-lookup"
 	}
 	thirdParty("com.miglayout:miglayout-swing:5.2") {
 		because "drop-simulator"
--- a/plugins/loot-lookup
+++ b/plugins/loot-lookup
@@ -1,2 +1,2 @@
 repository=https://github.com/donth77/loot-lookup-plugin.git
-commit=785e9e97c50700e431b2cbef0e6b2bb72b54186d
+commit=e5fdff4f97072a48b0be042367d0a381e5500bae
--- a/plugins/loot-lookup
+++ b/plugins/loot-lookup
@@ -1,2 +1,2 @@
 repository=https://github.com/donth77/loot-lookup-plugin.git
-commit=f36e577fbbffe4e459497d40c0dc3411d5397196
+commit=39e80920e12629e2c5b17c44282aac3f9046ff65
--- a/plugins/loot-lookup
+++ b/plugins/loot-lookup
@@ -1,2 +1,2 @@
 repository=https://github.com/donth77/loot-lookup-plugin.git
-commit=39e80920e12629e2c5b17c44282aac3f9046ff65
+commit=785e9e97c50700e431b2cbef0e6b2bb72b54186d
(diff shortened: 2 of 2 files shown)


Nightfirecat:
You will need to follow [the Third party dependencies section of the README](https://github.com/runelite/plugin-hub#third-party-dependencies) to add your new dependencies.

Nightfirecat:
This fails to build because it uses `String#strip()` which doesn't exist in Java 8, and hub plugins must compile on 8. You can instead use [`StringUtils#strip()`](https://commons.apache.org/proper/commons-lang/apidocs/org/apache/commons/lang3/StringUtils.html#strip-java.lang.String-) as that's available to you.

donth77:
I've removed some unnecessary 3rd party dependencies and added this plugin to the verification template for jsoup which parses the HTML of the OSRS Wiki pages.

> This fails to build because it uses `String#strip()` which doesn't exist in Java 8, and hub plugins must compile on 8. You can instead use [`StringUtils#strip()`](https://commons.apache.org/proper/commons-lang/apidocs/org/apache/commons/lang3/StringUtils.html#strip-java.lang.String-) as that's available to you.

Thanks for pointing that out. I'll make sure to get this updated for Java 8.

abextm:
the `compile` configuration is deprecated; you must use `implementation` instead.
you should use the client's existing OkHttpClient rather than creating your own or using HttpUrlConnection. Also you should set your user agent to include your plugin's name in this case (`RuneLite.USER_AGENT + " (loot-lookup)"` or something).
your config group shouldn't have spaces
you need to always check the menu Type before reading data such as the identifier out of it

donth77:
@abextm Thanks. I've pushed a [commit ](https://github.com/donth77/loot-lookup-plugin/commit/e5fdff4f97072a48b0be042367d0a381e5500bae)that addresses these changes.

- `implementation` instead of `compile` for jsoup. [link](https://github.com/donth77/loot-lookup-plugin/commit/e5fdff4f97072a48b0be042367d0a381e5500bae#diff-49a96e7eea8a94af862798a45174e6ac43eb4f8b4bd40759b5da63ba31ec3ef7L18)
- Replace `new OkHttpClient()` with `RuneLiteAPI.CLIENT`. [link](https://github.com/donth77/loot-lookup-plugin/commit/e5fdff4f97072a48b0be042367d0a381e5500bae#diff-67fa66a003ea2bbb210b8d94a874c93846ca9bc7e81b8c08da7623b042bb13ecL20)
- Update constant for user agent to include `RuneLite.USER_AGENT and the plugin name`. [link](https://github.com/donth77/loot-lookup-plugin/commit/e5fdff4f97072a48b0be042367d0a381e5500bae#diff-07b72feb9649d25aa7835016c030b9f9d4d5ff9f9ea2a7a1ce42583cc241752eR13)
- Add new constant for config group and replace space with hyphen. [link](https://github.com/donth77/loot-lookup-plugin/commit/e5fdff4f97072a48b0be042367d0a381e5500bae#diff-07b72feb9649d25aa7835016c030b9f9d4d5ff9f9ea2a7a1ce42583cc241752eR11)
- Check that menu type is `MenuAction.EXAMINE_NPC` or `MenuAction.NPC_SECOND_OPTION` before reading the identifier. [link](https://github.com/donth77/loot-lookup-plugin/commit/e5fdff4f97072a48b0be042367d0a381e5500bae#diff-e37addb33b084599af06f9ca8f187542da8cf268727dea17596d33a7717eab5cR91)

donth77:
@Nightfirecat Thanks!

PR #2271, opened 2022-01-05
Nexvoice

gord5500:
This adds the voice dialog from the original Nex fight in RS2. These were taken from the cache and whats on the rs3 wiki and the code to play sound files was from m0bile btw's C Engineer plugin.


Hydrox6:
A plugin to do this has already been added to the hub, and we try to avoid having two plugins with no meaningful differences.

PR #2272, opened 2022-01-06
Bank Tag Layouts 1.4.12

geheur:


geheur:
Apparently this needs to be compared with two periods instead of three due to some gross commit history i.e. https://github.com/geheur/bank-tag-custom-layouts/compare/c404c59be403872e7411404938d8485ae8091ca4..geheur:cd841bda236e5e00802d1267dbda65c4e97c6eb7. ty keyosk

PR #2277, opened 2022-01-07
Add Herb Farming Calculator plugin

LucasPickering:
This plugin calculates potential profit and XP from farming herbs. It's basically a more powerful and easier-to-use version of [the Wiki's calculator](https://oldschool.runescape.wiki/w/Calculator:Farming/Herbs). The idea is that you configure the plugin once based on what patches and items you use, then any time you open the panel, it'll tell you what herb you should be farming based on current prices to make the most profit.

Here's an example of the output:

image

### DIFF
--- a/plugins/herb-farm-calculator
+++ b/plugins/herb-farm-calculator
@@ -1,0 +1,2 @@
+repository=https://github.com/LucasPickering/runelite-herb-farm-calculator.git
+commit=528d8e83b5dfc001ad986e0715353aed80055e07


abextm:
You should not set `enabledByDefault` in your PluginDescriptor; it confuses users.
You must remove your navbutton in `shutDown`
otherwise LGTM

LucasPickering:
> You should not set `enabledByDefault` in your PluginDescriptor; it confuses users. You must remove your navbutton in `shutDown` otherwise LGTM

Thanks for the review, just fixed both of those things so it should be good to go know.

PR #2283, opened 2022-01-08
Update Bot Detector to v1.3.7.2

Cyborger1:
- Cleaned up a deprecated RuneLiteAPI object.
- Added a `Request-Epoch` header field to all http requests to our API, which will serve as a sanity check on our end to ensure sighting data is correct for people with offset system clocks.


abextm:
you should `@Inject` the client's OkHttpClient and build yours off of that

Cyborger1:
I would try that, but it's a static class, I'm not sure what would be the best way to `@Inject` it in there. Any suggestions?

Cyborger1:
I've made the changes, is this correct?

PR #2288, opened 2022-01-11
Add the Item Share plugin

vyxyl:
This plugin gives players the ability to share items from their Equipment, Inventory, Bank, and GIM Storage to other players within their invite-only group

https://github.com/vyxyl/runelite-item-share

install
view-items

### DIFF
--- a/plugins/item-share
+++ b/plugins/item-share
@@ -1,3 +1,3 @@
 repository=https://github.com/vyxyl/runelite-item-share.git
-commit=3d7ab17f2b78784bb710bef7ecba40ed189ab589
+commit=8d30a648ab941da9a7b62484dab52993f3e3415e
 warning=In order to share your items with other players, this plugin sends player data (including IP address) to an open-source 3rd party server. The server is not controlled or verified by the RuneLite Developers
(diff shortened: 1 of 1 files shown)


vyxyl:
I almost forgot to add the 3rd party server warning message

The server code is here:
https://github.com/vyxyl/runelite-item-share-aws

Is there a way to give Runelite developers the ability to control + verify the server?
ex. Adding them as a user to AWS

abextm:
There isn't a way for us to verify what is running on the server without administrating it ourselves, which we will not do.

You are doing a lot of accesses to client data off the client thread(I think, your code is incredibly difficult to read) , which is generally invalid. You should develop with `-ea` added to your vm arguments to check for this behavior

vyxyl:
@abextm Thank you for your review!

I was able to verify that `-ea` was added to the vm options during development + data accesses were under the client thread
Just to be extra sure, I wrapped client calls with `ItemShareState.clientThread.invokeLater`
I also removed the unused assets + moved the readme screenshots out of the resources folder

Let me know if you would like any other changes!

abextm:
you need to remove your navbutton on 'shutDown`.

also I don't require it, but generally things like
```
logger.error("Failed to save player: " + e.getMessage());
e.printStackTrace();
```
should be
```
log.warn("Failed to save player", e);
```
since `printStackTrace` goes to the void, also you should prefer using format placeholders `{}` to concats in logs

vyxyl:
@abextm Thank you again for your review, let me know if there needs to be any other changes

Committing:
This repo has been deleted. The question mark link leads to 404 on the plugin hub.

PR #2290, opened 2022-01-11
Create SkullSceptre

FreckledDad:
Skull Sceptre plugin to allow users to see the charges of their skull sceptre.


abextm:
You should not check the `build` folder into git; you may want to learn how to operate a git client, such as the one included in IntelliJ.
You cannot use the same config group as an existing plugin, also your config key generally shouldn't include parenthesis. Tracking this seems simple enough it would be fine to be in the core item charges plugin.

FreckledDad:
> You should not check the `build` folder into git; you may want to learn how to operate a git client, such as the one included in IntelliJ. You cannot use the same config group as an existing plugin, also your config key generally shouldn't include parenthesis. Tracking this seems simple enough it would be fine to be in the core item charges plugin.

Should I add my code to that of the core item charges plugin and close this pull then? Or would another individual add this? (Not overly familiar with contributing)

PR #2297, opened 2022-01-13
Clue steps

SkyBouncer:
Displays the amount of steps completed on each clue.

Also, why does it show my full commit log when adding a new plugin?


abextm:
it shows the whole log because you branched off a branch with other commits instead of runelite:master

PR #2301, opened 2022-01-14
Create ticktracker

Tatters654:
Tick tracker is a plugin that calculates the time difference of game ticks, and then draws an overlay showing you if the server is good or not.


Tatters654:
Thanks to @m0bilebtw, the plugin has been overhauled to use nanoTime instead of system time. It was briefly discussed as a problem in the RuneLite discord, and now has been fixed.

abextm:
your config group should just be `ticktracker`, otherwise LGTM

Tatters654:
config group name was changed to "ticktracker"

PR #2308, opened 2022-01-15
Add plugin Watchdog

adamk33n3r:
FYI @Bullish53 gave me permission to use his service for the TTS as long as I was able to cache the results so that it's not overwhelming it. This was back in November so if he has an issue with it now, hopefully he can guide me into setting up my own service.

### DIFF
--- a/plugins/watchdog
+++ b/plugins/watchdog
@@ -1,3 +1,3 @@
 repository=https://github.com/adamk33n3r/runelite-watchdog.git
-commit=9e0787b0876d6d40a93b1673c97126ef383f60d5
+commit=35c256bf554e6b05236ebbc119941eccf33e020f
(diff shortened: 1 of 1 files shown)


adamk33n3r:
Oh I forgot, do I need to remove the shadow jar stuff in the build file for plugin hub?

adamk33n3r:
Oh I forgot, do I need to remove the shadow jar stuff in the build file for plugin hub?

abextm:
You need to close resources you open, like `Clip`s, File streams, http connections, etc. see try-with-resources statements. You need to remove your overlay in shutDown. You should not be calling setSelected on your panel in startUp

adamk33n3r:
@abextm thanks for taking a look! :D I have pushed up the changes you requested. Let me know if there's anything else!

PR #2311, opened 2022-01-16
Update Collection log plugin to v2.0

evansloan:
- Remove old configs that cluttered runelite config file
- Save collection log data to file on logout
- Add new directory for collection log data exports
- Add config to upload collection log data to collectionlog.net
- Remove buggy collection log updating on new drop

[collectionlog.net](https://github.com/evansloan/collectionlog.net) site repo
[collection log api](https://github.com/evansloan/collection-log-api) api repo


abextm:
We don't have a way to add warnings to to existing plugins without renaming the plugin. However since your network io is locked behind a default-off config option, you can add a warning to the config instead, and explicitly check it before doing any requests (ideally in `apiRequest`, since that makes it trivial for us to verify its working as intended)

evansloan:
Updated

PR #2314, opened 2022-01-17
Update gim-plugin to v2.0

davidvorona:


abextm:
Since you are now doing network io by default this plugin needs a warning, and it will need to be renamed so that existing users don't get opted in without accepting the warning.
Also, using the `compile` gradle configuration is deprecated (use `implementation` instead).

davidvorona:
Thanks for the comment!

Curious what you mean by warning and what I should rename it to? Do you have a plugin I can use as an example?

adamk33n3r:
@davidvorona something like this in the plugin file in this PR
```
warning=This plugin submits your IP address to a 3rd party website not controlled or verified by the RuneLite Developers.
```

davidvorona:
If I made the public server opt-in, would I need the warning / new name?

davidvorona:
Fixed the usage of the deprecated `compile` dependency configuration and removed the default public server. If users want to use my public server, they must manually copy the address from the GitHub README.

Will this suffice in place of renaming the plugin and adding a warning?

abextm:
Yeah, if the config doesn't have a default value and you have to enter the url manually that is fine to not have a warning.

you should use HiscoreManager rather than creating your own OkHttpClient & stuff. Also we recommend injecting Gson rather than creating your own, it protects you from some easy mistakes.

stuff like <https://github.com/davidvorona/gim-plugin/blob/221b47a83a5f2ad69ccfc3b338ed3f05918d5128/src/main/java/com/gimp/map/GimWorldMapPointManager.java#L87> should be debug or lower

davidvorona:
@abextm Thanks for the feedback. I'm fixing all this now. 

I'm actually a bit confused by your comments on injecting Gson, probably because I'm a Java noob. I [use the `new` operator](https://github.com/davidvorona/gim-plugin/blob/221b47a83a5f2ad69ccfc3b338ed3f05918d5128/src/main/java/com/gimp/GimPlugin.java#L326) with the [class](https://github.com/davidvorona/gim-plugin/blob/221b47a83a5f2ad69ccfc3b338ed3f05918d5128/src/main/java/com/gimp/GimBroadcastManager.java) I'd want to inject Gson into. Doesn't this mean I can't use the @Inject annotation?

abextm:
You can inject it into another class and pass it into your constructor. Its not specifically injecting Gson that I care about, its that you are using the Gson that we configure (and happens to be provided by injection)

davidvorona:
Okay, all mentioned issues should be fixed.

Gson is now injected properly: https://github.com/davidvorona/gim-plugin/commit/1c9d2fff5f9152baf7b7dc1a48652da02210dce4
All info-level logs are now debug-level: https://github.com/davidvorona/gim-plugin/commit/1c9d2fff5f9152baf7b7dc1a48652da02210dce4

Hiscores are fetched via an injected `HiscoreManager`: https://github.com/davidvorona/gim-plugin/commit/ab4efb024f03688f2b14a2e6629e506f9ff074c7

PR #2327, opened 2022-01-21
Combat Tasks & League tracker plugin

tylerthardy:
### Features

1. Tracks league & combat task progress from the UIs
2. Mark specific tasks for tracking/visibility
3. Automatic updating for task progress
4. Export button to export data into json (to be used with any website, but notably [osleague.tools](https://www.osleague.tools/) & [reldo.net](https://reldo.net))

### Coming Soon

Section mostly because I'm ashamed
1. Cleaning up the spaghetti & TODOs
2. Exporting varbits instead of scraping UIs - will reduce a lot of complexity w saving & tracking
3. UI improvements

preview

### DIFF
--- a/plugins/tasks-tracker
+++ b/plugins/tasks-tracker
@@ -1,2 +1,2 @@
 repository=https://github.com/osrs-reldo/tasks-tracker-plugin.git
-commit=5b100854f00c3f231d06f7cbb6411394646ba641
+commit=d1d9ccb82663f96e851954c1a14fb596493654da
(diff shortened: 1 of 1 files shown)


tylerthardy:
We have at least 120 users ready to use the export data functionality in this plugin, with a Discord of ~300. (osleague.tools)

Is there anything we can do to help speed up review on this PR? Or alternatives?

Proposals:
- Continue refactoring to slim down complexity if it's bottlenecking review
- Separate plugin to simply export League task data (dump varbit values, and we will process in our app; since Wikisync closed their API)

PR #2343, opened 2022-01-25
Bump WikiSync to 1.1.1

andmcadams:
- Attempt to re-store data on failed submission so it goes through the next time
- Check `onGameTick` for profile type change to fix a bug where hopping to a new world type could carry over info from previous world type
- Add calls to a version endpoint to check if the manifest needs to be updated (every 20 mins)
- Add icon


andmcadams:
- Closing for now because we decided to do more server load testing on this before deciding whether to merge

PR #2344, opened 2022-01-25
Create autumngarden

Ailie-Luibh:
My plugin is an adaptation of Majesty373's Spring Garden plugin. It helps with the Autumn Garden portion of Sorceress' Garden instead as the name implies.

### DIFF
--- a/plugins/autumngarden
+++ b/plugins/autumngarden
@@ -1,0 +1,2 @@
+repository=https://github.com/Ailie-Luibh/autumn-garden.git
+commit=af0d815105b71118db3b25f74bf4530cfacee174


abextm:
Have you considered pring this to the existing summer garden plugin?

pwatts6060:
> Have you considered pring this to the existing summer garden plugin?

I think users are fine with separate ones, the gardens are close together and share the same teleport point when you finish a round probably, people definitely don't want to have multiple gardens at the same time with one plugin. 

If it was merged, at least a combo box to choose which garden you are doing.

Ailie-Luibh:
So far the summer and spring garden addons are already separated from one another. If I were to merge all of them would I push it to the summer garden one, my own branch? Not sure how that would work.

pwatts6060:
Needed this for leagues and cloned it. Works well, should be merged now imo.

PR #2349, opened 2022-01-27
Create roidie-collection-log plugin

Mrnice98:
Create roidie-collection-log plugin, simple plugin just plays a sound bite when you get a collection log addition


abextm:
You need to close your Clips. Also do you have permission to use this audio?

Nightfirecat:

PR #2351, opened 2022-01-29
Update Barbarian Assault Personal Bests to 1.4.0

SkylerPIlot:
This update adds the ability to submit all times saved using the plugin as well as username and gamble count to an external database for use in highscores by using an jwt for authentication.


SkylerPIlot:
Let me know if you'd like to see the server's source, it is deployed on heroku with automatic git deployment so idm letting one of the rl devs have access to see the code

abextm:
We don't have a way to add warnings to to existing plugins without renaming the plugin. However since your network io is locked behind a default-off config option, you can add a warning to the config instead, and explicitly check it before doing any requests (ideally in `submit_pb`, since that makes it trivial for us to verify its working as intended)

SkylerPIlot:
Is there any example I can look at for implementing the warning.

Unless you mean a config option that's sole purpose is to have the warning and the user would have to check it on before any network requests are made

Hannah-GBS:
ConfigItems can have a warning field, as can be seen [in the GPU plugin option for compute shaders.](https://github.com/runelite/runelite/blob/9458bd8d656beae0309d0995c591575b9b60607c/runelite-client/src/main/java/net/runelite/client/plugins/gpu/GpuPluginConfig.java#L107)

SkylerPIlot:
Perfect, would you like the warning tag removed from the plug-in hub file or should I leave that for any future people to download the plug-in?

abextm:
as long as all network io checks the option the whole plugin doesn't need the warning

SkylerPIlot:
Sounds good i'll get the fix pushed when I get home

SkylerPIlot:
Added the warning in the config item `submit_pb`
Also removed the warning from the pluginhub file

PR #2358, opened 2022-01-31
Update Barbarian Assault pbs 1.4.1

SkylerPIlot:
Adds a small feature !ba gc adding 1 more network IO for alt gamble count support

New network IO is located in grabgc() and implemented in gcLookup() and is limited behind by a config with a relevant warning


Adam-:
The core already has a !gc command - are you aware youre duplicating that?

SkylerPIlot:
I don't plan on adding anything to this PR. Is there anything I need to do to fix this PR? Otherwise when will a review be expected? Anything would be greatly appreciated.

PR #2366, opened 2022-02-03
Add fixed-mode-toggle-chat

ariiZona:
This is actually a fork of https://github.com/deathbeam/example-plugin/tree/fixed-mode-hide-chat but since that repo was a bit messy I've decided to do a "v2". Viewport can now be resized to normal size if chat is open (configurable).

If I didn't do this the right way, please tell me, the other option I have is making a pull request on the orig repo.


Nightfirecat:
Generally we prefer PRing features to hub plugins rather than making a "v2" for small changes. Deathbeam is active and will review changes if PRed there I'm sure.

PR #2375, opened 2022-02-05
disable zalcano for breaking core functionality

Alexsuperfly:
the overlay names for this hub plugin and the core zalcano plugin are the same, so changes to overlay position effect both, breaking dynamic markings for the core overlay


Nightfirecat:
Is this needed now that runelite/runelite@2d52f6d1b4a0271c27ab8a52b029e7cf732c8341 is merged?

Nightfirecat:
Fixed in #2395

PR #2388, opened 2022-02-09
Camdozaal v1.1 - Migrate Camdozaal plugin from old repo to new one.

TicTac7x:


Hydrox6:
You should really just not change the plugin's name, leave it in the `camdozaal` file

TicTac7x:
> You should really just not change the plugin's name, leave it in the `camdozaal` file

I would like to go forward with the name change.

Adam-:
I too would prefer if it wasn't renamed. There is no reason to rename it and it just causes headaches

abextm:
I've decided to disallow renames like this since it is entirely pointless. Do not rename the file in this repo

TicTac7x:
As a sign of protest I have decided to request disabling this plugin

PR #2404, opened 2022-02-15
Update ba-minigame to v1.6

BegOsrs:
This version includes:
- Create widgets to display info instead of editing existing ones.
- Add info about role total/round/session points.
- Change wave/round summary and points message.
- Add sessions.
- Add broken trap highlights.
- Fix runner tick timer.
- Write a chat message if teammate corrects call.


BegOsrs:
Not planning on adding new commits on this pr.
Been a while without feedback, when can a review be expected? Would be appreciated :)

PR #2429, opened 2022-02-26
Add TwitchEmotesPlugin

Jlao17:
A plugin that replaces text in the chat with the corresponding emote.


hex-agon:
Please do not open multiple pull requests for the same plugin, you can just update your original PR and the bot will refresh the changes. Also it is worth nothing that a plugin like this cannot be accepted into the hub as you do not own the copyrights to the emotes and twitch's license does not permit use outsite of it's site.

Alexsuperfly:
this is going to have the same issues as #43, #1158, and #1946

Jlao17:
Yep, already thought about copyright issues. I'll put this one on the back burner.

PR #2436, opened 2022-03-02
Add unidentified herbs plugin

hex-agon:


Nightfirecat:
You're missing a license. Code lgtm otherwise

hex-agon:
I've fixed the license and the broken plugin manifest

PR #2454, opened 2022-03-16
Remove Leagues III fragment presets plugin

SyntaxBlitz:
It looks like removing plugins isn't super common in this repository, but Leagues III is over, so we can clean this up. Let me know if there's anything else I should do to prep for the cleanup!


Alexsuperfly:
i believe its preferred to just set the disabled flag, or else someone could make another plugin with the same name and then everyone that had yours installed would automatically get theirs without them actually selecting to install it.

SyntaxBlitz:
Sounds good! I've seen both `disabled=true` and `disabled=unmaintained` in the repo. I went for `disabled=true` because `unmaintained` isn't quite the reason for disabling here. Thanks for the feedback!

PR #2465, opened 2022-03-21
Ba role rando v2

ransty:
Added UI for randomizing teams. 

Preferences can be selected via checkboxes.

BA Role Randomizer UI


Nightfirecat:
You need to remove your nav button from the client toolbar on plugin shutDown.

ransty:
> 

Ah okay will update it when I get a chance. Thanks.

ransty:
Have removed the navButton from the client toolbar on shutdown.

PR #2466, opened 2022-03-21
Update and re-enable custom-item-hovers

geel9:
- Update to latest RL version / dependencies
- Add chat command to open hover file directory
- Add variables and functions for more dynamic hover texts
- Add blockcomments to most functions
- Add real documentation for how to create custom item hovers

### DIFF
--- a/plugins/custom-item-hovers
+++ b/plugins/custom-item-hovers
@@ -1,3 +1,2 @@
 repository=https://github.com/geel9/runelite-custom-item-hovers.git
-commit=1d610586cea8876b83ec4eae32bfdb520e1874f5
-disabled=unmaintained
+commit=e90a5348633f02a590bf9273702d35ac7a86d6b8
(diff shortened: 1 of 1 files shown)


Nightfirecat:
Your item id to name mapping probably doesn't work, and we generally discourage reflection in hub plugins. You can instead copy the implementation from ChatboxItemSearch in the client: https://github.com/runelite/runelite/blob/2654975db7762e2676c9b83efc8ad5c13fd87a78/runelite-client/src/main/java/net/runelite/client/game/chatbox/ChatboxItemSearch.java#L296-L329

Adam-:
> Your item id to name mapping probably doesn't work, and we generally discourage reflection in hub plugins. You can instead copy the implementation from ChatboxItemSearch in the client: https://github.com/runelite/runelite/blob/2654975db7762e2676c9b83efc8ad5c13fd87a78/runelite-client/src/main/java/net/runelite/client/game/chatbox/ChatboxItemSearch.java#L296-L329

Specifically, the `ItemID` class is not available at runtime on the release builds, so you cannot reflect into it.

geel9:
Understood; I've opted to switch to using proper item names (eg, `Pineapple sapling`), which is both easier for users to understand and avoids usage of `ItemID` or reflection.

PR #2472, opened 2022-03-22
Clan activity tracker

Phobos97:
New plugin to track some clan data to csv file. Tracks when a player was last seen and how many messages player has send since start of tracking. 

This was mainly made for use in my own clan but im sure more clan leaders will get use out of this to track the activity of their members.

For an example see picture in the readme.


Nightfirecat:
FYI rather than [specifying the full path for your plugin's directory](https://github.com/Phobos97/clan-activity-tracker/blob/1c60ae25181fd9324905af4be52f6473b40883a7/src/main/java/com/clanactivitytracker/ClanActivityTrackerPlugin.java#L41), you can use `new File(RuneLite.RUNELITE_DIR, "clan-activity-tracker")`.

PR #2473, opened 2022-03-22
Add Jar Generator plugin

Jannyboy11:
Plugin features:
- Overlay displaying the number of charges in your impling jar generator.


Nightfirecat:
Your config group shouldn't have spaces

Jannyboy11:
Okay, I'll adjust that, but can I ask why not?

Nightfirecat:
I believe it's a safety measure to prevent possible errors with serialization. Thanks for making that change

PR #2486, opened 2022-03-25
Update Discord rare drop notificator to 1.4.0

MathieuDR:
* No more Osrsbox API
* Only api used is osrs wiki
* Sent to discord using user set endpoints
* **Removed warning,** _please advise_
* Cache a JSON on startup


MathieuDR:
I updated the version number to 1.4.0

PR #2491, opened 2022-03-27
Update discord-rare-drop-notificater to v1.4.1

MathieuDR:
Change on loading the json file


MathieuDR:
Is the json file in the Resource directory included when this gets compiled/shipped?

PR #2493, opened 2022-03-27
Add star calling assist

zodaz:
This plugin allows for crashed stars found around the game to be posted to a remote endpont when found.


Nightfirecat:
Use the client's instance of `OkHttpClient` instead of creating your own OkHttpClient here: https://github.com/zodaz/star-calling-assist/blob/202bcf484cc3b0d8f1062da173924a03ba1141c3/src/main/java/com/starcallingassist/CallSender.java#L35-L39

zodaz:
> Use the client's instance of `OkHttpClient` instead of creating your own OkHttpClient here: https://github.com/zodaz/star-calling-assist/blob/202bcf484cc3b0d8f1062da173924a03ba1141c3/src/main/java/com/starcallingassist/CallSender.java#L35-L39

Done!

PR #2495, opened 2022-03-28
Update ToG-Crowdsourcing - Add Get Error Message to UI

jcarbelbide:
Also update README to mention the website for non-runelite and mobile users.


jcarbelbide:
That was fast, thanks!

PR #2498, opened 2022-03-29
Add TheleRune Updater plugin

Theleruby:
This adds a plugin which I made to automatically update the stats tracker on my website on logout.

I don't expect it to be useful to anyone except me, but I wasn't able to figure out how to install the plugin without using the plugin hub (if that's even possible) so it seemed better to me just to do it this way and install it from the plugin hub afterwards.

### DIFF
--- a/plugins/thelerune-updater
+++ b/plugins/thelerune-updater
@@ -1,0 +1,2 @@
+repository=https://github.com/Theleruby/thelerune-runelite-plugin.git
+commit=5007eb90924ba8d687b7a8508550b9c86d394518


Adam-:
FYI if you have your own client build **running in --developer-mode** you can place plugins in the `sideloaded-plugins` directory.

Theleruby:
Hey, thanks, that's useful to know. Maybe this should be added to the documentation page explaining how to make plugins?

I think it would still be preferable for me to submit this to the plugin hub though (if I'm allowed)

Adam-:
Yes. It's fine to submit it if you want.

Theleruby:
Yes please :)

Nightfirecat:
I understand this plugin is intended for your use only, but because it submits web requests, you need to add a `warning` field, such as this: https://github.com/runelite/plugin-hub/blob/33de167aea1acba28ec23e2d60678ea7593bb191/plugins/f2pwiki#L3

Theleruby:
No worries, I have added the warning you requested.

PR #2503, opened 2022-03-31
add riftguardian plugin

rdutta:
Highlighting/overlays for Guardians of the Rift minigame.


rdutta:
> Don't clear the hint arrow if you're not in the minigame area: https://github.com/rdutta/runelite-riftguardian/blob/173144ec4a09faf10cdc1375a435daf9e4851f0c/src/main/java/ca/riftguardian/RiftGuardianPlugin.java#L145 (other plugins, and the client itself, use the hint arrow)

Updated. It will now check if the portal gameobject is present before remove the hint arrow.

PR #2506, opened 2022-03-31
Add Left Click Banking plugin

wizguin:
Same functionality as the built in left click menu entry customization, but in the bank interfaces.


Nightfirecat:
This is basically the same as #1883, which we decided we do not want to add.

PR #2521, opened 2022-04-03
Polywoof Plugin

furfy:
Hi, I made plugin to help my friend understand story in quests. Polywoof stands for «polyglot».
So what it does is just «subtitles» plugin that will Google-translate every NPC and non-NPC text, not players tho..

You can choose destination language, so it will be useful to anyone who has problems with english.

I'm very unexperienced Java-programmer, so I'm very-very sorry for stupid stuff..

### DIFF
--- a/plugins/polywoof
+++ b/plugins/polywoof
@@ -1,0 +1,2 @@
+repository=https://github.com/furfy/Polywoof.git
+commit=f2d769b6779d35336376d617d7671a51720eb5da
--- a/plugins/polywoof
+++ b/plugins/polywoof
@@ -1,2 +1,2 @@
 repository=https://github.com/furfy/Polywoof.git
-commit=f2d769b6779d35336376d617d7671a51720eb5da
+commit=92b36e3c6642c8275c91748f9f3793d3db4c5785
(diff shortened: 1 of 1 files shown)


Hydrox6:
A lot of people have thought about using Google translate for translator plugins in the past, and last we checked, a plugin using Google translate for something like this would be against their Terms of Service. Others have instead required users to provide an API key for their own Google Cloud Platform account (see https://github.com/simbleau/chat-translator).

Either way, given that this plugin only translates NPC text, it doesn't really make sense to make requests out to a translator service; the text is always going to be the same, so a prospective plugin could pre-translate all the text in the game and serve that as needed. Just a tip if you want to have a rethink alongside fixing the ToS violation.

furfy:
It does have temporary cache for translations, but I'm not sure how to store translations locally.
Anyway, I will change translation method to token-only way.

Nightfirecat:
Seems DeepL's ToS is pretty flexible, so that service should be fine to use. (and apparently fine to store the responses if needed) I'll review this tonight and let you know if anything else needs changing.

Nightfirecat:
[`assert`](https://github.com/furfy/Polywoof/blob/b7da455ab717109022a4005452cbcc89f17cdb99/src/main/java/com/furfy/PolywoofThread.java#L56) does not work on release as assertions are disabled (and will cause a hang when `-ea` is enabled). I recommend using a guard clause instead.

Besides that, LGTM

PR #2524, opened 2022-04-04
Add Chop Chop Stop plugin

call-me-maple:
This plugin will hide the overhead messages from the various tool special attacks.


Nightfirecat:
fyi you should be able to use the Chat Filter plugin to accomplish this

call-me-maple:
I plan on adding the option to hide the animations too.

PR #2527, opened 2022-04-05
Create Entrana Plugin

pwatts6060:
Plugin that highlights items that are not allowed on entrana when you are on the entrana dock in port sarim.


Nightfirecat:
Your plugin's overlay must be removed on plugin `shutDown()`

PR #2538, opened 2022-04-08
Update Collection log plugin to v2.1

evansloan:
* Add side panel with various features/utilities:
    * Add links to relevant websites (collectionlog.net/GitHub/Discord)
    * Display collection log entries that have not been loaded
    * Add buttons to recalculate totals, upload collection log data (if config is enabled), and reset collection log data
* Prevent collection log data from being overwritten when viewing another player's log when viewed from the adventure log in the POH

### DIFF
--- a/plugins/collection-log
+++ b/plugins/collection-log
@@ -1,2 +1,2 @@
 repository=https://github.com/evansloan/collection-log.git
-commit=f19a76c9aa0e42694e7ba77005a27f38042ba678
+commit=f8d05631785ed61750f131f3e94b0cc99b9361f4
--- a/plugins/collection-log
+++ b/plugins/collection-log
@@ -1,2 +1,2 @@
 repository=https://github.com/evansloan/collection-log.git
-commit=52171d9bf6a55d722a861fef94b3140cc003144b
+commit=f19a76c9aa0e42694e7ba77005a27f38042ba678


Nightfirecat:
[The user agent you use should include your plugin name.](https://github.com/evansloan/collection-log/compare/52171d9bf6a55d722a861fef94b3140cc003144b...evansloan:f19a76c9aa0e42694e7ba77005a27f38042ba678#diff-313bfb4dbe387e5414a30dd8c53e00d273620219346a88bae1d486cb5f3e0a89R216)
[The "Open collectionlog.net" button](https://github.com/evansloan/collection-log/compare/52171d9bf6a55d722a861fef94b3140cc003144b...evansloan:f19a76c9aa0e42694e7ba77005a27f38042ba678#diff-459be4208d9630c3572c366d6ee83e66c641bdd15ce2c071a5f4bdf0cf274f74R76) should create some kind of warning regarding sending IP addresses to a server not operated or controlled by RuneLite. Discord/GitHub are generally trusted services by users so it is not needed for those.

evansloan:
Right, forgot about that link. Would it be okay to only show that button if the uploadCollectionLog config is enabled since it shows that warning message when it is enabled? I'd rather do that than have a dialog box or something pop up every time the button is clicked.

Nightfirecat:
Yes that would be fine--just be sure the panel is properly rebuilt if the config is toggled so it is not showing (or absent) when switching between the panels.

abextm:
The button is fine as-is, since its obvious it will take you to another site

PR #2540, opened 2022-04-08
Update copy-clan-usernames

solowanna:
Changed my username in Github, so this has to be changed aswell.


Nightfirecat:
There's no change needed as GitHub redirects requests to your new username. If you have plugin updates, go ahead and update the username at the same time once you submit those updates.

PR #2549, opened 2022-04-13
Update Inventory  Setups to no longer use deprecated varbit API

dillydill123:
Updated to use newer varbit API


dillydill123:
Should be merged after 1.8.17 is released.

PR #2551, opened 2022-04-13
Update lootbag-utilities to use new varbit api

LootBagger:
Update varbit api
Also fix a few bugs where item variants options were not being changed

hold for release, but I was told release was today 👀


Adam-:
You have to use the lombok version specifically from https://github.com/runelite/example-plugin/blob/master/build.gradle

PR #2557, opened 2022-04-14
Create Bones

BenTapp753:
My plugin notifies the user when bones pop up on the ground. It notifies them both audibly and with text.


Hydrox6:
Closing since this seems abandoned.
For anyone who was interested in using this plugin, the same effect can be achieved entirely with the core `Ground Items` plugin, by adding bones to your manual highlight list and enabling notifications for highlighted items.

PR #2559, opened 2022-04-15
Change ownership and rewrite slash-swapper plugin

pwatts6060:
@nganzalo is no longer maintaining this plugin and has agreed to let me take ownership and fix it.


nganzalo:
I have not been maintaining, @pwatts6060 can take ownership of plug-in 👍

PR #2561, opened 2022-04-16
ComboRuneHelper

Vesticle:
New Combo Rune Helper plugin

### DIFF
--- a/plugins/combo-rune-helper
+++ b/plugins/combo-rune-helper
@@ -1,0 +1,2 @@
+repository=https://github.com/Vesticle/combo-rune-helper.git
+commit=02666b0e847da27bf46a72ba10b57a745c060226


LlemonDuck:
This is covered for the fire altar already in https://github.com/LlemonDuck/combo-runes-only, if you would like to PR an expansion to cover all altars, I would be more than happy to accept it in order to prevent duplicate plugins.

Vesticle:
> This is covered for the fire altar already in https://github.com/LlemonDuck/combo-runes-only, if you would like to PR an expansion to cover all altars, I would be more than happy to accept it in order to prevent duplicate plugins.

Ya I tried to message you about doing that but I don't know how to do that on git. All I did was apply the same code to all the elemental altars, so it's easier to craft combo runes in the new runecrafting minigame. I'd be happy to combine them or what have you.

LlemonDuck:
GitHub doesn't provide user messaging, but you can open issues within the repository to discuss changes. I've created an issue for this at LlemonDuck/combo-runes-only#1, and included a change to the plugin there which includes this and more. Take a look and let me know if that works for you, then I'll submit it here as an update to the existing plugin.

Vesticle:
Your changes seem great so I'll close this request

PR #2574, opened 2022-04-20
added plugin that handles tracking skills on logout/closing runelite …

Jmaasy:
Added new plugin which will automatically track your progress on either a logout or when you close runelite.

It has the option to track via crystalmathlabs and/or wise old man. It uses the account that is currently logged in.

### DIFF
--- a/plugins/skill-tracker
+++ b/plugins/skill-tracker
@@ -1,2 +1,2 @@
 repository=https://github.com/Jmaasy/skill-trancker.git
-commit=cff6e66b4c97be27e33eaa18d3a680ab91da1b5c
+commit=07a0d328908a8b9c404b1adf2b2c0835e9625704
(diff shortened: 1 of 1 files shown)


Hydrox6:
This is already functionality provided by the core plugin `XP Updater`.

Jmaasy:
Still a bit different, when I looked up XP Updated it showed me the F2P WIki plugin. When looking at F2P Wiki it's different from WOM and CML. This plugin also allows you to select which one you would report to ( or both ). Besides that this also actively updates when you close the client which that one won't do as well.

Hydrox6:
> Still a bit different, when I looked up XP Updated it showed me the F2P WIki plugin. When looking at F2P Wiki it's different from WOM and CML. This plugin also allows you to select which one you would report to ( or both ). Besides that this also actively updates when you close the client which that one won't do as well.

The XP Updater plugin does *all* of this. That's the entire point of the plugin.
image

Jmaasy:
Oh nevermind I'm blind -.-

PR #2578, opened 2022-04-21
Action progress 1.03

CalebWhiting:
1.03
* Rewritten a lot of core components.
* Many actions added/
* Many ingredient checks added, giving more accurate estimations of how many actions can be performed.
* Many customizations added, all actions can be enabled or disabled.
* Option to show skill icons instead of product icons (For example, Herblore icon instead of the potion being made).
* Updated for the latest RuneLite API changes


CalebWhiting:
Still new to this build system, sorry about the spam, trying to fix dependency verification.

CalebWhiting:
Having trouble getting org.jetbrains:annotations verified.
Is this something that I have the ability to fix, or would someone need to whitelist the library?

Adam-:
I've added it to the verification metadata

CalebWhiting:
> I've added it to the verification metadata

Thank you very much.
That one had me stumped :')

PR #2583, opened 2022-04-21
Create Star Info Plugin

pwatts6060:


pwatts6060:
I may close this PR and we might replace the Star Tier Indicator plugin. Coordinating with the owner of the other plugin and we'll see after he looks over the changes to it.

PR #2585, opened 2022-04-22
Update bank-tag-generation

CalebWhiting:
Fixed drop tags, code cleanup, test units, arguments with spaces (https://github.com/MitchBarnett/wiki-bank-tag-integration/pull/12)
* Fixed drop tags, cleaned up code, fixed various warnings and implemented test units.
* Removed the limitation of requiring underscores in the command arguments. Which also expands the possibilities of queries.

### DIFF
--- a/plugins/bank-tag-generation
+++ b/plugins/bank-tag-generation
@@ -1,2 +1,2 @@
 repository=https://github.com/MitchBarnett/wiki-bank-tag-integration.git
-commit=233f143f9994558bb44696411eccc6fd74d92d88
+commit=491d7054b3dfb34f66e7c30ed66712a973086128


Nightfirecat:
@MitchBarnett Are you fine with this update being released?

CalebWhiting:
Unfortunately @MitchBarnett is inactive, hence why I submitted the plugin instead.
He does pop up sometimes, hopefully he will soon.

Ben10164:
@CalebWhiting Since this is a very prominent feature of the plugin, as well as a lot of the code being redone by you, would it be allowed to reupload a "new" plug-in that is just "Better Bank Tag Generation"? It seems the original creator is inactive.

MESLewis:
Do not submit as another plugin because of an inactive author. We will transfer ownership to a new maintainer if the original does not respond.

CalebWhiting:
> @CalebWhiting Since this is a very prominent feature of the plugin, as well as a lot of the code being redone by you, would it be allowed to reupload a "new" plug-in that is just "Better Bank Tag Generation"? It seems the original creator is inactive.

I don't think that would be allowed by RuneLite, they're very strict about forks and requiring permission of the original developer. Besides, I'm not active right now.

PR #2587, opened 2022-04-22
Update PM Highlight to latest version

wtommyw:
Update to latest Runelite version


Hydrox6:
The version in the `build.gradle` is only for local testing; you don't need to make a PR unless you actually change code

PR #2589, opened 2022-04-23
Add Group Ironmen Tracker plugin

christoabrown:
Sends character information for group ironmen where other members can view. Groups will need to go to the site to create a new group and are provided with credentials for login and sending the data.

[groupiron.men](https://groupiron.men/)  
Demo of what it looks like logged in is also available: [Demo](https://groupiron.men/demo)

This version sends the following information, all of it is encrypted before being stored:
* Inventory, bank, equipment, and shared bank
* HP, prayer, energy, and world
* World position
* Skill XP
* Quest states  

Source code for the server and frontend is here:
[https://github.com/christoabrown/group-ironmen](https://github.com/christoabrown/group-ironmen)

It seems somewhat similar to the other tracker plugins, although I believe having a public server with a DB and login credentials sets it apart. I do also intend to add more features like more inventories, achievement diaries, etc.

### DIFF
--- a/plugins/group-ironmen-tracker
+++ b/plugins/group-ironmen-tracker
@@ -1,3 +1,3 @@
 repository=https://github.com/christoabrown/group-ironmen-tracker.git
-commit=027bb16b756fd70862504bbf29e3d3a85e83a94b
+commit=f61c6aa5454616145ea2796e028e90a71cc48fbb
 warning=This plugin pushes data about your character to an external server.<br>This data is encrypted and is only viewable by people you share the credentials with.
--- a/plugins/group-ironmen-tracker
+++ b/plugins/group-ironmen-tracker
@@ -1,2 +1,2 @@
 repository=https://github.com/christoabrown/group-ironmen-tracker.git
-commit=d8cdd5ce253cbfdbfa2da715cb06814221d7a256
+commit=3bbfeaee11a00b63db65ffa6facd7315ed16a8c0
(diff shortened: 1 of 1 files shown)


Nightfirecat:
As far as I'm concerned, the `warning` you've added isn't needed--no requests are sent without making configuration changes.

You should set a unique user agent for this plugin for requests it submits.

christoabrown:
I removed the warning and included the custom user agent.

Nightfirecat:
You don't need to make a whole new Okhttp client, just add `.header("User-Agent", USER_AGENT)` in your request builders.

christoabrown:
That's what I tried at first, but wherever this client is being built it is adding its own user-agent header before sending the request. I am not super familiar with the runelite code, but I am guessing it is here the interceptor is added which overrides the header:
https://github.com/runelite/runelite/blob/master/runelite-client/src/main/java/net/runelite/client/RuneLite.java#L439

Adam-:
> That's what I tried at first, but wherever this client is being built it is adding its own user-agent header before sending the request. I am not super familiar with the runelite code, but I am guessing it is here the interceptor is added which overrides the header: https://github.com/runelite/runelite/blob/master/runelite-client/src/main/java/net/runelite/client/RuneLite.java#L439

This is a bug which I've fixed https://github.com/runelite/runelite/commit/c40b33a1e988b344f3e59ff2659cf26cfb752422

christoabrown:
Thanks for the fix. I changed it so it now uses @Nightfirecat's suggestion with setHeader which I expect will start working in a future runelite release.

PR #2591, opened 2022-04-23
Replace Startierindicator with star-info plugin / change ownership

pwatts6060:
Star Info is a very large enhancement of the Star tier indicator plugin. I've coordinated with the original author @zodaz to replace the plugin with my version and he should comment his authorization below soon if all goes well.


zodaz:
Confirming ownership change

PR #2592, opened 2022-04-23
Action progress 1.03.1

CalebWhiting:
I made an oopsie.

Was in a rush to push the latest version due to API changes, and didn't test some changes I made before.

Fixed ItemClickDetector


CalebWhiting:
Replaced query API with constants.
Didn't know database classes didn't exist outside of the testing environment (ItemID, NpcID, AnimationID etc)

PR #2594, opened 2022-04-23
update imbued-heart-notifier

JuliusAF:
Fixed heart highlighting not being removed when it's on cooldown, and stopped notifier from notifying if heart isn't in the inventory


JuliusAF:
Wait new plugin? I think I might have done something wrong. This should be an update to an existing one.

Bram91:
You need to rebase your branch

PR #2596, opened 2022-04-23
Slayer Assistant Plugin

LeeOkeefe:
Plugin for searching for slayer tasks and showing information related to each task.

- Monster locations
- Items required/recommended
- Monster attack styles and attributes
- Slayer masters that can assign the task


Nightfirecat:
The "slayer" config group is already in use by the core slayer plugin, use a unique string here. Alternatively, given your plugin doesn't use any configuration, you could simply remove the config class entirely.

LeeOkeefe:
> The "slayer" config group is already in use by the core slayer plugin, use a unique string here. Alternatively, given your plugin doesn't use any configuration, you could simply remove the config class entirely.

Oops! I've removed the config file and updated the commit hash.

PR #2598, opened 2022-04-24
Add Time Tracking Reminder plugin

queicherius:
This plugin extends the built-in "Time Tracking" plugin to show an infobox when bird houses or farming patches are ready.

java_2022-04-24_00-52-25

Currently supports bird houses, herb patches, tree patches and fruit tree patches. Each infobox can be individually enabled and disabled.

### DIFF
--- a/plugins/time-tracking-reminder
+++ b/plugins/time-tracking-reminder
@@ -1,2 +1,2 @@
 repository=https://github.com/queicherius/runelite-time-tracking-reminder.git
-commit=72ab261f0dcbd5e746887c456647de7f8644e7c4
+commit=5016e04b7bed845a7764ad50c724978dfe5e5ed0
--- a/plugins/time-tracking-reminder
+++ b/plugins/time-tracking-reminder
@@ -1,0 +1,2 @@
+repository=https://github.com/queicherius/runelite-time-tracking-reminder.git
+commit=72ab261f0dcbd5e746887c456647de7f8644e7c4


Nightfirecat:
Do not access the time tracking plugin via reflection. You should be able to get the data you need by reading the time tracking data out of config instead.

queicherius:
> Do not access the time tracking plugin via reflection. You should be able to get the data you need by reading the time tracking data out of config instead.

This would mean quite a bit of duplicated parsing logic in the plugin, which I was trying to avoid. But if you'd be more comfortable with that, I can change it 👍 Essentially I would copy/paste the `loadFromConfig`/`loadCompletionTimes` and `updateCompletionTime` methods into the plugin.

Would the `ConfigChanged` event be the preferred method of listening to updates from the `TimeTrackingPlugin` or is there any better way?

Nightfirecat:
ConfigChanged is the right way to listen for updates to those keys, yes.

queicherius:
I removed reflection from the plugin and added the files for configuration parsing.

I opted to copy over the files from the `runelite/runelite` repository largely unmodified and only lightly patch them to make them usable. That should make the review easier (since you can diff the directory against a known good version in the client repo), and makes it a lot less of a hassle for me to maintain long-term.

I recognize that it's a lot of code, but when I removed the code I don't need (i.e. the `updateData` functions) it didn't change all that much about the total size, and made diffing and maintenance way harder.

In my opinion reflection would still be cleaner than this, but I understand if that's something that is scary to have in a plugin.

PR #2605, opened 2022-04-26
Post inventory rework update for herblorerecipes

skiclimbcode:


Adam-:
The newest lombok version we have verification metadata for is 1.18.20, so ideally just use that version unless you have some specific reason you require a newer version

PR #2610, opened 2022-04-26
Update Collection log plugin to v2.2

evansloan:
Contribution from @call-me-maple

## Add !log command
This adds a chat command !log PAGE_NAME. When the command is found the collection log data for the message sender is loaded from the collectionlog.net API and deserialized into class instances. Then the needed sprites are loaded into the client and the chat message is replaced with collection log data.


Nightfirecat:
You need to unregister the added chat command on plugin `shutDown()`.

PR #2619, opened 2022-04-27
Disable action-progress

CalebWhiting:
Temporarily disable the plugin until such a time as the update can be approved.


Nightfirecat:
I merged #2608, so this shouldn't be needed anymore.

PR #2648, opened 2022-05-03
Create custom-xp-drops

TylerHarding:
This plugin overrides the default OSRS XP Drops, and gives the player more customization options.

Features:
-   More customizable font options including - Font Name, size, color
-   Custom move speeds and fade out speed, can turn fade out off
-   Removal of default XP Tracker, replaced with a Custom XP Tracker dimension, allowing for it to be moved more freely, as well as greater control over the size of the XP Tracker
-   Addition of custom prefixes to the xp drop


raiyni:
How is this different from `customizable-xp-drops`

PR #2651, opened 2022-05-04
Add my lizardman shaman minion spawn sound plugin

LillieH001:
This plugin makes it where a sound plays whenever a minion from the lizardman shaman spawns and the plugin has an option to configure the volume.


LillieH001:
> Don't use `loadWhenOutdated`, as your plugin requires a client to function: https://github.com/LillieH001/Lizardman-Shaman-Minion-Spawn-Sound/blob/3ab63a7fb58107fb6f1628dd7cf9853323989f95/src/main/java/com/lizardmanshamanminionspawnsound/LizardmanShamanMinionSpawnSoundPlugin.java#L21

Fixed

PR #2656, opened 2022-05-05
added plugin file for salve-amulet-checker

Zabith4:
This plugin tells you which players are wearing their salve amulets in Cox's Mystic and Tob's Bloat room.  It also has a feature to overlay which room you are currently in in TOB and Cox.  The last feature is to call out players in the chat box who are not wearing their salve amulet in the rooms.

### DIFF
--- a/plugins/salve-amulet-checker
+++ b/plugins/salve-amulet-checker
@@ -1,0 +1,2 @@
+repository=https://github.com/Zabith4/Salve-Amulet-Checker.git
+commit=9b2027315a1218ecc2e5fcab5bd10ed1eddf501d
--- a/plugins/salve-amulet-checker
+++ b/plugins/salve-amulet-checker
@@ -1,2 +1,2 @@
 repository=https://github.com/Zabith4/Salve-Amulet-Checker.git
-commit=9b2027315a1218ecc2e5fcab5bd10ed1eddf501d
+commit=c5caf202df73e5c4f2ddf6e7ca4c2eba6444c6db


Nightfirecat:
Why are you doing reflection to get fields you defined in another class? <https://github.com/Zabith4/Salve-Amulet-Checker/blob/9b2027315a1218ecc2e5fcab5bd10ed1eddf501d/src/main/java/com/sac/panel/SalveAmuletCheckerPanel.java#L45-L48> Use an `enum` instead.
I recommend against [using spaces in your config group](https://github.com/Zabith4/Salve-Amulet-Checker/blob/9b2027315a1218ecc2e5fcab5bd10ed1eddf501d/src/main/java/com/sac/SalveAmuletCheckerConfig.java#L8) for serialization, and would suggest removing them or changing them to hyphens instead.
You shouldn't be [adding any overlays during `shutDown()](https://github.com/Zabith4/Salve-Amulet-Checker/blob/9b2027315a1218ecc2e5fcab5bd10ed1eddf501d/src/main/java/com/sac/SalveAmuletCheckerPlugin.java#L130-L131). I assume these were supposed to be `.remove()` instead?

Zabith4:
Thanks for the feedback I removed the `.add()` and changed them to `.remove()`.  I am using reflection so I do not have to change the code in `SalveAmuletCheckerPanel .loadMonsterDropDown()` when some one adds a new mob to check for.  The images I load have the mobs name + small.

Nightfirecat:
> I am using reflection so I do not have to change the code in `SalveAmuletCheckerPanel .loadMonsterDropDown()` when some one adds a new mob to check for. The images I load have the mobs name + small.

You can change `EntityNames` to be an enum, like the following:

```java
@AllArgsConstructor
@Getter
public enum EntityNames {
	BLOAT("Pestilent Bloat"),
	MYSTIC("Skeletal Mystic");
	private name;
}
```

Then iterate the entries via `EntityNames.values()`.

Zabith4:
> > I am using reflection so I do not have to change the code in `SalveAmuletCheckerPanel .loadMonsterDropDown()` when some one adds a new mob to check for. The images I load have the mobs name + small.
> 
> You can change `EntityNames` to be an enum, like the following:
> 
> ```java
> @AllArgsConstructor
> @Getter
> public enum EntityNames {
> 	BLOAT("Pestilent Bloat"),
> 	MYSTIC("Skeletal Mystic");
> 	private name;
> }
> ```
> 
> Then iterate the entries via `EntityNames.values()`.
> 
> Also, you must update the commit this PR points to for the change to be rebuilt.

I replaced my static EntityNames class with the Enum.  I also removed the reflection and looped over the values in the Enum. Thanks for all your help and feedback I appreciate it!  I also updated the commit.

PR #2663, opened 2022-05-07
 Update customizable-xp-drops to v1.5

l2-:


l2-:
Should the lombok and junit upgrades be reverted?

LlemonDuck:
JUnit should be fine, but Lombok will need to match one of the versions in https://github.com/runelite/plugin-hub/blob/master/package/verification-template/gradle/verification-metadata.xml

l2-:
Done. Thanks

PR #2675, opened 2022-05-12
Betteroverload

tdb48:
New plugin that is an improvement on the overload timer from the 'timers' plugin. The plugin accounts for world lag so the timer will never be off.


tdb48:
Found an issue where ovl conflics with enhance varp - working on a fix so its not ready atm

Nightfirecat:
For what it's worth, if you were interested in putting in some more time and effort to have it reviewed, I'm sure we'd be interested in accepting this as an enhancement to the core timers plugin.

PR #2681, opened 2022-05-14
Create sacredoil

KhaiQ:
Provides a charge counter and notification regarding Flamtaer Bracelets during the Shades of Morton minigame

### DIFF
--- a/plugins/sacredoil
+++ b/plugins/sacredoil
@@ -1,0 +1,2 @@
+repository=https://github.com/KhaiQ/sacredoil.git
+commit=99c494542d6c961b52aafc1238db5866830f0cb6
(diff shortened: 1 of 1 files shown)


Nightfirecat:
Don't use `"example"` as your plugin's config group: https://github.com/KhaiQ/sacredoil/blob/5f77e54d90c6097b49e50af6f1253062c1e39bfc/src/main/java/com/sacredoil/SacredOilConfig.java#L5

Nightfirecat:
We recommend against using [`enabledByDefault = false`](https://github.com/KhaiQ/sacredoil/blob/af920e66e003ef29e3bcae9cceea7b4efc2e8ffc/src/main/java/com/sacredoil/SacredOilPlugin.java#L30) as users get confused by an installed plugin not working upon installation.

PR #2683, opened 2022-05-14
created combat-level-calc

HerbsIO:
Enables a simple combat level calculator in the side panel


LlemonDuck:
You aren't assigning the value of `uiNavigationButton` here, so it won't be removed properly on plugin shutDown at the moment. https://github.com/HerbsIO/combat-calc/blob/fa667f1f19db7a81b869518d4fa47019d6f4eaa6/src/main/java/com/hzy/cmbcalc/CombatLevelCalculatorPlugin.java#L43

fwiw, and all optional changes, you can simplify some of your swing stuff by:
* Using an ActionListener for the calc button instead of a MouseAdapter
* Only overriding keyTyped for your number-only KeyAdapter, and checking for the character code there (which should also help with not intercepting mid-press hotkey combos), [here's an example](https://gist.github.com/LlemonDuck/ecef99f7bdde8c5bc7444af7cc127ad1)
* Using an [indexed for-loop here](https://github.com/HerbsIO/combat-calc/blob/fa667f1f19db7a81b869518d4fa47019d6f4eaa6/src/main/java/com/hzy/cmbcalc/CombatLevelCalculatorPanel.java#L154), instead of a for-each into indexOf

PR #2697, opened 2022-05-18
Add apm plugin

adam-bunce:
adds an overlay that shows average session apm and actions within the last minute


Nightfirecat:
Your [mouse and keyboard listeners](https://github.com/adam-bunce/apm/blob/3aba72dd9723a6fbb17f06ed49929fe71ba9a7c2/src/main/java/com/apm/ApmPlugin.java#L79-L80) need to be unregistered on `shutDown()`.

adam-bunce:
> Your [mouse and keyboard listeners](https://github.com/adam-bunce/apm/blob/3aba72dd9723a6fbb17f06ed49929fe71ba9a7c2/src/main/java/com/apm/ApmPlugin.java#L79-L80) need to be unregistered on `shutDown()`.

fixed now (i think), thanks.

Nightfirecat:
Don't shut down the executor service on plugin shutdown--other plugins use this service as well.

PR #2699, opened 2022-05-18
Create temple-osrs v1.0

SMaloney2017:
Fetch and display user's skilling/ bossing gains for the day/week/month/etc. using the [TempleOSRS API](https://templeosrs.com/api_doc.php).


koivel-admin:
Should change to in your build.gradle and compile with java 8. Need to remove stuff like `Map/List.of()` calls since they are from a higher version of java.
```
sourceCompatibility = '1.8'
targetCompatibility = "1.8"
```

PR #2707, opened 2022-05-20
Adding Ranking System plugin

Landathradon:
This plugin will show players if they can rank up in their clan according to ranks that have been setup by the clan admins.

It will check through the Inventory, bank, equipment, collection logs, combat achievements and prayers to know if they can be ranked up. Collection logs are also saved on disk.

Currently only my clan will use it but I'll work on making it available for everyone.

### DIFF
--- a/plugins/ranking-system
+++ b/plugins/ranking-system
@@ -1,2 +1,2 @@
 repository=https://github.com/Landathradon/Ranking-System.git
-commit=3f92eb576227c7afe3808a64f38711ff40aba023
+commit=0bdcf7ac747336ae7c56b31e0ef068b387a96119
--- a/plugins/ranking-system
+++ b/plugins/ranking-system
@@ -1,0 +1,2 @@
+repository=https://github.com/Landathradon/Ranking-System.git
+commit=3f92eb576227c7afe3808a64f38711ff40aba023
--- a/plugins/ranking-system
+++ b/plugins/ranking-system
@@ -1,4 +1,4 @@
 repository=https://github.com/Landathradon/Ranking-System.git
 commit=db5d0939aab6cd12c2ec61d19ddf470dcad1936b
 warning=This plugin submits the name of your clan and your IP address to a 3rd party website not controlled or verified by the RuneLite Developers.
-authors=Lore Bot
+authors=Landathradon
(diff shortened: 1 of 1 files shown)


Nightfirecat:
[Loading the ranks file like this](https://github.com/Landathradon/Ranking-System/blob/0bdcf7ac747336ae7c56b31e0ef068b387a96119/src/main/java/com/rankingsystem/RankHandler.java#L199) will not work in a shaded jar. It only works in development because you're running in your project's root. You should rethink how this loading will work.

Landathradon:
> [Loading the ranks file like this](https://github.com/Landathradon/Ranking-System/blob/0bdcf7ac747336ae7c56b31e0ef068b387a96119/src/main/java/com/rankingsystem/RankHandler.java#L199) will not work in a shaded jar. It only works in development because you're running in your project's root. You should rethink how this loading will work.

Oh ok that makes sense. The end product would be calling an api on my website to get the clan ranks in json format. Would something like this work fine?

Nightfirecat:
Yes, that's fine. Keep in mind, if this plugin makes requests by default (which would make sense), it will need a `warning` declaring that. (see other such plugins for reference) If its configuration is such that a URL must be entered in config before it would make outside requests, no such warning is needed.

plugins/ranking-system
```diff
@@ -0,0 +1,4 @@
+repository=https://github.com/Landathradon/Ranking-System.git
+commit=db5d0939aab6cd12c2ec61d19ddf470dcad1936b
+warning=This plugin submits the name of your clan and your IP address to a 3rd party website not controlled or verified by the RuneLite Developers.
+authors=Lore Bot
```
Nightfirecat, line 4: Who is this? I don't see any github user named  `Lore Bot`.

Nightfirecat (commented):
Code looks fine to me now.

plugins/ranking-system
Landathradon: This is me, I thought it was my runescape name I'd put there but I can change it back if thats the wrong way of doing things

Nightfirecat:
`authors` is used as a tool to help us know when other users are authorized to make changes to a plugin besides the author. You don't need to add it unless users besides yourself will be submitting PRs to this repository.

PR #2724, opened 2022-05-23
Add Recently banked items plugin

pwatts6060:


raiyni:
You should change the plugin class file to not be `BankPlugin`.

pwatts6060:
> You should change the plugin class file to not be `BankPlugin`.

Fixed. Changed to RecentBankPlugin

PR #2740, opened 2022-05-26
Change repo of Hue implementation plugin

pwatts6060:
@Nick-Anzalone has given me permission to take ownership of the plugin to maintain it.


Nick-Anzalone:
Hello! @pwatts6060 can take ownership of this plugin as I am no longer actively  keeping up with this project

amzn-paunort:
CI build failed: https://github.com/runelite/plugin-hub/runs/6617411108?check_suite_focus=true#step:6:79

LlemonDuck:
CI build failed: https://github.com/runelite/plugin-hub/runs/6617411108?check_suite_focus=true#step:6:79

pwatts6060:
Unsure how to fix the issue, would appreciate assistance.

pwatts6060:
I still don't know what caused that duplicate but changing the package task to have an include duplicate strategy fixes the issue but would result in overwriting of the duplicate which may be problematic, not sure what consequences that has. I've exhausted my options now though and not looking into this further without some pointers.

package/target_init.gradle
```diff
@@ -54,6 +54,7 @@ allprojects {
 	def buildDir = new File(System.getenv("runelite.pluginhub.package.buildDir"));
 
 	task runelitePluginHubPackage(type: Jar) {
+		duplicatesStrategy = DuplicatesStrategy.INCLUDE
```
Nightfirecat, line 57: We're not adding this so one plugin will build, especially if you don't understand what this change is or why it's needed. You should research the root cause of your plugin's build issue and resolve that instead.

pwatts6060:
@Jallah123 has permission to maintain the plugin if they can fix the build issues.

PR #2750, opened 2022-05-28
TypoFix and refactor create_new_plugin.py

Ferrariic:


abextm:
I am not interested in reformatting random files for no reason, and the supposed typo is correct as is. As much as I love type hints, I think require a fairly new version of python3, and I would like for this script to work basically anywhere any python3 is available within reason.

PR #2760, opened 2022-05-28
Add 'blindfold' plugin

stevenwaterman:
The blindfold plugin is a fork of the core GPU plugin which disables rendering for everything except your character and the UI layer. Created as a replacement for the greenscreen plugin for [Hanannie's YT series](https://www.youtube.com/watch?v=ifqgPJ8LhzQ). Compared to the greenscreen plugin, it provides anti-aliased edges, significantly improved performance, and allows tile markers, user chat, and other interfaces to still be shown.

image


stevenwaterman:
Sorry about slight commit spam, realised that I was still using the old `build.gradle` commands, updated those + squashed back into 1 commit to keep your history tidy

PR #2777, opened 2022-06-02
Ultimate nmz

iant89:


TheStonedTurtle:
You removed the class but not the references to that class so the build failed, re-add the class the 503 error wasn't caused by that code.

iant89:
The class I removed had no actual references in any of the other classes.

If someone could help me out it would be great!

If I am reading the build log correctly it is using java 11, and the actual issue is its saying there is no methods with the names "toSecondsPart", "toMinutesPart", "toHoursPart" which are defined in the java.time.Duration class.

Java 9 added those methods I believe, so why does this not compile in 11? Do i need to make the code capable to 1.8? or something else.

iant89:
> You removed the class but not the references to that class so the build failed, you can re-add the class. The previous 503 error wasn't caused by that code (or any of your code if I understand it correctly)

The class I removed had no actual references in any of the other classes. It failed the build on the first pull I done. and i thought I messed up, so i closed the pull request. and removed the class, and committed the change. I'm still learning git so I know very little about , but in the process of learning. 

If someone could help me out it would be great!

If I am reading the build log correctly it is using java 11, and the actual issue is its saying there is no methods with the names "toSecondsPart", "toMinutesPart", "toHoursPart" which are defined in the java.time.Duration class.
atleast it is according to these docs.
https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/time/Duration.html

Java 9 added those methods I believe, so why does this not compile in 11? When it will compile fine on my setup. 

Do i need to make the code capable to 1.8? or something else?

Nightfirecat:
Yes, plugin hub code must compile on Java 8 because not all users are using a launcher which ships Java 11.

iant89:
Fixed issues, Thank you!

PR #2789, opened 2022-06-04
Create rc-lap-tracker

JwowSquared:
Hello,

This plugin allows for tracking pouch degradation while training at ZMI.
Other plugins I have found to be unreliable, due to the pouch's inventory not always updating if you bank too quickly.
My goal with this plugin was to circumvent this issue by simply tracking the laps you perform at ZMI.

Config allows for selecting the highest tier pouch you are using, and the counter will reset to the proper value based on this setting.
Counter decrements the first time you craft runes after entering ZMI.
Counter resets on NPC Contact cast.


JwowSquared:
I updated this from a ZMI specific plugin to work with any other lap related RC methods.

JwowSquared:
Okay this is actually done now. Plugin state persists through world hops and logging out.

PR #2793, opened 2022-06-05
Create OWM-CHAT Plugin Submission

Cerisk:
Plugin takes the chat that npcs display in their dialog boxes and puts it in the chat box.

### DIFF
--- a/OWM-CHAT Plugin Submission
+++ b/OWM-CHAT Plugin Submission
@@ -1,0 +1,2 @@
+repository=https://github.com/Cerisk/OWM-Chat.git
+commit=ae57ab76a9b33e858efc134a8f1ae41813b73412
--- a/OWM-CHAT Plugin Submission
+++ b/OWM-CHAT Plugin Submission
@@ -1,2 +1,0 @@
-repository=https://github.com/Cerisk/OWM-Chat.git
-commit=ae57ab76a9b33e858efc134a8f1ae41813b73412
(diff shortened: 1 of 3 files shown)


OWM-CHAT Plugin Submission
```diff
@@ -0,0 +1,2 @@
+repository=https://github.com/Cerisk/OWM-Chat.git
```
Nightfirecat, line 1: This file should not be added. The only one submitted should be `plugins/owm-chat`.

Cerisk:
Also, I don't know why the checks were not successful so I thought it had to do with the changes I made to the gradle (putting in the information so I can create a shadowJar). I deleted those portions in the build.gradle so hopefully when it runs another check it can build it properly.

Nightfirecat:
[Adding the shadowJar plugin is causing the build failure](https://github.com/Cerisk/OWM-Chat/blob/b413b9334ea88c9958bec5ccbe8391ffffe63166/build.gradle#L3), remove it and it'll work fine.

Nightfirecat:
Your config group is still not really well namespaced. "User" isn't specific enough.

Cerisk:
> Your config group is still not really well namespaced. "User" isn't specific enough.

Not sure exactly how I should change this to be more specific, so I just set it to be "userChatFilterConfig" since in the future I will have users be able to set which npcs / messages they will want filtered in their config.

Nightfirecat:
It's often the same as your plugin name, so something like `owm-chat` is what I would've chosen. This is fine though.

PR #2799, opened 2022-06-07
Added new plugin ItemDropSound

Spongie:
Plugin that plays a sound when an item drops.
Which item to play sounds on is configurable


bigboobdude:
> Plugin that plays a sound when an item drops. Which item to play sounds on is configurable

Hey do you think it would be possible to make the sound played customizable? even just making the .wav or whichever file replaceable would be great. i think this could be a fantastic and massively popular plugin if that were possible.

I would have all the huge drops like harm orb/ely etc on list and it would really add to the moment : )

PR #2815, opened 2022-06-09
addition gamesense plugin file

LowieN-1848537:
The plugin makes it possible to send game related data (Hp,Prayer, last trained skill xp) to the steelseries gamesense client. This enables users to display in game stats on the keyboard (such as a bar correlating to your HP)

A short example of the result in the event of prayer is shown in the attached file.

https://user-images.githubusercontent.com/60964789/172820111-955e135a-edcc-427a-8b9d-73b3455e7464.mp4

### DIFF
--- a/plugins/gamesense
+++ b/plugins/gamesense
@@ -1,2 +1,2 @@
 repository=https://github.com/LowieN-1848537/SteelseriesGamesense.git
-commit=a04c8bc6994e2663bd31a9559ab4f2dfb07d97d7
+commit=5bdb4dd5789a13bc3ee5da99b5d3a9b8eb528572
--- a/plugins/gamesense
+++ b/plugins/gamesense
@@ -1,2 +1,2 @@
 repository=https://github.com/LowieN-1848537/SteelseriesGamesense.git
-commit=d45025d8995472f11dadc1d9baaa273a2a222b59
+commit=efd084c68b2e8ad30b59fcb9c80fb8901df741dd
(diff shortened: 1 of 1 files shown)


Nightfirecat:
Instead of adding a dependency on `org.json`, you can use GSON to submit JSON in requests.

LowieN-1848537:
ah is that why the checks failed? I'll replace it once i find the time to do so.

LowieN-1848537:
ah finally, took me way to long to figure out that verification metadata

LlemonDuck:
Do you really need to String->fromJson->toJson->String? https://github.com/LowieN-1848537/SteelseriesGamesense/blob/d45025d8995472f11dadc1d9baaa273a2a222b59/src/main/java/com/example/GamesensePlugin.java#L75

LlemonDuck:
* Ideally you would define these message types as POJOs and then serialize with Gson to prevent any formatting/interpolation errors https://github.com/LowieN-1848537/SteelseriesGamesense/blob/efd084c68b2e8ad30b59fcb9c80fb8901df741dd/src/main/java/com/example/GamesensePlugin.java#L68
* You can use `OSType` https://github.com/LowieN-1848537/SteelseriesGamesense/blob/efd084c68b2e8ad30b59fcb9c80fb8901df741dd/src/main/java/com/example/GamesensePlugin.java#L159
* Same here for a POJO message type but on the parsing end https://github.com/LowieN-1848537/SteelseriesGamesense/blob/efd084c68b2e8ad30b59fcb9c80fb8901df741dd/src/main/java/com/example/GamesensePlugin.java#L182
* You should inject and use the client's existing OkHttpClient with a custom user agent https://github.com/LowieN-1848537/SteelseriesGamesense/blob/efd084c68b2e8ad30b59fcb9c80fb8901df741dd/src/main/java/com/example/GamesensePlugin.java#L248
* Your config group should not contain a space (since you are not using a config you can remove it instead) https://github.com/LowieN-1848537/SteelseriesGamesense/blob/efd084c68b2e8ad30b59fcb9c80fb8901df741dd/src/main/java/com/example/GamesenseConfig.java#L7

PR #2817, opened 2022-06-09
Update Clan Events Plugin

cmsu224:
Added HTML Tables, Buttons, and Links support to the plugin panel.


Nightfirecat:
Why is the Sheets Service API updated for this change? I don't see you doing anything new or differently with it in this change.

cmsu224:
> Why is the Sheets Service API updated for this change? I don't see you doing anything new or differently with it in this change.

'com.google.api.client.json.jackson2.JacksonFactory' was deprecated so, I switched to using 'com.google.api.client.json.GsonFactory' instead.  That was the only API change.

Brianmm94:
Yeah, I had changed it when making a pull request to his repository. The only reason I did that was because Intellij was telling me that JsonFactory was deprecated. I'm guessing I probably didn't need to change it yet seeing as the deprecated package likely still works, but might as well just change it now rather than later.

Nightfirecat:
There's no need to change it at all, it won't suddenly stop working.

Brianmm94:
It won't? I was looking everywhere on stack overflow and all I saw were answers like this that say although it's available now it likely won't be in the future and that it's likely unsafe to keep using it. I can change it back though if you really want me to. https://stackoverflow.com/questions/2941900/is-it-wrong-to-use-deprecated-methods-or-classes-in-java

Brianmm94:
I think I understand now, it's only for future releases of the packages that they may not be usable, but they likely will be still for backwards compatibility. I'll look into why it was deprecated.

cmsu224:
> There's no need to change it at all, it won't suddenly stop working.

This has now been reverted to the old API version.

package/verification-template/gradle/verification-metadata.xml
```diff
@@ -887,4 +887,4 @@
          </artifact>
       </component>
    </components>
-</verification-metadata>
+</verification-metadata>
```
Nightfirecat, line 890: Reverting this EOF line ending removal to make the bot labels happy
```suggestion
</verification-metadata>

```

PR #2823, opened 2022-06-11
Update `custom-item-hovers` to v1.1.2

geel9:
- Fixes hovers not showing in the inventory when the default action on an item is "Use"
- Bumps RL version


geel9:
Closing this PR to fix the massive diff caused by whitespace changes

PR #2824, opened 2022-06-11
Update `custom-item-hovers` to v1.1.2

geel9:
- Fixes hovers not showing in the inventory when the default action on an item is "Use"
- Bumps RL version


LlemonDuck:
https://github.com/geel9/runelite-custom-item-hovers/compare/440db3f8c5b9325cb94d841b2047393d43219ca5...geel9:a9ccfaf1259187a4910e286094e7e2832811363b#diff-ac5622e5ea23beaa9ea1b4a1a68ad04d4ea3831be39e3b9231c6635ba938fc5eL1

If this code was originally licensed to "Charlie Waters", you should be retaining this copyright notice or providing attribution elsewhere (e.g. in your `LICENSE` file) to comply with the license terms, unless you have a permission from the original author.

geel9:
The license existed in that file because I copied another plugin's overlay file -- and then removed all actual code present -- solely for the skeleton of expected method signatures that an `Overlay`-implementing class must have.

That is, the only code that I didn't write entirely by hand was effectively the following:

```java
class CustomItemHoversOverlay extends Overlay
{
	@Override
	public Dimension render(Graphics2D graphics)
	{
	}
}
```

I accidentally included the license as well, of course.


I removed the license because it did not seem, to me, that the declaration of an override of a method -- with no actual code -- was in any way... real code.

If you feel strongly that I need to re-add the license, I suppose I can do that.

LlemonDuck:
I see... I suppose that should be fine.

PR #2833, opened 2022-06-14
Add ba-call-highlight plugin

rugg0064:
A plugin that highlights the current barbarian assault call
image
image


LlemonDuck:
Your config group should be more distinctive instead of "main", usually the plugin name in all lowercase with no spaces: https://github.com/rugg0064/ba-call-highlight/blob/8b47dcea94e5ef915cad37314f9d7e81db1ff037/src/main/java/com/ba/call/highlight/baCallHighlightConfig.java#L9

You should remove the ExamplePlugin and ExampleConfig files: https://github.com/rugg0064/ba-call-highlight/tree/8b47dcea94e5ef915cad37314f9d7e81db1ff037/src/main/java/com/example

PR #2835, opened 2022-06-14
Update ba-minigame to 1.66

equirs:
Fixes the currently broken menu entry reorder for the "tell-defensive" option on the collector horn.


LlemonDuck:
You can add an `authors` field as is here: https://github.com/runelite/plugin-hub/blob/master/plugins/osrs-wiki-crowdsourcing

PR #2852, opened 2022-06-17
add stat track plugin

MatteoVH:
Logs stat changes to a server


Nightfirecat:
Instead of creating your own OkHttpClient, inject the one from RuneLite. Furthermore, especially if this server is not owned by you, I recommend setting a user agent so it's clear traffic is coming from this plugin.
Your plugin needs a `warning` field such as the following:
```
warning=This plugin submits your skill experience and IP address to a 3rd party server not controlled or verified by the RuneLite Developers.
```

MatteoVH:
Thanks for the review @Nightfirecat!

It's my own server, so I skipped adding the user agent, but I did add the warning and injected the OkHttpClient vs. constructing a new instance.

PR #2855, opened 2022-06-17
tob-qol-improvements: Add simple plugin for QoL improvements at ToB

WLoumakis:
Add a simple plugin to add the following QoL improvements in ToB:
* Swap supply chest 'value' option with configurable 'buy-1', 'buy-x', 'buy-all', and 'value' (in case they want to keep it off)
* Add an outline and overlay arrow over the loot chest outside the Theatre
* Remove forced right-click menu when banking loot

The first option is depicted in [this gif](https://i.imgur.com/NWVs67v.gif).

The latter two are depicted in [this gif](https://i.imgur.com/o380nJj.gif).

### DIFF
--- a/plugins/tob-qol-improvements
+++ b/plugins/tob-qol-improvements
@@ -1,2 +1,2 @@
 repository=https://github.com/WLoumakis/tob-qol.git
-commit=028af268d38efb30f733b15af575a9fefd28191b
+commit=0cc477f80d2f1bcb95d12e8ebcea3718c4ba5908
--- a/plugins/tob-qol-improvements
+++ b/plugins/tob-qol-improvements
@@ -1,2 +1,2 @@
 repository=https://github.com/WLoumakis/tob-qol.git
-commit=a7db3fcbe2ee2694d71054757a3826b7d30b2989
+commit=028af268d38efb30f733b15af575a9fefd28191b
--- a/plugins/tob-qol-improvements
+++ b/plugins/tob-qol-improvements
@@ -1,0 +1,2 @@
+repository=https://github.com/WLoumakis/tob-qol.git
+commit=a7db3fcbe2ee2694d71054757a3826b7d30b2989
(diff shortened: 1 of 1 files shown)


geheur:
fyi there is a PR for swapping the chest's buy-1 option in vanilla MES. It's not accepted yet and idk if it will be but it might be good to be aware of: <https://github.com/runelite/runelite/pull/15108>

WLoumakis:
Good to know @geheur. I'll leave this up and see what happens; can always take those out and just make this a loot reminder plugin.

geheur:
On second thought that one is on shift-click and yours seems to be on left-click so I guess it's not the same thing after all. nevermind then, mb.

WLoumakis:
Either/or work tbh. I just thought about it, but a left-click bank loot option would be nice to have as well...i'll work on that next.

geheur:
excited to see what you come up with because I've tried that before and couldn't figure out how to make it happen.

WLoumakis:
Edited the description already, but here's the showcasing of the last two portions: https://i.imgur.com/o380nJj.gif

osrscommits:
I don't mind abandoning https://github.com/runelite/plugin-hub/pull/2855 in favor of your swap; I don't think it needs both, and it seems nice to have this grouped with TOB. In theory the MES is more general, but I'm not sure who would ever want to swap buy-all or buy-x.

WLoumakis:
@osrscommits if you're alright with it, I could yoink ur code and replace my static buy-1 option and then we could close #2855

osrscommits:
For sure! Sounds good to me

geheur:
My 2 cents: as long as you can still swap it without shift held. There is no other option than buy-1 that has any practical use in that menu.

WLoumakis:
agreed, but also doesn't really hurt to have the option in case some folks want to use it

WLoumakis:
Alright here's demo of the buy-1 as well: https://i.imgur.com/NWVs67v.gif. Didn't think in the moment to showcase buy-x or buy-all, but it should work the same.

PR #2856, opened 2022-06-17
Update customizable-xp-drops to v1.5.5

l2-:


LlemonDuck:
It looks like the author field of your commit is set incorrectly. It appears as `unknown <you@example.com>`

l2-:
> It looks like the author field of your commit is set incorrectly. It appears as `unknown <you@example.com>`

Thanks for notifying but that is my intention.

PR #2857, opened 2022-06-18
Create PartyHealthStatus plugin.

1Defence:
Visual representation of party members hitpoints over the player model as both a colored text and optionally a colored hull

https://user-images.githubusercontent.com/63735241/174417566-5df021b9-b665-4519-a459-ce76305ed239.mp4

### DIFF
--- a/plugins/party-health-status
+++ b/plugins/party-health-status
@@ -1,2 +1,2 @@
 repository=https://github.com/1Defence/party-health-status.git
-commit=d78912133a65f9e6d0317d99ea7a43e7ca602511
+commit=929c559604e19edffffa762c73ff9559d6b162bd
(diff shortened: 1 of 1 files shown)


LlemonDuck:
There is already a health packet in core that you can reuse via the `PartyData`, which you can pull from `PartyPluginService`, rather than creating your own health packet. Reusing the data from this packet will reduce redundant network load.

1Defence:
> There is already a health packet in core that you can reuse via the `PartyData`, which you can pull from `PartyPluginService`, rather than creating your own health packet. Reusing the data from this packet will reduce redundant network load.

Done, rewrote the plugin to reflect said changes.

LlemonDuck:
You may find interesting: there is a proposal to bring this information and more into the core party plugin as well, either to contribute there or use the information to expand your plugin: runelite/runelite#15168

PR #2862, opened 2022-06-19
Create Party Defence Tracker

MoreBuchus:
Calculates the defence of a boss based off special attacks from members of a party. Displays infoboxes for the calculated defence level of a boss and if vulnerability successfully lands.

This is very useful when boosting, pvming with multiple alts, or pvming with friends in a party.

partydefencetracker


LlemonDuck:
Will you be updating the plugin to use the event info you requested from core? https://github.com/runelite/runelite/commit/004a52159f622fdf1d3197471125649215bc03d9

MoreBuchus:
Changes:
+ Calculate defence off of Special Attack Counter
+ Queue specs if the boss has not been set/has not cached the NPC
+ Tracks boss index to work in instances
+ Move the plugin over to buchus-plugins repo

PR #2866, opened 2022-06-20
Add Nightmare timers plugin

dey0:
Time tracking for Nightmare phases and kill times


LlemonDuck:
This PR contains the object-hider changes as well. You should split that into a separate PR.

PR #2868, opened 2022-06-21
Create dice

bogstandard:
Adds plugin to introduce on-screen dice rolls to the game for a table-top like experience.

### DIFF
--- a/plugins/dice
+++ b/plugins/dice
@@ -1,2 +1,2 @@
 repository=https://github.com/bogstandard/dice.git
-commit=99ecea1bb41eab1ced717fd1f298037fa63a18f3
+commit=5841a593b16937fa087c0b2517ff74a7354a4857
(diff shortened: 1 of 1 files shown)


LlemonDuck:
This does not truly remove the widget from the parent: https://github.com/bogstandard/dice/blob/5841a593b16937fa087c0b2517ff74a7354a4857/src/main/java/com/dice/DicePlugin.java#L101

LlemonDuck:
You can use Widget#getChildren -> copy the array without your widget -> Widget#setChildren

abextm:
just `parent.getChildren()[child.getIndex()] = null;`

bogstandard:
Thanks both, hopefully that's done it? I assumed that the arrays were working by reference to their child objects, so nullifying the child would in-turn cause the reference within the array to be nullified too.

LlemonDuck:
> I assumed that the arrays were working by reference to their child objects, so nullifying the child would in-turn cause the reference within the array to be nullified too.

fwiw, java objects are "by-reference" but those references are not shared, so setting some variable/field to null will not also null out other references to the object i.e. in an array somewhere or in another variable

PR #2871, opened 2022-06-21
Create phoenix-necklace-jingle

Nope1252:
Plays a sound when an equipped phoenix necklace breaks

### DIFF
--- a/plugins/phoenix-necklace-jingle
+++ b/plugins/phoenix-necklace-jingle
@@ -1,0 +1,2 @@
+repository=https://github.com/Nope1252/phoenix-necklace-jingle.git
+commit=b77e18a556b1ea3a3fa5d71368e93286de6342be


Skretzo:
What do you think of turning this into a more general plugin to also support e.g ring of recoil?

Nope1252:
> What do you think of turning this into a more general plugin to also support e.g ring of recoil?

I could if somebody needed that. The main use case for adding a sound effect when a pneck breaks to help time actions for healer accounts during pvm boosting; I'm not sure there is a use case for a subtle sound effect when a ring of recoil breaks.

LlemonDuck:
There is at Zulrah for those that don't have a ring of suffering, but there is already a plugin for this: https://runelite.net/plugin-hub/show/ring-of-recoil-notifier

PR #2875, opened 2022-06-22
Update object-hider

dey0:
Disables functionality at Sotetseg regions to prevent abuse

### DIFF
--- a/plugins/object-hider
+++ b/plugins/object-hider
@@ -1,3 +1,2 @@
-repository=https://github.com/Leo288/object-hider.git
-commit=3d6f915c7f17db61ae0e6664dc7524bd28033f18
-disabled=tob abuse
+repository=https://github.com/dey0/ground-object-hider.git
+commit=38e414ccf7756083fdf88ac912823ad2b1e1c1bb


pwatts6060:
Would help easier review if you could not reformat the entire plugin code. 
Relevant changes for reviewer is in hideGroundObjectOnTile method

Old code:
```
final int[] disallowedObjIds = {41750, 41751, 41752, 41753};  // TOB
int objId = obj.getId();
if (Arrays.stream(disallowedObjIds).anyMatch(v -> v == objId)) {
	log.debug("hiding of a disallowed Ground Object was prevented ({})", objId);
	return false;
}
```

New code:
```
for (int i : client.getMapRegions()) {
  if (i == 13123 || i == 13379) {
	return false;
  }
}
```
image
The region ids are the sotetseg shadow realm and regular sotetseg room region

pwatts6060:
Additionally you aren't the original author of the plugin. Usually you need to show that you have permission to maintain the plugin from the original author, in this case @Leo288

Leo288:
I am ok with this - thanks for maintaining.

PR #2881, opened 2022-06-23
Update and take ownership of Health Notifier.

geheur:
Permission granted here: https://github.com/Callum-A/health-notifier/issues/5

### DIFF
--- a/plugins/health-notifier
+++ b/plugins/health-notifier
@@ -1,2 +1,2 @@
-repository=https://github.com/Callum-A/health-notifier.git
-commit=c4a04e960e4347bf79d171b256fcdbcde29b818d
+repository=https://github.com/geheur/health-notifier.git
+commit=c4f546e33b4497f2fb2f636347797ed6dbf010b3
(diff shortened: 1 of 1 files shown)


LlemonDuck:
ClientTick seems aggressive here, do you want to use GameTick instead? You're hitting the config on each event until the notification is fired. https://github.com/Callum-A/health-notifier/compare/c4a04e960e4347bf79d171b256fcdbcde29b818d...geheur:c4f546e33b4497f2fb2f636347797ed6dbf010b3#diff-6bfb765a4dd45c59cea4dbe7ff774f8bbbd137d9dec92fdfe4613ecf5eecbe0cR77

geheur:
hmm that's a good point. I'm not aware of any situations where an npc's health would change outside of a game tick.

Also, is hitting the config 50 times per second a significant cost? I thought the config was super fast.

geheur:
I have updated the plugin to use game tick and cache the config value. I also fixed a bug where the config was not read on startup and where changing the config might produce unnecessary notifications.

PR #2882, opened 2022-06-23
Update friend-live-locations

TiboDeMunck:
Now able to display Clan Title icon

- Plugin will not work unless they update the API as well (is there a way to notify this to the user or will they just land on my GitHub page figure it out?)

### DIFF
--- a/plugins/friend-live-locations
+++ b/plugins/friend-live-locations
@@ -1,2 +1,2 @@
 repository=https://github.com/TiboDeMunck/runelite-live-friend-locations-plugin.git
-commit=fd62f0fe02ea17c0e2620ac1acad00ecf13aae46
+commit=ca12629f2c99e8068fd3f53abd5c59b632296937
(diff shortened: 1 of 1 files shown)


LlemonDuck:
* `ImageUtil#loadImageResource` already returns a BufferedImage, you don't need to draw loaded resources onto other BufferedImage instances, just `private static final BufferedImage ZENYTE = ImageUtil.loadImageResource(...)` will suffice. https://github.com/TiboDeMunck/runelite-live-friend-locations-plugin/blob/ca12629f2c99e8068fd3f53abd5c59b632296937/src/main/java/com/livelocationsharing/LiveLocationSharingIcons.java
* Considering all these icons are coming from the game anyway, you may want to look into the SpriteManager instead
* "is there a way to notify this to the user or will they just land on my GitHub page figure it out?" you can notify the user on first plugin start

Nightfirecat:
The sizing issue can be fixed by using [`ImageUtil#resizeImage()`](https://static.runelite.net/api/runelite-client/net/runelite/client/util/ImageUtil.html#resizeImage(java.awt.image.BufferedImage,int,int)). Presently, this looks fine though so I will go ahead and merge it.

PR #2892, opened 2022-06-25
Update PartyHealthStatus to v1.1

1Defence:
Various bug fixes in overlay, fixed major oversight regarding X user hopping between clients & Added more customizability.

**New features**

- Vertical offset for stacked players can now be modified
- Font can now be bold or not rather than forced bold
- Added healthy offset, in which players missing X number of hitpoints won't render their hitpoints
- Added multiple variations of Color calculations; Lerp2D,lerp3D,Static and Color Thresholds

### DIFF
--- a/plugins/party-health-status
+++ b/plugins/party-health-status
@@ -1,2 +1,2 @@
 repository=https://github.com/1Defence/party-health-status.git
-commit=018e2d87bf21c5db1b2615e0a89c3226c42d53af
+commit=e8f0b6e008eded638de85a8d98fa419050923119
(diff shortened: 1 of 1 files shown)


LlemonDuck:
You should avoid hitting the config multiple times per frame as you are currently in the overlay render and instead cache the config values and use the ConfigChanged event.

1Defence:
> You should avoid hitting the config multiple times per frame as you are currently in the overlay render and instead cache the config values and use the ConfigChanged event.

Thanks, removed all config calls from render as requested.

PR #2893, opened 2022-06-25
Update npc-idle-timer-plugin v1.2

lejeffe:


LlemonDuck:
Do you have permission from @vonpawn to submit this change?

vonpawn:
@lejeffe has been added as a collaborator on the plugin's  repo and is the new maintainer of this plugin. Is there something I need to change to make them show as the author from now on?

LlemonDuck:
> Is there something I need to change to make them show as the author from now on?

Between you (@vonpawn) and @lejeffe, either you can:
* transfer the repository to @lejeffe's name if they will be the owner from now on
* keep the repository in your name, and add @lejeffe to the authors line of this plugin hub file, [e.g. as here](https://github.com/runelite/plugin-hub/blob/master/plugins/osrs-wiki-crowdsourcing)
@vonpawn will need to confirm which of the above should be true and @lejeffe can update the PR accordingly

vonpawn:
Ah cool, I did not know an author line could be added. @lejeffe could you update the PR with an added author line?

lejeffe:
> > Is there something I need to change to make them show as the author from now on?
> 
> Between you (@vonpawn) and @lejeffe, either you can:
> 
> * transfer the repository to @lejeffe's name if they will be the owner from now on
> * keep the repository in your name, and add @lejeffe to the authors line of this plugin hub file, [e.g. as here](https://github.com/runelite/plugin-hub/blob/master/plugins/osrs-wiki-crowdsourcing)
>   @vonpawn will need to confirm which of the above should be true and @lejeffe can update the PR accordingly

Updated the authors

PR #2900, opened 2022-06-27
create rseye-connector

BasketBandit:
rseye-connector gives the ability to send player data such as position, inventory, equipment, login state etc out to a configurable endpoint in json format for external manipulation.

There is a similar plugin to this but I feel it has a limited scope to only send out data reactively whereas I require data proactively so this plugin aims to fill that gap and allows me to flesh out further features in the future.

### DIFF
--- a/plugins/rseye-connector
+++ b/plugins/rseye-connector
@@ -1,2 +1,2 @@
 repository=https://github.com/osrseye/rseye-connector.git
-commit=22920c58c8fa182e7e2db14f53ded632874613a1
+commit=4a6e7fd4cd00ba8054859dbdc1f52fe9c5a6a7f5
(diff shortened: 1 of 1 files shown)


LlemonDuck:
Since the plugin lives in an organization, you should add yourself an author for this plugin [like as done here](https://github.com/runelite/plugin-hub/blob/master/plugins/osrs-wiki-crowdsourcing)

BasketBandit:
> Since the plugin lives in an organization, you should add yourself an author for this plugin [like as done here](https://github.com/runelite/plugin-hub/blob/master/plugins/osrs-wiki-crowdsourcing)

Thanks for the heads up!

PR #2902, opened 2022-06-28
Add toggle chat plugin - needs 1.8.25

MoreBuchus:
Allows the user to set a hotkey to open/close the chatbox. Also disables the flashing notification on the chat tabs.


LlemonDuck:
* You should use Client#setVarcIntValue https://github.com/MoreBuchus/buchus-plugins/blob/11309f10a29a2db335574195c1d4e1918f14e656/src/main/java/com/togglechat/ToggleChatPlugin.java#L87
* You should look into HotkeyListener built specifically for this purpose, or just KeyEvent#consume to prevent the chat box receiving the key press https://github.com/MoreBuchus/buchus-plugins/blob/11309f10a29a2db335574195c1d4e1918f14e656/src/main/java/com/togglechat/ToggleChatPlugin.java#L119

PR #2908, opened 2022-06-29
updated readme

zibdie:


zibdie:
If you jump to the second option, you may miss the step about the VM. I have updated the README to tell the user accordingly

PR #2927, opened 2022-06-30
Add clan-tracker Plugin

Talados:
This is a plugin to track the data of my clan. the data is being transferred to a 3rd party server that forwards it from there to discord using webhooks and stores the data for analytics.


Nightfirecat:
[Don't use `enabledByDefault = false`](https://github.com/Talados/ClanTracker/blob/9b45ed2d86e83557078f9cc61fdeeae5949541da/src/main/java/com/clantracker/ClanTrackerPlugin.java#L33), it's confusing for users.

Talados:
> [Don't use `enabledByDefault = false`](https://github.com/Talados/ClanTracker/blob/9b45ed2d86e83557078f9cc61fdeeae5949541da/src/main/java/com/clantracker/ClanTrackerPlugin.java#L33), it's confusing for users.

Removed

PR #2934, opened 2022-07-02
Enhanced-Discord-Notifications

oliverpatrick:
An extended version of Discord Notifications forked from the latter to include: Collection logs and Valuable drops for a nice all in one discord webhook plugin for runelite. Keep motivated and keep track of your progress, all from one place!

image


LlemonDuck:
You have unrelated files attached to this PR which you should remove.

PR #2942, opened 2022-07-02
Add Better NPC Highlight

MoreBuchus:
A more customizable way to highlight NPCs that includes:
+ 9 Different styles of highlighting
+ Each highlight style has a name and ID list, color picker, width adjustment, and some offer an option to turn off/on anti-aliasing
+ Exclusion list of what NPCs highlights you do not want to hide when Ignore Dead NPCs is enabled
+ Respawn timer in ticks or seconds
+ Turbo mode for rave enthusiasts

example

### DIFF
--- a/plugins/better-npc-highlight
+++ b/plugins/better-npc-highlight
@@ -1,0 +1,2 @@
+repository=https://github.com/MoreBuchus/buchus-plugins.git
+commit=db1fab3b479e5c38bd9941c0a6b8225e285f543b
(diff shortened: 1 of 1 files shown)


Nightfirecat:
What use cases are there for highlighting NPCs by ID? We've been bitten in the past by other x-by-ID plugins having the ability to do things which don't abide by Jagex's statement, and suspect the same could be done via this plugin.

Nightfirecat:
At this time I'm inclined to agree that it should be fine. Of course if it's discovered there are statement-breaking issues, we will contact you requesting changes as needed.

PR #2943, opened 2022-07-02
Update InventorySetups to Vv1.13

dillydill123:
Adds sections


dillydill123:
Diff is incorrect, probably because the repository was switched when the API change broke inv setups. Here's the diff I made

https://github.com/dillydill123/inventory-setups/compare/b6a4551b0d804dbd7cc438e30017945dd00a1fb3..980eda8d2f14dd34bac99d4fce21c47d4586e330

PR #2946, opened 2022-07-03
ToB QoL - v1.0.2 - Small Fixes

damencs:
- adjusts config for xarpus to properly reflect which sound should be muted
- adds credits for xarpus area sound effects played


nbulischeck:
@damencs You might have to fix this, the link has a second commit of "unnamed"

RSDonny:
Some small feedback collected from people so far. *They are too lazy to make a github account*

1) Nylo instance timer stays above your head the whole room
2) Menu colors for Nylo we're a bit too faded ex: green looked more blueish
3) Can only pick 1 attack option of role selector (Duo's/North freeze would be beneficial to be able to have more than 1 selected)
4) Sote instance timer shows green to click off 0 but really need to click off on 3
5) Reports of Bloat configs not consistently working every raid

Nightfirecat:
@RSDonny please use [the issue tracker for the plugin's repository](https://github.com/damencs/tob-qol/issues) for feedback instead of requesting features/bugfixes on plugin-hub PRs.

LlemonDuck:
You'll need to remove the whitespace around the commit hash on this PR

abextm:
You need to remove the html from your plugin descriptor and configs. Also as I told you on discord you are not allowed to remove clickboxes such as the bloat tiles

damencs:
> You need to remove the html from your plugin descriptor and configs. Also as I told you on discord you are not allowed to remove clickboxes such as the bloat tiles.
> 
> Also you cannot set `enabledByDefault = false` since its just confuses people

When did you tell me that removing null tiles wasn't allowed? NFC/Lemon did not have a problem with it when reviewing. It should not even be a part of the game, hence why it's a QoL feature. It doesn't give you an advantage within the room. It doesn't strike against guidelines (which I reviewed HEAVILY prior to submission). It's just like the Ver Sinhaza bank with the big entrance where if you barely even click into it's area it sends you to Narnia.

As for the enabledByDefault = false, I don't intend to force QoL functions onto people. There are so many config items, it would be an abundance of features instantly enabled which may not be desired by people. If you would like me to default some of the basic and less obtrusive items, then I can enable some. I don't feel comfortable forcing all on everyone on initial load.

Adam-:
> When did you tell me that removing null tiles wasn't allowed? NFC/Lemon did not have a problem with it when reviewing. It should not even be a part of the game, hence why it's a QoL feature. It doesn't give you an advantage within the room. It doesn't strike against guidelines (which I reviewed HEAVILY prior to submission). It's just like the Ver Sinhaza bank with the big entrance where if you barely even click into it's area it sends you to Narnia.


This almost certainly falls under `Any movement or resizing of click zones for 3d components` in the boss fights section from https://secure.runescape.com/m=news/third-party-client-guidelines?oldschool=1

Adam-:
> I plan to remove the html from the title so it can fall into place. As for the config items, I would like to provide some kickback to as it does not do anything besides help separate the abundance of config items. It does not provide performance issues just due to colorations. It's a plugin hub item and if people don't like the attempt to assist with separating it, they can uninstall the plugin. They're most likely only going to stare at the config ONCE or TWICE on initial download to setup and then close the panel away once they're happy with the settings.

Personally I dislike this because it makes it visually inconsistent with the rest of the plugins. While the config panel is part of the plugin, its really part of the larger config system which is part of the base client. I would prefer to see it changed to be visually consistent with the rest of the configs.

PR #2948, opened 2022-07-03
Update inventory-value-overlay

wikiworm:
User reported plugin inactive. New commit addresses old runelite-client and other outdated dependencies. Replaces old gradle method testCompile with testImplementation.
Removes disabled=unmaintained flag.


Hydrox6:
Don't make new PRs if the build fails. Keep the PR open and push new commits to it.

wikiworm:
> Don't make new PRs if the build fails. Keep the PR open and push new commits to it.

Sorry about that. Changed my process.

PR #2972, opened 2022-07-07
update clan-member-list-sort

jodelahithit:
Added additional checks to counter and find specific issues.


jodelahithit:
Fix needs more work. Something changed on Jagex's side

PR #2979, opened 2022-07-08
Add hue-ambiance plugin

Jallah123:
This PR adds the hue-ambiance plugin.

The plugin will connect with Philips Hue lights and change them based on in-game events.

Currently supported features:
- Skybox
- Level up notifier
- Low hp notifier
- Low prayer notifier
- Drop notifier
  - Drops above X price
  - COX drops
  - CG drops
- Custom zulrah colors


Hydrox6:
Curious as to why this isn't an edit to the existing `phillip-hue-integration` plugin

PR #2981, opened 2022-07-08
Add hue-ambiance plugin

Jallah123:
This PR adds the hue-ambiance plugin.

The plugin will connect with Philips Hue lights and change them based on in-game events.

Currently supported features:

- Skybox
- Level up notifier
- Low hp notifier
- Low prayer notifier
- Drop notifier
  - Drops above X price
  - COX drops
  - CG drops
- Custom zulrah colors

### DIFF
--- a/plugins/phillip-hue-integration
+++ b/plugins/phillip-hue-integration
@@ -1,3 +1,2 @@
-repository=https://github.com/nganzalo/Hue_Implementation.git
-commit=f0aabcb8bc4ad53fdf6980adcee2414ed16cbec6
-disabled=unmaintained
+repository=https://github.com/Jallah123/hue-ambiance.git
+commit=c260c0fc3e4ce5cff13fed7052d92367dda01102


package/target_init.gradle
```diff
@@ -54,6 +54,7 @@ allprojects {
 	def buildDir = new File(System.getenv("runelite.pluginhub.package.buildDir"));
 
 	task runelitePluginHubPackage(type: Jar) {
+        duplicatesStrategy = 'warn'
```
LlemonDuck, line 57: Why is this included?

Jallah123: Basically when creating a fat jar it copies all the files into a single jar. Whenever there are duplicates like LICENSE or NOTICE it will break.

Now it will keep both duplicates and log a warning that it happened. This way there is no loss of files and the build will succeed.

Jallah123:
@LlemonDuck any more questions or do you think it is good to go?

package/target_init.gradle
Nightfirecat: Why is this put here (which would affect all plugins) instead of in the build script for your plugin?

Jallah123: ah thanks didn't know that was possible. fixed it in my branch, will clean up the git history when I'm home :)

Jallah123:
@Nightfirecat any other comments?

PR #2988, opened 2022-07-09
Added no blue chat categories plugin

GraysonSpencer:
Stops the chat categories from turning blue when chat is hidden and new messages appear.
Provides options to allow users to decide which chat categories to hide.


MoreBuchus:
This already exists in the Toggle Chat plugin

GraysonSpencer:
I attempted to merge my implementation with Buchu's.

PR #3004, opened 2022-07-13
Update mlm-upper-level-markers to v1.3

Cyborger1:
Added a config option to accommodate for the new ore respawn mechanics at Upper MLM.
By default, after 120 seconds of activating a marker, that marker will be deleted, now assuming the rock is fully respawned.


Cyborger1:
Please hold on this for now, there's something weird going on with onSpawn events with the rocks

Cyborger1:
This is good to merge

PR #3021, opened 2022-07-15
Add Touch Bar Plugin

PatrickKocken:
(If you have any questions about this plugin feel free to ask them in this conversation)

Runelite Touch Bar
=============

A plugin that adds Macbook Pro Touch Bar support for Runelite.

Screenshots
-----------

image
image

Config 
------

Make sure the keybinds in Runescape and in the Touch Bar Plugin config match.

image

Native Library
--------------

This plugin uses a Dynamic Link Library in order to use the Macbook's Touch Bar. The source code of the native library can be found [here](https://github.com/PatrickKocken/runelite-touch-bar-native-library).

### DIFF
--- a/plugins/touch-bar
+++ b/plugins/touch-bar
@@ -1,2 +1,2 @@
 repository=https://github.com/PatrickKocken/runelite-touch-bar.git
-commit=8ed8131a325bdd9266546d4a08c0e8dbd92ce38c
+commit=671cfc425c206b8c584d9ec42d6cdf1b6128ec89
--- a/plugins/touch-bar
+++ b/plugins/touch-bar
@@ -1,0 +1,2 @@
+repository=https://github.com/PatrickKocken/runelite-touch-bar.git
+commit=8ed8131a325bdd9266546d4a08c0e8dbd92ce38c


Adam-:
We have no existing plugins which bundle native libraries. Previously we have just not allowed it. Our issue with this is we don't have a good way to verify they aren't malicious. This is a problem for us as we've had people submit malicious code to us before for verification to try and hack players.

I am not sure what the best solution to this is. If the library is not updated too often we could vet the code and produce our own build for you to use. Or if you could show that the build was produced from actions, that might work too.

PatrickKocken:
> We have no existing plugins which bundle native libraries. Previously we have just not allowed it. Our issue with this is we don't have a good way to verify they aren't malicious. This is a problem for us as we've had people submit malicious code to us before for verification to try and hack players.
> 
> 
> 
> I am not sure what the best solution to this is. If the library is not updated too often we could vet the code and produce our own build for you to use. Or if you could show that the build was produced from actions, that might work too. 

Hey Adam,

Thank you for your response. I was aware that the native library could be seen as malicious and I completely agree with your statement. I will try to make the build process of the native library more transparent by building it with Github Actions. I will also look into possibilities to verify the integrity of the native library.

Feel free to close this PR. Thank you for your feedback and have a nice weekend.

PR #3034, opened 2022-07-18
added altar-of-net plugin

GabrielReyes136:
Clan plug in designed to "enhance the user's clan experience"
As of version 1.0.1 the following commands have been added:
!kill !event !slap !pre !bgs !help !ranks !greet !mute !beg !give !gratz

EDIT:
Originally the plugin checked to make sure that you are part of the "altar of net clan" in order to be used.

That is no longer the case, and anybody can use the features provided by the plugin regardless of whether they are in my clan or not.

In the future, members of my clan will receive unique features, but the plugin will always have features available the the general public as well as that of my clan


GabrielReyes136:
*disclaimer*
there will eventually be features on the plugin that are clan only and others that the user can use if the user happens to download it while also not being part my clan.

the documentation on my github page will make note of features that are specific to my clan only, while emphasizing the features that are designed for every other clan/user

PR #3049, opened 2022-07-21
Create shift-remapping

WhatATopic:
Allows the remapping of the shift key since key remapping has yet to add this functionality.


Bram91:
For future reference, dont close/open new PRs and just push changes to the same branch.

Nightfirecat:
[Don't use `enabledByDefault = false`.](https://github.com/WhatATopic/Shift-Remapping/blob/d137b38ab0efd837325ea96032f67a2d80313b96/src/main/java/com/shiftremapping/ShiftRemappingPlugin.java#L55) It's confusing for users to install a plugin and not have it work right away.

PR #3052, opened 2022-07-22
Update task-jingle plugin

maximonster:
Update task-jingle plugin with options for custom paths for jingles


LlemonDuck:
Needs rebase

PR #3065, opened 2022-07-25
Update music-replacer

alowaniak:
Add mp3 support, file chooser for local, and presets

### DIFF
--- a/plugins/music-replacer
+++ b/plugins/music-replacer
@@ -1,3 +1,3 @@
 repository=https://github.com/alowaniak/music-replacer.git
-commit=9cec8eb48f9097e5d8bf56c902215f5d0a9071e7
-warning=This plugin will (when using the YouTube functionality) download .wav files from (and send your IP address in the process, to) a 3rd party website not controlled or verified by the RuneLite Developers.
+commit=31cd34d417c70d89535a7b452ed40e243badb7a6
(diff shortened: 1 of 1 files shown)


LlemonDuck:
You should be using the Maven library system where possible to import libraries instead of storing entire JAR files in your repository. I'm not sure if JACo is available through any Maven repositories, but you are free under the terms to include the source fully within your own source code. This is a large library and will depend on one of our reviewers taking the time to read through that entire source as well.

Also, your use of the library currently does not abide by its LGPL licensing terms. At minimum you need to include a copy of the original licensing terms, provide attribution, and provide the source code. You also may not re-license the code under your own BSD license that you are currently using.

alowaniak:
JaCo isn't available in a maven repo AFAIK.
I took the same approach as the [fire-beats plugin](https://github.com/RKGman/fire-beats), and in fact used the same library assuming that could expedite the review since it's been accepted before.

For abiding by the LGPL licensing terms would it be enough to:
- Add a copy of the lgpl in `music-replacer` repo
- Provide attribution in the README
- Add a link to jaco's repo in the README's attribution section
?

alowaniak:
I've added a copy of the LGPL license, linked to the JaCo source and gave attribution.

(Also fixed the plugin since it broke after widget ID change last osrs update.)

PR #3073, opened 2022-07-26
Update World Heatmap (just added an image file that I forgot)

GrandTheftWalrus:
I forgot to add the osrs_world_map.png template image to /resources, so the plugin wasn't working in practice, but now it should be all good.


Hydrox6:
You need to rebase this branch off of `upstream/master`

PR #3087, opened 2022-07-27
Update NeverScapeAlone to v2.0.0-alpha

Ferrariic:
Please review this PR as if it were a new plugin, due to the number of changes that were made.

+ Added Quick Match, Search Match, Create Match
+ Added Basic Group functionality
+ Switched to Websocket


Ferrariic:
I am going to close this PR for now, as I'm working on changes which would allow for more features and better flow. These features are currently WIP -- but it would be better to release when it is all complete and stable.

PR #3090, opened 2022-07-28
Update House Advertisement Plus to 1.0.2

salverrs:
Fixed non-local adverts being assigned incorrect widgets
Fixed pinned favourites order not respecting config ordering
Fixed overlay drawing on top of the container when scrolling


salverrs:
Closing due to a new bug, will open once fixed.

PR #3092, opened 2022-07-28
Create resting

ScreteMonge:
Creates a "Resting" plugin that gives the player a resting idle, much like that in RS2/RS3, performed by either right-clicking the run orb or a player-made fire and selecting "rest." Also allows other players (using the plugin) to appear as resting as well if they use the spin emote.

### DIFF
--- a/plugins/resting
+++ b/plugins/resting
@@ -1,0 +1,2 @@
+repository=https://github.com/ScreteMonge/TestPlugin.git
+commit=16042d7eafd98e9e793c7812e09f4587bc4537c8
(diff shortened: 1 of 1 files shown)


Nightfirecat:
You need a unique 	ConfigGroup name: https://github.com/ScreteMonge/TestPlugin/blob/0c26e0f32a68552bf473e66fa53ffa2541ec84bb/src/main/java/com/example/RestingConfig.java#L7

ScreteMonge:
@Nightfirecat Thank you, should be resolved

PR #3107, opened 2022-08-02
Updated my github name

NickSpaghetti:
Updated my Github username from Zabith4 to NickSpaghetti


Nightfirecat:
GitHub redirects requests for changed usernames, so this isn't needed. Do feel free to change it when your plugin(s) next have an update though!

PR #3123, opened 2022-08-05
Update nm-autosplitter

framboid:
fixed some errors and added a description


Nightfirecat:
I don't see any error fixes, did you miss some commits here?

framboid:
> I don't see any error fixes, did you miss some commits here?

I think I got confused, it showed me commits from July so I wasn't sure how up to date the version that got added to the plugin hub was. Sorry about that, the most recent commit just added a description because I forgot to change that.

PR #3128, opened 2022-08-06
Add wintertodt-highlight

47Y:
Adds various tile highlights for events within Wintertodt.

### DIFF
--- a/plugins/wintertodt-highlight
+++ b/plugins/wintertodt-highlight
@@ -1,0 +1,2 @@
+repository=https://github.com/47Y/wintertodt.git
+commit=760aface5d3a7d7e01c5ded62059c34233f429bc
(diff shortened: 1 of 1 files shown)


Nightfirecat:
We are discussing whether we believe this would be acceptable given the client guidelines. While Wintertodt isn't much of a boss, it is technically still a skilling boss, and at any other boss these changes would not be allowable.

Nightfirecat:
Sorry for the delayed response here. It's been some time since this was discussed, but the conclusion we came to is that while it's not much of a boss in its own right, Wintertodt is still a boss, hence this would run afoul of [Jagex's Third Party Client Guidelines](https://secure.runescape.com/m=news/third-party-client-guidelines?oldschool=1), so we're sorry to say we're not accepting this submission at this time.

PR #3129, opened 2022-08-06
Update Default Tab - v1.0.1

damencs:
- allows plugin functionality in the wilderness ... still restricted for PvP worlds

### DIFF
--- a/plugins/defaulttab
+++ b/plugins/defaulttab
@@ -1,2 +1,2 @@
 repository=https://github.com/damencs/default-tab.git
-commit=675b80e634ef049f568e4b6db7e3bea1bec31cd1
+commit=08d592e92c85ac724cb4ec4e700275e10b35ab8e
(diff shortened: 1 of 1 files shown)


Nightfirecat:
I'm a bit confused. Why is it enabled in the wilderness, but not PvP worlds? How is it slower than a keypress from the user?

damencs:
> I'm a bit confused. Why is it enabled in the wilderness, but not PvP worlds? How is it slower than a keypress from the user?

@Nightfirecat - I guess we could enable it in PvP worlds as well. There's requests to be able to use this in the wilderness for alt'ing wilderness bosses but there's not really the same experience with PvP worlds - in my mind that is. - I guess if we enable it in Wildy, we could do the same for PvP worlds.

As for the delay, nevermind. I thought there was a more noticeable delay- but disregard that point.

Nightfirecat:
If it has a couple of ticks delay for wildy/pvp, that's fine.

damencs:
> If it has a couple of ticks delay for wildy/pvp, that's fine.

@Nightfirecat - Added a 3-tick delay and config option for swapping when in wilderness or pvp worlds.

PR #3131, opened 2022-08-06
Add bank-menu-entry-swapper

MarkHarms:
Adds config for the new bank-menu-entry-swapper plugin which provides shift-right click menu config capabilities within the bank interface. This works in the same manner as the new npc/inventory shift-right click menu configuration.


MarkHarms:
I've have some people recommend I make a PR for this to the actual menu-entry-swapper plugin, so i'll be trying that!

PR #3134, opened 2022-08-07
Add seaweed-spore-notifier plugin

rbbi:
Notifies the user when [seaweed spores](https://oldschool.runescape.wiki/w/Seaweed_spore) spawn. 

Useful for gathering seaweed spores while playing on other accounts.

The "Ground Items" plugin is able to notify for _all_ highlighted drops, but not specific ones (like seaweed spores).

### DIFF
--- a/plugins/seaweed-spore-notifier
+++ b/plugins/seaweed-spore-notifier
@@ -1,0 +1,2 @@
+repository=https://github.com/rbbi/seaweed-spore-notifier.git
+commit=2628c32283098c35b8108313bc5537784bcd71c0


Nightfirecat:
Is the use case for this plugin to effectively create a "highlighted items 2" list, containing only seaweed spores, so that you don't get notifications for items already highlighted via the Ground Items plugin?

rbbi:
> Is the use case for this plugin to effectively create a "highlighted items 2" list, containing only seaweed spores, so that you don't get notifications for items already highlighted via the Ground Items plugin?

Yes

Nightfirecat:
No, that doesn't really matter. Ideally we will make some changes to plugins to allow multiple/additional highlight groups or something of the sort so that different color/notification settings can be applied to them, but that isn't in the works right now.

PR #3165, opened 2022-08-12
Update 117HD to 1.1.3

aHooder:
This update is primarily to add https://github.com/117HD/RLHD/pull/20 which fixes issues with loading LWJGL in some cases.


aHooder:
Hmm, it does not like `com.sun.nio.file`. My bad, will get that sorted.

aHooder:
Should be good to go now :+1: The switch to sRGB makes darker areas slightly brighter, so we may revert that in a few days if enough people complain, even though it should be better overall.

PR #3174, opened 2022-08-13
Update NeverScapeAlone to v2.10.0-alpha

Ferrariic:
Changes:
+ Minor Refactor of UI
+ Changed Group Leader symbol to flat 2d crown
+ Changed Match Notes to show in panel without hover

image


Ferrariic:
image
Collapsible panels added in v2.9.0-alpha, works for match panels and search panels.

Ferrariic:
Resolves fatal bug when sending fancy discord names like '𝓦𝓸𝓸๔𝓼𝓴𝓲'.

PR #3186, opened 2022-08-15
Accustom mult to further dev usage

tainn:
Includes multiple changes towards the `create_new_plugin` root python script.

Consists of:
- Adhering to the PEP-8 guidelines
- Avoiding the shadowing of in-built keywords (`str`, `dir`)
- Avoiding shadowing of global variables outside of the function's scope (`strfun`)
- Replacing comparison with `None` being performed with equality operators; this change also increases performance
- Preventing potential of undefined names (`infi`)

Overall, this change slightly improves performance and alleviates bad practices -- such as undesired overshadowing -- allowing for safer continuation should the script see further development.


Nightfirecat:
An alternative which uses yapf formatting (matches https://github.com/runelite/runelite-wiki-scraper) and incorporates only the equality operator change is https://github.com/Nightfirecat/plugin-hub/commits/python-cleanup
I don't have strong feelings on this either way

PR #3190, opened 2022-08-15
Add basqh

BA-Services:
Plugin to send all kinds of information about the current state of the 'BA Services' friends chat to an external server, where the info is processed to automatically manage our online queue. Uses a key as a security measure, handed out separately to members.


raiyni:
Inject `OkHttpClient ` instead of using RuneliteAPI

raiyni:
You should also turn `log.info` calls into `log.debug` or remove them

BA-Services:
Should be okay now!

PR #3191, opened 2022-08-15
Created fight-cave-helper

runkev:
This is a panel plugin that predicts and displays Fight Cave wave spawns after a few button inputs from the user.


Nightfirecat:
FYI you can move the readme files (and any other files not actually loaded via Java code) out of your resources dir as they don't need to be packaged to be visible on the readme.

runkev:
> FYI you can move the readme files (and any other files not actually loaded via Java code) out of your resources dir as they don't need to be packaged to be visible on the readme.

So would moving them out of the src directory work or are you saying to move them outside of the repo entirely?

PR #3204, opened 2022-08-18
enriath author updates

Enriath:
It's rebrand time!
These are all the plugins without any code changes attached to them. There are another 5 with actual code changes, and there'll be another PR with those in.


Nightfirecat:
`shattered-relic-xp` didn't get a commit update, idk if that was intentional or not

Enriath:
That was intentional; it should have been disabled when leagues ended but I forgot.

PR #3214, opened 2022-08-19
Add Jekyll herb reminder v1.0.0

AmrikSD:


AmrikSD:
This started off as an addition to the random events plugin https://github.com/runelite/runelite/pull/15525 but was told it'd be more suitable as a plugin hub plugin because it's so niche

LlemonDuck:
You have both an MIT license (/LICENSE.md) and a BSD 2-Clause license (/LICENSE) in your repository. You should remove whichever one you do not wish to use (and also maybe take the time to add /bin to your .gitignore while you're at it, but not strictly necessary)

AmrikSD:
> You have both an MIT license (/LICENSE.md) and a BSD 2-Clause license (/LICENSE) in your repository. You should remove whichever one you do not wish to use (and also maybe take the time to add /bin to your .gitignore while you're at it, but not strictly necessary)

My bad - I was just fighting to get the CI to pass so thought it was beting picky about BSD seems the actual issue is that it needs /LICENSE (without .md) to work

Good spot though, is everything else in order?

PR #3215, opened 2022-08-19
Update Inventory Setups to v1.14 and Bank Tag Layouts to 1.4.18

dillydill123:
Inventory Setups:
- Added Icon Mode
- Optimized how inventory setup data is stored 

Bank Tag Layouts:
- Updated to deal with new inventory setup data layout

@geheur will confirm


geheur:
looks good to me 👍 .

PR #3219, opened 2022-08-20
AFK Movement Alerter plugin

ConorJS:
Gives a visual indicator when your character stops moving.

This can be useful if you're training skills (or doing some other activity) involving constant movement, such as Agility or Runecrafting, especially if the game is on a side monitor.

The game window will glow a color (configurable). You can also configure the speed of the glow effect, and how long your character needs to stand still for the alert to trigger. You can also configure a separate alert window when the character is at a bank (to compensate for extra time banking takes compared to just moving your character)


LlemonDuck:
This is achievable with Idle Notifier + RuneLite Notification settings, have you tried using those? Let me know if you'd still like this plugin added to hub.

ConorJS:
> This is achievable with Idle Notifier + RuneLite Notification settings, have you tried using those? Let me know if you'd still like this plugin added to hub.

I didn't know you could do that, neat.

Still, it has use because getting desktop notifications for all movement events is annoying, and it's also a loss to disable desktop notifications for all plugins just because you don't want movement idle notifiers to trigger them. The visual notification is intended to be a lower priority than other notifications, but RuneLite doesn't allow you to do this (at least, AFAICT).

Also, I find the 'breathing' visual affect of the notifier on my plugin less harsh than the flashing in RuneLite's notification settings, which I did intentionally to avoid this, I think I started with the same effect.

It's also nice being able to extend the delay when the bank is open, which my plugin allows you to do, to compensate for the extra time it takes to use a bank (as your character is stationary during).

Also, I keep catching the RuneLite alerter break like this: https://gfycat.com/grizzledimaginarygharial (I have no other notifications active in this, so it shouldn't be active while moving). The delay timing seems off sometimes too, but I'm not sure. I never observed this, or anything similar, with mine.

So in short, yeah, I'd still like the plugin added. Thanks for pointing that out though.

LlemonDuck:
Two things:
* Whatever you're doing in the overlay class and WindowUtil can be replaced with `graphics.fill(client.getCanvas().getSize())`
* Your `displayName` in /runelite-plugin.properties should be more descriptive, it is the name that appears on the Plugin Hub listing

ConorJS:
Hey @LlemonDuck - just giving you a friendly ping in case this fell off your radar

PR #3223, opened 2022-08-21
Create yewww

lafffta:
Plays Soulja Boy sound effect when chopping yew logs, also plays when someone nearby (within 5 tiles) chops a yew tree.


LlemonDuck:
Funny plugin idea, but the sound clips are not public domain and would require permission from the copyright holder to include here.

LlemonDuck:
Closing due to the above. Feel free to reopen if you want to use another (copyright-free/license-granted) sound, or some other change that addresses the above.

PR #3231, opened 2022-08-22
Update startierindicator - Fix Failed to build for hiscore api changes

pwatts6060:


Nightfirecat:
FYI changing a config's `keyName` will cause its value to be reset for users. Are you sure you want to change `xpPerHour` -> `dustPerTick`? (I'm aware the old name was incorrect for what the config should have been storing)

Nightfirecat (approved):

pwatts6060:
> FYI changing a config's `keyName` will cause its value to be reset for users. Are you sure you want to change `xpPerHour` -> `dustPerTick`? (I'm aware the old name was incorrect for what the config should have been storing)

this is fine, please go ahead with the merge

PR #3257, opened 2022-08-28
Remove abandoned and broken plugin

nofatigue:
Hey, my plugin is REALLY buggy and is not gonna be fixed any time soon... I think it's better to remove it for now.


Nightfirecat:
Instead of removing the file, add the following line to the plugin file:

```
disabled=removed by author because unmaintaned
```

nofatigue:
> Instead of removing the file, add the following line to the plugin file:
> 
> ```
> disabled=removed by author because unmaintaned
> ```

done

PR #3263, opened 2022-08-29
Taskman plugin

remcowesterhoud:
## Description
Taskman is a gamemode aiming to progress an account by focusing purely on the collection log. Items in the collection log are categorized into different tiers. A new account would generate a new task. This will randomly pick an item from the "Easy" tier. The player must now collect this item. Once the player has obtained the item he/she can complete the task and re-generate the next one.

We have a small community playing these snowflake accounts, which was started by [GenerateTask on Youtube](https://bit.ly/2vgt1Od).

### How does it work?
The Taskman community works with a Google spreadsheet. This plugin takes advantage of that. It will read/modify the spreadsheet that already exists. It does this by communicating with [a small API](https://github.com/remcowesterhoud/osrs-taskman). This API will use the Google Sheets API to interact with a player's spreadsheet, allowing it to read the current progress and generate/complete tasks. This small API is necessary as it uses a Google Service Account. The secrets to that account are configured there.

A player has to configure his spreadsheet key and a random passphrase in the Taskman plugin configurations. In addition to this the player must give spreadsheet access to a Google Service Account. The key identifies the spreadsheet so the Service Account know which spreadsheet to read/edit. The passphrase is used so not every random person with the spreadsheet key can read/edit the spreadsheet.

### DIFF
--- a/plugins/taskman
+++ b/plugins/taskman
@@ -1,0 +1,3 @@
+repository=https://github.com/remcowesterhoud/taskman-plugin.git
+commit=50266dd9b9f52bd888c32efe072d86a425406cd8
+warning=This plugin submits your Taskman spreadsheet key and passphrase to an external server not controlled or verified by RuneLite developers. This server will use this information to read/update your spreadsheet.
--- a/plugins/taskman
+++ b/plugins/taskman
@@ -1,3 +1,3 @@
 repository=https://github.com/remcowesterhoud/taskman-plugin.git
 commit=50266dd9b9f52bd888c32efe072d86a425406cd8
-warning=This plugin submits your Taskman spreadsheet key and passphrase to an external server not controlled or verified by RuneLite developers. This server will use this information to read/update your spreadsheet.
+warning=This plugin submits your Taskman spreadsheet key and passphrase, and your IP address to a server not controlled or verified by the RuneLite developers.


Nightfirecat:
This plugin needs a `warning` field as it makes web requests to `https://osrs-taskman.herokuapp.com`
[Your overlay](https://github.com/remcowesterhoud/taskman-plugin/blob/99866c88507b9c4f2078017513e9683295de2347/src/main/java/com/westerhoud/osrs/taskman/TaskmanPlugin.java#L56) must be removed on plugin `shutDown()`

Also, not a requirement, but I'd recommend [not reloading your sidebar panel on _every config change_](https://github.com/remcowesterhoud/taskman-plugin/blob/99866c88507b9c4f2078017513e9683295de2347/src/main/java/com/westerhoud/osrs/taskman/TaskmanPlugin.java#L93), but doing a check to make sure the config changed group matches your plugin's config group.

Nightfirecat:
I'd suggest following the pattern used by [other plugins in the hub](https://github.com/runelite/plugin-hub/blob/0ae471edecf9de6c117251761482dc46822fa09a/plugins/temple-osrs):

```
```

remcowesterhoud:
Thanks @Nightfirecat! I have applied your suggestions!

PR #3270, opened 2022-08-30
Added kephri-number-puzzle-solver plugin

EliasLahham:
Description:
https://github.com/EliasLahham/kephri-number-puzzle-solver

This plugin utilizes a commonly used set of tile markers for the Kephri number puzzle in the new raid and provides a solution in the chat window. Most people, including myself, have the solutions listed in a notepad/Runelite notes, so this plugin makes it a bit easier to see the solution without looking at another monitor/window.

image

image

### DIFF
--- a/plugins/kephri-number-puzzle-solver
+++ b/plugins/kephri-number-puzzle-solver
@@ -1,0 +1,2 @@
+repository=https://github.com/EliasLahham/kephri-number-puzzle-solver.git
+commit=ae80d2cefdfddca6020da284d889817dcffb5baf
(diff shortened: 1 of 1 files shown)


Adam-:
Mod Arcane has asked us to allow a months grace period without having any automatic gameplay-impacting toa plugins. So we can add this in about ~3 weeks.

EliasLahham:
> Mod Arcane has asked us to allow a months grace period without having any automatic gameplay-impacting toa plugins. So we can add this in about ~3 weeks.

Sounds good, thanks for the heads up

PR #3279, opened 2022-09-03
Add forsen-lines

Unmoon:
Plugin plays text to speech generated sound clips in reaction to certain events in (mostly) Tombs of Amascut: 
- Deaths
- Entry/exit
- Starting/completing challenges
- Activating invocations
- Party creation etc

### DIFF
--- a/plugins/forsen-lines
+++ b/plugins/forsen-lines
@@ -1,0 +1,2 @@
+repository=https://github.com/Unmoon/forsen-lines.git
+commit=2a8dc3f9d1bb64c40ece1bb69b05ac8a036311fb


Nightfirecat:
What's the source of the audio files here? It's clearly TTS--are they created by a generator which allows sharing/use of the audio in this way?

Unmoon:
> What's the source of the audio files here? It's clearly TTS--are they created by a generator which allows sharing/use of the audio in this way?

It's default Windows text-to-speech, and the files were generated using with pyttsx3 using Microsoft Speech API.

PR #3282, opened 2022-09-03
Discord Account Hook

Cyx-Development:
Sends Discord Notification with Client Authorization Key. 
For Users who can not claim Webhook URL or Bot Token.


Enriath:
This is against Discord's ToS, as it's what they call a ["self-bot"](https://support.discord.com/hc/en-us/articles/115002192352-Automated-user-accounts-self-bots-).

PR #3284, opened 2022-09-04
customizable-xp-drops to v1.5.9

l2-:


l2-:
Using client.getVar(VarPlayer) is not allowed because it is deprecated however using the successor function client.getVarpValue(VarPlayer) does not compile because it is not in release...?

LlemonDuck:
Not sure if you're mixing up some definitions here but the current methods are `getVarpValue(int)` and `getVar(VarPlayer)`, the deprecated method is `getVar(int)`.

l2-:
I am using `getVar(VarPlayer)` currently but the checks fail saying it is terminally deprecated (see first log in my earlier comment). I can see that it is marked as such on master branch https://github.com/runelite/runelite/commit/a6f1a7794979b016106a23b8a9ca3a18ad6e36d7#diff-ab3d2b8a0e4587c939f887eeaa7ccd90ec8d3eaae571e5d9f9618404f7b453a3R806. The logic i employ is copied from the attackstyles plugin and for this plugin `getVar(VarPlayer)` was replaced by `getVarpValue(VarPlayer)` yet I cannot do the same as `getVarpValue(VarPlayer)` is not in release yet (see second log in my earlier comment). I could maybe use `getServerVar(VarPlayer)` to pass the checks as of this moment but then it will require another update for next runelite release when the api will be renamed?

LlemonDuck:
I see, you can change your build.gradle to reference the snapshot version (1.8.33-SNAPSHOT), then submit a PR for the changes and we will wait to merge until 1.8.33 is released.

I'll mark this PR as hold for release now as well

l2-:
@LlemonDuck builds correctly against release now :). Also for some reason using `latest.release` or `latest.integration` as runelite version always tries to use version 1.8.33-SNAPSHOT and it always fails to fetch it. This started about 2 days ago. Is it just for me? Anyone know how to fix it?

Nightfirecat:
You probably have an old version cached locally, try clearing your gradle/maven caches.

l2-:
I tried clearing it before but i missed some files. Seems to work again thanks.

PR #3286, opened 2022-09-04
Create maxhitalert

doppy2023:
This plugin allows you to receive discord notification whenever the player gets a max hit. You can set the discord webhook url in the plugin settings.

### DIFF
--- a/plugins/maxhitalert
+++ b/plugins/maxhitalert
@@ -1,2 +1,2 @@
 repository=https://github.com/doppy2023/maxhitalert.git
-commit=c31d1836ed5f2319eacdf6a6d8bc7bc90c50a655
+commit=6f0af347319ddfb2ec7cd909e45e2bb375f46418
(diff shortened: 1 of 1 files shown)


raiyni (changes requested):
1. Use okhttp to send the request
2. Change your config group to `maxhitalert`
3. Remove calls to `log.info`

raiyni:
1. Inject our okhttp client instead of making your own
2. Use enqueue instead of execute

See https://github.com/Adam-/runelite-plugins/blob/discord-loot-logger/src/main/java/info/sigterm/plugins/discordlootlogger/DiscordLootLoggerPlugin.java for reference

raiyni:
Use clientThread invokeLater to wrap your addMessage call on https://github.com/doppy2023/maxhitalert/blob/5e8362e93c993918fe03157250fe9df1c0a477b9/src/main/java/com/maxhitalert/MaxHitAlertPlugin.java#L86 and it will be good

https://github.com/raiyni/example-plugin/blob/bank-screenshot/src/main/java/io/ryoung/bankscreenshot/BankScreenshotPlugin.java#L75

doppy2023:
thank you for your patience boss

PR #3290, opened 2022-09-05
Minecraft Eating Sound Plugin

Ellakazam:
A simple plugin that plays the Minecraft eating sound whenever you eat food


LlemonDuck:
You'll need permission from the copyright holder or the sound to be licenseable in order for us to distribute it as a hub plugin. I don't see the sound file listed on https://www.minecraft.net/en-us/attribution/sound as one you might be able to use but feel free to prove me wrong. Otherwise we won't be able to distribute this as-is (although you can submit a plugin which plays any custom user-provided sound on eating).

LlemonDuck:
Going to close this since we can't accept the plugin in its current state and it's been a few weeks. Feel free to open another PR if you change the plugin to no longer include the copyrighted materials.

PR #3298, opened 2022-09-06
Update butler plugin to point to new repo

NickMckloski:
I have created a new plugin which covers the old plugins functionality plus more. @pwatts6060 had suggested I take over maintaining the plugin instead of creating a new one. New repo https://github.com/NickMckloski/butler-info-plugin


pwatts6060:
I give permission for @NickMckloski to takeover the butler plugin

raiyni:
Change your config group to `butler-info` please

NickMckloski:
@raiyni Updated the `ConfigGroup` to `butler-info`

Nightfirecat:
[This call to `eventBus.register()`](https://github.com/NickMckloski/butler-info-plugin/blob/9fbeef6ac22f44ff20c3105d6d931c39a6169d0c/src/main/java/com/butlerinfo/ButlerInfoPlugin.java#L117) should be `eventBus.unregister()` instead.
It also looks like `shutDown()` is missing a call to `removeAll()` to ensure the infoboxes/counters created by the plugin are removed when it is disabled.

NickMckloski:
@Nightfirecat thanks for taking a look, fixed the type and added a call to `removeAll()`

PR #3299, opened 2022-09-06
Update Animation Transmog (v2.0.0)

ArtimosPenguidor:
This update adds use of a database that tracks the plugin settings of all user with the plugin installed. With this, the plugin can now detect players that are rendered, get their plugin settings, and play custom animations supported by the plugin on those players.

This essentially adds "multiplayer" to the plugin :) (Big)

I also tweaked the region blacklisting (previously added to fix a bug in ToA where the Blessed Crystal Scarab would trigger a custom teleport animation since the scarab breaking animation is the same as a teletab) to only disallow certain animations in those regions (namely the Teleport animation).

-----------------

This is a revision to my previous PR. I have made the following changes to comply with Runelite:

- Changed the package used to make HTTP requests to `okhttp3`.
- Changed the way the HTTP requests are made to not block the main thread.
- Added a config option to the plugin so that players have to opt in to using the new database feature that makes HTTP requests, with a warning on the new config toggle that states the player's game name and IP address will be collected by enabling it.

### DIFF
--- a/plugins/animation-transmog
+++ b/plugins/animation-transmog
@@ -1,2 +1,2 @@
 repository=https://github.com/ArtimosPenguidor/animation-transmog.git
-commit=3cd44adfd05b78e0700ba767084f8fca8c655654
+commit=3c861f96548c6a8cd6e6ca910f05c3170338bb7a


Enriath:
Instead of making new PRs when the build fails, just update the existing branch that the PR used; this PR didn't need to exist.

ArtimosPenguidor:
Yeah I apologize about that. After some initial discussion in the discord, I closed the PR and was just going to scrap the idea of this new feature. However after thinking about it, I made the needed changes to finish this feature in my plugin and created a new PR because I didn't want to cause any confusion by reopening the previous PR.

I'll keep that in mind in the future, thanks!

PR #3325, opened 2022-09-14
Create clock-blocker

Ditiae:
Plugin to silence specifically the grandfather clock as I find it to be extremely extremely annoying and would rather not mute all ambient sounds entirely.

### DIFF
--- a/plugins/clock-blocker
+++ b/plugins/clock-blocker
@@ -1,0 +1,2 @@
+repository=https://github.com/Ditiae/Clock-Blocker.git
+commit=37a22e97d2d59b5ff5e957b4c17feaa9d0424e21


LlemonDuck:
What is the purpose for this? https://github.com/Ditiae/Clock-Blocker/blob/37a22e97d2d59b5ff5e957b4c17feaa9d0424e21/src/main/java/com/clockblocker/ClockBlockerPlugin.java#L34-L39

Ditiae:
> What is the purpose for this? [Ditiae/Clock-Blocker@`37a22e9`/src/main/java/com/clockblocker/ClockBlockerPlugin.java#L34-L39](https://github.com/Ditiae/Clock-Blocker/blob/37a22e97d2d59b5ff5e957b4c17feaa9d0424e21/src/main/java/com/clockblocker/ClockBlockerPlugin.java#L34-L39)

I was looking to see how the music plugin handled ambient sounds and it seemed necessary, otherwise, the client wouldn't properly clear the current ambient sound list on plugin start. At least in my testing it didn't seem to want to work without adding that, if that doesn't seem expected I can do some more testing with it if necessary.

PR #3332, opened 2022-09-15
Create modicons

lewislarsen:
This plugin adds fake crowns to your local client and is intended to be harmless fun, based upon the Fake Iron Icons plugin.


LlemonDuck:
Thank you for the submission, but we are not interested in hosting this plugin.

PR #3337, opened 2022-09-16
Update tasks-tracker

tylerthardy:
Uses the new VarbitChanged API

Also uses latest.release in gradle so that we don't have to re-release for version bumps.

Diff for our changes:
https://github.com/osrs-reldo/tasks-tracker-plugin/commit/f444a124b7215cdf6563a11153d61f0fab232de8#diff-d2d7c47f16e86e69f6312203a2bbe7f2119572efdbf43768de2ac0b7d2cb065f

### DIFF
--- a/plugins/tasks-tracker
+++ b/plugins/tasks-tracker
@@ -1,2 +1,2 @@
 repository=https://github.com/osrs-reldo/tasks-tracker-plugin.git
-commit=689115c06e84ee958791e62cab52299198f0d61a
+commit=f444a124b7215cdf6563a11153d61f0fab232de8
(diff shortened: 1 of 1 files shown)


plugins/tasks-tracker
```diff
@@ -1,2 +1,2 @@
 repository=https://github.com/osrs-reldo/tasks-tracker-plugin.git
-commit=689115c06e84ee958791e62cab52299198f0d61a
+commit=f444a124b7215cdf6563a11153d61f0fab232de8
```
Nightfirecat, line 2:
```suggestion
commit=f444a124b7215cdf6563a11153d61f0fab232de8
authors=tylerthardy
```

PR #3354, opened 2022-09-20
Drop Sounds plugin

MitzDK:
The drop sound plugin reads chat messages, specifically system messages, GIM messages or Clan messages, to determine if the player for a valuable drop (if it shows up in the chat), or whether they get an "Untradeable drop", that meets the list in the config. 
This plugin was heavily based off of m0bile btw's plugin, C Engineer, and I've had contanct with him and have had plenty help from his side, to develop this. :)

### DIFF
--- a/plugins/dropsounds
+++ b/plugins/dropsounds
@@ -1,0 +1,2 @@
+repository=
+commit=
--- a/plugins/dropsounds
+++ b/plugins/dropsounds
@@ -1,2 +1,2 @@
 repository=https://github.com/MitzDK/DropSounds.git
-commit=109bf0d47e4e7779d0b4992f0f70fc3cce21afe5
+commit=037ec6f7b810664a622ce28f2026a3b50597d43a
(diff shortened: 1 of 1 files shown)


LlemonDuck:
Your build is failing since the FQDN specified in runelite-plugin.properties (`com.github.mitzdk.dropsounds.DropSoundsCompletedPlugin`) doesn't match your actual plugin class (`com.github.MitzDK.DropSoundsCompletedPlugin`)

MitzDK:
> Your build is failing since the FQDN specified in runelite-plugin.properties (`com.github.mitzdk.dropsounds.DropSoundsCompletedPlugin`) doesn't match your actual plugin class (`com.github.MitzDK.DropSoundsCompletedPlugin`)

Ty! I've tried fixing that now :) I had totally missed this

Nightfirecat:
This plugin probably should not use C Engineer's logo as its icon, especially since it's not the C Engineer plugin

MitzDK:
> This plugin probably should not use C Engineer's logo as its icon, especially since it's not the C Engineer plugin

Very good shout.. Ill fix that, another thing I totally blanked on. Will fix it immediatly.

Nightfirecat:
[Change your config group as well](https://github.com/MitzDK/DropSounds/blob/037ec6f7b810664a622ce28f2026a3b50597d43a/src/main/java/com/github/MitzDK/DropSoundsConfig.java#L11)

PR #3368, opened 2022-09-25
Update zom-ignore-list-updater

JZomDev:
Use local players around you for gametick, got spammy otherwise. Whoops. 

I didn't realize that when I had tested it. Lol.


LlemonDuck:
You should update the url to use your new GitHub name to remove the "non-author plugin change" label

PR #3369, opened 2022-09-26
Create ToACustomPresets

DenchyRS:
- Allows the user to customize the names of the presets in ToA


LlemonDuck:
Is this a feature you would be interested in building into the existing Tombs of Amascut plugin?

PR #3371, opened 2022-09-26
add alch-blocker plugin

robrichardson13:
# Alch Blocker
##### A plugin for [RuneLite](https://runelite.net/)

Allows you to block items from being alched

## Notes
- The plugin just checks the list of items configured and hides the item widgets when alching. It then unhides them after the alch cancels or completes.
- Adds quality of life (especially for UIM players), who always have near full inventories.
- This plugin can be used as a safeguard to ensure you never alch specific items. (The warning dialog doesn't have the fine controls to warn in all cases, where this plugin solves that.)

## Preview

Demo

### DIFF
--- a/plugins/alch-blocker
+++ b/plugins/alch-blocker
@@ -1,2 +1,2 @@
 repository=https://github.com/robrichardson13/alch-blocker.git
-commit=be2c3fbbaf7853d6bc67690f65c53b08751c3cda
+commit=f37c6a374debf6aa62579dfc00c8d35999d5a0ce
--- a/plugins/alch-blocker
+++ b/plugins/alch-blocker
@@ -1,2 +1,2 @@
 repository=https://github.com/robrichardson13/alch-blocker.git
-commit=d4999c8e5e266e74c8cbbb37a5de0784e83a1e59
+commit=a6cde22f30e9a3912dfe4e8fbfa5b7ae4dc1dcae
--- a/plugins/alch-blocker
+++ b/plugins/alch-blocker
@@ -1,2 +1,2 @@
 repository=https://github.com/robrichardson13/alch-blocker.git
-commit=fef10c3ce52ba9316c927798f8b0450ec476872b
+commit=be2c3fbbaf7853d6bc67690f65c53b08751c3cda
--- a/plugins/alch-blocker
+++ b/plugins/alch-blocker
@@ -1,0 +1,2 @@
+repository=https://github.com/robrichardson13/alch-blocker.git
+commit=d4999c8e5e266e74c8cbbb37a5de0784e83a1e59
--- a/plugins/alch-blocker
+++ b/plugins/alch-blocker
@@ -1,2 +1,2 @@
 repository=https://github.com/robrichardson13/alch-blocker.git
-commit=a6cde22f30e9a3912dfe4e8fbfa5b7ae4dc1dcae
+commit=fef10c3ce52ba9316c927798f8b0450ec476872b


Nightfirecat:
This plugin should reset the inventory's state if on `shutDown()` so that inventory items don't remain hidden

robrichardson13:
> This plugin should reset the inventory's state if on `shutDown()` so that inventory items don't remain hidden

Good catch.

New commit I implemented the `shutDown()` method and changed the string comparing to use `Text.removeTags()`.

robrichardson13:
- Implemented startUp and onConfigChanged methods
- Moved hiding widgets to after the inventory is drawn to solve race condition
- Using string list to search blocked items instead of for loop

pwatts6060:
I'd move the blockedItems list creation outside of the loop. Otherwise, you recreate it the list every loop iteration. I'd also consider using a Set instead of a List for constant time contains calls.

Additionally, you could keep only update the blockedItems set when the player changes the config. 
https://github.com/robrichardson13/alch-blocker/blob/fef10c3ce52ba9316c927798f8b0450ec476872b/src/main/java/io/robrichardson/alchblocker/AlchBlockerPlugin.java#L97

robrichardson13:
> I'd move the blockedItems list creation outside of the loop. Otherwise, you recreate the list every loop iteration. I'd also consider using a Set instead of a List for constant time contains calls.
> 
> Additionally, you could store the blockedItems set as a field in the plugin and only update the set when the player changes the config. https://github.com/robrichardson13/alch-blocker/blob/fef10c3ce52ba9316c927798f8b0450ec476872b/src/main/java/io/robrichardson/alchblocker/AlchBlockerPlugin.java#L97

- Changed ArrayLists to HashSets
- Sets `blockedItems` on `startUp` and `configChanged`

pwatts6060:
With the recent change of hiddenItems from List to Set. I believe there will be a subtle bug now in the loop in the showBlockedItems method. 

```
for (Widget inventoryItem : Objects.requireNonNull(inventory.getChildren())) {
	if(inventoryItem.isHidden() && hiddenItems.contains(inventoryItem.getItemId())) {
		inventoryItem.setHidden(false);
		hiddenItems.remove((Integer) inventoryItem.getItemId());
	}
}

hiddenItems.clear();
```

There's only one occurence of items in a set so if there are 2 items hidden, the set will only have the id once and in the first occurence of that item in the loop it will remove it entirely from the set. Then on the next occurence of it in the loop the hiddenItems.contains will return false and you won't do setHidden(false) on that item.
Right after the loop you clear the hiddenItems set anyway though, so you can fix this by removing the hiddenItems.remove line inside the loop.

robrichardson13:
> With the recent change of hiddenItems from List to Set. I believe there will be a subtle bug now in the loop in the showBlockedItems method.
> 
> ```
> for (Widget inventoryItem : Objects.requireNonNull(inventory.getChildren())) {
> 	if(inventoryItem.isHidden() && hiddenItems.contains(inventoryItem.getItemId())) {
> 		inventoryItem.setHidden(false);
> 		hiddenItems.remove((Integer) inventoryItem.getItemId());
> 	}
> }
> 
> hiddenItems.clear();
> ```
> 
> There's only one occurence of items in a set so if there are 2 items hidden, the set will only have the id once and in the first occurence of that item in the loop it will remove it entirely from the set. Then on the next occurence of it in the loop the hiddenItems.contains will return false and you won't do setHidden(false) on that item. Right after the loop you clear the hiddenItems set anyway though, so you can fix this by removing the hiddenItems.remove line inside the loop.

Removed!

PR #3380, opened 2022-09-27
Create dragon-wahammer-plugin

petertalbanese:
Dragon WAHammer replaces the Dragon Warhammer standard special attack sound with some fun sounds . On hit, it plays a Wario "WAH". On miss it plays a Yoshi "Mlem".

### DIFF
--- a/plugins/forsen-lines
+++ b/plugins/forsen-lines
@@ -1,0 +1,2 @@
+repository=https://github.com/Unmoon/forsen-lines.git
+commit=2a8dc3f9d1bb64c40ece1bb69b05ac8a036311fb


Nightfirecat:
FYI, we're not going to be able to distribute a plugin with copyrighted sounds. (I love the concept though)

petertalbanese:
How can I update my open PR with a fix? There was a small build error. Sorry, I haven't contributed to a public project before haha.

These .wav files were pulled from public sound bytes, are they still not allowed? :(

Enriath:
Make the change, push it to your repo, then make a new commit to this branch updating the commit.

The sound files are definitely not public domain. Even if you got them from a public site, that doesn't mean that they're free to use.

petertalbanese:
Damn I should've checked on that before I got going. I'll let the lads know :( I suppose I'll close the PR then, and maybe re-open with some different sounds.

Enriath:
alternatively, because it's going to get asked about later (we've seen it happen with similar sound plugins), you could make it load a sound from `.runelite`, like how the core custom notification sound works. That way, you don't have to distribute copyrighted content with the plugin, and the users can choose whatever they want (such as Wario saying "Oh I missed!" for missing instead)

petertalbanese:
I was actually just writing up a reply to ask about that! I'll take a look and see if I can figure that out.

petertalbanese:
While I have you, has there been any effort to port the sound list from the wiki into a huge enum or map? It could be pretty helpful to have. I had people asking me to write a generic sound swapper plugin, where they could enter a list of sounds and it would replace each with a same-name sound file in your RuneLite dir. Could be a nice, generic solution to sound swaps.

I would still like to keep this plugin, as it targets a shared sound for a specific scenario, but would be nice (and pretty easy now) to do as a second plugin.

Enriath:
Questions like that are better to ask on the [Discord](https://runelite.net/discord) really.

petertalbanese:
Sorry, I was pushing from my master branch on my forked repo and wanted to PR something else as well, so I had to close this and fixup my branches on my fork. Will open this as a new PR since the functionality changed from the original description anyway.

PR #3384, opened 2022-09-28
Create Sound Swapper Plugin

petertalbanese:
A generic Sound Swapper plugin.

Allows the user to enter a list of comma-separated sound names (e.g. `blood_cast,warhammer_crush,fire_bolt_all`) and replaces those sounds with matching `.wav` files in the RuneLite root directory (e.g. `blood_cast.wav`).

This is based on the list currently on the OSRS wiki here: https://oldschool.runescape.wiki/w/List_of_in-game_sound_IDs

### DIFF
--- a/plugins/sound-swapper
+++ b/plugins/sound-swapper
@@ -1,2 +1,2 @@
 repository=https://github.com/petertalbanese/SoundSwapper.git
-commit=fed02a22a328b2d56f04f78fd0376a5713abcf47
+commit=0e651520375bc9b12755ea39f6ad3a57ddf94d83
--- a/plugins/sound-swapper
+++ b/plugins/sound-swapper
@@ -1,0 +1,2 @@
+repository=https://github.com/petertalbanese/SoundSwapper.git
+commit=fed02a22a328b2d56f04f78fd0376a5713abcf47


Nightfirecat:
I'm not really sure why you've split SoundIds into two maps instead of just one. (You mention hitting some "max size"... Maps don't have a max size defined.)
Rather than [having users put sound files directly inside `~/.runelite`](https://github.com/petertalbanese/SoundSwapper/blob/fed02a22a328b2d56f04f78fd0376a5713abcf47/src/main/java/com/soundswapper/SoundSwapperPlugin.java#L91), create a directory under there which is specific to this plugin.

petertalbanese:
It was actually a Java issue. There is a byte limit (65535) on the amount of code that can make up any single function, object, etc. Never heard of it before this, but also never dealt with a multiple thousand line map before.

Will do for the directory. Should I be generating this directory with the plugin or just ask the user to drop the files in a directory with the plugin name?

Nightfirecat:
You can build the maps using an immutable map builder or something instead. We use that pattern in numerous places in the core client.
This directory should definitely be created by the plugin so that users can't make one with the wrong name or anything like that by mistake.

petertalbanese:
Got it, thanks!

petertalbanese:
Updated to use a folder in the .runelite directory to hold sound files.

I looked into the ImmutableMap.Builder and decided to stick with the double HashMap instead. There is no "ImmutableHashMap" and I prefer to use hashing here for 0(1) searches since the list is so long.

PR #3385, opened 2022-09-28
Update BAS Queue Helper plugin

BA-Services:


BA-Services:
Important fix to prevent the code from overloading our server

PR #3391, opened 2022-09-30
Add Ground Item Sounds hub plugin

ThePharros:


Nightfirecat:
What is the source of the packaged sounds?

ThePharros:
> What is the source of the packaged sounds?

I made them

PR #3395, opened 2022-10-01
Toa custom presets

DenchyRS:
Allows the user to create custom preset names in the tombs of amascut


Nightfirecat:
Would you be interested in contributing this to the existing [Tombs of Amascut plugin](https://runelite.net/plugin-hub/show/tombs-of-amascut)?

DenchyRS:
> Would you be interested in contributing this to the existing [Tombs of Amascut plugin](https://runelite.net/plugin-hub/show/tombs-of-amascut)?

I'd rather leave it stand alone for now but may consider merging it with the existing ToA plugin in the future :)

LlemonDuck:
fwiw, I also readded the previous custom presets feature to the ToA plugin, which allows custom names for those presets, but doesn't modify the jagex preset widgets like this one does

PR #3403, opened 2022-10-02
Update nylo-death-indicators to version 1.0.5

InfernoStats:
Update to work with the new spawn/despawn code since Nylocas respawn every tick when walking into the room.


Nightfirecat:
FYI, the change in NPC spawn events is being reverted due to some other features being broken. Will this PR need to be merged with that in mind?

nbulischeck:
> FYI, the change in NPC spawn events is being reverted due to some other features being broken. Will this PR need to be merged with that in mind?

If the change is reverted, this PR can be declined

InfernoStats:
> Our release today includes said NPC spawn event change being reverted. Now that that's happened, can you confirm whether this is still needed?

Can confirm that the issue is resolved and this PR is no longer needed.

PR #3406, opened 2022-10-02
Create glory-counter

JorVaCoding:
a small simple plugin to count how many glories you make at the fountain of rune


raiyni:
There's like 0 reason to add this plugin.

PR #3408, opened 2022-10-03
Update world-heatmap

GrandTheftWalrus:
-Significantly reduced memory usage
-customizable heatmap colour gradient sensitivity
-customizable image buffer size
-Instead of the heatmap values being stored as a big ass 2752 x 1664 int matrix, it is stored in a Hashmap<Point, Integer>, so that only the tiles that have been stepped on at all need be stored
-Fixed missing ".png" in image filenames, which causes Windows file explorer not to recognize the file as a PNG
-Hopefully, the .heatmap files of previous users, which are serialized versions of the old heatmap data structure style, are automatically converted to the new style
-Heatmap Type A's visualization is now logarithmic, as has been previously Type B
-Heatmap data can now be exported as a .CSV so users can do ye olde fancy statistical analysis in Excel or R or Quicken or whatever
-Introduced an "Are you sure?" popup for when you perhaps accidentally click the "Restart Heatmap" button in the panel
-Refactored the code so that it doesn't do the original retarded method of using the Executor via predefined Runnables

### DIFF
--- a/plugins/world-heatmap
+++ b/plugins/world-heatmap
@@ -1,2 +1,2 @@
 repository=https://github.com/GrandTheftWalrus/RuneLite-World-Heatmap.git
-commit=de1fe32c0b0daf98d168aea0ff32687718bdbd30
+commit=92d0c3d551704a232d0d6c79928ea60f2f35ce2a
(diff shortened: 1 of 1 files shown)


GrandTheftWalrus:
cancelling this PR because it turns out the included version of BigBufferedImage can't be used as-is

PR #3411, opened 2022-10-03
Update PersonalCurrencyTracker to v1.1

CWorldEnder:
- Implements coin rewards for NPC kills and gained XP
- Added more Coin Types to choose from for Infobox


Nightfirecat:
FYI I wouldn't recommend [using `assert`](https://github.com/CWorldEnder/personalcurrencytracker/compare/3d087f29ec2bd75a10051ba4afff4f0b406a4ba3...CWorldEnder:b57c8044e853ef03cce8da5292a0430df1ed1ed8#diff-d4d44efe14564bbf86aef9187be3f6cbc5c7422e16f500437f8bdc5fa9e1e880R384) as it will hang the running client if being run with `-ea`. `log.error()` with a guard clause would be a better way to ensure that case is not executed.

CWorldEnder:
Oh yeah you're absolutely right. That was just a sanity check that I overlooked. Sorry about that, I have removed the assertion

PR #3437, opened 2022-10-08
Update dps-helper 

LHPlugins:
I have updated some small bugs in the code, and added feature to skip op to 2 animation ids. This makes it possible to use spec animation without it counts.


LHPlugins:
Forgot to write this, when I push the new update. But I have added an infobox to display ticks since last attack instead of the chat, this prevents chat spamming.  Changed summary from avg ticks for x attacks to ticks lost for x attacks, to give better feedback to the player.

PR #3458, opened 2022-10-15
Add nex instant transmission plugin

HSJ-OSRS:


Nightfirecat:
As much as I enjoy the plugin concept, we wouldn't be able to distribute that audio clip.

PR #3460, opened 2022-10-15
Created Clan Broadcast Filter

andreasx23:
Filters clan broadcasts. E.g. filter PBs from clan broadcast if you find them annoying or filter level ups below a threshold!


Nightfirecat:
Please downgrade all your log messages to `log.debug()` level instead of `log.info()`, as that will clutter user's logs.

andreasx23:
> Please downgrade all your log messages to `log.debug()` level instead of `log.info()`, as that will clutter user's logs.

I have been away on vacation so I just now noticed your comment, I have done what you requested :-)

PR #3468, opened 2022-10-16
Create more-bank-fillers

UnExploration:
Adds a new plugin called More Bank Fillers. 
Adds options for overlays for bank fillers to make your bank easier to look at if you use bank fillers for organization. 

There is bank tags but I much prefer using fillers and having my sorted tabs so I created this 

Video of all the current overlay options available on the readme. 
All images were made by the linked artist's Instagram for use with this plugin.

### DIFF
--- a/more-bank-fillers
+++ b/more-bank-fillers
@@ -1,0 +1,2 @@
+repository=https://github.com/UnExploration/more-bank-fillers.git
+commit=22e94c29c2184a57ab92827aebd2d7b24de924ac


geheur:
file needs to be in the `plugins` directory.

UnExploration:
> file needs to be in the `plugins` directory.

woops

Enriath:
Don't make a new PR, just edit this one

PR #3469, opened 2022-10-17
Add fullscreentoggle plugin

willthehuman:
Add a plugin to toggle fullscreen mode.


raiyni:
What's the difference in this plugin and the one that exists already?

willthehuman:
You can toggle it via the sidebar, and it works on my steam deck. The existing one is buggy and you have to disable the custom chrome frame.

Nightfirecat:
I'd recommend you not use Ctrl + F11 as your hotkey of choice, as that's already in use by the RuneLite configuration as the default sidebar toggle key.

Nightfirecat:
Closing due to inactivity

PR #3471, opened 2022-10-18
Create more-bank-fillers

UnExploration:
Adds a new plugin called More Bank Fillers.
Adds options for overlays for bank fillers to make your bank easier to look at if you use bank fillers for organization.

There is bank tags but I much prefer using fillers and having my sorted tabs so I created this

Video of all the current overlay options available on the readme.
All images were made by the linked artist's Instagram for use with this plugin.

File is in the correct place this time 😅, had to make a new PR because the last PR could not be reopened.

### DIFF
--- a/plugins/more-bank-fillers
+++ b/plugins/more-bank-fillers
@@ -1,0 +1,2 @@
+repository=https://github.com/UnExploration/more-bank-fillers.git
+commit=22e94c29c2184a57ab92827aebd2d7b24de924ac
--- a/plugins/more-bank-fillers
+++ b/plugins/more-bank-fillers
@@ -1,2 +1,2 @@
 repository=https://github.com/UnExploration/more-bank-fillers.git
-commit=22e94c29c2184a57ab92827aebd2d7b24de924ac
+commit=e79f9ffce5f993f0613a48caf3b20b65b1259419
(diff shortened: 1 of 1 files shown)


raiyni (changes requested):
1. The config inside of the  overlay should be injected with the ctor. You can see that in core runelite
2. Make the config group `morebankfillers`

geheur:
I think this is gonna look odd with resource packs on, since it looks like you draw in the bank's brown background to cover up the bank fillers. This will look very strange if a resource pack changes it from brown. Maybe you can change the bank filler's item graphic so that it's invisible, maybe there's some unused item that doesn't have an inventory graphic that you could change bank fillers to, kind of like how unidentified herbs plugin changes herb appearances? (Just a suggestion of course).

UnExploration:
> I think this is gonna look odd with resource packs on since it looks like you draw in the bank's brown background to cover up the bank fillers. This will look very strange if a resource pack changes it from brown. Maybe you can change the bank filler's item graphic so that it's invisible, maybe there's some unused item that doesn't have an inventory graphic that you could change bank fillers to, kind of like how the unidentified herbs plugin changes herb appearances? (Just a suggestion of course).

That's a good point, I'll be honest I didn't know resource packs were a thing😅 . I think I went with an overlay because I couldn't figure out how to change the sprite/model with a custom image.  Do you know any plugins that swap models/sprites with custom images that I can reference?

geheur:
Custom images no, I don't think that's possible. What I had in mind was just removing the bank filler graphic by modifying the item composition in some way that nothing renders, and then drawing your own graphic (minus the brown background) over them.

PR #3479, opened 2022-10-21
Create low-detail-chambers

JacobLindelof:
Create a plugin for enabling low detail mode in Chambers of Xeric specifically instead of togging the entire low detail mode everywhere.


Nightfirecat:
[Don't use `enabledByDefault = "false"`](https://github.com/JacobLindelof/runelite-plugins/blob/99a61d802e8b127b412fe7205d6649ffce82fff9/src/main/java/com/lowdetailchambers/LowDetailChambersPlugin.java#L24), it's confusing for users to install a plugin which isn't turned on by default.

JacobLindelof:
All fixed. Sorry about that. Meant to remove it.

PR #3488, opened 2022-10-23
Add Jungle Japes Indicator plugin

TheKlodster:
Indicates when the banana peels have spawned and when a player steps on a banana peel during the Ba-Ba fight in the Tombs of Amascut, by playing a sound effect. There are two sound effects that are played, one is for when the banana peels spawn, and one is for when a player steps on a banana peel.


Nightfirecat:
Is there already a sound effect for bananas appearing on the floor? If not, [that would run afoul of this line of the Third Party Client Guidelines:](https://secure.runescape.com/m=news/third-party-client-guidelines?oldschool=1)

A sound effect on a player slipping on bananas should be fine regardless.

TheKlodster:
From what I tested I don't hear a sound effect played upon the bananas spawning. [Here is a clip indicating how my sound effect is played https://streamable.com/vukf8r](https://streamable.com/vukf8r)

PR #3499, opened 2022-10-26
Create PvMTickCounter

Chrisioman:
Adds total damage dealt for the local player/session

Also calculates combat ticks based on weapons used for local player as well as others in same instance.


oeg91:
Hello Would be possible to also tell how many times you did a max hit?.

Chrisioman:
> Hello Would be possible to also tell how many times you did a max hit?.

Yes, I've was able to modify the code to do this. Not sure if I should wait for review in order to push this.

Chrisioman:
Made too many modifications, will resubmit when ready

PR #3506, opened 2022-10-29
Toa points tracker

AbusiveTuna:
Tracks ToA points, based on the ToA blogpost and wiki calculator.


leejt:
Adam asked me to take a look at this before it gets merged, and I have a couple comments:

- This is also missing puzzle points, which were not really specified in the blog post but we have a decent idea of how they work
- MVP points are going to be very hard to estimate because if you want to do it right, you'll need to account for everyone else's room points
- Calculating based on hitsplats is going to be somewhat inaccurate since multiple people can get points for overkill damage (basically, damage you get XP for, but it doesn't splat because the NPC has already hit 0). If possible, I would suggest reworking to base it on XP drops rather than hitsplats
- There are a few monsters that give no points that weren't given in the blogs – a prominent one is the Ba-Ba boulders.
- Het's Seal (weakened) appears to have a space in it ingame, but not in your code
- The blog wasn't clear on this, but the 5000 points you get at the beginning are subtracted off at the end – they're only there to make early deaths more painful. I think I would suggest displaying the overlay as `totalPoints - 5000`, which means the display would start at 0. IMO that's better than the alternative of just overestimating the final points by 5000 (or subtracting off 5000 when they enter the reward room)

TylerBesecker:
> Het's Seal (weakened) appears to have a space in it ingame, but not in your code

Thanks for the follow up, currently i'm reworking it to use IDS instead of NPC names to solve some of the above issues.

> This is also missing puzzle points, which were not really specified in the blog post but we have a decent idea of how they work

Is there information for this I can find somewhere? 

>  There are a few monsters that give no points that weren't given in the blogs – a prominent one is the Ba-Ba boulders.

Is there a way you've been able to see this? I've only been going off of what was posted in the blog

> The blog wasn't clear on this, but the 5000 points you get at the beginning are subtracted off at the end – they're only there to make early deaths more painful. I think I would suggest displaying the overlay as totalPoints - 5000, which means the display would start at 0. IMO that's better than the alternative of just overestimating the final points by 5000 (or subtracting off 5000 when they enter the reward room)

Same thing as above, where can I see that happening? 

> Calculating based on hitsplats is going to be somewhat inaccurate since multiple people can get points for overkill damage (basically, damage you get XP for, but it doesn't splat because the NPC has already hit 0). If possible, I would suggest reworking to base it on XP drops rather than hitsplats

Noted, going to change this over once I'm happy with everything else as a few extra points being calc'd is fine with the lack of MVP.

Leigham:
A discussion we had on discord yesterday is about Kephri, how does it calc shield damage, are you able to farm it by letting it heal ? or does it have a total cap ? does it count the newest or oldest damage if it is capped.

GeChallengeM:
> Is there information for this I can find somewhere?

You can see the specifics by looking at how the wiki calculator was made: https://oldschool.runescape.wiki/w/Module:Tombs_of_Amascut_loot
For example, looking at line 108, it seems like it's 100 Points per Scabaras path puzzle completed.

> Is there information for this I can find somewhere?

Since the wiki calculator was ready right as jagex released the official blog, it's safe to assume they got the info to make the calculator from another source than that blog. It seems like the blog misses some of that info.

> A discussion we had on discord yesterday is about Kephri, how does it calc shield damage, are you able to farm it by letting it heal ? or does it have a total cap ? does it count the newest or oldest damage if it is capped.

Yes, you can farm it by letting it heal. Not worth doing though, looking at a loot/hr perspective.

TylerBesecker:
> You can see the specifics by looking at how the wiki calculator was made: https://oldschool.runescape.wiki/w/Module:Tombs_of_Amascut_loot
For example, looking at line 108, it seems like it's 100 Points per Scabaras path puzzle completed.

Thanks! Added a few updates already, when I get some time today I'll go through and add changes based on this.

TylerBesecker:
Took out the free 5k points
Added an MVP assumption config option while MVP is being worked on.
Fixed Wardens points going crazy
Added Palm watering points.

Next:
Add monkey room vent/pillar/dd points
Add scarab puzzle points
Add Better MVP tracking
Add better warden p2 tracking

TylerBesecker:
Added points for missing puzzle rooms, currently assumes 300 (3puzzles x 100) + 450 (75 x 6traps).  So overall this should be extremely accurate for solos, and decently accurate for teams. Few more changes and improvements could be made, but i'm happy with how this is now

LlemonDuck:
@TylerBesecker Are you interested in adding a points tracker to the existing [Tombs of Amascut](https://github.com/LlemonDuck/tombs-of-amascut) plugin to keep it all in one place? I'm sure users would appreciate having it all-in-one as well.

It would also be easier for us to review rather than a whole new plugin.

TylerBesecker:
@LlemonDuck Not at the moment, personally I like the way CoX plugins are where each functionality is a different plugin. Obviously this is a personal thing, but I find I ignore the config options of large plugins where there's just to much going on.

pwatts6060:
At some point this good integrate with the party plugin to help calculate mvp damage and communicate points among team members

TylerBesecker:
> At some point this could integrate with the party plugin to help calculate mvp damage and communicate points among team members

That's the plan! At least that's the only reasonable way I can think of calculating team MVP accurately

AbusiveTuna:
Changed my github name (and the plugin properties to match) which added the non-author plugin change label. So it can be removed if needed.

Nightfirecat:
You should update the plugin manifest file to match as well

AbusiveTuna:
> You should update the plugin manifest file to match as well

Not sure how I missed that, thanks!

Im2be:
Hey, when could we get this merged please?

LlemonDuck:
I only have two points to revisit:
* VarbitChanged is a *very* hot event, it can be called many many times per tick, and you're doing a lot in this event without actually seeming to care that a varbit has changed. At the very least it would be preferred that you use onGameTick here https://github.com/AbusiveTuna/ToA_Points/blob/b5d41944b1f54b82b03b21f2401350b325f292dc/src/main/java/com/tuna/toa/ToAPointsPlugin.java#L122
* While you're changing things, you should also consider using a matching config group name like `toapointsoverlay` or just `toapoints` https://github.com/AbusiveTuna/ToA_Points/blob/b5d41944b1f54b82b03b21f2401350b325f292dc/src/main/java/com/tuna/toa/ToAPointsConfig.java#L7

AbusiveTuna:
Updated varbitchanged to ongametick, changed config group name to match project

PR #3513, opened 2022-11-01
Create 3D-Weather

ScreteMonge:
A plugin that adds 3D weather to the game. Weather can be selected through the config or left to dynamically change. Includes ambient weather sounds


Nightfirecat:
Did you forget to include your resources when you pushed your plugin? [This seems to be trying to load images which don't exist in your repo right now.](https://github.com/ScreteMonge/3D-Weather/blob/917149e6193ed2226516f26d899f0d2f412b558c/src/main/java/com/weather3d/Condition.java#L94)
Code LGTM otherwise, just figured you would probably want to fix that before this went live.

ScreteMonge:
Can't believe I forgot to add the resources folder. Thank you very much for catching that.

Everything should be ready now

PR #3526, opened 2022-11-04
Add OSRS player count plugin

protje:
This is a simple plugin that shows the amount of players currently in the game (OSRS only):

image

To retrieve the amount of players it does the following:
- Checks if the amount of players were already fetched within the pre-defined refresh interval.
  - Minimum and default is 60 seconds, can be changed to a maximum of 3600 seconds (1 hour).
- If enough time has passed, it fetches the OSRS homepage (async) to retrieve the player count.
  - Does some web scraping to retrieve the correct amount of players.
  - Stores this amount of players in memory
- If not, it simply retrieves what is currently stored in memory.
- Shows this in-game.


protje:
Created a new PR since I removed the third party dependency as it was unneeded: #3527

PR #3535, opened 2022-11-06
Added Memoir Charges plugin

IvanovitchOS:
This plugin takes the number of charges inside your Kharedst's memoirs or Book of the Dead from the chat and displays them in an info box. It will also push a notification if your charges reach 0.


Nightfirecat:
This has been failing to build for 2 months with no follow-up.

PR #3547, opened 2022-11-09
PvMTickCounter

Chrisioman:
Adds a variety of tools to calculate individual damage dealt based of seconds or combat ticks.
Can keep track of how many max hits are landed
Also counts individual combat ticks versus others in the same instance based off weapon used.

### DIFF
--- a/plugins/pvmtickcounter
+++ b/plugins/pvmtickcounter
@@ -1,2 +1,2 @@
- repository=https://github.com/Chrisioman/PvMTickCounter
+ repository=https://github.com/Chrisioman/PvMTickCounter.git
  commit=e0d51ede94a3315532c6df3d06e3834021cdd190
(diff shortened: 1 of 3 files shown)


LlemonDuck:
You need to place your plugin file within the plugins folder, rather than in the root of the repository. Also, use lower-kebab-case for the file name.

Chrisioman:
my bad, should have doubled checked that. I think I made the appropriate changes

Nightfirecat:
Bot seems confused, closing and reopening to see if it'll wake up

plugins/pvmtickcounter
```diff
@@ -0,0 +1,2 @@
+ repository=https://github.com/Chrisioman/PvMTickCounter.git
+ commit=85ad93e11fe34a1857e21bba4bc8176f2b638e82
```
Nightfirecat, line 2: You have a leading space before each of these lines, remove that.
```suggestion
repository=https://github.com/Chrisioman/PvMTickCounter.git
commit=85ad93e11fe34a1857e21bba4bc8176f2b638e82
```

PR #3568, opened 2022-11-11
Updated Clan Spam Filter to version 1.1

andreasx23:
Clan Spam Filter now has an option to set a personal best mode. The modes exists of include all pb's except and exclude all pb's except based on a CSV list the user can enter. We now also support to exclude self which always includes own player in broadcasts. Also added an option for custom filtering in CSV format


Nightfirecat:
Instead of closing your PRs and opening new ones, please leave it open and push new commits to it

andreasx23:
> p

I just noticed it was buggy.. again.. and I have no idea how to "wait for author". I will not open a new PR before I am sure it is bug free

andreasx23:
Bugs has been fixed

PR #3570, opened 2022-11-11
Add crop-circle-tracker plugin

mattjrumble:
This plugin crowdsources crop circle tracking. Nearby crop circles are detected and reported to a backend server (default server: https://github.com/mattjrumble/crop-circle-tracker-server). The server combines sightings to calculate the location of crop circles in different worlds. The plugin gets these locations from the server and displays them in a panel so users can quick-hop to a world where a crop circle is present. Mainly useful for area-locked/snowflake accounts.

### DIFF
--- a/plugins/crop-circle-tracker
+++ b/plugins/crop-circle-tracker
@@ -1,2 +1,3 @@
 repository=https://github.com/mattjrumble/crop-circle-tracker-plugin.git
-commit=d1ef0dcd56629cd3bf5ecb5980dd0c85a1e9dd22
+commit=a7f9820ff03187f548fc26435a159b0c10380096
+warning=This plugin submits your IP address to a server not controlled or verified by the RuneLite developers.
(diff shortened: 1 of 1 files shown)


plugins/crop-circle-tracker
```diff
@@ -0,0 +1,2 @@
+repository=https://github.com/mattjrumble/crop-circle-tracker-plugin.git
+commit=d1ef0dcd56629cd3bf5ecb5980dd0c85a1e9dd22
```
Nightfirecat, line 2: Since your plugin defaults to sending data to an external server, it requires a `warning` field.
```suggestion
commit=d1ef0dcd56629cd3bf5ecb5980dd0c85a1e9dd22
warning=This plugin submits your IP address to a server not controlled or verified by the RuneLite developers.
```

Nightfirecat (changes requested):
You add a panel button [on plugin startup here](https://github.com/mattjrumble/crop-circle-tracker-plugin/blob/d1ef0dcd56629cd3bf5ecb5980dd0c85a1e9dd22/src/main/java/cropcircletracker/CropCircleTrackerPlugin.java#L104), but it is not removed on shutdown, please see to it it's properly removed then.

mattjrumble:
@Nightfirecat I've addressed your two comments - thanks for reviewing :)

PR #3574, opened 2022-11-13
add quest Splits plugin

jonas-fo:
This is a plugin for the speedrunning worlds, which allows autosplits based on items, like LiveSplits


Nightfirecat:
Don't [unnecessarily spam the logs with debug items](https://github.com/jonas-fo/quest-splits/blob/905f2ba63b019756ff4b09938eece1f25555ed07/src/main/java/com/questsplits/QuestSplitsPlugin.java#L130). If you wish to keep your logging in, use `log.debug()` instead.

PR #3590, opened 2022-11-19
Impling saver

AggieNateHarris:
- Added missing easy clue id
- Added nature implings for hard clues
- Added crystal implings for elite clues
- Added beginner clues


Nightfirecat:
You need to resolve the merge conflicts in this branch. I recommend always fetching from upstream's master branch and using that as the branch source for plugin-hub pull requests to avoid this problem.

PR #3350, opened 2022-09-20
Create hpdisplay

austinwblake:
This plugin allows you to make a real life LED health bar using an Arduino, LED strip and a few other electronic components.

PR #2795, opened 2022-06-05
update dps-calculator to v2.0.0

LlemonDuck:
Sorry :)

This rewrites pretty well the entire plugin in order to make future expansions much easier but ~~aims for feature parity only at this point~~. I recommend reviewing the plugin as an entirely new plugin, rather than the diff. I hope the code is fairly straightforward to read through, and a huge portion of the massive diff can be attributed to test cases.

The plugin now has some really solid test coverage on the compute module:  
[codecov](https://codecov.io/gh/LlemonDuck/dps-calculator)

And as a neat side effect, the plugin now allows for much more insight into result computation, like graphs:  
image

And a live overlay in-game of DPS:

Closes LlemonDuck/dps-calculator#49
Closes LlemonDuck/dps-calculator#46
Closes LlemonDuck/dps-calculator#38
Closes LlemonDuck/dps-calculator#37
Closes LlemonDuck/dps-calculator#36
Closes LlemonDuck/dps-calculator#18
Closes LlemonDuck/dps-calculator#17
Closes LlemonDuck/dps-calculator#14

PR #3596, opened 2022-11-20
Create hpdisplay

austinwblake:
This plugin allows you to make a real life LED health bar using an Arduino, LED strip and a few other electronic components.


austinwblake:
I thought I was waiting for my dependency (jSerialComm) to be reviewed. My apologies if I did not correctly add it to the third party metadata verification list correctly.

Nightfirecat:
Apologies, I missed that context from #3350. [Per Adam's comment](https://github.com/runelite/plugin-hub/pull/3350#issuecomment-1312858429), you still need to update plugin-hub's `verification-metadata.xml` in this PR, similar to how this is done in #2795, rather than being added to your repository.

Nightfirecat:
Hi, sorry for the exceptionally long delay in response on this PR. Unfortunately we're not going to accept it to the plugin hub at this time because we would need to either validate the natives' sources and build the natives ourselves in a way this plugin could make use of them, or implicitly trust that they are safe and secure. The latter is not an option we are comfortable with [given our promise to users that we have verified plugins on the hub](https://github.com/runelite/runelite/blob/312776c5fc979365ce2e4218ee9d7b289414038a/runelite-client/src/main/java/net/runelite/client/plugins/config/PluginHubPanel.java#L513-L516), and the former option would require more effort than we are able to commit at this time.

PR #3600, opened 2022-11-23
Create Level 3 Collection Log

HoorayForJay:
Hides all the log slots and categories that level 3's cannot complete.


Nightfirecat:
Your plugin manifest should be placed in the `plugins/` directory, and its filename should be all lowercase and with words separated by hyphens, not spaces. You can look at any of the existing plugin manifest files for reference

HoorayForJay:
My bad, yesterday afternoon I had never used github or java so my brain was a bit fried from all the instructions I've followed everywhere haha. Should be in the right place now.

Nightfirecat:
This has been failing to build for nearly 2 months with no follow-up.

PR #3614, opened 2022-11-30
Create daily-impling-jars

Samuel-Fox:
Plugin that send a chat message when the player can buy impling jars from Elnock Inquisitor in Puro-Puro. Works like the daily task indicator plugin.


Nightfirecat:
[You shouldn't use the config group "example"](https://github.com/Samuel-Fox/Daily-Impling-Jars/blob/a726a10b5225d8ae9d72abb0b37fc1f2e2389b22/src/main/java/com/dailyimplingjars/DailyImplingJarsConfig.java#L6). In fact, the config can be removed entirely since it is not used.

PR #3625, opened 2022-12-02
Create inventory-dvd-logo plugin

RGamberini:
Plugin that bounces your inventory around like the DVD logo.


Adam-:
we don't allow use of reflection in pluginhub plugins.

PR #3643, opened 2022-12-08
Create client snapshot plugin

PDunder:
This plugin will periodically take snapshots of client data and save it to JSON files that can be used for data analysis


raiyni:
Influxdb plugin already exists for this kind of thing and forcing people to manually parse files saved every X ticks sounds not nice

PR #3664, opened 2022-12-14
Add ImaDrum Money plugin

EwwItsMike:
A simple plugin that plays a random sound clip of content creator ImaDrum when you receive a good drop.

### DIFF
--- a/plugins/imadrum-money
+++ b/plugins/imadrum-money
@@ -1,2 +1,2 @@
 repository=https://github.com/EwwItsMike/ImaDrum-Money.git
-commit=1c69f4f37c4a8e5d203f32dcec7cb050d03e5fc3
+commit=a27356976d7ac4b7780c8ca7d6bfa1080d3a8259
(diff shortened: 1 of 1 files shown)


Nightfirecat:
Do you have permission from ImaDrum to use their voice clips in this plugin?

EwwItsMike:
> Do you have permission from ImaDrum to use their voice clips in this plugin?

I do! He recorded them specifically for this plugin and sent them to me. Need me to provide any proof of that?

Also, seems like build failed, did I mess up something simple?

Nightfirecat:
> Need me to provide any proof of that?

Yes please.

Your build is failing because your plugin properties file has `plugins=com.imadrummoney.MoneySoundPlugin`, but your plugin is actually located at `com.imadrummoney.ImadrumMoneyPlugin`

EwwItsMike:
Updated plugin properties file in the plugin repo, and updated commit hash in this fork.
Hopefully enough proof of Drum providing the sound clips himself for this plugin: https://gyazo.com/4a49736d51708eb3d282a3a176ae2d41
If this is not sufficient, feel free to @ me or DM me on discord (Ew Its Mike#1946) so I can get it sorted out

Nightfirecat:
The "bruh.wav" sound is something we can't distribute without permission from the maker of the sound. (it looks like your plugin doesn't actually use it at all? should be able to remove it without changing the behavior at all) The rest looks fine to me.

PR #3675, opened 2022-12-16
Add Moleman Mode

JZomDev:
Snow flake account game mode.

How to play: 
You have a timer. The timer decrements based on where you are at. 
Spending time below ground and gaining XP, you gain time to spend above ground. The amount you gain per threshold is customized.
There are options to manually give yourself or remove time to spend above ground. 
Time is tracked in ticks.

The tracking of "below ground" is done by checking the X,Y of the player as *most* things that are below ground are NORTH of the wildy so it checks that. 

Can manually add below ground zones for where the above logic fails and can manually add above ground zones for when you're underground but it doesn't fit the normal scenario. 

Looking to gather feed back from anybody but this plugin was made for a content creator, so I don't expect it to be wildly popular.

### DIFF
--- a/plugins/zom-moleman-mode
+++ b/plugins/zom-moleman-mode
@@ -1,0 +1,2 @@
+repository=https://github.com/JZomerlei/zom-external-plugins.git
+commit=66b8c831a3e9070181f8dadc006eecbd9cf6640d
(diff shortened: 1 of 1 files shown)


plugins/zom-moleman-mode
```diff
@@ -0,0 +1,2 @@
+repository=https://github.com/JZomerlei/zom-external-plugins.git
```
Nightfirecat, line 1:
```suggestion
repository=https://github.com/JZomDev/zom-external-plugins.git
```

PR #3683, opened 2022-12-19
Update Alternate MLM Overlay, no longer unmaintained

BuzzfeedBlue:
was unmaintained due to update elsewhere, now working with current version


Nightfirecat:
Plugin class names must be unique, so you can't use `MotherlodeMine` because the core plugin uses the same class name. Also, the overlay class name might cause issues if it is the same as the one in core as the overlay configs for storing position are based on the overlay's class name.

BuzzfeedBlue:
Ive changed things and checked that its still working locally, if I still haven't done whats needed please direct me because I'm not very experienced and may have misinterpreted what needed renaming.

PR #3691, opened 2022-12-22
Update bot-detector to v1.3.9.1

Cyborger1:
- Fix for broken Manual Detect functionality: API call was missing a parameter indicating the upload was manual, so the API treated it as an automatic upload.


Cyborger1:
Closing as we decided to change the API instead to accept the data format from the plugin directly.

PR #3693, opened 2022-12-22
Add fire-timer plugin

autumn-smellegy:
This plugin creates a timer over player-created fires in game. It counts in ticks starting at 200. It changes color after 100 ticks have elapsed to indicate that the fire can now disappear at anytime. The timer colors are configurable.


raiyni:
Remove your `log.info` calls

autumn-smellegy:
> Remove your `log.info` calls

done @raiyni

PR #3694, opened 2022-12-22
Adds max-skill-trim plugin.

NathanQuayle:
Simple plugin to add a trim (PNG border) to all skills that a player has achieved 99 in.

Adds buttons in the panel to direct users to a separate branch where they can install/download new trims.

### DIFF
--- a/plugins/max-skill-trim
+++ b/plugins/max-skill-trim
@@ -1,2 +1,2 @@
 repository=https://github.com/NathanQuayle/max-skill-trim.git
-commit=8401772feb4d93eaaccf8395d6c42a7bf99e5aaf
+commit=8d88bab4ff043f96270d6baeef2cce72e760306c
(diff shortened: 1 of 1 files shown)


Nightfirecat:
[This instance of `ImageIO.read()` must also be synchronized](https://github.com/NathanQuayle/max-skill-trim/blob/8d88bab4ff043f96270d6baeef2cce72e760306c/src/main/java/com/maxskilltrim/MaxSkillTrimPlugin.java#L104), like the one you have in your plugin startUp(()

NathanQuayle:
> [This instance of `ImageIO.read()` must also be synchronized](https://github.com/NathanQuayle/max-skill-trim/blob/8d88bab4ff043f96270d6baeef2cce72e760306c/src/main/java/com/maxskilltrim/MaxSkillTrimPlugin.java#L104), like the one you have in your plugin startUp(()

Thanks, @Nightfirecat. 

I _think_ I've done that now - my Java is a bit slacking, sorry.

PR #3697, opened 2022-12-23
Add remaining-casts

salverrs:
Add new plugin: Remaining Casts

- Shows the number of remaining casts through tooltips and infoboxes
- Tracks multiple spell costs at once
- Can filter spells affected using whitelist/blacklist


salverrs:
Sorry about the amount of commits, it's in a stable state so I will hold off adding any new features for now.

PR #3703, opened 2022-12-24
Create wiki-dialogue

AshleyThew:
This plugin allows for easier copying of text from chat for external use, such as the wiki.

Java-WebSocket dependency to talk to an external editor to automatically export text.


Adam-:
1) the "wiki" is sort of an officially recognized thing, and we don't think you are affiliated with them since they don't know who you are. Maybe name this something else like dialog-transcriber
2) instead of adding a new dep for websockets, prefer using the okhttp client's websocket support if possible. It is already a dependency.

AshleyThew:
> 1. the "wiki" is sort of an officially recognized thing, and we don't think you are affiliated with them since they don't know who you are. Maybe name this something else like dialog-transcriber
> 2. instead of adding a new dep for websockets, prefer using the okhttp client's websocket support if possible. It is already a dependency.

1. I can change the name no problem, the idea was to put much of the extracted dialogue onto the wiki in some form.

2. The reason for the package is to host a websocket server. The external editor is a react application which is unable to host it's own websocket, otherwise would require a third party middleman websocket server.
The build failed for some reason, I attempted to add the correct packages, but the latest build still refused it. Any suggestions here?

Adam-:
The wiki already has a bunch of existing tooling for this, including other plugins. If you really want to do that, it may be best to coordinate with them. You appear to be rewriting large potions of most of the verification files which is probably why the build is failing. You can look at the "files changed" tab and see what it is you have submitted.

PR #3711, opened 2022-12-29
New plugin Untrimmed Warning

pelletier2017:
Helps player protect their untrimmed cape

- Overlay popup when getting close to a new 99 to remind player to hide their untrimmed cape.

- Overlay also describes where the cape is located and whether that spot is safe or not according to the wiki.

- Hide left click pickup for untrimmed cape to avoid auto-trim on pickup.


raiyni:
Use a dash in the config name instead of a period

pelletier2017:
I could write a github action build that can verify any conventions you guys have if you would like it.

PR #3727, opened 2023-01-02
Create xp-relic-sound plugin

dabolink:
Add Plugin that has a random chance of playing the leagues relic sound each time you gain xp.
- how often the sound triggers is configurable as well as if maxed skills should be used


Nightfirecat:
[Don't use "example" as your plugin's config group](https://github.com/dabolink/example-plugin/blob/afdb3be8c7cdc446c3cac405b47d61e34a9ec5d2/src/main/java/com/dabolink/XPRelicSoundConfig.java#L7)

dabolink:
updated config group to xprelicsound

PR #3750, opened 2023-01-10
Create Discord Multi Notifications

NafizHaque:
This Plugin Provides the ability to parse multiple Discord server webhooks to simultaneously send multiple notifications at a time. 

Features include: multiple webhooks for death screenshots and multiple webhooks for Collection log slots and Loot!

This is intended to provide an All-in-One experience for the user that would want their notifications posted across multiple servers they are apart of.

This eliminates the need for any redundancy as many users use multiple plugins with the same features to have the ability to post to multiple servers.


Felanbird:
Is there any feature your plugin adds that [Dink](https://runelite.net/plugin-hub/show/dink) does not already have? 
We've been trying to truly make a centralized notification plugin, and we already support your main feature of sending notifications to multiple webhooks, for all of our notification types.

NafizHaque:
Ah yes i see , currently the plugin does not add anything unique to set it apart from Dink at this current stage. Thank you for the comment!

PR #3760, opened 2023-01-12
Update Unpotted reminder to 1.0.4

AnkouOSRS:
- Added optional cooldown for notifier.
- Only alert for imbued heart when it is actually available. Credit to JuliusAF on GitHub for most of this logic.


Nightfirecat:
FYI there is a varbit which tells you how long until you can re-activate the imbued heart. You can see it referenced in the core timers plugin. If you'd rather I merge this as-is, I can do that, but I figured I'd let you know about that first.

AnkouOSRS:
Ah that's great to know. Using that would simplify the code quite a lot so I'll make that change whenever I get around to it. I can raise another PR for that change when it's done. Thank you for providing a helpful suggestion instead of just merging and calling it a day!

PR #3764, opened 2023-01-13
Fix slash swapper

JZomDev:
Apparently the original owner left the project.


Nightfirecat:
@pwatts6060 I know you said as much in #3752, but I assume you're cool with Zom taking ownership of this plugin? From what I see here, it'll be with an entirely fresh codebase

pwatts6060:
> Provided it works yes that is fine

pwatts6060:
Tested it now and it works well. Haven't tested for group ironman though who may have slightly different behavior.

JZomDev:
1.9.8 is now out, so this is ready for release.

PR #3767, opened 2023-01-14
Fun Sounds Plugin

Defqonlord:
Just a plugin that plays sounds when certain items are dropped and when certain NPCs spawn.


LlemonDuck:
It looks like you are using a lot of copyrighted sounds, which we cannot distribute on the plugin hub. One such way you can retain the features of the plugin without this is by allowing users to provide their own sound files and not distributing them as part of the plugin.

LlemonDuck:
Closing as stale since it has been over a month. Feel free to open a new PR when the above comments are addressed.

PR #3771, opened 2023-01-14
Add visible wealth plugin

Emmsii:


Nightfirecat:
I'm concerned that this plugin would be used by scammers or lurers to scout potential targets.

Emmsii:
Is it not similar to the [equipment inspector plugin](https://runelite.net/plugin-hub/show/equipment-inspector) which allows you to see and value the equipment individual players are wearing?

For example
image

Felanbird:
Being able to see someones wealth and immediately being told when a rich person walks by, are 2 completely different things.

Emmsii:
Ah that's very true. Fair enough

PR #3773, opened 2023-01-14
Add ezhopper plugin

i:
Adds ability to type `!hop 420` in chatbox to easily switch between worlds.


Enriath:
You can already do `::hop 420` for the exact same effect

PR #3774, opened 2023-01-14
Create Max Hit sound effect plugin

maximonster:
Created a plugin that plays a sound effect when you hit your max hit


maximonster:
Superior version Hit Sounds already exists

PR #3783, opened 2023-01-17
Create Boss Name Replacer Plugin

Mrnice98:
replaces bosses names with funny things, currently replaces Phantom Muspah with 'The Grumbler'


LlemonDuck:
Remove the config if you are not using it yet.

You can re-add it later when you are using it.

PR #3784, opened 2023-01-17
wilderness-clan-callouts

Gerbzz:
A RuneLite Plugin That will automatically type the name of a person you are 'Attacking' Or 'Following' When The Plugin is On and you press the hotkey.


JZomDev:
This is auto typing, which is botting and not allowed.

PR #3813, opened 2023-01-22
Comment out prints

TicTac7x:


LlemonDuck:
You should replace these [println calls](https://github.com/TicTac7x/runelite-plugins/compare/deff04ebfdfe9dfd5722c174c04f2e8887bc5edf...TicTac7x:90cb0cfbaa96b99c530e56f72110f1d594960293#diff-e5a09f21db8fa1a8631273b75e83e85b45c2abb9778c19fa76614de127e828d4R154-R160) with proper slf4j logging. In this case, remove your debug variable and use `log.debug("MESSAGE | type: {}, message: {}, sender: {}", event.getType().getName(), ...)`

TicTac7x:
I prefer to use unreachable "sout" instead, so i can debug my plugin without needing to enable the debug argument, which would mean lots of bloat logs from others plugins. If this is an issue, i can comment these lines out...

PR #3817, opened 2023-01-23
Scowled plugins

JaccodR:
Time alive plugin; shows how many seconds an NPC has been alive for within the players view. Useful for various skilling activities such a chinchompas to see how long they have been alive for.


raiyni:
Don't close and open new prs. Push to the same branch in this pr.

LlemonDuck:
This plugin can be used to give additional information to the player in ToB, which breaks a few things in the [Third Party Client Guidelines](https://secure.runescape.com/m=news/third-party-client-guidelines?oldschool=1). So we cannot accept the plugin as it is now.

JaccodR:
@LlemonDuck Hi, I have a quick question;
I have read the third party guidelines before making the plugin, but I didn't really see anything that it wouldn't comply with, apart from maybe "Indicates the time where a boss mechanic may start or end". Although the plugin realistically could only give an advantage in the Nylocas puzzle room (not boss) which is not really a boss fight nor do they have 'mechanics' I thought this would be fine.

Either way, do you think if I for example changed the plugin to instead of having counter above NPCs, have tile indicators under given NPCs, and the colors change based on tick thresholds for how long they've been alive for, would that comply, or is nothing of the sort allowed? Thanks in advance.

Nightfirecat:
This would be an indicator for boss mechanics at both bloat and nylocas in tob.

PR #3830, opened 2023-01-26
Regular expression item counter

jammy-dodgers:
Counts the number of items in your inventory / equipped that matches regular expressions.
An example of a place where this is useful: 
You need 1250 more mole parts of either kind to finish making all your saradomin brews. 
You can match on "Mole (claw|skin)" and it will display the total number of things in your inventory that match the regex
(Though I didn't get a screenshot of that specific example, I'm sure you get the idea of it)

alt text


jammy-dodgers:
https://github.com/runelite/plugin-hub/pull/3848

PR #3836, opened 2023-01-28
Add 1Password login plugin

elgbar:
Log into your account using the 1Password Command Line Interface (CLI).
This plugin is using the 1Password Command Line Interface (CLI) to enter your password for you, just
like it is done in with the 1Password browser plugin or mobile app.

## Security considerations

* Your user ID and the item ID will be uploaded the runelites server.

1Password is designed to minimize the damage to its user in the case of a breach of an attack.
As only the IDs of the account and the item are stored using this plugin will leak no usable
information to a potential hacker.

* You must trust the author of this plugin to not include malicious code

As long as you are using the official runelite client, you will be using a plugin
reviewed the runelite team.
Essentially, if you trust the other community plugins to not be malicious, then you should trust
this plugin.

### DIFF
--- a/plugins/1password-login
+++ b/plugins/1password-login
@@ -1,0 +1,2 @@
+repository=https://github.com/elgbar/1password-login
+commit=675c732680fd6a38ef58b14900bfd10662fd53f1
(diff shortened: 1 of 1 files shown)


raiyni:
onepassword-plugin already exists

elgbar:
Oh, it didn't when I made the initial version of this plugin two months ago. Well, I guess this can be closed then.

PR #3837, opened 2023-01-28
Add npc-collection-log plugin

ElOsoGroso:
NPC Collection Log

This plugin shows a panel when you initiate combat with any NPC, showing information about which items are possible to receive from them, and which ones you've already got in your collection log, or in your bank.

### DIFF
--- a/plugins/npc-collection-log
+++ b/plugins/npc-collection-log
@@ -1,0 +1,2 @@
+repository=https://github.com/ElOsoGroso/NPCCollectionLog.git
+commit=0b93eb20bcac598158feab34da672ba557607d7b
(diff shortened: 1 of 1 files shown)


raiyni:
Would be helpful if you scaled down the icon size.

ElOsoGroso:
> Would be helpful if you scaled down the icon size.

The nav-icon.png in resources or icon.png? I assume the nav-icon one. Yeah I can do that.

ElOsoGroso:
> Would be helpful if you scaled down the icon size.

That's been resized.

plugins/npc-collection-log
```diff
@@ -0,0 +1,3 @@
+repository=https://github.com/ElOsoGroso/NPCCollectionLog.git
+commit=988dce503948f84f6d9f8826309ead808c3a3294
+
```
Nightfirecat, line 3:
```suggestion
warning=This plugin submits your account hash and IP address to a server not controlled or verified by the RuneLite developers.
```

Nightfirecat (changes requested):

ElOsoGroso:
After this merge... it seems to be immediately crashing upon login or installing the plugin. Locally I don't see the same behavior at all @Nightfirecat

PR #3848, opened 2023-01-31
Create regular expression item counter 

jammy-dodgers:
Sorry for duplicating this, Github didn't warn me that renaming the branch on my fork would automatically close the pull request. Couldn't find a way to change the PR to the newly renamed branch.

Counts the number of items in your inventory / equipped that matches regular expressions.
An example of a place where this is useful: 
You need 1250 more mole parts of either kind to finish making all your saradomin brews. 
You can match on "Mole (claw|skin)" and it will display the total number of things in your inventory that match the regex
(Though I didn't get a screenshot of that specific example, I'm sure you get the idea of it)

alt text

### DIFF
--- a/plugins/item-counter
+++ b/plugins/item-counter
@@ -1,2 +1,2 @@
 repository=https://github.com/jammy-dodgers/item-counter.git
-commit=2879eade523ee4abe61c68aa973187f4513c8029
+commit=8fc90c96cc7cacfeb9557e9463d7d1086d4f1af7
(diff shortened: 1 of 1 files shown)


raiyni:
Remove the `log.info` calls

raiyni:
There's really no need for you to do this btw https://github.com/jammy-dodgers/item-counter/blob/b514be630d9ec95733c459d8f08951de02665360/src/main/java/cc/jambox/ItemCounterPlugin.java#L140

jammy-dodgers:
> There's really no need for you to do this btw https://github.com/jammy-dodgers/item-counter/blob/b514be630d9ec95733c459d8f08951de02665360/src/main/java/cc/jambox/ItemCounterPlugin.java#L140

It's creating a type-safe generically typed array without compiler warnings. I don't think there are other ways to do this in Java because of type erasure

raiyni:
I'm tell you that you don't need generics for two Item arrays.

jammy-dodgers:
I was writing it with the intent of being reusable, but yeah I guess so

PR #3849, opened 2023-01-31
Create any-rsn-hider

jammy-dodgers:
Requested.
I'm sure it could be pretty useful for content creators, too. Like when you make RSMVs.


raiyni:
We aren't really interested in having a plugin that potentially allows impersonation for other people on streams/videos.

PR #3852, opened 2023-02-01
Updated Guild Spam Filter

andreasx23:
We now support PK broadcasts with a GP treshold and users can now decide which collection log pages they want to remove from broadcasts


andreasx23:
Ended up removing the package "Jackson" and just parse the collection log data myself from a giant string. The data is scraped from the wiki and stored within the CollectionLogHandler class

Adam-:
> Ended up removing the package "Jackson" and just parse the collection log data myself from a giant string. The data is scraped from the wiki and stored within the CollectionLogHandler class

The preferred way of parsing json is using Gson fyi.

andreasx23:
> > Ended up removing the package "Jackson" and just parse the collection log data myself from a giant string. The data is scraped from the wiki and stored within the CollectionLogHandler class
> 
> The preferred way of parsing json is using Gson fyi.

I tried using gson but it kept making my arraylist of data null but Jackson didn’t. But ultimately I ended up just handling it myself :-). Thanks for the heads up though :-D

PR #3853, opened 2023-02-01
Added Splitter plugin

gustav901:
Splitter adds a chat command to get the split amount of the gold (coins or platinum) in the players inventory or by searching by item.
Functionallity does NOT overlap with Loot Splitter.

### DIFF
--- a/plugins/splitter
+++ b/plugins/splitter
@@ -1,2 +1,2 @@
 repository=https://github.com/gustav901/splitplugin.git
-commit=a43fcf460bc58afcf8a726c76b85199b03993bea
+commit=cec9bb5e5bf71d116b2bfb43675d32e1bc8337cd
--- a/plugins/splitter
+++ b/plugins/splitter
@@ -1,2 +1,2 @@
 repository=https://github.com/gustav901/splitplugin.git
-commit=197ab26cef22d8a0c11dea8176bfab2c78ce1f96
+commit=54225f11b79d665fc8401f8470db7b00f68fd206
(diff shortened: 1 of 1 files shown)


Enriath:
Given that chat commands are for sharing information, and this doesn't share information, this should be a `::command`. if you intended it to share information, you've not done that correctly.

gustav901:
> Given that chat commands are for sharing information, and this doesn't share information, this should be a `::command`. if you intended it to share information, you've not done that correctly.

I did not consider if it should be shared or private information. Personally I would only use it for personal information but I could see how it might be good to have it as public information.
Could you point me in the right direction to read more about `::command`?

Enriath:
I mentioned it since quite often, newer devs will make `!commands` either for things that are just for the local player, which spams up chat, or they assume that it'll magically share to other clients despite them not implementing a server component for it.

`::commands` are awesome since they never even get into the chatbox, let alone to other clients; perfect for what you want to do.
Have a look at usage of the `CommandExecuted` event in the core client. Notable plugins that use it are Devtools and Clue Scroll, who both have dev commands that are *somewhat* simple.

gustav901:
> I mentioned it since quite often, newer devs will make `!commands` either for things that are just for the local player, which spams up chat, or they assume that it'll magically share to other clients despite them not implementing a server component for it.
> 
> `::commands` are awesome since they never even get into the chatbox, let alone to other clients; perfect for what you want to do. Have a look at usage of the `CommandExecuted` event in the core client. Notable plugins that use it are Devtools and Clue Scroll, who both have dev commands that are _somewhat_ simple.

Alright, thanks! I'll look into and update it to use `::commands`.

gustav901:
Changed from "!command" to "::command".

Only other change was to add an error message if no item was found (instead of having the message simply diappear).

raiyni:
You need to change your config group and remove `log.info` calls

PR #3865, opened 2023-02-04
Create discord-chat-logger

PJGJ210:
Plugin will output private messages/group messages currently to a discord webhook.

### DIFF
--- a/plugins/discord-chat-logger
+++ b/plugins/discord-chat-logger
@@ -1,0 +1,2 @@
+repository=https://github.com/PJGJ210/Discord-Chat-Logger.git
+commit=45e49f5bc7c97eb4ae8e9b0b37bb0131944e003e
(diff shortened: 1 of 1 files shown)


Felanbird:
As much as we would like to centralize discord-related plugins, this specific one does not feel like it fits well into Dink.

1. Likely this feature will expand to needing a single webhook for each message type, for cleanliness sake, which from looking at the README will end up being about 5 separate webhooks. 
Dink has the functionality to use 13 separate webhooks for 13 different notifiers, if the user wants to, and this would eventually add 5 for a single one.
2. As far as I'm aware mirroring chat to discord sounds like it's going to be awful to do, but if the plugin is accepted I wish you luck doing it. 👍

PJGJ210:
Intention here is primarily for group chat etc. The other variants I would
only program for the niche use cases. Obviously per Readme 'all' chat would
be way too much.

If there is another place for this (or none at all) I can accept that.

Honestly wish there was an osrs chat similar to many other games.

Group chat helps other members of group/clan keep included while some
member is online. Else private for keeping tabs/log. There are already
those that write to file. In particular for those messages lost while doing
spammy activities.

PR #3868, opened 2023-02-04
Create did-I-compost plugin

AbusiveTuna:
Adds a compost bucket over patches you've composted. So you don't try to add compost to them 16 times in a row.


raiyni:
Include your image in the resources folder and read it like every other plugin does

AbusiveTuna:
> Include your image in the resources folder and read it like every other plugin does

Updated to correct way. Just based it on how quest helper pulls them so hopefully that's right.

PR #3877, opened 2023-02-05
Add Hold Left Click

Jbleezy:
Reclicks once every tick while holding left click

Doesn't move the mouse at all, just simply reclicks without having to lift finger up and down to decrease repetitive strain injury


Adam-:
automating input is against TOS

PR #3897, opened 2023-02-10
Submission of plugin: email-notifications

boiwantlearncode:
Sends an email on notification event. Sender email has to be hosted on gmail.


dekvall:
We do not want to handle email credentials.

PR #3898, opened 2023-02-10
For email-notifications plugin: Added third party dependencies

boiwantlearncode:
Added third party dependencies to plugin-hub repo and removed the jar files in the email-notifications plugin repo. Updated metadata as per https://github.com/runelite/plugin-hub/blob/master/README.md.


Adam-:
We don't want to be storing/handling gmail credentials

PR #3902, opened 2023-02-11
Update cox-qol

MoreBuchus:
+ Updated prayer enhance formula to be accurate for all prayer levels (Thank you, Nick Ints and Mod Ash)
 + Added CM thieving chest group highlights
 + Added party grub counter
 + Added prep overlay - all configs to 0 does not display the overlay
 + Added CC warning - Highlights CoX entrance if you are not in a cc or if you are in a cc without a party being created (Made by Deo)


MoreBuchus:
> It doesn't look like the newly-added infoboxes are removed on plugin shutDown(), please explicitly remove them so that disabling the plugin doesn't leave infoboxes persisting

Updated it. Thank you

PR #3907, opened 2023-02-12
Create hide-cast-on-pets

jammy-dodgers:
I got tired of the game trying to cast spells on my pets instead of the enemies walking near. This hides cast options on all pets.
image


Nightfirecat:
This violates [Jagex's Third Party Client Guidelines](https://secure.runescape.com/m=news/third-party-client-guidelines?oldschool=1), so we can't accept it. Specifically this line:

> Makes it easier to target 3D entities with a spell by removing some options

PR #3919, opened 2023-02-14
Toggle plugin hotkeys

dampfhub:


antero111:
You can achieve this already with Plugin Presets. See https://github.com/antero111/plugin-presets/wiki/General-Features#keybinds

dampfhub:
> You can achieve this already with Plugin Presets. See https://github.com/antero111/plugin-presets/wiki/General-Features#keybinds

Ah nice, much more feature complete too. Closing.

PR #3926, opened 2023-02-17
Create tzhaar-death-indicator

JaccodR:
This plugin makes NPCs within the Inferno & Fight caves dissapear on xp drop, which is a lot faster than with entity hider. This plugin works the same as the Nylo death indicator.


JaccodR:
Apparantly others have been working on a similiar plugin that works everywhere, so will close this and let them finish theirs.

PR #3927, opened 2023-02-18
Afk threshold plugin

Velite12:
This plugin allows you to AFK while skilling by setting a minimum time you want to be AFK for and sends you a notification if you are idle at that time, or if you are not idle, it will wait until you become idle and then send you a notification. This plugin basically combines the AFK Timer plugin and the Idle Notifier plugin into one for specifically skilling activities, but also allows for more time to AFK after the threshold. 

For instance, say you set the threshold to be AFK to 30 seconds. You mine something and are done in 15 seconds, and idle. The plugin would wait until 30 seconds, see you are idle and notify you. You then mine something else and it takes 45 seconds. The plugin checks at 30 seconds, sees you're still mining, and waits until you become idle to notify you, that way you always have a minimum time to AFK before receiving a notification, and if it takes longer you're only notified until you need to take action.

The motivation behind this plugin was to create a way to train skills which have variable completion time for actions (gathering skills) while focused on other tasks such as work or school, to match the AFKness of skills which have fixed completion times (most bankstanding skills). The existing plugins available did not meet the needs of this problem, so I created this one. The AFK Timer plugin does a better job than the Idle Notifier plugin, but I found that too often the regular interval checks meant I didn't need to actually do anything sometimes and I could have waited longer to respond.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,7 +1,7 @@
 ![](https://runelite.net/img/logo.png)
 # plugin-hub [![Discord](https://img.shields.io/discord/301497432909414422.svg)](https://discord.gg/mePCs8U)
 
-This repository contains markers for [RuneLite](https://github.com/runelite/runelite)
+This repositorya contains markers for [RuneLite](https://github.com/runelite/runelite)
 plugins that are not supported by the RuneLite Developers. The plugins are
 provided "as is"; we make no guarantees about any plugin in this repo.
 
--- a/README.md
+++ b/README.md
@@ -1,7 +1,7 @@
 ![](https://runelite.net/img/logo.png)
 # plugin-hub [![Discord](https://img.shields.io/discord/301497432909414422.svg)](https://discord.gg/mePCs8U)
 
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -1,7 +1,7 @@
 ![](https://runelite.net/img/logo.png)
 # plugin-hub [![Discord](https://img.shields.io/discord/301497432909414422.svg)](https://discord.gg/mePCs8U)
 
-This repository contains markers for [RuneLite](https://github.com/runelite/runelite)
+This repositorya contains markers for [RuneLite](https://github.com/runelite/runelite)
```
adamk33n3r, line 4: Accidental change?

Velite12: @adamk33n3r  My changes were not showing up on github even after I pushed until a day later, I could change this back but it was intentional.

Velite12: To clarify I was pushing changes, even force pushing, and this PR was not updating, that's why I did this, seems now it's instantly updating 🤷

adamk33n3r: Oh, weird. Never seen that before.

PR #3928, opened 2023-02-18
Afk threshold plugin

Velite12:
This plugin allows you to AFK while skilling by setting a minimum time you want to be AFK for and sends you a notification if you are idle at that time, or if you are not idle, it will wait until you become idle and then send you a notification. This plugin basically combines the AFK timer plugin and the Idle Notifier plugin into one for specifically skilling activities, but also allows for more time to AFK after the threshold.


Felanbird:
Don't open new PRs, keep pushing your changes to the same one

PR #3939, opened 2023-02-21
Add Health Notifications Plugin

TommyAndy:
Adds a configurable visual notification in the form of an overlay when total hitpoints breach a user-specified value.


adamk33n3r:
For what it's worth, I think my plugin [Watchdog](https://runelite.net/plugin-hub/show/watchdog) supports everything this one does.

PR #3941, opened 2023-02-21
Create OSRSParaTodos

Bwana-Ian:
This translates some of the common menu options to a different language(currently just Spanish) in order to help be more inclusive to non-English speaking OSRS players. Just uses a CSV for translations instead of an API to remove the need for an API key or extra web traffic.


iProdigy:
You should move the `es.csv` file to `src/main/resources` and read it via `getResourceAsStream`: https://github.com/runelite/plugin-hub/blob/master/README.md#plugin-resources

Bwana-Ian:
Alright I added that getResourceAsStream before my BufferedReader. Is that the correct way?

iProdigy:
You don't need `getClassloader` & aren't closing the reader. Here's an example using try-with-resources: https://github.com/pajlads/DinkPlugin/blob/88331ec8f35cface89444d6c4575a0e5127c439e/src/main/java/dinkplugin/util/NpcDropTable.java#L44

Bwana-Ian:
Updated to mimic your code to use try-with-resources and removed getclassloader

PR #3956, opened 2023-02-27
Create clan-chat-webhook

pascalla:
Plugin allows Clan Chat messages and broadcasts to be sent via a Webhook. It also natively supports Discord webhooks.
image


Nightfirecat:
[Not only is "clanchat" not very specific to what your plugin actually does](https://github.com/pascalla/clan-chat-webhook/blob/f40ce9d71da627358b358a0835eb80bd3762cafc/src/main/java/com/clanchatwebhook/ClanChatConfig.java#L7), that config group actually overlaps with that of the core chat channels plugin. Please change it.
While not a hard requirement, I'd recommend renaming your config and plugin classes similarly to be more specific. (ie. add "Webhook" in the name)
LGTM otherwise

pascalla:
Updated the commit hash to include changes suggested above :)

PR #3959, opened 2023-02-28
Create world-finder

Cgende:
A modified version of the world hopper plugin which adds a column to sort by when a world was last visited. This is useful to avoid hopping to worlds you've already checked when looking for a free world. Also works with the party plugin to show when anyone in the party was last in the world.

### DIFF
--- a/plugins/world-finder
+++ b/plugins/world-finder
@@ -1,0 +1,2 @@
+repository=https://github.com/Cgende/World-Finder.git
+commit=97e5397dd0f5a1389def57b4f4e2a1377447d3c2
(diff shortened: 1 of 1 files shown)


Nightfirecat:
LGTM, however I'd strongly recommend adding `conflicts = { "World Hopper" }` to your plugin's descriptor as I'm not sure how this would interact if enabled alongside the core plugin with similar settings--nothing good, I imagine. See the following for documentation on that property: https://github.com/runelite/runelite/blob/60677858a55e7e23e693c30af53694e85af678b0/runelite-client/src/main/java/net/runelite/client/plugins/PluginDescriptor.java#L56-L61

Cgende:
OK I've added the conflict in the plugin descriptor.

PR #3965, opened 2023-02-28
Use client.getItemCount and ItemComposition instead of the ItemId class (reverting changes to build.gradle as well)

ElOsoGroso:
The ItemId class was not available at runtime due to the 'compileOnly' setting, causing crashes on login and gameStateChanged


Nightfirecat:
This isn't the right way to solve this problem. You should instead use [`client.getItemCount()`](https://static.runelite.net/api/runelite-api/net/runelite/api/Client.html#getItemCount()) and fetch the name of each item from its composition.

ElOsoGroso:
Okay thanks, @Nightfirecat, I'll do that.

PR #3969, opened 2023-03-01
Create Loot console logger

EmerycP:
A very simple plugin that logs the drops in the OSRS's console. 
Pretty useful when playing with multiple accounts on limited screen space.
I plan to add more type of logs, color & message customization

You can add some drops in an ignore list.
You can also highlight specified drops.
java_RitZB8IDoP


raiyni:
If this repository is private, it needs to be made public.

EmerycP:
> If this repository is private, it needs to be made public.

Woops. I thought I made it public. Thank you.

PR #3981, opened 2023-03-04
OWM-Chat Plugin Whitelist Update

Cerisk:
Updated the plugin to include new features:
- Overhead messages are not automatically pushed to the OSRS chatbox
- Config now includes a whitelist formatted similarly to the NPC Indicators plugin = ex. (NPC Name), (NPC Name)
- Includes a reset button
- Comments in code for what methods are doing
- Commented out code for potential feature of adding a sub-right click menu option that would tag and untag NPCs from the whitelist.

### DIFF
--- a/OWM-CHAT Plugin Submission
+++ b/OWM-CHAT Plugin Submission
@@ -1,0 +1,2 @@
+repository=https://github.com/Cerisk/OWM-Chat.git
+commit=ae57ab76a9b33e858efc134a8f1ae41813b73412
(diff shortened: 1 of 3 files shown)


Cerisk:
Not sure if it saved properly, here is the updated file information:

repository: https://github.com/Cerisk/OWM-Chat.git
commit: b4c4c7a33c460b41d6559530808c17c312689906

Cerisk:
If there are any further updates/information needed, let me know.

PR #3984, opened 2023-03-04
Updates max-skill-trim

NathanQuayle:
Changes how we handle Overlaying the trim. Originally was using OverlayManager, but the trim needed to sit between the skill & the exp tooltip. OverlayManager would appear above both.

Moved to creating a graphics widget inside each skill and then applying a custom sprite.


Nightfirecat:
You need to [synchronize `ImageIO.read()`](https://github.com/NathanQuayle/max-skill-trim/compare/6375a58752977b4d4225876910a500eccaba66b8...NathanQuayle:f18e96ef84ee80ac63f0331bddca3b0cdd969873#diff-9063b6ba273acc093ca0954047650399319ea0a71d264bd80b657bc76007bf6dR219) like you had previously been doing in startUp()

NathanQuayle:
Think this is resolved now.

I've also added people to the licence whose work I've copied functions from. I think this is how it works.

PR #3985, opened 2023-03-04
Create browser-wiki

gkazan:
An alternative to the wiki plugin which adds an embedded browser.


Adam-:
We do not have support for javafx, and, probably would not want this regardless since it executes arbitrary javascript.

gkazan:
JavaScript can be disabled, which I will do.
` webEngine.setJavaScriptEnabled(false);`
As for jfx, am I not able to add a build dependency to my build.gradle or it is not supported entirely?

PR #3989, opened 2023-03-05
Item Charges Improved v0.2.3

TicTac7x:
* Celestial ring added, select infoboxes to hide from settings (so new items are visible automatically in the future)
* Improve item charges reset logic


jiminysasda:
Could this possibly be updated to include Tumeken's shadow and the log basket? Thanks tho great plugin.

TicTac7x:
https://github.com/TicTac7x/runelite-plugins/issues/66

PR #4016, opened 2023-03-08
Created PokemonLowHPSound

truenosus:
simple program that will play the pokemon low hp sound when player is below threshold


raiyni:
Where is this sound from?

truenosus:
> Where is this file from?

the .wav? its from yt video converter

raiyni:
We can't distribute sounds/IP from nintendo.

truenosus:
> We can't distribute sounds/IP from nintendo.

ah shit i didnt even think about that, what if i replace it with a fan made recreation? that sounds very familiar

truenosus:
well i ended up recreated the sound effect very quickly in audacity, so it is no longer ripped from the official game, and instead my own creation with no copywrite. hopefully that suffices, otherwise just close the pull request

Nightfirecat:
[This isn't going to work for a deployed .jar file](https://github.com/truenosus/PokemonLowHPSoundPlugin/blob/88ac02a86927f1e6595b75253eee76bb44f190d0/src/main/java/com/pokemonsound/PokemonSoundClip.java#L6). The .wav file isn't going to exist on end user's computers, so this won't find any actual files. You should instead include the file in your resources directory, and load it as an input stream relative to your plugin package path.

Nightfirecat:
Closing due to inactivity. Feel free to re-open this PR or open a new one once you've addressed the feedback from my comment above.

PR #4036, opened 2023-03-10
Create wilderness-player-alarm

adhansen:
### Overview
This PR adds a new plugin called `Wilderness Player Alarm` to the plugin hub that alerts the user to the presence of others by flashing the whole screen with a bright color. The intention is mainly for scout accounts that camp outside wildy boss lairs while a ~~prey animal~~ player with risk inside the arena fights the boss.

### Legality
I referenced this rules page while implementing the plugin:
and don't believe that I'm infringing on any of the rules. There is an option in the plugin to refrain from flashing when the detected player is a friend/clanmate but I don't think that's infringing on the rules any more than the "Player Indicators" plugin which highlights friends/clanmates/others etcetera differently with even more granularity. Logic depending on a players' skull status is new but that's not mentioned in the rules at all.

### DIFF
--- a/plugins/wilderness-player-alarm
+++ b/plugins/wilderness-player-alarm
@@ -1,2 +1,2 @@
 repository=https://github.com/adhansen/plugin-repo.git
-commit=afdeacba85a3cc7cb807e00ffdec7342775864c8
+commit=64c26e1e81738f6e20958f339384988c8922e8ce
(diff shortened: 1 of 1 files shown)


Nightfirecat:
FYI, the "Ignore unskulled" option will not work, [as `Player#getSkullIcon()` returns `null` for players who are not the local player](http://static.runelite.net/api/runelite-api/net/runelite/api/Player.html#getSkullIcon()).
Additionally, this plugin's overlay must be removed on plugin `shutDown()`, otherwise it would remain active with no way to remove it.

adhansen:
Updated to remove that setting and override `shutDown()`, thanks for looking.
I've also gone ahead and remade the icon using OSRS wiki-only assets, having seen the other guy's PR with the Pokémon sound.

PR #4053, opened 2023-03-15
Add No-Adviser plugin

Daylend:
Disables the Activity Adviser button next to the minimap when in combat, or optionally, all the time.

### DIFF
--- a/plugins/no-adviser
+++ b/plugins/no-adviser
@@ -1,0 +1,2 @@
+repository=https://github.com/Daylend/disable-adviser.git
+commit=c69e6bdd83807ce01d0a082b38f853db030010a6


Daylend:
First time PR. Changing from `com.example` to `com.noadviser` broke it somehow. Will look into it later.

Nightfirecat:
You should be using `plugins=com.noadviser.NoAdviserPlugin`, not just `com.noadviser`.

Nightfirecat:
Side note, can't the activity adviser be disabled ingame? I'm not sure who this plugin caters to

iProdigy:
It should be `com.noadviser.NoAdviserPlugin` in runelite-plugin.properties

Skretzo:
Yes, it is possible to remove the activity adviser orb in the in-game settings
image

Daylend:
> Yes, it is possible to remove the activity adviser orb in the in-game settings image

Thanks for this, I wasn't aware of this feature.

By default the plugin only disables the activity adviser when the player is in combat, so there is still some merit to using this if you don't want to re-enable the button every time you want to use it. For me that's the largest annoyance since it can get you killed if you misclick it during a raid, for example. It's a small benefit, but it's also a small plugin. So I'll leave it up to whoever decides these things if they think it's worthy of a spot or not.

Thanks for the help everyone!

PR #4059, opened 2023-03-15
[region-locker] Update GPU

slaytostay:
Tested and working

### DIFF
--- a/plugins/region-locker
+++ b/plugins/region-locker
@@ -1,2 +1,2 @@
 repository=https://github.com/slaytostay/region-locker.git
-commit=7c2b2cb717d06fbdc447ef86b127c6d2e32a8e8c
+commit=b14e27896518dfaa5e68c8884c2f124b13f8456c
--- a/plugins/region-locker
+++ b/plugins/region-locker
@@ -1,2 +1,2 @@
 repository=https://github.com/slaytostay/region-locker.git
-commit=b14e27896518dfaa5e68c8884c2f124b13f8456c
+commit=ac106b7a82abee9eabf9f4d3b740b9f613917214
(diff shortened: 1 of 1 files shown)


raiyni:
We'd prefer you to not submit something that you haven't built and tested.

slaytostay:
> We'd prefer you to not submit something that you haven't built and tested.

That makes sense, sorry. Currently I don't have a development environment I can build and test on, I should probably just find a new maintainer.

slaytostay:
> Is there any appetite for removing the need to copy the entire GPU plugin here? It would probably be easier to maintain if it could instead extend RuneLite's GPU plugin class and make the few modifications it wants. The GPU portion of the plugin could then be reduced to this: https://github.com/aHooder/region-locker/blob/80786c940a3deae405a6ac6cd62e16cc1f5c2ac7/src/main/java/com/gpu/RegionGpuPlugin.java
> 
> It would require the following small change to RuneLite core though: [aHooder/runelite@29b5fb2](https://github.com/aHooder/runelite/commit/29b5fb275192e6992615b7d9883e043e183a2017)
> 
> [This small bit of reflection code](https://github.com/aHooder/region-locker/blob/80786c940a3deae405a6ac6cd62e16cc1f5c2ac7/src/main/java/com/gpu/RegionGpuPlugin.java#L114-L121) would also be unnecessary if those two things were made public in the core GPU plugin.
> 
> [Here's a full diff](https://github.com/aHooder/region-locker/compare/08168b7fba136be63926c2b37de0d75d901fdb44..extend-gpu-plugin) of the region locker changes, if that's more interesting.

That would be perfect, thank you for this. Was looking for something like this when the plugin came out, had no idea this would be possible.

aHooder:
I should add, I also ported this to LWJGL in the commit before the changes in my previous comment. Here's a link to the commit: https://github.com/aHooder/region-locker/commit/d47f9843156ea7b3d0531a7aa94a48d2fb86d11f
I can confirm that one works on my end.

aHooder:
@slaytostay I can help with maintaining region locker if you'd like. [I currently maintain 117HD](https://github.com/runelite/plugin-hub/blob/master/plugins/117hd#L3), so it's somewhat closely related.

PR #4063, opened 2023-03-15
Created toxic-npcs plugin details

TheJamesLJ:
### Plugin Name
Toxic NPCs on Player Death

### Overview
Heavily inspired by IdylRS and his 'bad plugins' idea for toxic npcs that will be toxic to the player if they die.

When a nearby player dies, all nearby npcs will display overhead text that makes fun of them for dying.

Depending on the npc that displays overhead text, there is unique dialogue that will display for them, i.e. God Wars Dungeon bosses and Jad.


raiyni:
We have denied plugins similar to this before https://github.com/runelite/runelite/issues/12683 https://github.com/runelite/plugin-hub/pull/704

PR #4073, opened 2023-03-17
Update Region Locker

aHooder:
@slaytostay has added me as a maintainer and asked me to submit a PR here: https://github.com/slaytostay/region-locker/pull/31#issuecomment-1474084401

This removes the dependency on JOGL and brings it in line with RuneLite's current GPU plugin.


Lorentary:
This only happens when the HD plugin is also enabled

PR #4094, opened 2023-03-21
Update account-chat-filter

evaan:
Not my plugin, pushed to the original plugin repo but was never pushed to the actual plugin hub.


Felanbird:
@MrSableye seems to be active on Github, did you attempt to ask them to update the plugin with your changes?

MrSableye:
Looks good to me, change was made a while back but I forgot to submit it here

PR #4097, opened 2023-03-22
Create crosshair-cursor

Thomas-gonzalez:
Minimal plugin that switches your OS' cursor type to crosshair.

### DIFF
--- a/plugins/crosshair-cursor
+++ b/plugins/crosshair-cursor
@@ -1,0 +1,2 @@
+repository=https://github.com/Thomas-gonzalez/crosshair-cursor.git
+commit=1a8c5da7e40a78beb307829a6c3bb0511ee38ceb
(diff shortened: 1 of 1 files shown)


raiyni:
Custom Cursor already accepts custom images?

Thomas-gonzalez:
But not cursor files, Custom Cursor only works with pngs.

Thomas-gonzalez:
> Custom Cursor already accepts custom images?

And with images you are restricted to top-left click alignment, which doesn't work for something like a crosshair.

iProdigy:
Avoid using `example` as your config group: https://github.com/Thomas-gonzalez/crosshair-cursor/blob/1a8c5da7e40a78beb307829a6c3bb0511ee38ceb/src/main/java/com/crosshaircursor/CrosshairCursorConfig.java#L7

PR #4101, opened 2023-03-23
Create toa-gear-check

Need-Femboy:
Plugin that adds a side panel that displays the gear of people applying to a party in ToA.


Nightfirecat:
[Don't use `enabledByDefault = false`](https://github.com/Need-Femboy/ToAGearCheck/blob/b80e0847c1d789d62dc7986e7fc7621b4d1cea8d/src/main/java/com/toagearcheck/ToAGearCheckPlugin.java#L38), it's confusing for users to install a plugin and have it not be active immediately.

PR #4103, opened 2023-03-23
An alternate implementation for the Acreuus spellbook timers.

Mantic-says-hi:
Not something I want to try implement in RuneLite within the Timers plugin but it's a small feature me, a few clan mates and hopefully others will use.

### DIFF
--- a/plugins/arceuus-timers
+++ b/plugins/arceuus-timers
@@ -1,0 +1,2 @@
+repository=https://github.com/Mantic-says-hi/arceuus-timers.git
+commit=e74883d52df32e1459d440ddc77277457a2ae55b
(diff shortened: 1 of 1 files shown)


Mantic-says-hi:
> What's the difference to the core timers plugin here? Slightly larger icons?

Core timers plugin only shows the sprite for a 'Reurrect Superior Skeleton' not matter what thrall you summon, my edit gives a custom icon I made that will show either a skeleton, ghost or zombie, no matter what tier of thrall is cast. 

Also the time formatting does not have the 0:00 formatting and just shows the seconds total and the cooldown icons are colorful unlike the use of the sprite which comes from the icon in the spellbook when you don't have the runes to cast a spell.

Death charge is a little different, easier to see the changes in the picture on the README . But basically the cooldown icon is easier to see and the death charge active infobox does not have the unnecessary countdown on it.

PR #4111, opened 2023-03-24
Create RL Local Time Plugin

mbrewerton:
A very basic plugin by request of friends.

Displays the local time as a movable overlay in the client. Customisation to be added as a future feature.

image


raiyni:
There's a setting for this in the Report Button plugin

mbrewerton:
> There's a setting for this in the Report Button plugin

There is, however it's locked to the report button, however people I know uses that in time-logged mode and wants something that they can have as an actual overlay to be moved, and customisable in future.

I also failed the first build check which I think may have been because the repo was private? It's now public.

Nightfirecat:
You should read the build logs to see why it failed to build.

mbrewerton:
> You should read the build logs to see why it failed to build.

I did, so I made the repo public and now the build failure is different. This is a different issue that needs resolving.

PR #4119, opened 2023-03-25
Create Custom Drops Sounds

daanbom:
Plugin for custom sound for different tier of value drops.
Also custom sounds for highlighted items, pets and clues.
More in the future if people want.

Check the readme for a detailed description

### DIFF
--- a/plugins/custom-drop-sounds
+++ b/plugins/custom-drop-sounds
@@ -1,2 +1,2 @@
-repository=https://github.com/daanbom/Custom-Drop-Sounds
+repository=https://github.com/daanbom/Custom-Drop-Sounds.git
 commit=7d1d77a92b7190e90524fa3f784d9cbb90f5e03b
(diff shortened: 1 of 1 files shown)


Nightfirecat:
[Do not use "example" for your config group.](https://github.com/daanbom/Custom-Drop-Sounds/blob/b2db8bea6f364c915628d4793d855ed25c2e51ee/src/main/java/org/CustomDropsSounds/CustomDropSoundsConfig.java#L5)
Also, we cannot distribute copyrighted audio. Please choose different audio files, or change your plugin's design to not distribute the files packaged into the jar.

daanbom:
> [Do not use "example" for your config group.](https://github.com/daanbom/Custom-Drop-Sounds/blob/b2db8bea6f364c915628d4793d855ed25c2e51ee/src/main/java/org/CustomDropsSounds/CustomDropSoundsConfig.java#L5) Also, we cannot distribute copyrighted audio. Please choose different audio files, or change your plugin's design to not distribute the files packaged into the jar.

I have altered the example and changed to non copyrighted audio

daanbom:
reopening PR with small change to make support link accurate

PR #4134, opened 2023-03-27
Create global-chat

RusseII:
This creates a global chat that is shared across all players in a world that have installed the plugin. Useful for making sure you don't accidentally walk too far away and miss each others messages.

### DIFF
--- a/plugins/global-chat
+++ b/plugins/global-chat
@@ -1,0 +1,2 @@
+repository=https://github.com/RusseII/region-chat.git
+commit=0ca8ee7bcced5608b3764d98396f447afc9b8ac3


plugins/global-chat
```diff
@@ -0,0 +1,2 @@
+repository=https://github.com/RusseII/region-chat.git
+commit=f02f30f7cf2fa232c8a4d653a4c4bf02920d3e71
```
Nightfirecat, line 2:
```suggestion
commit=f02f30f7cf2fa232c8a4d653a4c4bf02920d3e71
warning=This plugin submits all sent chat messages and your IP address to a server not controlled or verified by the RuneLite developers.
```

Nightfirecat (changes requested):
[Do not call `Thread.sleep()` on the client thread.](https://github.com/RusseII/region-chat/blob/f02f30f7cf2fa232c8a4d653a4c4bf02920d3e71/src/main/java/com/regionchat/RegionChatPlugin.java#L99) This would delay execution for the entire client.

RusseII:
@Nightfirecat all issues have been fixed! Let me know if you need anything else.

raiyni:
Please rename your package and plugin class.

RusseII:
> Please rename your package and plugin class.

Fixed in a7d42e4f7f4039185139a14fe1020c3656028d02

PR #4138, opened 2023-03-28
Add no-bad-alchs plugin

CreativeTechGuy:
Nice little demo and write up about the plugin in the README if you want more details: https://github.com/CreativeTechGuy/no-bad-alchs

Hide items when casting alchemy spells that you likely don't want to accidentally alchemize (such as untradeables or items which alch for less than their GE value).


CreativeTechGuy:
@Nightfirecat Thanks!

PR #4139, opened 2023-03-28
Add DMWatch plugin.

DMWatch:
It's a hybrid of player indicators (indicate party member tiers) runewatch but for death matching community
party plugin to share information between users to reduce scams and risk

Allows people to be marked as specific roles like scammer or regular or streamer.

### DIFF
--- a/plugins/dmwatch
+++ b/plugins/dmwatch
@@ -1,2 +1,2 @@
 repository=https://github.com/DMWatch/DMWatch.git
-commit=0598b7bc50a28a48a34abe516b83e07b4091c377
+commit=7e45ae8ae1d3ba0454ba910376d44ae4660eb924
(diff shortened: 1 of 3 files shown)


Nightfirecat:
Why are you modifying line-markers and radius-markers?

zguilt:
When I originally created this commit, I accidently had it added as an amend to a commit rather than one on it's own (fat fingered a check box in my git gui app). I didn't realize until you responded that I had done that. 

Should be cleaned up now. 

How can I issue the rebuild work flow?

zguilt:
Is there a way for me to make sure on the rebuild workflow issued that it’ll work? I corrected the gson issue but am not sure about any other issues that the workflow checks. 

Also does it only rebuild when an admin tells it to or is there a work flow button i am blind to? Thanks!

Nightfirecat:
There's a button we need to press since this is a submission from an account which hasn't previously had any contributions to this repository

zguilt:
> There's a button we need to press since this is a submission from an account which hasn't previously had any contributions to this repository

Good to know. My community friends are excited what we have put together. Thanks for issuing the rebuild, looks like it worked this time. 

Now to practice patience. :)

PR #4143, opened 2023-03-30
Add time-played plugin

muffyn:
Displays your time played per profile down to the millisecond in multiple styles of widgets. Special attention paid to world hopping.

The default style is intended to mimic LiveSplit's aesthetic, as this plugin was initially made for speedrunners.


Nightfirecat:
[Do not use `enabledByDefault = false`](https://github.com/muffyn/time-played/blob/ffa2c4a482b3512060ca89f69cf7f4c5ce8b8397/src/main/java/com/timeplayed/TimePlayedPlugin.java#L24), it is confusing for users to install a plugin which is not enabled when installed.
LGTM otherwise

muffyn:
Fixed

PR #4144, opened 2023-03-30
Create duck-duck-goose

skeldoor:
Adds life to the world by putting ducks in bodies of water that don't currently have them.


skeldoor:
Forgot that some places have different spellings for LICENCE LICENSE but that's fixed now

PR #4167, opened 2023-04-05
Dry streak

kriops:
# Dry Streak
Makes it easy to apply the inverse function of the cumulative [geometric distribution](https://en.wikipedia.org/wiki/Geometric_distribution) to calculate the expected dry streak for any given drop rate.

## Why?
If something has a drop rate of 1/100, the probability of seeing your first drop exceeds 50% after approximately 68 kills. This is a question that commonly arises during gameplay, and this plugin makes it easy to get an answer from within RuneLite. While this functionality is available in other plugins, this one is easier to use, and the output is significantly easier to understand.

## Usage
1. Input the drop rate. If the drop rate is `1 in 100`, you can enter `100`, `1/100` or `0.01`.
2. Click "Calculate".
3. Read the result.


kriops:
I will deal with the unverified dependencies after Easter.

plugins/dry-streak
```diff
@@ -0,0 +1,2 @@
+repository=https://github.com/kriops/dry-streak.git
+commit=ea6b9a0b7d56a7b5a88491c9406ee1ace93939fa
```
iProdigy, line 2:
```suggestion
commit=4d712a73166f04a66c689846b797581ed59d7f33
```

kriops:
As far as I know, there is nothing else I need to do for this to be merged? If I need to do anything at all, please let me know :)

PR #4169, opened 2023-04-05
Update ImaDrum Money plugin

EwwItsMike:
Fixed SoundclipManager.
Now uses BufferedInputStream to stream soundclips from resources. 

It was throwing IOExceptions for not supporting mark/reset in regular InputStream.

### DIFF
--- a/plugins/imadrum-money
+++ b/plugins/imadrum-money
@@ -1,2 +1,2 @@
 repository=https://github.com/EwwItsMike/ImaDrum-Money.git
-commit=65e94ba6042cf38bea86ecbe982197e959cb7c96
+commit=4abf8e35cd7edfe78e520af9d403ff3294e9a78b


LlemonDuck:
You should wrap [all three of these](https://github.com/EwwItsMike/ImaDrum-Money/compare/65e94ba6042cf38bea86ecbe982197e959cb7c96...EwwItsMike:4abf8e35cd7edfe78e520af9d403ff3294e9a78b#diff-d597e189c578a19e1d6da72ee4cac1962e2ee64811fe53904e80b2c74f960f3cR47-R50) in a try-resource block, it's not as nest-y as you had it before either:
```
try (InputStream s = getClass().getResourceAsStream(sound.getFileName());
     InputStream bufferedIn = new BufferedInputStream(s);
     AudioInputStream audioStream = AudioSystem.getAudioInputStream(bufferedIn)) {
     // ...
}

PR #4175, opened 2023-04-05
Update quest-helper 4.0.0

Zoinkwiz:
This adds the basis for player made quests, such as:
* RuneliteNpcs to talk to and examine
* Quest steps for NPCs
* Tracking of player quest states
* Interactable items

This is demonstrated in Cook's Helper.

Main things I think worth considering with this PR:
* Item pickup changes the text of "Walk" to something else (Pick up). Doesn't interact with the actual behaviour of walk
* NPC replacement keeps the original clickbox of the original NPC, but the menu option is still moved about


Nightfirecat:
Does this PR supersede #4067? (I'm aware both are `hold-for-release` since they're awaiting API changes to be released)

Zoinkwiz:
It will supersede it yes, I can close that one now.

PR #4180, opened 2023-04-06
Create resizable-modern-chat

UNDEAD-B0T:
Resizeable/Modern Layout only: collapses chat to an icon that shows DMs and all chat


UNDEAD-B0T:
ugyhgh im sorry

PR #4182, opened 2023-04-06
Add Item-Hunter plugin

MakingStan:
# Item Hunter

<p>Adds a side-panel to your client where you can generate random items.</p>


Nightfirecat:
FYI we have [`ImageUtil.resizeImage()`](http://static.runelite.net/api/runelite-client/net/runelite/client/util/ImageUtil.html#resizeImage(java.awt.image.BufferedImage,int,int)) if you ever want to use that [instead of this borrowed-from-SO code](https://github.com/MakingStan/Item-Hunter/blob/e79543616dfb1dfa71086f365b04c77c7004422d/src/main/java/org/makingstan/ItemHunterPanel.java#L230).

PR #4192, opened 2023-04-07
PC Rewards

marvinb16:
PC Rewards is a plugin that allows the user to customize the Pest Control rewards shop by giving them the ability to pick and choose what shop options they would like to have displayed to prevent any misclicks if turning in large sums of points.

Resubmitting due to file error.


Nightfirecat:
Please just push new commits to your branch instead of closing and opening new PRs.

PR #4200, opened 2023-04-10
Create anime-feet

Ondra-D:
An NSFW plugin that overlays a picture of anime feet on the screen.


geheur:
I don't think you can write to the imagine in src/main/etc since that ends up in the jar file. You have to put it in a file in .runelite, e.g. `.runelite/tax_documents`.

Nightfirecat:
We're not interested in accepting a nsfw plugin.

Ondra-D:
Understandable <3

PR #4201, opened 2023-04-10
Create nekos

Ondra-D:
A plugin that overlays a picture of a cute neko on the screen.

<3

### DIFF
--- a/plugins/nekos
+++ b/plugins/nekos
@@ -1,3 +1,3 @@
 repository=https://github.com/Ondra-D/nekos.git
-commit=e0d540d31cee892b276e6d9da13f64ae163a6229
+commit=b2f3e65af79cfab4ff1a5601504e707b525d8ad1
 warning=This plugin submits your IP address to a server not controlled or verified by the RuneLite developers.
--- a/plugins/nekos
+++ b/plugins/nekos
@@ -1,0 +1,2 @@
+repository=https://github.com/Ondra-D/nekos.git
+commit=e78ea2d01d4a4baeef9754159aa6238c5bb47b8e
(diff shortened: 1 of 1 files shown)


plugins/nekos
```diff
@@ -0,0 +1,2 @@
+repository=https://github.com/Ondra-D/nekos.git
+commit=6c73f2d4c60ee98b0e2b6db490b458caf917ec3a
```
Nightfirecat, line 2:
```suggestion
commit=6c73f2d4c60ee98b0e2b6db490b458caf917ec3a
warning=This plugin submits your IP address to a server not controlled or verified by the RuneLite developers.
```

Nightfirecat (changes requested):
[You should fix your plugin's support link](https://github.com/Ondra-D/nekos/blob/6c73f2d4c60ee98b0e2b6db490b458caf917ec3a/runelite-plugin.properties#L3)
The provided image would be fine to display, but do you/we have permission to distribute it? It would still be owned/copyrighted by whoever drew it. You could also remove it and fetch an image from the server on plugin startup instead.
[This code will not work from a jar file](https://github.com/Ondra-D/nekos/blob/6c73f2d4c60ee98b0e2b6db490b458caf917ec3a/src/main/java/com/ondrad/nekos/nekoOverlay.java#L58). Rather than writing the downloaded image to a file, you should just set `image` to the resulting BufferedImage from `ImageIO.read()`.

Nightfirecat:
Image reading should also be synchronized, eg.

```java
synchronized (ImageIO.class)
{
	image = ImageIO.read(imageUrl);
}
```

LGTM otherwise.

PR #4213, opened 2023-04-12
Create aerial-tick-colors

JadeGerbitz:
My plugin allows the user to select different colors to highlight the fishing spots at aerial fishing based on the number of ticks it will take to catch the fish. The fishing plugin does this already, but it only has colors for 2-tick spots and not 2-tick spots. I think this should help visualize the distance to aerial fishing spots better.

### DIFF
--- a/plugins/aerial-tick-colors
+++ b/plugins/aerial-tick-colors
@@ -1,2 +1,2 @@
 repository=https://github.com/JadeGerbitz/Aeirial-Tick-Colors.git
-commit=00774b950795acf3fbe6566b8c3f26f13089c40d
+commit=06dd86c1a101ad04dc4bbad695d646e9c3ad0a76
--- a/plugins/aerial-tick-colors
+++ b/plugins/aerial-tick-colors
@@ -1,2 +1,2 @@
 repository=https://github.com/JadeGerbitz/Aeirial-Tick-Colors.git
-commit=70b959a1fabdbb6b9685f3a223dc29ba4dec7da0
+commit=54dfbe479804d5fe78e0246692ff36e987c85ba1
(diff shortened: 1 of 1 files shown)


skeldoor:
Hi, did you forget to commit/push your changes? This looks like a blank template example plugin. Only the readme and the license is changed from the original.

These steps are worth following to get everything set up. https://github.com/runelite/plugin-hub#creating-new-plugins

Also is the repo name "Aeirial-Tick-Colors" a misspelling of "Aerial-Tick-Colors"

JadeGerbitz:
The plugin code has been pushed to the repository and the typo in the repository name has been fixed.

Nightfirecat:
You need to update the commit in this PR as well, as that's what would be built for the hub.

JadeGerbitz:
I've updated the commit.

Nightfirecat:
There already exists a hub plugin with the name "Aerial Fishing": https://runelite.net/plugin-hub/show/aerial-fishing. Your config group doesn't conflict with it, but we'd prefer you change it to be more specific to the full plugin name so as not to be confused, eg. `@ConfigGroup("aerial-tick-colors")`

JadeGerbitz:
I've changed the config group to "Aerial-Tick-Colors".

plugins/aerial-tick-colors
```diff
@@ -0,0 +1,2 @@
+repository=https://github.com/JadeGerbitz/Aeirial-Tick-Colors.git
```
Felanbird, line 1: This is just a nit-pick but the repo url probably should be the post name-changed version
```suggestion
repository=https://github.com/JadeGerbitz/Aerial-Tick-Colors.git
```

PR #4223, opened 2023-04-13
Combatlogouttimer

LuxOG:
Create simple plugin to display the 10 second timer that prevents logout upon being hit in combat


Nightfirecat:
Your plugin and config classes conflict in name with the core client's logout timer plugin. Additionally, your config group is the same as well, which would cause a mix of the two plugin's configs; please change the name of all of these. Ideally you should name your classes/config groups as specifically as possible--ie. *Combat* Logout Timer, not just Logout Timer.

PR #4229, opened 2023-04-15
Updated Dynamic inventory tags (v1.1.0)

BraveH:
### What's New:
- [Added Panel](https://github.com/BraveH/gear-switch-alert#panel)
- [Added the ability to add items by searching for them. No need to have them equipped or in inventory anymore!](https://github.com/BraveH/gear-switch-alert#searching-mechanism)
- Added check to ensure that items being added are equippable items (can be disabled in config if you wish to set tags for unequippable item for some reason)


BraveH:
### Update - Added:
- Ability to import/export profiles! (Right-click any profile to export and use the "Import profile from clipboard" button at the bottom of the panel to import)

Nightfirecat:
You [add your plugin's nav button here](https://github.com/BraveH/gear-switch-alert/blob/423f743e2abfe1ebcb5c71759748211546af3d2b/src/main/java/com/gearswitch/GearSwitchAlertPlugin.java#L118) during startup, but it is never removed; it must be removed during shutdown so that toggling the plugin doesn't add multiple buttons.

BraveH:
@Nightfirecat thanks for the catch, fixed :)

PR #4237, opened 2023-04-17
Update quest-helper to 4.0.2

Zoinkwiz:
Small change to add a disable option for fake npcs.


Nightfirecat:
Needs conflict resolution

PR #4243, opened 2023-04-18
Update bot-detector to v1.3.9.2

Cyborger1:
Changed a deprecated variable from Runelite's API.

Changed the plugin's API URL to point to our new subdomain (`api.prd.osrsbotdetector.com`).

### DIFF
--- a/plugins/bot-detector
+++ b/plugins/bot-detector
@@ -1,4 +1,4 @@
 repository=https://github.com/Bot-detector/bot-detector.git
-commit=5ffabc78ba6329cbf612cb7778257a7c57f33a08
+commit=db306de03f9f505d83bfceeadeb0db130286f859
 warning=This plugin submits the names, locations, and gear of encountered Players, and your IP address to a 3rd party website not controlled or verified by the RuneLite Developers.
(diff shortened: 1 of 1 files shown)


Nightfirecat:
Unrelated to this PR specifically, [please remove `enabledByDefault = false` from your plugin descriptor](https://github.com/Bot-detector/bot-detector/blob/db306de03f9f505d83bfceeadeb0db130286f859/src/main/java/com/botdetector/BotDetectorPlugin.java#L129) since it's confusing for users to install plugins which aren't then automatically enabled.
This change LGTM otherwise.

Cyborger1:
> Unrelated to this PR specifically, [please remove `enabledByDefault = false` from your plugin descriptor](https://github.com/Bot-detector/bot-detector/blob/db306de03f9f505d83bfceeadeb0db130286f859/src/main/java/com/botdetector/BotDetectorPlugin.java#L129) since it's confusing for users to install plugins which aren't then automatically enabled. This change LGTM otherwise.

Removed, that was a leftover from when we first published the plugin as some bit of extra assurance. Players are always wary of plugins that have the IP warning so we just wanted to be safe.

PR #4245, opened 2023-04-19
Add GIM Improve Plugin for GIM Collection Logs of Split Chat

TheSimpleBlue:
Allows GIM to view visual indicators in their collection log of their teammates and allows users to puts GIM chats into split PMs chat and


plugins/gimimprove
```diff
@@ -0,0 +1,2 @@
+repository=https://github.com/TheSimpleBlue/gimimprove.git
+commit=255e0aaf55143d604b9a2da0a76be6eea4e97b48
```
Nightfirecat, line 2: This needs a warning as it has a default configuration which submits requests to non-verified servers.
```suggestion
commit=255e0aaf55143d604b9a2da0a76be6eea4e97b48
warning=This plugin submits your IP address to a server not controlled or verified by the RuneLite developers.
```

Nightfirecat (changes requested):
LGTM otherwise

TheSimpleBlue (commented):
Let me know if I'm missing anything else

PR #4246, opened 2023-04-19
Add Boss Health Indicators plugin

rugg0064:
Boss Health Indicators is a basic plugin that adds markers on boss health bars.

ingame
panel

### DIFF
--- a/plugins/boss-health-indicators
+++ b/plugins/boss-health-indicators
@@ -1,0 +1,2 @@
+repository=https://github.com/rugg0064/boss-health-indicators.git
+commit=2ae2df2c32aaa198a2e50aa7a3f032ae80e61a99


Nightfirecat:
Where are your icons sourced from? Are they licensed such that we can distribute them?

Nightfirecat:
Also, your plugin should remove the bars when it is shut down so they don't persist when it's disabled.

rugg0064:
> Also, your plugin should remove the bars when it is shut down so they don't persist when it's disabled.

The image is from the wiki. However, I couldn't recall the exact image used. I went ahead and remade it with a wiki image and included a link on the readme. I asked the wiki folks, they said the image is fine to use.

> Also, your plugin should remove the bars when it is shut down so they don't persist when it's disabled.

I have removed the widgets on plugin shut down.


Additionally, changed the priority of the sidebar icon.

PR #4254, opened 2023-04-21
Update pvp-performance-tracker to v1.5.7

Matsyir:
Decently major fixes:
- Voidwaker spec support, dwh spec support (dmg only not def reduction)
- Support opal dragon bolts(e) in LMS for stronger cbows instead of always forcing regular diamond bolts(e)
- Viggora's chainmace & Inquisitor's mace tracking (coded awhile ago but never updated)
- Fixed deprecated GSON initialization


Matsyir:
Sorry about the two commits, had to fix the gson initialization, either I just suck at github (I do) or it didn't let me rebase cause the commit was in a PR, let me know if I have to remake

PR #4255, opened 2023-04-21
Add gotr-leech

QuestingPet:
Add Guardians of the Rift Leech

Plugin for helping to leech/afk in GotR by obtaining the minimum required points for rewards (150) and then mining the rest of the round. Useful while working.

I'm sorry.

### DIFF
--- a/plugins/gotr-leech
+++ b/plugins/gotr-leech
@@ -1,0 +1,2 @@
+repository=https://github.com/QuestingPet/gotr-leech.git
+commit=8aa324d848e3054b7ad2bb861b11ba2ab89d7ccf


Felanbird:
Doesn't this plugin already exist?
https://runelite.net/plugin-hub/show/zom-afk-gotr

QuestingPet:
That plugin seemed largely unusable. It has very little to no documentation on how to use it, especially what the "Notify on selection" means (I wasn't able to get it to work). Also, it doesn't handle binding necklaces which are useful for doing a 150 elemental energy in a single trip, maximizing afk-ness. It also doesn't make you aware if you're not properly AFK mining.

Also, from what I can tell, that plugin is just straight up broken. I personally can't figure out how to make it do anything (notify my on my selection, show the "info box", etc.)

I created this plugin as a result of that other one both not working for me and not having all the features I think are necessary.

Examples of features my plugin has:

* An overlay panel describing useful info, color-coded for easy peripheral viewing like if you're mining, if you have enough binding necklaces/charges, if you have enough uncharged cells
* Ability to flash the overlay (as seen below) if not mining
* Notification settings for game start, game end (for missing requirements), and at the 60% threshold if not enough points
* Configurations for all the above

image

All of this is explained in each configuration option name and hover desciption:

image

QuestingPet:
> Also, from what I can tell, that plugin is just straight up broken. I personally can't figure out how to make it do anything (notify my on my selection, show the "info box", etc.)

Okay so trying it for a bit longer, it looks like it doesn't work if you turn it on mid-game. Even after turning it on before a game, I still wasn't getting notifications for the "selection" items, but perhaps I'm just using it wrong.

PR #4258, opened 2023-04-22
Create Pickpocket Safe Guard

Orrezon:
Pickpocket Safe Guard allows users to specify a HP Threshold in the settings. Once this threshold is reached (current HP equal to or less than the threshold), it prevents the user from performing the "Pickpocket" option and alerts them with a game message stating to eat some food. Designed originally for hardcore ironmen to help prevent deaths while pickpocketing.


Orrezon:
Closing this as I figured it might be better to add it to the existing Pickpocket Helper plugin instead. Created PR https://github.com/RDieleman/runelite-pickpocket-helper/pull/8 there for it

PR #4270, opened 2023-04-24
Update visual-sounds

Box-Of-Hats:
- Fix memory leak
- Add option to show only tagged sounds


raiyni:
@Box-Of-Hats You need to update your config group from example to something else

PR #4285, opened 2023-04-28
Create no-new-items

timwestwood:
No New Items is a plugin for players who want to maximise their nostalgia by avoiding items which have been added to Old School RuneScape (OSRS) since its initial release. It does so by changing the visual appearance of new items to be bright red, acting as a visual warning. There is also an option to change the in-game name of new items.

By default, the plugin will consider items from the original God Wars dungeon and the armour reward from the easy Varrock diary as new because they weren't present in the game at the time of the backup used to create OSRS. But as these were actually added to the main game in 2007 and may be considered "old school" by many players, there is a toggle to allow such items. Similarly, the plugin will default to recolouring the coin pouches from pickpocketing that replaced coins as part of an update to tackle botting/autoclicking, but this can be toggled too.

### DIFF
--- a/plugins/no-new-items
+++ b/plugins/no-new-items
@@ -1,2 +1,2 @@
 repository=https://github.com/timwestwood/NoNewItems.git
-commit=b547a7482962c6404c0e276341dc7246ca85ede6
+commit=579931e2cfbbd775404fd19ed50464b44a882091
--- a/plugins/no-new-items
+++ b/plugins/no-new-items
@@ -1,2 +1,2 @@
 repository=https://github.com/timwestwood/NoNewItems.git
-commit=d7e27e4530f0077aefc4ed6dd774c2d8945af085
+commit=eed318cc219b8a629266853ab3bd3d65f9f736c8
--- a/plugins/no-new-items
+++ b/plugins/no-new-items
@@ -1,0 +1,2 @@
+repository=https://github.com/timwestwood/NoNewItems.git
+commit=d7e27e4530f0077aefc4ed6dd774c2d8945af085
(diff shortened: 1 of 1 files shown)


Nightfirecat:
[Don't use `enabledByDefault = false`](https://github.com/timwestwood/NoNewItems/blob/d7e27e4530f0077aefc4ed6dd774c2d8945af085/src/main/java/com/NoNewItems/NoNewItemsPlugin.java#L20), it's confusing for users to install a plugin which doesn't turn itself on.

Also, it looks like you don't reset item textures when the plugin is disabled. Does that leave a player's items permanently recolored? If so, you should make an attempt to reset that

timwestwood:
I've changed it to be enabled by default -- thanks.

The colour change isn't permanent because the ItemComposition objects that I modify only exist for the current 'logged in' session; simply logging out and back in will fully enact any change to the config, including disabling the plugin. It is true though if a player disables the plugin while logged in, no new items encountered by the player will be recoloured, but it won't retroactively undo colour changes the plugin had already made that session. I could add an in-game message if the player does this, advising them that they may need to log out and back in for it to take full effect?

timwestwood:
OK, I've made it so that you can change most options in the config (including enabling and disabling the plugin) and have the appearances adapt immediately in game. This works for items in the inventory, in shops, in the bank, in the GE, on the floor etc. as well as those worn by players (both the user's character and others).

In a small number of cases, the appearance of items worn by specific other players won't change and will remain as they were when the game first loaded them in. I can't work out what it is about these players that causes this, but the original solution of simply logging out and back in will still work to fully propagate these changes and the plugin will print a message to the chat advising the player that they may want to do this.

Hopefully this adequately addresses all of your concerns.

raiyni:
Generally prefer if you don't have lines 21 and 22 defined

timwestwood:
Done @raiyni. Anything else I can do?

PR #4294, opened 2023-04-30
Add ivegotwood plugin

adamdyer028:
This is a simple plugin that allows the user to display a default or customizable message above the character's head while woodcutting, fishing, mining, or constructing objects.


Nightfirecat:
You need [real details in your plugin properties file](https://github.com/adamdyer028/ivegotwood/blob/6d6323101bdf47e31ee506f6701c4739d5b7ba5e/runelite-plugin.properties), not just "Example", "Nobody", etc.
LGTM otherwise

adamdyer028:
@Nightfirecat should be good to go now! thank you

PR #4295, opened 2023-04-30
Add mute-all plugin

CreativeTechGuy:
Enables plugins to get Sound Effect events without the sounds audibly playing.

### DIFF
--- a/plugins/mute-all
+++ b/plugins/mute-all
@@ -1,2 +1,2 @@
 repository=https://github.com/CreativeTechGuy/mute-all.git
-commit=fb0f64799f7d2d53c49b4dd728f453df77372e7d
+commit=a28cab258a72da91ee387b9ad5a8a935fe97b474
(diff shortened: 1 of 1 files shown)


CreativeTechGuy:
Is there a way to query the volume/mute state of Sound Effects and Area Sounds channels so I can have more granular alerts to notify users when their settings make this plugin not work? I was not able to figure out a way to make it work given the existence of the RuneLite Music plugin which overrides how these settings are set.

Nightfirecat:
[You don't need to provide `configName`](https://github.com/CreativeTechGuy/mute-all/blob/a28cab258a72da91ee387b9ad5a8a935fe97b474/src/main/java/com/creativetechguy/MuteAllPlugin.java#L22). Is there some reason you're using this here?

CreativeTechGuy:
@Nightfirecat , force of habit. I did it in my previous plugin because I saw other people do it in their plugins too. I assumed it was there for a reason. Is that field a no-op or when should it be included?

CreativeTechGuy:
Ah okay. Thanks for letting me know! Is there any harm in keeping it?

PR #4301, opened 2023-05-04
Update inventory-setups

Azeirah:
Update to most recent version with sections

I'm not sure if it's ok for me to make this PR. But I do really want to see this new feature for the plugin: https://github.com/dillydill123/inventory-setups/pull/204

@dillydill123


Nightfirecat:
We reached out to @dillydill123 on Discord and confirmed they don't want to release this just yet; they will make a PR themselves when they're ready to do so.

PR #4306, opened 2023-05-05
Update boss-health-indicator

rugg0064:
Change display name from "Boss Health Indicator" to "Boss Health Indicators"
Add boss reordering, as well as collapsing tabs, as shown below:
image


rugg0064:
- Officially set to version 1.0
- Update percent pickers to automatically append % sign

PR #4319, opened 2023-05-07
Create lazy-farming

Speaax:
Farming plugin where user can decide what patches to include, teleport options to use and highlights to show the user what to bring and where to click. Currently lacking a few functions and need a lot of polish, but herb run works well enough that I want to personally use it.

Sorry in advance if the code is a bit messy :d

### DIFF
--- a/plugins/lazy-farming
+++ b/plugins/lazy-farming
@@ -1,2 +1,2 @@
 repository=https://github.com/Speaax/Farming-Helper.git
-commit=35c8dce6ee014d5775524b88bb3f6e44c66b3889
+commit=5ed301339c9f1b72cbd3ae19817d546796a387de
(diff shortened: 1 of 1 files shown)


Speaax:
I'm new to github. Hopefully the way I tried to fix the build errors I had with my commits were the correct way to do it.

geheur:
You didn't open a new PR for each change like most people do so you know more than many others 😆.

Nightfirecat:
[Your plugin icon is extremely large in dimensions](https://github.com/Speaax/Farming-Helper/blob/35c8dce6ee014d5775524b88bb3f6e44c66b3889/src/main/resources/com/farminghelper/speaax/icon.png), which will cause a lot of memory to be used by this plugin when Java loads it. Please scale it down to prevent that from happening.
[Don't use `enabledByDefault = false`](https://github.com/Speaax/Farming-Helper/blob/35c8dce6ee014d5775524b88bb3f6e44c66b3889/src/main/java/com/farminghelper/speaax/FarmingHelperPlugin.java#L33), it's confusing for users to install a plugin which doesn't turn itself on

LGTM otherwise

PR #4329, opened 2023-05-09
Create Max Hit Calculator Plugin

j-cob44:
Max Hit Calculator Plugin.
based on the original Max Hit plugin which has not been updated in 2+ years but rewritten mostly from scratch.

Automatically calculates Max Hit, Max Spec Hit, Max Hit Vs Type, and Max Spec Vs Type and displays them to the player. A tooltip will be displayed when hovering over the panel to show what stats will help the player achieve their next max hit. Includes configuration toggles to allow the player to customize what they see in the panel.

### DIFF
--- a/plugins/max-hit-calculator
+++ b/plugins/max-hit-calculator
@@ -1,2 +1,2 @@
 repository=https://github.com/j-cob44/max-hit-calc.git
-commit=061338a5e137b0f7269cb20ee6752d5f7d20eb38
+commit=72c626268c02979ff32c7219f458e7659c8d60a1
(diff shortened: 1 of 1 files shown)


Adam-:
You must remove your overlays at shutdown.

j-cob44:
> You must remove your overlays at shutdown.

You are right, that was a typo. Fixing now.

j-cob44:
> You must remove your overlays at shutdown.

Fixed, let me know if there is anything else.

j-cob44:
I think this is ready for 1.0 release. I'm trying to confirm with Mod Ash on twitter how type modifiers are ordered. Otherwise, I think I've ironed out any other calculation problems and added all bonuses I can think of. Let me know what you all think here or on discord @jacob#4444.

LlemonDuck:
Are you aware of the existing [Max Hit plugin](https://runelite.net/plugin-hub/show/maxhitplugin) and would you like to contribute there? It already has nearly 100k users, so it may be more convenient for users if it is all in one place.

j-cob44:
Yeah, the github hadn't been touched in over 2 years. The code was way more complicated for not really any reason that was clear to me, so I decided to start fresh.

PR #4363, opened 2023-05-15
Create osrs-clan-notifications

ividrine:
plugin to capture clan notifications and send then to a web api

### DIFF
--- a/plugins/osrs-clan-notifications
+++ b/plugins/osrs-clan-notifications
@@ -1,0 +1,2 @@
+repository=https://github.com/ividrine/osrs-clan-notifications.git
+commit=95df21621ded11992c0614792fe5541cc63b521c


iProdigy:
Would [Dink](https://github.com/pajlads/DinkPlugin) fulfill your use case? It's also built to facilitate custom handling beyond discord. (disclosure: i'm a maintainer of dink)

iProdigy:
Hey @ividrine, tried to reach out on discord, but not sure if I have the right username, so just leaving my comments here:

&nbsp;

> This is specifically for capturing clan notifications rather than individual user achievements.

I'm not sure I follow; doesn't your plugin [also](https://github.com/ividrine/osrs-clan-notifications/blob/95df21621ded11992c0614792fe5541cc63b521c/src/main/java/com/osrsclannotifications/OSRSClanNotificationsPlugin.java#L90-L91) only trigger on the local player's achievements? Dink also uses some clan events, and is often used by clan discords (or custom event handling servers).

&nbsp;

> One of the big things we wanted this for was to get notifications for clan coffer deposits / withdraws. Let me know if you think there is a workaround for this using Dink

If coffer notifications would allow Dink to satisfy your use case, we can add that as a new notifier in the plugin! If so, feel free to create a PR or issue on the repo (& I'll get to it shortly)

ividrine:
After giving it some thought I think it makes the most sense to use Dink for my use case. Even though clan only notifications would be neat, I think its a very niche thing most people probably aren't concerned with and since its an already well maintained project I'd rather just add contributions there where I can. Thanks again @iProdigy

PR #4368, opened 2023-05-16
Update effective-level to v1.3

XrioBtw:
- Added missing void equipment variations


Nightfirecat:
Changes LGTM. Can you [remove `enabledByDefault = false` from your plugin descriptor](https://github.com/XrioBtw/effective-level/blob/07404510fb653db14c9bdb1b7688400904a80b87/src/main/java/xrio/effectivelevel/EffectiveLevelPlugin.java#L35)? It is generally confusing to users to install a plugin which isn't enabled.

PR #4374, opened 2023-05-17
Update annoyance-mute

GraceThree:
Updated to include new Hunllef sounds. Original project hasn't been updated in 2+ years, so I forked it and am gonna try to maintain it now.


raiyni:
@Broooklyn is still around, you should be able to pr to their repo.

PR #4375, opened 2023-05-17
Update clan-chat-webhook

pascalla:


raiyni:
Remove `log.info` calls

pascalla:
> It's not possible to add a warning to an existing plugin and retroactively warn users who have already installed it.

reverted this change

iProdigy:
Warning would still be useful for new installs, or are you implying a new plugin should be created because retroactive warn isn't possible?

Nightfirecat:
The issue isn't that this plugin shouldn't have a warning--if it were submitted as a new plugin today with its current code, it would require one. The issue is that we cannot retroactively add one. I recommend updating your code so that it does not make calls to 3rd party servers by default, [such as by changing this code to early return if  `config.secretKey()` is null or empty](https://github.com/pascalla/clan-chat-webhook/blob/d50900ba53a6dd208569540c9d8b7fa4c1928b8b/src/main/java/com/clanchatwebhook/ClanChatWebhookPlugin.java#L190), for instance.

pascalla:
Now early returning on empty webhook/secret

PR #4378, opened 2023-05-19
Update tog-crowdsourcing - Nav icon too large

jcarbelbide:
Fix nav icon too large

Was taking 5MB in memory. Resized to 30x30 px

### DIFF
--- a/plugins/tog-crowdsourcing
+++ b/plugins/tog-crowdsourcing
@@ -1,3 +1,3 @@
 repository=https://github.com/jcarbelbide/tog-crowdsourcing.git
-commit=fff30f469b7a6cd2f2fbebd8d77f040ecbf31106
+commit=9fbecd05efff386d5374ebb85517210ab80575ec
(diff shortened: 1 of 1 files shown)


Adam-:
It looks like your plugin icon.png is also too large, in addition to the nav button being changed here.

jcarbelbide:
Got it, thank you! Let me fix that as well. Will get to it in a bit

jcarbelbide:
Looks like I need to update some deprecated stuff. I don't have time right now but I can come back to this later

jcarbelbide:
nevermind, it seemed like it was a simple fix

jcarbelbide:
CI passes, but when I run this locally it no longer works. Can we hold off on merging for now? I want to look into this more

Nightfirecat:
You should either close this PR or convert it to a draft in the meantime.

jcarbelbide:
good point, thank you!

jcarbelbide:
turns out the runelite version was just out of date. seems like it works fine after updating the version, so I'm good on my side. Please let me know if there are any other changes I should make. Thanks!

PR #4394, opened 2023-05-23
Update bank-tag-generation

samami-code:
Fixed drop tags, code cleanup, test units, arguments with spaces


raiyni:
It appears this PR has become stale, thus will be closed. If you are still looking to make changes, you are welcome to resubmit a pr.

Note that you should look to collaborate with the original author to fix build issues and obtain their permission to submit changes on their behalf.

PR #4396, opened 2023-05-23
Add modern tick counter to plugin registry

Shawneh:
Looking to revive the dead Tick Counter plugin by updating it with the missing weapons.


Shawneh:
I am a javascript/typescript developer so any help with debugging builds would be appreciated.

PR #4398, opened 2023-05-24
Update herblorerecipes to v1.7

climbridecode:
Refactored the plugin.
Using item IDs rather than Strings and preloading Tooltip cache on plugin startup. Better organization of the configuration options...


climbridecode:
Woops, had wrong language level set in IDE.

PR #4403, opened 2023-05-25
Update PC Rewards 1.2.2

marvinb16:
Version 1.2.2
Changes:
- Added option to hide exp calculation (both on startup and running)
- Replaced enum in config with new class for abbreviations
- Fixed some issues being able to click hidden options
- fixed enum naming per Raiyni


raiyni:
If your config options aren't the same for `PcRewardsConfigOptions` as `RewardOptions`, it will fail to load config. So you need to use new keynames for them.

PR #4405, opened 2023-05-26
Add Damage Taken Logger

LittleBitwise:
This plugin displays damage taken in chat, as well as XP gained per kill.

Both are optional, and colors can be configured.

Total XP per kill

XP can be displayed as a "per second" rate as well.

XP per second per kill

If anyone has a suggestion for a better name _(since this started as just the damage log part)_, I'm open to changing that before this is merged.


raiyni:
Plugin seems fine, is this a final name for this plugin?

LittleBitwise:
> Plugin seems fine, is this a final name for this plugin?

Yup. All good.

PR #4413, opened 2023-05-28
Update multicolor-highlights

patmizi:
Hello 👋 

I'm looking to take ownership of the multicolor-highlights plugin as [the original owner has abandoned it](https://github.com/sgfost/multicolor-highlights/blob/master/README.md?plain=1#L15-L17) 

* Feature - Allows NPC IDs to be configured for highlights instead of just NPC names ( [fefb9551415845c606b4579110ef0e47935c9f6d](https://github.com/patmizi/multicolor-highlights/commit/fefb9551415845c606b4579110ef0e47935c9f6d))
* Maintenance  - Updated author and support link in Runelite plugin properties ([4d91b584a8909b7959a30129a71c844747521e23](https://github.com/patmizi/multicolor-highlights/commit/4d91b584a8909b7959a30129a71c844747521e23))

### DIFF
--- a/plugins/multicolor-highlights
+++ b/plugins/multicolor-highlights
@@ -1,2 +1,2 @@
-repository=https://github.com/sgfost/multicolor-highlights.git
-commit=0ad4513c87b28e755ed1717a660bfab93232c439
+repository=https://github.com/patmizi/multicolor-highlights.git
+commit=cbbb0dfa133e0ac7cf1fe08150d51ea515b87177


Nightfirecat:
@sgfost Just double-checking, is this transfer fine with you? We'd also be fine with you adding @patmizi as a contributor to your repository and adding the to the `authors=` field of this plugin entry if you'd prefer that.

sgfost:
Sure, I'm assuming there can be multiple entries in the `authors` field here (and in `plugin.properties`?)

I've thought about updating things after recently trying to configure the "Radius Markers" plugin to achieve the by-ID tagging added here.. But I'm not a huge fan of the 7 different plugins drawing overlays on NPCs approach and would be more likely to tackle it as a merge request to the built-in highlighter if anything so transferring is fine, worst case I can contribute to @patmizi's fork.

raiyni:
> Sure, I'm assuming there can be multiple entries in the authors field here (and in plugin.properties?)

Yes, it's a csv field

Nightfirecat:
If you were to add patmizi as a contributor, the only field which would be important to add would be `authors=sgfost,patmizi` in this file. (The one in `plugin.properties` is for display purposes only)

I assume from your comment you're fine with this being transferred?

sgfost:
Correct, I'm fine with either. I invited @patmizi to the original repo in case that option is preferred so I'll leave it up him and whatever works best for his plans

Nightfirecat:
@patmizi how do you want to go from here then?

LlemonDuck:
closing since this is very stale, feel free to re-open if needed

PR #4417, opened 2023-05-28
Create DropTracker plugin

joelhalen:
**DropTracker**

The DropTracker is intended to be used by clans to maintain an overall "loot leader", based on drops received by their members.

Clans can specify whether they'd like to track **all drops**, or only drops over a specified value. The drops are automatically sent to a Discord webhook, where they're read by the DropTracker discord bot, and added to the clan loot.

Every player is assigned a unique identifier, so that bad actors can't hijack your webhook channel with fake submissions.

Below is the current side panel that is displayed once a player is logged in & their identifier is correct:

Panel

Having had little knowledge with Java before starting this project, it's been quite a learning experience. I borrowed some of the code to get started from https://github.com/BossHuso/discord-rare-drop-notificater.

Without having a server set up, with your account registered in discord, you won't really be able to test / use the plugin; it's intended for clans--so in the event you would like to test it before adding it; please DM me here or on Discord so I can make you a dummy account!

Here's what a clan's loot leaderboard looks like:
image

### DIFF
--- a/package/verification-template/build.gradle
+++ b/package/verification-template/build.gradle
@@ -39,7 +39,7 @@ dependencies {
 		because "discord-rare-drop-notifier"
 	}
 	thirdParty("org.json:json:20200518") {
-		because "discord-rare-drop-notifier"
+		because "discord-rare-drop-notifier, droptracker"
 	}
 	thirdParty("org.jsoup:jsoup:1.13.1") {
 		because "discord-rare-drop-notifier, drop-simulator, fire-beats, loot-table, loot-lookup"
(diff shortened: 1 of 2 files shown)


Felanbird:
Don't close your PRs and keep creating new ones, push to the existing PR to resolve issues.

joelhalen:
> Don't close your PRs and keep creating new ones, push to the existing PR to resolve issues.

Not sure what I've done wrong here; I'm fairly new to Java and even more so to working on GitHub collaboratively; so I apologize for my arrogance. 
Looks like something to do with libraries based on the build output?
If you (or anyone) could point me in the right direction I'd appreciate you!

geheur:
You could use Gson instead since runelite uses it already. i.e.
```java
@Inject
private Gson gson;
```
then you won't have to deal with adding libraries.

joelhalen:
Looked through a few other pull requests and noticed that outbound connections usually mean the plugin requires a warning on install; hopefully I've added that where it belongs.

Nightfirecat (changes requested):
LGTM otherwise

PR #4418, opened 2023-05-29
Added pre-ge

plankst:
A simple plugin that simulates pre-Grand Exchange era by disabling any Grand Exchange related menu options.


Nightfirecat:
We [recently removed jogl from our allowed dependencies for hub plugins](https://github.com/runelite/plugin-hub/commit/f7eca6b9696b38864063054e30b0c84dc93dd7ae) as it didn't work well and caused a lot of crashes. We would recommend using lwjgl instead.

PR #4426, opened 2023-05-31
remove magic-secateurs. No longer needed

PISTOLCUPCAKES:
Remove magic-secateurs plugin as it is [no longer needed after Poll 78 changes](https://oldschool.runescape.wiki/w/Magic_secateurs#cite_note-1).


Adam-:
Instead of deleting the plugin, set `disabled=true`

Nightfirecat:
You can just re-open this PR and push a new commit instead of opening a fully new PR

PR #4427, opened 2023-05-31
stardust per hour calculator plugin

kevindarbydev:
Simple plugin to calculate stardust gained per hour. First runelite plugin


Felanbird:
@kevindarbydev Don't delete your branch/open a new PR - just update the existing branch with commit hash of your most recent change.

PR #4429, opened 2023-05-31
Stardust per hour Plugin

kevindarbydev:
Calculates and displays the hourly rate the player is accumulating stardust. First RL plugin to get the hang of things

### DIFF
--- a/plugins/stardust-per-hour
+++ b/plugins/stardust-per-hour
@@ -1,0 +1,2 @@
+repository=https://github.com/kevindarbydev/stardust-per-hour-calculator.git
+commit=4c6f53c3fee1123bcd60855d69601f793cda9882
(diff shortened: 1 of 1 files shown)


Nightfirecat:
FYI [plugins are automatically registered to the event bus, there's no need to manually do it](https://github.com/kevindarbydev/stardust-per-hour-calculator/blob/b4dd09574d9a14c9d33d62a99bcb6342f380f836/src/main/java/com/stardust/StardustPlugin.java#L78). Also [I have no idea how a config method which accepts arguments would work](https://github.com/kevindarbydev/stardust-per-hour-calculator/blob/b4dd09574d9a14c9d33d62a99bcb6342f380f836/src/main/java/com/stardust/StardustConfig.java#L17), but it seems like it's not used anyway.

PR #4434, opened 2023-06-01
Create check-for-spores

123z:
Notifies the user if seaweed spores spawn, can enable to only show the notification when underwater


123z:
im sorry for spamming these

PR #4445, opened 2023-06-03
Update action-progress

guillaume009:
Add a vertical mode for the resize and support for plank make spell


guillaume009:
As it might be visible, I am not sure of what I am doing. I tried to follow an example but it's obviously not working. Can someone help me?

abextm:
Delete your `javax.annotation.Generated` annotations, they are not available in Java 11+ and do nothing.

PR #4453, opened 2023-06-05
Update tasks-tracker plugin

tylerthardy:
Fixed by refactoring for new VarbitChanged event
Added new Combat Achievement varps (we are working on a future refactor to control this from external JSON instead)


tylerthardy:
Fixed GSON issue. PR has been open a while, so commenting for bump. Thanks!

PR #4458, opened 2023-06-06
Change authors for Osrs Profile

mathiasHillmann:
I hadn't realized that it would use my github username as the author for the plugin if not supplied.

Added a tag for the correct name


mathiasHillmann:
I will be creating another PR with some other changes.

PR #4460, opened 2023-06-06
Update RastaXP bar misplacement.

Beardedrasta:
Updated the bar placement causing the bars to be in middle of screen and text at the top.  Supports run, special attack as well.


Beardedrasta:
Not sure why it posted the previous commits.

PR #4473, opened 2023-06-09
Update Standalone ToA Points tracker

AbusiveTuna:
Adds pets, cleans up some files, adds an icon.


raiyni:
Are you using `src/main/java/com/tuna/toa/icons/Tumeken's_Guardian.png` and `src/main/java/com/tuna/toa/icons/Tumeken's_shadow.png`? They are very large images that should be resized if used or deleted if not.

AbusiveTuna:
> Are you using `src/main/java/com/tuna/toa/icons/Tumeken's_Guardian.png` and `src/main/java/com/tuna/toa/icons/Tumeken's_shadow.png`? They are very large images that should be resized if used or deleted if not.
whoops, I am not. Was added for an update I never ended up doing. Removed

PR #4476, opened 2023-06-09
Create bingo-event-automated

triplejingle:
https://github.com/triplejingle/Bingo-Event-Automated

Plugin is used to automate bingo events. It sends the loot(actiondata) to the server so the server can process it. It also sends an image when the loot is needed for the corresponding bingocard.


Nightfirecat:
Please use [a more descriptive config group than simply "event"](https://github.com/triplejingle/Bingo-Event-Automated/blob/12330a12f2c7539df462b42a574bb786575a78c6/src/main/java/com/bingoeventautomated/config/IEventConfig.java#L7)
LGTM otherwise

triplejingle:
Changed the key name of ConfigGroup from "event" to "bingoEvent"

PR #4506, opened 2023-06-17
Add Fun Reminders

erversteeg:
Fun reminders are configurable and each reminder can be stylized differently.

### DIFF
--- a/plugins/fun-reminders
+++ b/plugins/fun-reminders
@@ -1,2 +1,2 @@
 repository=https://github.com/erversteeg/Proximity-Prompts.git
-commit=4b97f8d5d02ed64bd5f3abd7f86b32ea04632ac5
+commit=f77bd49f53f8e6466d055941c439ce8044716e6d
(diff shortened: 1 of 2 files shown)


raiyni:
It would greatly improve the user experience and our time looking at your plugin if you used `ConfigManager::setConfiguration` instead of  generating a massive config file with python.

erversteeg:
So the issue I'm running into is if someone wasn't comfortable with json and wanted to create more than a few prompts, I can't see a way for them to do this without having the items already setup in the config. I can look into switching over to setConfiguration but I'm not seeing a way to create sections or add configurations to a specific section this way.

erversteeg:
Was just thinking if you all need a quick way to verify the integrity of the large config then grab the first 93 lines (everything before "Prompt 1" section) from the config file. Then run the updated copy of config.py and add the output onto the end. Finally do a diff between the file you created and the config file and they will come out the same. This way you know what is defined for each prompt section in config.py is the same and isn't all that different than doing a bunch of setConfiguration calls in a loop.

raiyni:
We aren't going to run a python script to verify your plugin isn't harmful.

erversteeg:
I updated the number of reminders to 10 and renamed the plugin to Fun Reminders.

PR #4512, opened 2023-06-20
Update wintertodt-scouter

nucleon:
changed plugin network latency from 5 seconds to 2 seconds due to decreased popularity


nucleon:
first commit failed build because _the plugin_ created a new instead of gson instead of using the clients

PR #4517, opened 2023-06-21
Create log-master

ImTedious:
Added file in plugins for log-master

Log-master is a plugin that adds a task list in a dashboard on the collection log interface which I will use to further my youtube series (Tedious' GenerateTask).

### DIFF
--- a/plugins/log-master
+++ b/plugins/log-master
@@ -1,0 +1,2 @@
+repository=https://github.com/ImTedious/generate-task.git
+commit=de5d07de19b298c4f642ba1ba11c3444803336aa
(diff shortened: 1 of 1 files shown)


Nightfirecat:
[Don't use "example" as your config group](https://github.com/ImTedious/generate-task/blob/de5d07de19b298c4f642ba1ba11c3444803336aa/src/main/java/com/logmaster/LogMasterConfig.java#L7). In fact, you could delete your config interface entirely as it's not used, if you like.
LGTM otherwise

ImTedious:
I have changed "example" to "log-master" and updated the commit hash.
Do I need to do a new pull request or will you be able to push it through
now?
Thanks!

Kind regards

Op do 29 jun 2023 om 03:57 schreef Jordan ***@***.***>:

PR #4523, opened 2023-06-23
Create Trouver Parchment Alerts plugin

zlight97:
This is a plugin that will notify you when you are carrying an unlocked (unparchmented) item. It can be configured to always be on, or only in pvp zones. It can also be configured to flash.


raiyni:
Don't use `ClientThread::invokeLater` on every client tick. Use `onGameTick` to do all of your logic without invoking.

zlight97:
> Don't use `ClientThread::invokeLater` on every client tick. Use `onGameTick` to do all of your logic without invoking.

Thank you for the tip! I have updated my PR with the commit hash with this change implemented.

PR #4525, opened 2023-06-24
Update skilling-notifications

erwinvc:
Added Lunar spells notification.
Added notification flash setting to panel.


erwinvc:
The label 'non-author plugin change' was added but I am actually the author. Changed my Github username recently.

PR #4527, opened 2023-06-25
Create TranslationPlugin

PlayerCoder1:
Translates everything in the chatbox from NPC dialogs to Players talking on the side panel for now it's English to Spanish but I'm going to add more langages in the futur


raiyni:
1. This plugin should adhere to their api usage and handle api limits correct https://mymemory.translated.net/doc/usagelimits.php
2. Don't spin up new threads every chat message, use an executor pool or something. 
3. Inject our okhttp client

PlayerCoder1:
> 1. This plugin should adhere to their api usage and handle api limits correct https://mymemory.translated.net/doc/usagelimits.php
> 2. Don't spin up new threads every chat message, use an executor pool or something.
> 3. Inject our okhttp client

is everything good now?

Nightfirecat:
We have not re-checked yet, please be patient

PlayerCoder1:
Oh I didn't know it had to be in lower caps sorry

PR #4528, opened 2023-06-25
Add Random Screenshot plugin

neeerp:
This plugin takes a screenshot with a 1 in `n` chance every game tick, where `n` is configurable.

The base of this plugin was lifted directly from the core Screenshot plugin and simplified significantly.

(This plugin was suggested in https://github.com/runelite/runelite/discussions/16608. Here, it's suggested that this should _not_ be part of the core Screenshot plugin, hence why I've made this plugin hub plugin)


raiyni:
This looks fine. Your startup code is unnecessary to have.

neeerp:
I'll update the plugin to get rid of the startup code then. I just verified that the folder gets re-created when a screenshot is saved.

PR #4532, opened 2023-06-27
Update Old School Snitch to 1.0.1

wkpatrick:
Add support for exporting player model (You can see how this would look [here](https://oldschoolsnit.ch/character/2vnluqXHkexeaWgQgkuv2s)).
Check that API Key exists before sending data.


wkpatrick:
The build is currently failing due to:
Caused by: org.gradle.api.InvalidUserCodeException: Entry META-INF/LICENSE is a duplicate but no duplicate handling strategy has been set. Please refer to https://docs.gradle.org/7.4/dsl/org.gradle.api.tasks.Copy.html#org.gradle.api.tasks.Copy:duplicatesStrategy for details.

Due to Jackson Annotations

wkpatrick:
Closing, it will be faster to write my own GLTF encoder and save dealing with the dependency mess.

PR #4534, opened 2023-06-27
Fix client freezing issue

MakingStan:


Nightfirecat:
In the time since this plugin's creation we've changed our policies to disallow reflection. This means you should change this code:

```java
c = Class.forName("com.code.OefPlugin");
URL url = c.getClassLoader().getResource(soundFilePath);
```

to something more like this:

```java
URL url = OefPlugin.class.getResource(soundFilePath);
```

LGTM once that change is made.

MakingStan:
Should be updated now, thanks for the code example too!

PR #4535, opened 2023-06-27
Create reword-manager

ejaz-karim:
Manage lists to reword Chat messages, Items, NPCs, Objects


Nightfirecat:
Code LGTM, but we're concerned that this plugin could be used to create faked video/stream footage by editing player chat messages client-side. Please either remove chat message editing or add proper checks to ensure it will not modify player messages.

ejaz-karim:
> Code LGTM, but we're concerned that this plugin could be used to create faked video/stream footage by editing player chat messages client-side. Please either remove chat message editing or add proper checks to ensure it will not modify player messages.

@Nightfirecat Would it be acceptable if I made modified chat messages have a warning header in red text, and each modified word be in red text or the entire message in red text to make it obvious that the message was modified by the plugin.
For example:
Bob: **_[modified]_** Hi my name is **_Wise Old Man_**.
I could even add the warning emoji or any other requests the Runelite team would feel is appropriate.

Nightfirecat:
What use case is there for modifying player messages?

ejaz-karim:
This list is by no means exhaustive, just what I came up with.
- A clan in the wilderness or castle wars may want to prevent non-clan members from seeing/understanding their tactics in public chat. This also prevents scouts from spying in clan chats or from standing and waiting across various points in the wilderness waiting for clan information in public chat messages
- For immersive gameplay, E.g. Players can use creative names for items/objects, and see them being used in chat. This can help content creators with storytelling and other cinematic purposes
- For abbreviating or unabbreviating words, e.g. dwh turns into dragon warhammer. This is so that players who are forgetful of item names or who are new to the game may understand what other players are talking about.
- To create a custom word filter, so that players may prevent seeing certain words and replace them with their own words of choosing. This is different to the Chat Filter plugin as that doesn't have the option to do a word replacement filter type.
- Having certain words be translated to the language of your choice. Yes, there are translation plugins but what if I only want a couple words in a sentence to be translated only. This can assist with language learning.
- Disabled players may communicate with friends easier as they can share their personal Chat List.
- Quick chat message presets that can be shared with clan members or friends
- Personal notes embedded into messages. E.g. If a player says "Crandor", you can make it so that it will return "Crandor (Go through Karamja dungeon)
- Allows players to set challenges for themselves. E.g. a snowflake account that has chat messages scrambled etc.
- Friends can have fun with inside jokes. E.g. if they said "_[name of friend]_" it would return "noob"

Nightfirecat:
I'm willing to accept that reasoning--your proposal of adding a `[modified]` prefix should be sufficient for this. If you also want to add some kind of styling to the modified text, that is fine, but not necessary in my opinion.

PR #4543, opened 2023-06-28
Update Gauntlet map

StickySerum:
Updated the package name and moved the resources into a new folder structure to hopefully fix a bug in the plugin.

Also updated the repo name from "Gauntlet Map" to "gauntlet map" and updated the repo url to match


StickySerum:
Sorry, fixing an issue with the branch. It wouldn't let me push changes for whatever reason.

PR #4550, opened 2023-06-29
Create blackjackv2

janbeenham:
This plugin automatically swaps between knock-out and pickpocket, no more RSI when doing blackjack.


Bram91:
This is against the statement found here: https://secure.runescape.com/m=news/third-party-client-guidelines?oldschool=1

PR #4551, opened 2023-06-29
Vital issue with low-value drops for clan events

joelhalen:
Previously, we selected to remove the npc or event name from low value drops to reduce the size of each embed sent to discord.

This presents an issue for clan events such as the one we are about to run in a few days, where players can obtain the items required for their task by opening implings, etc.

As a result, this update is quite small as it simply adds an extra field to the embed for each message sent; and a speedy approval would be greatly appreciated!


Felanbird:
I'd suggest not doing your changes on master, to prevent this issue from happening again.

PR #4561, opened 2023-07-01
Added Forestry Helper Plugin

Diabolickal:


Nightfirecat:
You need to [remove your added overlay(s) when the plugin shuts down](https://github.com/Diabolickal/forestry-plugin/blob/9f8eec674a8343144c20c75358c5897e901a96ee/src/main/java/com/diabolickal/forestryplugin/ForestryHelperPlugin.java#L102) so they do not persist when disabling the plugin.

Diabolickal:
> You need to [remove your added overlay(s) when the plugin shuts down](https://github.com/Diabolickal/forestry-plugin/blob/9f8eec674a8343144c20c75358c5897e901a96ee/src/main/java/com/diabolickal/forestryplugin/ForestryHelperPlugin.java#L102) so they do not persist when disabling the plugin.

Fixed

Diabolickal:
Runelite now has essentially all the features this plugin has. No need for a separate plugin. Closing this.

PR #4562, opened 2023-07-01
Add Tree Count plugin

Infinitay:
With OSRS's first part of the forestry update, they added invisible boosts to woodcutting depending on the number of players chopping a given tree. This plugin helps users distinguish how many players are chopping a tree by displaying a counter on the tree indicating how many players are currently chopping it.

 
image


Infinitay:
Last push completed. Initially had an issue with not including a license, but got carried away updating the plugin as well. Should be done updating things until the plugin gets released and there are updates to be made.

For now I've hardcoded the ids as suggested to release the plugin sooner. I'll submit an update reverting to referencing the ids within `ObjectID` instead once the snapshot version `1.10.8-SNAPSHOT` or later has been released as the latest RuneLite version.

PR #4567, opened 2023-07-01
Highlight stackable items

horse4lunch:


horse4lunch:
Highlights any drops that can be stacked in your inventory.
Example. If your carrying fire runes and get a fire rune drop they will

PR #4568, opened 2023-07-01
Create highlight-stackable-items

horse4lunch:
Highlights any drops you get that can be stacked with items in your inventory
Example. If you have fire runes in your inventory and get a fire rune drop they will be highlighted.

### DIFF
--- a/highlight-stackable-items
+++ b/highlight-stackable-items
@@ -1,2 +1,2 @@
 repository=https://github.com/horse4lunch/Highlight-Stackable-Items.git
-commit=a9bfca1d8d45df1922a2c6683dc6eb0020b251c4
+commit=12e6610961bdd6948db4fd98873c9a90dab27152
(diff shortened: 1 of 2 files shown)


raiyni:
You need to move your file into the plugins for

horse4lunch:
> You need to move your file into the plugins folder

Damn sorry, Got it fixed now I think.

Nightfirecat:
Please don't [check spawned items and config values every client tick](https://github.com/horse4lunch/highlight-stackable-items/blob/b2fb024ab4963fb7b7c0aba6e71c7e61d5c9b14b/src/main/java/com/HighlightStackables/HighlightStackablesPlugin.java#L92), run this code on game ticks instead.

Also it looks to me like this plugin will leave the ground items plugin's configuration in an incorrect state if you close your client, since `shutDown()` will not be called. You should take care to reset it on ClientShutdown as well as plugin shutDown.

horse4lunch:
> Please don't [check spawned items and config values every client tick](https://github.com/horse4lunch/highlight-stackable-items/blob/b2fb024ab4963fb7b7c0aba6e71c7e61d5c9b14b/src/main/java/com/HighlightStackables/HighlightStackablesPlugin.java#L92), run this code on game ticks instead.
> 
> Also it looks to me like this plugin will leave the ground items plugin's configuration in an incorrect state if you close your client, since `shutDown()` will not be called. You should take care to reset it on ClientShutdown as well as plugin shutDown.

Made the needed changes and added a loop to remove items from Uniquelist. Should be good to go now.

PR #4576, opened 2023-07-03
Create attack-sound-notifications

DominickCobb-rs:
Plays audio depending on the results of attacks on hitsplat.
All audio played is configurable and customizable*
Ex1: When Dragon Warhammer spec hits 0 a squeaky toy sound is played.
Ex2: When a basic attack is missed, a fart noise is played.
Audio is customized manually via file system.

### DIFF
--- a/plugins/attack-sound-notifications
+++ b/plugins/attack-sound-notifications
@@ -1,0 +1,2 @@
+repository=https://github.com/DominickCobb-rs/Attack-Sounds-Notifications
+commit=21811000c4bdc7a619d998c8e64a17a6452940eb


DominickCobb-rs:
This plugin is very similar to the Hit-Sounds plugin, but this one is much more versatile.

LlemonDuck:
* Your license file is attributing copyright to another developer (Ferrariic). If this is using modified code from him you should keep that, but also add your own name to the list.
* Some of the sound files provided as defaults are copyrighted. You'll need to either replace or remove them. You can instead provide no defaults and require users to find their own sounds.

DominickCobb-rs:
Is the Windows error sound okay to keep, or does that one also need to go?

DominickCobb-rs (commented):
Removed default audio aside from the Windows error sound

Nightfirecat:
You'll also need to remove the windows error sound; we can't distribute that as it's definitely copyrighted.

DominickCobb-rs:
Should've clarified, the plugin contains no audio resources at all anymore and expects the user to provide their sounds. The windows error sound has been replaced with `java.awt.Toolkit.getDefaultToolkit().beep()`

Nightfirecat:
Please [remove `enabledByDefault = false` from your plugin descriptior](https://github.com/DominickCobb-rs/Attack-Sound-Notifications/blob/d5218f20abc06357d8a56cfc4e915089bb5c9305/src/main/java/com/AttackSoundNotifications/AttackSoundNotificationsPlugin.java#L70); it's confusing for users to install a hub plugin which isn't enabled upon installation.
Please use `ImageUtil.loadImageResource()` [instead of `ImageIO.read()`](https://github.com/DominickCobb-rs/Attack-Sound-Notifications/blob/d5218f20abc06357d8a56cfc4e915089bb5c9305/src/main/java/com/AttackSoundNotifications/AttackSoundNotificationsPlugin.java#L123) as it provides synchronization for ImageIO

PR #4577, opened 2023-07-03
Create AColdOne Clogging

OmbudRov:
A plugin that announces Collection Logs with the voice of AColdOne. I've copied code from m0bilebtw's C Engineer plugin and have given him the appropriate attribution 

Here is permission to use the audio clips from the Big Man himself: image


Nightfirecat:
[Don't use "example" as your config group](https://github.com/OmbudRov/AColdOne-Clogging/blob/d64004d76dbaa479f0c99641f083b05f74792348/src/main/java/com/acoldoneclogging/AColdOneCloggingConfig.java#L8)

OmbudRov:
i have fixed my oopsie, terribly sorry about that

PR #4585, opened 2023-07-04
Create runecost

MJensen03:
Adds a simple panel to the game that allows you to calculate how much it would cost to purchase runes from certain rune shops in the game.

### DIFF
--- a/plugins/runecost
+++ b/plugins/runecost
@@ -1,2 +1,2 @@
 repository=https://github.com/MJensen03/Rune-Cost-Calculator.git
-commit=d4f973c5bd21eb40a02d11d1f0fa039b5562c4c2
+commit=6c2923fc695ad5a19c922f87bdc1bf7149ecb845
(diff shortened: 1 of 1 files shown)


MJensen03:
Added License to repo

Nightfirecat:
It looks like [you forgot to remove `com/example` from your source?](https://github.com/MJensen03/Rune-Cost-Calculator/tree/6c2923fc695ad5a19c922f87bdc1bf7149ecb845/src/main/java/com/example)
Your plugin [adds a panel during startup](https://github.com/MJensen03/Rune-Cost-Calculator/blob/6c2923fc695ad5a19c922f87bdc1bf7149ecb845/src/main/java/com/runecost/RuneCostPlugin.java#L93) but never removes it. Please remove it on shutdown so disabling this plugin would actually remove the sidebar panel button

MJensen03:
I think it should be updated with the requirements.  Let me know if I need to adjust more to the requirements.

PR #4596, opened 2023-07-07
Create Autocasting

zbruennig:
`autocasting` is a ground-up rework of `autocast-utlities`. Thus I removed the unlisted and unmaintained `autocast-utilities` and replaced it with `autocasting`.

JC, the original author and still-maintainer and repo-hoster, and I have worked together to fix its broken features and add many new ones. Both of us are active and will be responding to issues.

For a quick summary, this plugin is similar to attack styles but for autocasts. It will track your current spell being autocast, as well as number of casts remaining. If you brew your magic level down to the point where the autocast is unusable, or you run out of runes, it will detect that and send notifications.

A more complete list of features is available on its [README](https://github.com/jcarbelbide/autocasting#readme).


Nightfirecat:
Don't remove the old plugin file; it is already disabled, so it's not being built anyway

Nightfirecat:
If this is the same plugin as it was before, it may be better to remove the `disabled` line from the existing autocast-utilities plugin entry, as that would re-activate the plugin for any past users who had it installed, letting them continue to use it without seeking it out on the hub again. That said, if the rework you've done is significant enough that it would be better to be built up with a fresh userbase, the way you have it set up now would be fine.

Some brief notes here before I eventually get around to doing a proper review of the code:
* This file should have an `authors=` line including all the GitHub usernames of plugin contributors to remove the "non-author plugin change" label
* @jcarbelbide I see Zach has contributed a good amount to the repo but you're still the owner of the project, can you confirm he has your blessing to submit this plugin (and any future updates to it, if the above `authors` line is added to the plugin entry?
* [You're registering `subscriptions` with the event bus on startup](https://github.com/jcarbelbide/autocasting/blob/83cd2559154390a4d57ca3c3b0af8f1431206974/src/main/java/com/autocasting/AutocastingPlugin.java#L56), but it is never unregistered. Please make sure it's removed on plugin shutdown

zbruennig:
> If this is the same plugin as it was before, it may be better to remove the `disabled` line from the existing autocast-utilities plugin entry, as that would re-activate the plugin for any past users who had it installed, letting them continue to use it without seeking it out on the hub again. That said, if the rework you've done is significant enough that it would be better to be built up with a fresh userbase, the way you have it set up now would be fine.
> 
> Some brief notes here before I eventually get around to doing a proper review of the code:
> 
> * This file should have an `authors=` line including all the GitHub usernames of plugin contributors to remove the "non-author plugin change" label
> * @jcarbelbide I see Zach has contributed a good amount to the repo but you're still the owner of the project, can you confirm he has your blessing to submit this plugin (and any future updates to it, if the above `authors` line is added to the plugin entry?
> * [You're registering `subscriptions` with the event bus on startup](https://github.com/jcarbelbide/autocasting/blob/83cd2559154390a4d57ca3c3b0af8f1431206974/src/main/java/com/autocasting/AutocastingPlugin.java#L56), but it is never unregistered. Please make sure it's removed on plugin shutdown

We've also renamed the repo, but I'm assuming that doesn't matter for the build, correct? If so then the file in plugin-hub would be different than both the plugin display name and its repo name. Is that intended?

jcarbelbide:
Hey @Nightfirecat, appreciate the fast response on this! Yes, thanks for confirming :) @zbruennig has done a lot of great work to bring this back to life and add some core features that were missing from the original plugin. Really, all credit for this release goes to him! Hopefully with these changes, it will be much more useful to people

Echoing Zach's questions above, please let us know how a rename should happen. I think it would be great if this took the spot of the old plugin and just worked for users of the original plugin

Nightfirecat:
That would be fine; you could change the repository link in the existing file if you like, but so long as GitHub's redirect works it wouldn't matter. (for longevity I'd recommend updating that link) As for the name not exactly matching that of the plugin entry's filename, that doesn't matter.

jcarbelbide:
Got it, so to retain users of the old plugin, we would need to leave the old file name unchanged, and change the fields in there? And this file name wouldn't show up anywhere else?

Nightfirecat:
Correct. The files in `plugins/` are simply a reference for the hub's build process--their name is used only for building the plugins (a key for that process), and for the URL you'd find the plugin listed at under https://runelite.net/plugin-hub. Eg. I'm the author of the Wilderness Lines plugin, which is found at <https://runelite.net/plugin-hub/show/wilderness-multi-lines>--named so because it originally only displayed multi-combat lines, and only later had a more general lines display feature added.

jcarbelbide:
Got it, thank you so much for the explanation! That makes complete sense!

PR #4599, opened 2023-07-07
Update statussocket

The-Jani:
Updated plugin to work with the most recent version of RuneLite, removed Skill.OVERALL from the fetch list, as of it's deprecated since v1.10.6.

Plugin seems to be abandoned.


The-Jani:
Original creator merged. Closing :)

PR #4603, opened 2023-07-08
Update boss health indicators

rugg0064:
Update boss health indicators to add two new features:
Config option to hide the side bar panel
image

Trigger runelite notifications whenever an hp % threshold is reached on enabled indicators
image
notifications


Nightfirecat:
["main" is not an acceptable config group name](https://github.com/rugg0064/boss-health-indicators/blob/33b116479bba5fefa98ce91785409aabe9f4decc/src/main/java/com/boss/health/indicator/BossHealthIndicatorConfig.java#L7)
LGTM otherwise

PR #4604, opened 2023-07-08
Update statussocket

DStatIO:
Updated plugin to work with latest changes since it was broke.


Nightfirecat:
This code change looks good on its own. Can you also [remove `enabledByDefault = false` from your plugin's descriptor?](https://github.com/DStatIO/StatusSocket/blob/65f969d9ee9d7a99912782583f66db7e3019d55b/src/main/java/com/statussocket/StatusSocketPlugin.java#L31) It's confusing for users to install plugins which don't enable themselves upon installation.

LlemonDuck:
closing as stale, feel free to reopen if you resolve Nightfirecat's comment above

PR #4609, opened 2023-07-08
Update equipment-inspector

Waistcoat:
Temporarily storing equipment info when doing a right-click/menu-open so that you don't need to be in draw distance if you're inspecting someone that runs by. https://github.com/botanicvelious/Equipment-Inspector/pull/20

### DIFF
--- a/plugins/equipment-inspector
+++ b/plugins/equipment-inspector
@@ -1,2 +1,2 @@
 repository=https://github.com/botanicvelious/Equipment-Inspector.git
-commit=27f88a17b4c8d38d216020073b1ee699597e9009
+commit=4b0f5f29ae2a769203cfad6150698c9c70927555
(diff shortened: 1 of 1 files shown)


Nightfirecat:
Code change LGTM. Besides the changes here though, [can you remove `enabledByDefault = false` from your plugin descriptor](https://github.com/botanicvelious/Equipment-Inspector/blob/27f88a17b4c8d38d216020073b1ee699597e9009/src/main/java/equipmentinspector/EquipmentInspectorPlugin.java#L39)? It's confusing for users to not have it active upon installation. With that and the `authors` addition to this plugin entry, this should be fine to merge.

Waistcoat:
> Code change LGTM. Besides the changes here though, [can you remove `enabledByDefault = false` from your plugin descriptor](https://github.com/botanicvelious/Equipment-Inspector/blob/27f88a17b4c8d38d216020073b1ee699597e9009/src/main/java/equipmentinspector/EquipmentInspectorPlugin.java#L39)? It's confusing for users to not have it active upon installation. With that and the `authors` addition to this plugin entry, this should be fine to merge.

@Nightfirecat Made the changes you requested. Those automatic tags are nifty

PR #4612, opened 2023-07-09
Add GOTR Essence Counter Plugin

kentor:
Plugin for counting guardian essences crafted or mined during a game of Guardians of the Rift.


kentor:
actually. i might make a party version of this

PR #4618, opened 2023-07-11
Dmwatch 1.7 + base repo owner change

zguilt:
The DMWatch org account owner is no longer around, might as well do the development from the dev account anyhow.

The website that the plugin will now interact with is a live end point that gets the name updates of scammers, relying on github  has a 5 minute cache delay.

By offering closer to real time data by default we will help protect people from serial scammers that name change 20 times in a day.


DMWatch:
Do not approve this.

PR #4620, opened 2023-07-11
Birdhouse

cavine:
A plugin that helps the user know when their birdhouse run is available. Writes to a local file, so it should persist across sessions.


Felanbird:
Doesn't `Time tracking reminder` plugin already do this by just consuming the native `Time tracking` plugin?

cavine:
> Doesn't `Time tracking reminder` plugin already do this by just consuming the native `Time tracking` plugin?

Oh, it totally does, lol. My bad didn't see it earlier.

cavine:
Closing due to Time Tracking Reminder doing this better.

PR #4623, opened 2023-07-13
Add Bone Luck

Christian97Lewis:
This plugin will detect if you are offering bones at the Chaos Altar (wilderness) and offer some helpful information such as:

- bones used
- bones saved
- saved percentage
(optional configs)
- total XP gained in session
- expected XP gained (based on average luck)

### DIFF
--- a/plugins/bone-luck
+++ b/plugins/bone-luck
@@ -1,2 +1,2 @@
 repository=https://github.com/Christian97Lewis/bone-luck.git
-commit=c0deb635c7976f6d68e08a52348cae75dc140b30
+commit=2b61fd01484c112bef0ca289bb4c6aa9c10abcff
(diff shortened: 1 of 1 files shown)


Nightfirecat:
Don't use `enabledByDefault = false` in your plugin descriptor; it's confusing for users to install a plugin which isn't enabled after installation

Nightfirecat:
[Don't use `loadWhenOutdated = true`](https://github.com/Christian97Lewis/bone-luck/blob/c0deb635c7976f6d68e08a52348cae75dc140b30/src/main/java/com/boneluck/BoneLuckPlugin.java#L23); your plugin cannot work when the client is on an old revision.
Code LGTM otherwise

Christian97Lewis:
@Nightfirecat fixed

PR #4638, opened 2023-07-15
Add left-click-bank-actions plugin

evanblaine:
Adds configurable item equipping, eating, and drinking directly from the bank interface using only the mouse.


raiyni:
Wouldn't it be easier to tell people to use the normal menu swapper?

evanblaine:
Apologies, I was unaware of Runelite having this functionality.

PR #4639, opened 2023-07-16
Add charge calc

vikke1234:
Add buttons to fill a sang/shadow etc. either fully or `total - N`. This is mainly used for when you're smuggling runes in a chargable staff.


vikke1234:
https://youtu.be/APf4lY63G-0 here is a video of it

PR #4647, opened 2023-07-17
Update wilderness warnings

LuxOG:
remove bounty hunter worlds from list as they are no longer different in the wilderness


plugins/wilderness-warnings
```diff
@@ -1,2 +1,2 @@
 repository=https://github.com/CalumMagee/wildernesswarnings.git
```
Felanbird, line 1: This should be updated to your new name, so the bot doesn't mislabel your PRs
```suggestion
repository=https://github.com/LuxOG/wildernesswarnings.git
```

Nightfirecat:
Added in dbfe39b8d1989430a18c90cc98ceb0eafcb931a1

PR #4650, opened 2023-07-17
Adding zeah rc helper plugin

fSunday:
Highlights Zeah Runecrafting Altars(Blood/Soul)

### DIFF
--- a/plugins/zeah-rc-helper
+++ b/plugins/zeah-rc-helper
@@ -1,0 +1,2 @@
+repository=https://github.com/fSunday/zeahrc.git
+commit=22fbab9d002b6bb64a9498ac386f300bdebf0c15


Bram91:
Does this plugin do anything that <https://runelite.net/plugin-hub/show/zom-dense-essence> doesn't? You could make a PR to their repo to prevent clutter on the hub. Though it seems that `zom-dense-essence` covers all the functionality already.

fSunday:
That plugin highlights the runestones when you can mine them - this one highlights the altars when you have dark essence fragments/blocks in your inventory. I guess I can make a PR to that one instead?

Bram91:
That would be preferred yes, it would make sense for a feature like this to be included in there.

fSunday:
Sounds good - thanks.

PR #4653, opened 2023-07-18
Add Bank multisearch plugin.

geheur:
plugin to search by equipment slot and for multiple items at once (e.g. "chisel sapphire" to search for items containing "chisel" or "sapphire").


geheur:
@raiyni It's so that I can tell in a future update if the plugin was previously installed. e.g. if I want to add some feature that should be on in new installations by default but I don't want it to be enabled by default for people who already have it installed and are already used to existing functionality. Since there is no "plugin update" notification they might not know why some new behavior is happening.

PR #4658, opened 2023-07-19
Update konar-milestone-reminder with slayer master hiding

michael-gutman:
Adds the ability to hide non-Konar slayer masters when Konar reminder is active.
Props to @Henness0666


MrPryor:
Is this update live? I don't see a option to hide non-Konar slayer master in settings.

PR #4664, opened 2023-07-22
Add "dmsafe" plugin

DMSafe:
DMSafe - Keeping track of scams that occur in the PvP Worlds


DMWatch:
This is blatantly copied off DMWatch, Please be cautious approving this.

Adam-:

PR #4667, opened 2023-07-23
Add "ABC Alchemy" plugin. 

vartan:
This plugin includes 2 major features:

1. An overlay that highlights the intersection between the alchemy spell and the optimal inventory slot.
2. A sidebar panel that shows the most profitable items to alch. All prices use RuneLite APIs.

### DIFF
--- a/plugins/abc-alch
+++ b/plugins/abc-alch
@@ -1,3 +1,3 @@
 repository=https://github.com/vartan/abc-alch.git
-commit=c0e1952800e810bc3144dd2217d51afb1b77517e
+commit=75b7d27b40e20ca34625ece352d62bc4820b9f31
 authors=Vartan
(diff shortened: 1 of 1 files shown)


majed08 (approved):

Nightfirecat:
[Your panel icon is too large](https://github.com/vartan/abc-alch/blob/2d95b078a5d90ea78387ed0ce5f302e5d67bc81f/src/main/resources/icon.png). Although it is resized visually for the sidebar button, it takes up memory per pixel when read by Java. Please scale it down.
Code LGTM otherwise

vartan:
Done! thanks for taking the time to review :)

Nightfirecat:
You resized the wrong icon. You resized [assets/icon.png](https://github.com/vartan/abc-alch/blob/master/assets/icon.png) which isn't packaged with your plugin. [src/main/resources/icon.png](https://github.com/vartan/abc-alch/blob/master/src/main/resources/icon.png), which is actually shipped with your code, remains too large.

vartan:
you are correct, sorry.

PR #4690, opened 2023-07-27
Added soul runes to magic guild shop

MJensen03:
Updated Typo


Nightfirecat:
Stop opening new PRs; just update the one which was already open

PR #4696, opened 2023-07-29
Update lazy-farming

Speaax:


Speaax:
Is this the way to update a plugin?

geheur:
Looks fine to me. You might want to change the version in build.gradle as well, but that is optional.

Speaax:
Ok thanks, I will try and remember this for the next update then.

PR #4700, opened 2023-07-31
Update Nexus Menu Map

Antipixel:
Updated spelling of 'Carrallanger' to reflect change in latest game update, which was causing the teleport spell to become inactive even when unlocked.


raiyni:
Don't close, open new prs. Push to the same branch and the bot will pick up the update.

PR #4707, opened 2023-08-01
Add PVP Protection

pvpProtection:
PVP Protection is a spin off & rebrand of the now defunct DMWatch project. The goal is the same, to allow people to be tagged as trusted and to allow people to be tagged as scammers/rushers/ahkers and cheaters. 

This project's features will start off as equal as to what DMWatch 2.0 version was going to launch as and will diverge as the project matures.


Adam-:

PR #4716, opened 2023-08-03
Twitch emotes

skotizhoe:
sorry I closed my last one, I realized I didn't have the correct link in my repository. Anyway, I created a plugin that uses Twitch emotes in Chat. Thanks

### DIFF
--- a/plugins/twitch-emotes
+++ b/plugins/twitch-emotes
@@ -1,0 +1,2 @@
+repository=https://github.com/skotizhoe/Twitch-Emotes
+commit=cf589aa96e3443570d0a43d59c538722e5b168c7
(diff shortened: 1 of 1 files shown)


Nightfirecat:
These images aren't licensed such that we can distribute them

jbpratt:
hey @Nightfirecat, sorry this isn't something that we thought of. The custom made emotes can be licensed CC by the original authors but the ones pulled from global twitch/bttv are going to be difficult/impossible to track down to get licensed. I assume if we moved forward with only images that we have explicitly licensed, there wouldn't be an issue with it?

Thanks for taking a look!

raiyni:
Also, in the future don't close and open new prs. Commit the changes to the same branch and the bot will detect the changes.

Nightfirecat:
Correct. You can search closed hub PRs for other twitch emote plugins; Twitch simply doesn't give permission for those to be used elsewhere. As for custom emotes, any which you're able to obtain permission to reuse and distribute we would be fine with shipping.

Adam-:
Open a new PR if you want to make a version with custom emotes. Otherwise, we cannot use or distribute the twitch emotes.

PR #4718, opened 2023-08-03
Re-enable dmwatch

DMWatch:
Due to popular demand, our community has encouraged us to re-release the plugin. 

Since the deactivation of the plugin we have seen many people scamming again due to no ‘Runewatch’ equal for Deathmatching.
 
For the public good of the Deathmatching scene, we would like to see DMWatch back on the hub to be available to help keep scammers accountable like we originally set out to.


Adam-:

PR #4726, opened 2023-08-05
Update tob-mistake-tracker to v2.3

QuestingPet:
Fix issue where Verzik mistakes were not being detected

Looks like there was a recent change to when NpcSpawned event gets fired, which causes the PoseID to no longer be the same as what it was (it's now -1 at the time of event being fired). Using Verzik NPC IDs instead now.


QuestingPet:
Here is the diff of this commit, since the diff above is wonky (since it's going across repos):

https://github.com/QuestingPet/TobMistakeTracker/commit/835f0ee6bb251d5b335c5eb7040be690ebd60a9f

PR #4744, opened 2023-08-07
Watchdog: v2.14.1 - Remove SoundFiredAlerts, alert filter, mp3 support, and eleven labs tts

adamk33n3r:
Jagex no likey the sound fired alerts, so this removes them.

This also adds the jaco mp3 player, did I do it correctly?

### DIFF
--- a/plugins/watchdog
+++ b/plugins/watchdog
@@ -1,3 +1,2 @@
 repository=https://github.com/adamk33n3r/runelite-watchdog.git
-commit=07ec39cd52727b39a4d1868e27d4156d3ce6cb1d
-disabled=Disabled at Jagex's request
+commit=1eb7abf1397dcf5d2502a08cb2cba12c53581d69


Nightfirecat:
The HTTP calls to `https://api.elevenlabs.io` should be done only if the API key config is not null or empty, otherwise this plugin requires a warning for calls to external sites.
Code LGTM otherwise

adamk33n3r:
> The HTTP calls to `https://api.elevenlabs.io` should be done only if the API key config is not null or empty, otherwise this plugin requires a warning for calls to external sites. Code LGTM otherwise

Technically the code that calls the api doesn't run unless you have enabled the TTS config option that gives that warning. I can make the change you suggested if you think I should too though. 
https://github.com/adamk33n3r/runelite-watchdog/blob/master/src/main/java/com/adamk33n3r/runelite/watchdog/WatchdogConfig.java#L93
https://github.com/adamk33n3r/runelite-watchdog/blob/master/src/main/java/com/adamk33n3r/runelite/watchdog/ui/notifications/panels/TextToSpeechNotificationPanel.java#L50

AnthonyPoschen:
any new feedback on this PR @Nightfirecat? im just a user who likes using this plugin and following along wondering when it will make a return.

PR #4752, opened 2023-08-08
Create name-changer

mlgudi:
Allows users to change specific player names to names of their choosing using a side panel.

Requested by Soup for use during Gielinor Games recordings.


LlemonDuck:
Thank for the submission but we are not interested in hosting this on the plugin hub.

PR #4769, opened 2023-08-13
Update npc-accessibility-tagger

R-Y-M-R:
- No longer can tag "invisible" npcs. Tested @ Duke.
- Reduce logging, update logging messages


LlemonDuck:
You should switch this to `if (npc.getName() == null || npc.getName().equals("") || npc.getName().equals("null"))`

https://github.com/R-Y-M-R/NpcAccessibilityTaggerPlugin/compare/3a96b1a3faedb00349514bd45fb269a7c0ac6a1b...R-Y-M-R:b374750b39b5a6b6b0cb770451660833e4659b4a#diff-c76ed7ec5a0ca66b3c8c97dc601bcc0bae4fcc5807cee699b5fc1f65401941a5R87

R-Y-M-R:
A pedantic for my tastes but noted. I could also check name.length == 0 instead of .equals("") but not bothered by this as the purpose of this update was not to optimize rather strictly to keep the plugin in line with Jagex rules.

LlemonDuck:
This isn't an optimization thing, a null string is not the same as an empty string nor a length-zero string.

R-Y-M-R:
In my code if the npc.name string is null, it returns false

R-Y-M-R:
You're right @LlemonDuck I **had** made a mistake and thought it was optimizations when I looked on my phone, but its not actually just that. Thanks for catching it!

PR #4779, opened 2023-08-16
Add Boss Kicker

JZomDev:
https://streamable.com/gij93p

Yeet!


raiyni:
need to update here https://github.com/JZomDev/zom-external-plugins/blob/9009f05f81d2399d7e41d491a6d21b76e1d254e9/runelite-plugin.properties if you intend to use the Boss Kicker name

PR #4788, opened 2023-08-20
Create tk swamp helper

PlayerCoder1:
Helps you when you get the swamp room in the temple trekking, tells you which way to go without losing time for nothing since the pathing is terrible


PlayerCoder1:
Is there something wrong that it's not being checked on by the mods?

MosheBenZacharia:
> Is there something wrong that it's not being checked on by the mods?

I think there's just a big backlog since the mods that review hub PRs have been busy

Nightfirecat:
FYI we originally closed access to the developer tools plugin from the launcher because players were using it to solve the temple trekking swamp puzzle. That said, that was a change made _years_ ago, so this is probably allowable under current client guidelines. I'm prioritizing plugin updates in the queue at the moment, but will make sure to get back to this in time.

Nightfirecat:
We've decided this is probably fine now under current client guidelines.

PR #4790, opened 2023-08-20
Update max-hit-calculator to v1.1.8

j-cob44:
- Bugfixes on Plugin Install and plugin enabling while already logged in (j-cob44/max-hit-calc#15)


j-cob44:
Version 1.1.7 was also squashed appropriately before this pull request.

PR #4791, opened 2023-08-21
Projectile changer

Defqonlord:
This only changes the projectile ID of certain bosses. Currently it only changes the projectiles of Vardorvis to a couple of drop down options. In future this plugin will have more variety of projectiles to be changed to if the user wishes for ease of use. This will also expand to more bosses that will also be able to have their projectiles changed.

This plugin does not add any sounds or display anything that isn't already currently in the game. This only provides a way for the user to choose which projectile they wish to see when the attack occurs. It is not altered in any other way other than from being one projectile to another. No size, speed or additional visuals are added.


raiyni:
I don't think we're interested in modifying boss mechanics like this, especially for newly released content.

Defqonlord:
> I don't think we're interested in modifying boss mechanics like this, especially for newly released content.

This doesn't change any mechanics only the visual. I can see two plugins that already do that same thing "ToB QoL" and "Vardorvis Projectiles". The only difference is that mine you can add more bosses but at no point is it changing what happens just the visual projectile to another more appealing easier to see projectile that already exists in game.

LlemonDuck:
we're not interested in hosting this, closing

PR #4806, opened 2023-08-27
Update quest-helper to 4.1.1

Zoinkwiz:
Adds the first half of Desert Treasure II, along with various bug fixes.


Zoinkwiz:
Updated the PR to include the remaining sections of DT2, so the PR now includes the entirety of the quest.

PR #4808, opened 2023-08-27
Playermarker plugin

derfurkan:
PlayerMarker is a plugin that gives you the option to mark players. You also can specify a Sharing Channel in the configuration where your marks will be posted to. If another player uses the same Sharing Channel each others markers will be shared.

It uses a Redis Database to communicate between clients. 
Each packet contains the raw player name and a list of marked players formed into a json using Gson.


raiyni:
What is this used for? It looks like it would be used for pvp scouting.

derfurkan:
> What is this used for? It looks like it would be used for pvp scouting.

It should give groups of players a better visualization of crowded regions and better localization in them. 
Also every marked player is being unmarked after disabling the plugin or changing a channel.

raiyni:
[This would fall under multiple guidelines.](https://secure.runescape.com/m=news/third-party-client-guidelines?oldschool=1)

derfurkan:
> [This would fall under multiple guidelines.](https://secure.runescape.com/m=news/third-party-client-guidelines?oldschool=1)

Which guideline violations are we talking about?

raiyni:
Scouting, marking for attacking, marking opposing clan?

derfurkan:
> Scouting, marking for attacking, marking opposing clan?

Its like your friend telling you that you should look out for player "xyz". 
So you have to manually check every player you encouter if they are "xyz". 

This plugin just uses the information you are being provided with, and also can access, in a better way and wont give you any further informations. I dont see where this would be an unfair advantage against other players without the plugin or how this should violate anything at all? Maybe you could give me an more clear example on how this would be an clearly unfair advantage.

Felanbird:
> Its like your friend telling you that you should look out for player "xyz". So you have to manually check every player you encouter if they are "xyz".

Player Indicators would be perfect for this

raiyni:
In earnest, there is a difference between verbally calling out an opposing player and offering a coordinated system that marks a specific player's location on screen. The former, you have to verbally communicate where they are on screen accurately enough for everyone to comprehend what you said, either with looks or location. The latter, said player is visually marked on screen and you are brainlessly told where to click--it takes all of the processing out of it. 

As such, I still don't see a use for this outside of PVP that is within 3rd party client guidelines.

derfurkan:
> In earnest, there is a difference between verbally calling out an opposing player and offering a coordinated system that marks a specific player's location on screen. The former, you have to verbally communicate where they are on screen accurately enough for everyone to comprehend what you said, either with looks or location. The latter, said player is visually marked on screen and you are brainlessly told where to click--it takes all of the processing out of it.
> 
> As such, I still don't see a use for this outside of PVP that is within 3rd party client guidelines.

In my opinion its not a unfair advantage (looking at other plugins) but since its purpose is not usable enough i can understand that. Its violating the guidelines as you say so i cant do anything about that i will close this pull request then.

PR #4810, opened 2023-08-28
Update VoiceScape

derfurkan:
This pull request is an update of the in-game voicechat plugin Voicescape which was recoded from ground up and is now using redis as communicator between voice clients.

For redis communication a 3rd party dependency called Jedis (v4.4.3) is being used which can be found here:
https://github.com/redis/jedis

The class JedisPool from the Jedis dependency is not being used anywhere in the plugin. Instead the RedisPool class from the RuneLite-API is being used as replacement which also can be found here:
https://github.com/runelite/api.runelite.net/blob/master/http-service/src/main/java/net/runelite/http/service/util/redis/RedisPool.java


Felanbird:
Keep pushing changes to the same branch to fix your build errors, you don't need to open new PRs

derfurkan:
> Keep pushing changes to the same branch to fix your build errors, you don't need to open new PRs

I opened a new PR because i thought the "../gradlew --write-verification-metadata sha256" command broke my branch so i deleted it and created a fresh one to manually verify the dependencies. Which also deleted my PR before so i had to create a new one.

raiyni:
That's what `git reset` command is for.

derfurkan:
> That's what `git reset` command is for.

Noted. Still new to the git and gradle stuff.

Nightfirecat:
It looks like [this change adds a new config which defaults to making external connections.](https://github.com/derfurkan/VoiceScape/blob/17d278f5637356d1ddab7c55546cfe31b992163b/src/main/java/de/furkan/voicescape/VoiceScapeConfig.java#L102) At present, we do not have a system in the client to request acceptance of a new or changed `warning` on an already-installed plugin, so this cannot be accepted. For the time being, the only acceptable default for that config is CUSTOM.
This PR looks good to me otherwise.

derfurkan:
> It looks like [this change adds a new config which defaults to making external connections.](https://github.com/derfurkan/VoiceScape/blob/17d278f5637356d1ddab7c55546cfe31b992163b/src/main/java/de/furkan/voicescape/VoiceScapeConfig.java#L102) At present, we do not have a system in the client to request acceptance of a new or changed `warning` on an already-installed plugin, so this cannot be accepted. For the time being, the only acceptable default for that config is CUSTOM. This PR looks good to me otherwise.

Added the server type NONE and set it as default so players need to choose between DEFAULT and CUSTOM before they connect to any server.

PR #4813, opened 2023-08-28
Add citizens plugin

gc:
Repo: https://github.com/gc/citizens
Description: Adds 'fake NPCs' (Citizens for clarity) around the world, which walk/talk/animate. 

If you are loading the plugin into your client to run it locally, you will find the most citizens in varrock.
They despawn when they are too far away from the player (25~ tiles distance from the player).

### DIFF
--- a/plugins/citizens
+++ b/plugins/citizens
@@ -1,3 +1,3 @@
 repository=https://github.com/gc/citizens.git
-commit=8bbf3857bc3df18e3495ceebffbf2bd58013e30e
+commit=971baa0ce9252fdeff11e2be9d750661831cdadb
 authors=Magnaboy,Diabolickal
--- a/plugins/citizens
+++ b/plugins/citizens
@@ -1,3 +1,3 @@
-repository=https://github.com/gc/citizens
+repository=https://github.com/gc/citizens.git
 commit=8bbf3857bc3df18e3495ceebffbf2bd58013e30e
 authors=Magnaboy,Diabolickal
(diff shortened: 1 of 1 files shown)


plugins/citizens
```diff
@@ -0,0 +1,3 @@
+repository=https://github.com/gc/citizens
```
raiyni, line 1:
```suggestion
repository=https://github.com/gc/citizens.git
```

raiyni (changes requested):

plugins/citizens
gc: @raiyni Oops sorry. Fixed.

gc:
Fixed the error relating to the gson usage by using the injected gson instance.

Nightfirecat:
Have you tried running this plugin outside of your development environment? [It looks like it's trying to load files from your filesystem rather than from the plugin's resources](https://github.com/gc/citizens/blob/187271ee3523039f33bb0b83a6c150fb4107c757/src/main/java/com/magnaboy/scripting/ScriptLoader.java#L10), which would not work for other users. (This same problem seems to exist in Util.java and CitizenRegion.java)
I know the panel is only added for development [but you add it here and don't ever remove it, even on plugin shutdown](https://github.com/gc/citizens/blob/187271ee3523039f33bb0b83a6c150fb4107c757/src/main/java/com/magnaboy/CitizensPlugin.java#L91). Please fix that so there's no chance it ends up being an issue later on when it comes out of development.
Code generally LGTM otherwise.

Nightfirecat:
Your shutDown code is adding the nav button a second time, not removing it. Please test this by toggling your plugin off then back on after starting your client.

gc:
Apologies for that, not sure how I messed that up. I fixed that, and tested it making sure it gets removed when turning off the plugin. (and that it doesnt appear at all unless in development).

PR #4826, opened 2023-08-30
Add Canvas plugin 

cantrick:
added repository and commit info to plugin-hub/plugins


raiyni (changes requested):
You need to remove overlays that you added on shutdown

cantrick:
> 

gotcha, should be all good now.

PR #4830, opened 2023-08-31
Add Tick Fixer for Mac plugin

jonathangarelick:
This plugin simply pings the local gateway every 200ms in order to prevent a Mac energy saving feature from creating extra packet latency.

You can see more information here
https://www.reddit.com/r/2007scape/comments/165nhch/huge_increase_in_tick_quality_on_mac_with_this/
and here
https://apple.stackexchange.com/questions/357075/latency-spikes-on-wifi-power-saving-measure

Thanks!


raiyni:
Don't close and open new prs. Push commits to the same branch.

raiyni:
Do macs come with zsh pre-installed?

jonathangarelick:
> Do macs come with zsh pre-installed?

Yes. `zsh` has been the default shell for macOS since late 2019. [Source](https://support.apple.com/en-us/HT208050)

LlemonDuck:
* remove the config class since it's unused or change its config group to not be `example`
* use log.debug 
  * https://github.com/jonathangarelick/tick-fixer/blob/31d77d28e6213fc850d3cf8eef7b68fc120cdb2e/src/main/java/com/jonathang/TickFixerPlugin.java#L66
  * https://github.com/jonathangarelick/tick-fixer/blob/31d77d28e6213fc850d3cf8eef7b68fc120cdb2e/src/main/java/com/jonathang/TickFixerPlugin.java#L55

optionally also:
* RL has its own os detection via `OSType.getOSType() == OSType.MacOS` rather than using the Apache Commons

really neat idea for a plugin, and a nice creative solution

jonathangarelick:
@LlemonDuck the plugin has been updated based on your feedback, and is ready for review.

Thank you to @orion-git and @phife-dawg

phife-dawg:
Hi @jonathangarelick @cube-god unfortunately this does not seem to make the same improvements as running the command from the reddit post:

ping `route get default | grep gateway | awk '{print $2}'` -i 0.2

I am monitoring tick health using https://runelite.net/plugin-hub/show/ticktracker. 

When I enable this new tick fixer plugin my tick health does not change, it remains around 60%. When I run the aforementioned command, it immediately increases to >99%.

I would love to provide more detail on this, but I am not sure how to capture any data. Please let me know what other info I can provide!

Felanbird:
You should open an issue here with this information https://github.com/jonathangarelick/tick-fixer/issues

phife-dawg:
Done!

PR #4833, opened 2023-09-01
Add Tooltips Extended

iipom:
This plugin just takes the base plugin Mouse Tooltips and adds the ability to show the extra right click options as a tooltip, similar to how the mouse tooltips are in the C++ client.


iipom:
this plugin is stupid idk why i made this lmao

PR #4835, opened 2023-09-01
Add Shade Chest Highlighter

eemkukko:
Highlights chests in the Mort'ton shade catacombs based on which keys a player has in their inventory.


raiyni:
You need to remove your overlay on shutdown

eemkukko:
> You need to remove your overlay on shutdown

Done.

PR #4841, opened 2023-09-03
Scenery Hider plugin

THEufn:
Allows the player to hide any scenery object by ID. Bloat pillar is disabled.

### DIFF
--- a/plugins/scenery-hider
+++ b/plugins/scenery-hider
@@ -1,0 +1,2 @@
+repository=https://github.com/THEufn/RunelitePlugins.git
+commit=0fc123eea9149bad8f798a9a81ed56a72c4569e4


Nightfirecat:
We're still discussing this plugin internally, but I will say that we've encountered nothing but trouble from plugins which allow users to modify/hide arbitrary NPCs, sounds, objects, etc. by ID. I would be much more in favor of adding specific scenery you wish to be hidden to a plugin instead of allowing an open list of IDs.

glsm-github:
> Something that sticks out as a bit janky is that restoring (aka un-hiding) game objects is done through forcing an entire map reload via `client.setGameState(GameState.LOADING)` . Instead, consider holding an in-memory map between location and game object, such that you can restore these without needing to adversely affect the client. You may see how this is done [in this file](https://github.com/dey0/ground-object-hider/blob/master/src/main/java/com/objecthider/ObjectHiderPlugin.java#L54). Yes, it's a bit more involved, but it'll have a far better level of cooperation between this and other plugins.
> 
> Unsure of etiquette regarding new plugins, but that's my 2c.

The canonical way this was done for GroundObjects is to hide and unhide via `tile.setGroundObject(null)` and `tiles[plane][x][y].setGroundObject(this.hiddenObjects.get(location))`, respectively. However, for GameObjects, there exists only `client.getScene().removeGameObject(gameObject)` to hide them, with no complementary unhide function. This also seems to be the case with Tiles, with `Scene.removeTile`.

Is there an alternative then to unhiding GameObjects and Tiles other than via `client.setGameState(GameState.LOADING)` ?

LlemonDuck:
following up on what nfc said above, allowing arbitrary ids invariably leads to a game of cat and mouse. feel free to resubmit this plugin with a specific, fixed list of scenery objects to hide instead.

PR #4844, opened 2023-09-03
Create BrewTracker

codykoscielski:
Brew Tracker is a simple plug in to help players keep track of the number of brew sips they have used. Once the player hits 3, there will be a flashing restore icon to indicate when to sip a restore. As long as you are brewed down by 3 sips, the restore icon will flash.


Nightfirecat:
Instead of loading brew and restore images via ImageIO, use [`ItemManager.getImage()`](https://static.runelite.net/api/runelite-client/net/runelite/client/game/ItemManager.html#getImage(int))
Also [you don't need to use a logger factory to get a logger instance](https://github.com/codykoscielski/BrewRestoreTracker/blob/ca26908d83986e7cd3ae8361fda7a89b70040829/src/main/java/com/brewtracker/BrewTrackerOverlay.java#L26), you can annotate the class with `@Slf4j`, which will inject a `log` instance.

LlemonDuck:
closing since the above comments have not yet been resolved, feel free to resubmit when they are

PR #4845, opened 2023-09-04
Market Watcher Plugin

BobTabrizi:
Simple item tracker that displays three sets of price histories (1 week, 1 month, 3 months) for each item in the game. Items can be saved and can optionally be stored in tabs.


raiyni:
You should be setting some user agent for wiki requests that identifies this plugin as the source.

PR #4848, opened 2023-09-04
Update charge calc

vikke1234:
In order to make it easier to charge items, add a button that will effectively "press enter" for you. 
This is still 1:1 as you will need to click a separate button.


Nightfirecat:
Our feeling on this feature is that it would be too strong as it completely removes the need to use the keyboard for this interface. We're not comfortable accepting it to the hub at this time.

LlemonDuck:
closing due to above

PR #4851, opened 2023-09-05
Create altered-stats

JasonDahn:
Adds altered-stats, a plugin that shows any stats that are altered (not boosted) from their current base level.

### DIFF
--- a/plugins/altered-stats
+++ b/plugins/altered-stats
@@ -1,2 +1,2 @@
 repository=https://github.com/JasonDahn/altered-stats.git
-commit=52ab7a1d5fa17c30aeeb1eed01094046c375ed4b
+commit=50087efc816ce5424e412baeb1318bcc41d82804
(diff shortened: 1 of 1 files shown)


Nightfirecat:
[Don't use "example" as your config group](https://github.com/JasonDahn/altered-stats/blob/52ab7a1d5fa17c30aeeb1eed01094046c375ed4b/src/main/java/com/alteredstats/AlteredStatsConfig.java#L7)
Code LGTM otherwise

JasonDahn:
> [Don't use "example" as your config group](https://github.com/JasonDahn/altered-stats/blob/52ab7a1d5fa17c30aeeb1eed01094046c375ed4b/src/main/java/com/alteredstats/AlteredStatsConfig.java#L7) 

Thank you, I updated the config group to "alteredstats" in https://github.com/JasonDahn/altered-stats/pull/3/files

Nightfirecat:
You need to update this PR to point to your latest code, as-is that has no effect on what would be packaged and deployed to the hub.

PR #4862, opened 2023-09-07
Update the-gauntlet to v1.9

rdutta:


Nightfirecat:
Please remove [the reflection used in ResourceManager](https://github.com/rdutta/runelite-gauntlet/blob/3a3b693d355b6e8c0f8439f7c7dfa6ecdf0d8711/src/main/java/ca/gauntlet/module/maze/ResourceManager.java#L210)
Also [you should remove `enabledByDefault = false`](https://github.com/rdutta/runelite-gauntlet/blob/3a3b693d355b6e8c0f8439f7c7dfa6ecdf0d8711/src/main/java/ca/gauntlet/TheGauntletPlugin.java#L50) as it's confusing for users to install a plugin which doesn't turn on right away.

PR #4876, opened 2023-09-09
Create chat-filter-updater

Psicoses:
A plugin to automatically update runelite's chat filter regex. It pulls from a url after every plugin start and sets the chat filter's config to what's stored in the url.

### DIFF
--- a/plugins/chat-filter-updater
+++ b/plugins/chat-filter-updater
@@ -1,2 +1,2 @@
 repository=https://github.com/Psicoses/ChatFilterUpdater.git
-commit=626a0a20b9f79104b3f9a9da444af7ce1ec8c010
+commit=d762217a5a7c1e65dbbe44737551aed1c0eb7fc4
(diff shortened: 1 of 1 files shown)


raiyni:
Inject our okhttp object and use that to fetch your url.

Psicoses:
> Inject our okhttp object and use that to fetch your url.

Gotchya, done.

raiyni:
Two more things:

* Use enqueue instead of execute so requests aren't on thread
*  If you care about saving the current chat filter a user has, it will be overwritten permanently if they toggle config settings in your plugin

Psicoses:
Okay its queuing the request for the url now.

As for the config, it does store the user's regex as the plugin starts so that changes can be reverted by toggling. Changes are only permanently lost if the user restarts the client.

I added a warning to notify the user that the filter will be permanently overwritten, and any changes while the plugin is on will be lost.

Thank you for your patience and let me know if there is anything else.

PR #4883, opened 2023-09-09
Create Thrall Damage Counter

BraveH:
Adds a counter that tallies the total amount of damage done by your summoned thralls towards enemy units. The tally can be reset by shift clicking the overlay and selecting the RESET menu option.

### DIFF
--- a/plugins/thrall-damage-counter
+++ b/plugins/thrall-damage-counter
@@ -1,0 +1,2 @@
+repository=https://github.com/BraveH/thralldamagecounter.git
+commit=d278f99b20f5da7184b252821958d2ad0695c608
--- a/plugins/thrall-damage-counter
+++ b/plugins/thrall-damage-counter
@@ -1,2 +1,2 @@
 repository=https://github.com/BraveH/thralldamagecounter.git
-commit=d278f99b20f5da7184b252821958d2ad0695c608
+commit=613f15f1e43c77edeb1e5cd4079708ccb05a4062


raiyni:
What's with all of this animation code that's not for thralls?

BraveH:
> What's with all of this animation code that's not for thralls?

So the way the the code identifies whether the hitsplat is owned by the thrall or by the player is to first identify if the player is attacking this tick. For that, I check the player's animation data to confirm the player has an attack animation running. If that's the case then there will be 2 hitsplats expected, and as such, one of the hitsplats (the one that doesn't have any xp drops) will be the thrall's hitsplat

BraveH:
Any updates on this? Seems to be a bit stale...

Nightfirecat:
Please don't "bump" PRs, they'll be reviewed and accepted in due time. As you can see, there's a significant queue for new plugins and plugin updates to the hub, so be patient.

BraveH:
@Nightfirecat it's just I saw PR's merged today for stuff that was opened yesterday so I just wanted to make sure this wasn't forgotten behind that's all.

Felanbird:
Changes to 117 often get more priority since it's a plugin used by so many users, and the plugin itself is insane so when something needs to be fixed, it probably _needs_ to be fixed.

Also as mentioned above, adding a new plugin requires much more review than updating an existing one (while those are still properly reviewed)

BraveH:
> Changes to 117 often get more priority since it's a plugin used by so many users, and the plugin itself is insane so when something needs to be fixed, it probably _needs_ to be fixed.
> 
> Also as mentioned above, adding a new plugin requires much more review than updating an existing one (while those are still properly reviewed)

Yes of course no don’t get me wrong, I use 117 and sure I would love it to be prioritised as well. The only reason I thought this went unnoticed or forgotten was because my other PR for a plug-in update (https://github.com/runelite/plugin-hub/pull/4911) has also been stale for almost the same duration so I assumed what I assumed. But since they’re both still gonna be reviewed my bad for bumping these pr’s.

PR #4903, opened 2023-09-15
Add essence-pouch-health plugin

Blackberry0Pie:


YvesW:
Please use `enabledByDefault = true`. It confuses users when a plugin is downloaded but not activated.

raiyni:
> Please use `enabledByDefault = true`. It confuses users when a plugin is downloaded but not activated.

Removing the line is the same thing and less to read

Blackberry0Pie:
Should be updated

PR #4907, opened 2023-09-16
Update action-progress

guillaume009:
Fix progress bar not showing when smithing due to the fact that the spacebar is treated differently than a click on the item to smith


raiyni:
listener needs removed on shutdown

guillaume009:
@raiyni I added a commit that I think should do the trick, but not 100% sure. Let me know if it should be done differently.

PR #4910, opened 2023-09-17
New plugin - Sololegends - temple-trek-bog-helper

sololegends:
# Runelite Temple Trek Bog Helper
Highlights firm bog ground as you find them

During the bog event in Temple Trekking, as you discover tiles that are "firm" they will be highlighted automatically for you. These highlights will only persist for that single event and be reset when the bog puzzle resets.  

There is also a full on solver which automatically highlights all firm tiles without having to poke around in the bog.


MosheBenZacharia:
is this different than https://github.com/runelite/plugin-hub/pull/4788 ?

sololegends:
> is this different than #4788 ?

A bit, yeah. That linked plugin just renders out the "solid" tiles by ID. 
While this plugin also can do that, the main goal is to allow people to still play the mini game by keeping track of the solid tiles they have found using the testing sticks.

PR #4911, opened 2023-09-17
Update Dynamic inventory tags to make compatible

BraveH:
As the title suggests, fixes incompatibility with new runelite updates


BraveH:
Any updates on this? Seems to be a bit stale...

PR #4930, opened 2023-09-20
Create Bank Tag WikiTagger file for submitting a plugin

Manieri:
Hello,

I'd like to start this off stating this project is based off an abandoned project, Wiki Bank Tag Integration, by a fellow named Mitch. It was left roughly 2 years ago now, but the functionality for the **Categories** continues to work as intended, but the **Monster Drops** functionality has been broke for 2+ years now.

What I am submitting is a rendition of what was -- I have only added Drops functionality to not takeaway from the original project and the Drops functionality has the added feature of not requiring _'s between spaced out monster names like it's predecessor. What it does exactly is create bank tags based on specific monster drops through data from the Old School RuneScape Wiki by calling ::drops [monster name] in game.

### DIFF
--- a/Bank Tag WikiTagger
+++ b/Bank Tag WikiTagger
@@ -1,0 +1,2 @@
+repository=https://github.com/Manieri/Bank-Tag-WikiTagger.git
+commit=afb219fa3a4725f768ececb1161f8ab945991603
--- a/Bank Tag WikiTagger
+++ b/Bank Tag WikiTagger
@@ -1,2 +1,0 @@
-repository=https://github.com/Manieri/Bank-Tag-WikiTagger.git
-commit=afb219fa3a4725f768ececb1161f8ab945991603
--- a/plugins/bank-tag-wikitagger
+++ b/plugins/bank-tag-wikitagger
@@ -1,0 +1,2 @@
+repository=https://github.com/Manieri/Bank-Tag-WikiTagger.git
+commit=afb219fa3a4725f768ececb1161f8ab945991603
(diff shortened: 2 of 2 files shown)


Nightfirecat:
The file you're creating in this PR should be moved to `plugins/bank-tag-wikitagger` for it to be correct.

Manieri:
> The file you're creating in this PR should be moved to `plugins/bank-tag-wikitagger` for it to be correct.

I followed what it stated to do under [Submitting a plugin](https://github.com/runelite/plugin-hub#submitting-a-plugin). Would this not be correct? 
> Then go back to the [plugin-hub](https://github.com/runelite/plugin-hub) and click New pull request in the upper left

iProdigy:
Step 3 at that link states `Create a new file in the plugin-hub/plugins directory`

Manieri:
> Step 3 at that link states `Create a new file in the plugin-hub/plugins directory`

Ahh. My mistake. Allow me a moment.

raiyni:
If this is filling the same role as `bank-tag-generation` we would rather you become a collaborator of that repo. If mitch is unresponsive for a while, then we can transfer ownership.

Manieri:
> If this is filling the same role as `bank-tag-generation` we would rather you become a collaborator of that repo. If mitch is unresponsive for a while, then we can transfer ownership.

How would I go about that exactly? I'm fine doing so, but it has been two years and there has not been any work put towards fixing bugs.

geheur:
Try to contact the author and ask if they will accept your PR or if they can make you a collaborator so you can accept it yourself. If they don't respond, then make a PR to the plugin-hub repo that updates the plugin to your own fork of it and say that the author was not responsive, and that you want to take over the plugin.

PR #4939, opened 2023-09-21
readme: Update submission instructions to indicate new commits will move your PR to the back of the queue.

MosheBenZacharia:


Nightfirecat:
This really isn't official, and also isn't true for all cases, so I don't support adding this to the README

gc:
> This really isn't official, and also isn't true for all cases, so I don't support adding this to the README

It could be worth saying "Generally, pushing new commits will..." or "Pushing new commits *may*...", just as a helpful disclaimer without claiming it to be an absolute fact.

raiyni:
It's really up to the reviewer and we try to give priority to things that have been in queue longer and/or have addressed feedback recently. We don't want to discourage people from updating their plugins if they are sitting on code that should be going in, or if they forgot something.

PR #4941, opened 2023-09-22
add idle-flasher

poi56iop:
Hello!

This is a rather simple one.
No current plugin that I can find will flash the screen continuously as long as you remain idle.
A common workaround is to have idle notifications and make every notification flash the screen, but that's far from ideal.

This plugin allows you to tint/flash the screen in configurable colors (with configurable flashing durations) when idle.
Users can also configure delays before the flashing starts, and have keypresses/clicks temporarily disable the idle flashing/tint without resetting the main idleness timer.

There is also an option to tint the screen a different color when not idle.

Thank you for your time!

### DIFF
--- a/plugins/idle-flasher
+++ b/plugins/idle-flasher
@@ -1,0 +1,2 @@
+repository=https://github.com/poi56iop/idle-flasher.git
+commit=bf227fd6df4761450a37d380214809a115499fac


YvesW:
Skilling notifications does this, I think.

poi56iop:
After much discussion, Yves found another plugin ([Watchdog Alerts](https://runelite.net/plugin-hub/show/watchdog)) that is a great candidate to get this functionality. While it'd be less clearly useful to most people, (The same way Idle Alarm is more clearly useful despite Watchdog supporting its features and more) it'd enable a vast superset of this functionality, which is good enough for me.

There's a risk of needing to reopen if Watchdog Alerts' author does not want to add these features, but I imagine it's safe to close this.

PR #4951, opened 2023-09-24
Create toa-insanity-helper

DocGitGud:
This plugin displays an infobox which shows the amount of time you have remaining to attack the skulls, and recommends which direction to stand afterwards.


LlemonDuck:
This violates the Third Party Client Guidelines in two ways:

PR #4952, opened 2023-09-24
Add fuzzy-bank-search plugin

i:
Uses fuzzy search when searching your bank to allow for spelling errors and shorthand queries.

For example searching for `msb` or `bofa` does what you'd expect it to do.


YvesW:
Please don't close and open new prs. Push to the same branch.

PR #4971, opened 2023-09-27
Update Creator's Kit to v1.1.2

ScreteMonge:
- Added ability to search cache for an object or npc by id to directly add a new Custom Model or send the object/npc to Anvil
- Spelling correction of Creators' to Creator's
- Added config option to toggle the added right-click menu options
- Added "immersion" tag
- Bugfix: Removed text tags from PlayerGetter
- Bugfix: Added missing chat notification for Storing players
- Bugfix: Spawn button and hotkey now properly set the object to the player's location in instances if no location is set
- Bugfix: plane changes will now properly trigger location resetting
- Bugfix: Sending player to anvil now correctly applies kit recolours
- Removed unused operations


raiyni:
Inject our okhttp client and use enqueue with callbacks instead of starting new threads

ScreteMonge:
Thank you.

Removed the threading entirely as its impact on performance turned out to be entirely negligible anyway

raiyni:
My feedback is still valid because we don't want request done on client/swing thread

ScreteMonge:
Fair enough.

Given that this patch is fairly small, fixes a few important bugs, and the plugin has already been performing these requests on client/swing thread, would you be willing to approve this change in the interim?

Okhttp is a fairly steep learning curve for me (I'm a player more than programmer) and, from a cursory glance, looks like it'll require some substantial changes to how the plugin works. I don't mind learning but it will take some time.

I'll respect your request either way

ScreteMonge:
A fix for this has been applied.

Let me know if there's anything else.

PR #4980, opened 2023-09-29
Create pk-loot-splitter plugin v1.0.0

pwatts6060:
image
image

### DIFF
--- a/plugins/pk-loot-splitter
+++ b/plugins/pk-loot-splitter
@@ -1,0 +1,2 @@
+repository=https://github.com/pwatts6060/runelite-plugins.git
+commit=813108d0e0babbc36493d94e1cea9db8430b7bb2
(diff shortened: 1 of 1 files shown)


Nightfirecat:
[Your nav button icon is too large](https://github.com/pwatts6060/runelite-plugins/blob/813108d0e0babbc36493d94e1cea9db8430b7bb2/src/main/resources/lootsplit_icon.png). While nav buttons scale it down, it Java keeps it loaded in memory at the full size. Please scale it down so that's not an issue.
Speaking of nav buttons, [instead of removing it on shutdown, you're adding it a second time](https://github.com/pwatts6060/runelite-plugins/blob/813108d0e0babbc36493d94e1cea9db8430b7bb2/src/main/java/com/lootsplit/LootSplitPlugin.java#L78).

pwatts6060:
> [Your nav button icon is too large](https://github.com/pwatts6060/runelite-plugins/blob/813108d0e0babbc36493d94e1cea9db8430b7bb2/src/main/resources/lootsplit_icon.png). While nav buttons scale it down, it Java keeps it loaded in memory at the full size. Please scale it down so that's not an issue. Speaking of nav buttons, [instead of removing it on shutdown, you're adding it a second time](https://github.com/pwatts6060/runelite-plugins/blob/813108d0e0babbc36493d94e1cea9db8430b7bb2/src/main/java/com/lootsplit/LootSplitPlugin.java#L78).

Fixed both of these issues I believe

PR #4981, opened 2023-09-29
adding lowest xp plugin

HoustonBass:
new plugin!
shows the difference in xp between your lowest skill and all the others
Helpful for planning around your Tears of Guthix weekly

### DIFF
--- a/plugins/lowest-xp-order
+++ b/plugins/lowest-xp-order
@@ -1,0 +1,2 @@
+repository=https://github.com/HoustonBass/lowestSkill4Tears
+commit=20384a1d69b5fc2105e6231509d70d90be7e6707
--- a/plugins/lowest-xp-order
+++ b/plugins/lowest-xp-order
@@ -1,2 +1,2 @@
 repository=https://github.com/HoustonBass/lowestSkill4Tears.git
-commit=738e1074e143dcc50fbe4a9ba3a25a1b1fde86b1
+commit=7a2851a7797f30c9afbf8aeee129a81e09a24b2a
--- a/plugins/lowest-xp-order
+++ b/plugins/lowest-xp-order
@@ -1,2 +1,2 @@
 repository=https://github.com/HoustonBass/lowestSkill4Tears.git
-commit=20384a1d69b5fc2105e6231509d70d90be7e6707
+commit=738e1074e143dcc50fbe4a9ba3a25a1b1fde86b1
--- a/plugins/lowest-xp-order
+++ b/plugins/lowest-xp-order
@@ -1,2 +1,2 @@
-repository=https://github.com/HoustonBass/lowestSkill4Tears
+repository=https://github.com/HoustonBass/lowestSkill4Tears.git
 commit=20384a1d69b5fc2105e6231509d70d90be7e6707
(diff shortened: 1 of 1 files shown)


CreativeTechGuy:
It looks like your repo is missing a lot of the metadata. For example:

* You didn't fill out the `group` or `version` in your `build.gradle` file
* Your README has no information about your plugin and is currently incorrect
* Your `settings.gradle` is unedited
* Your plugin description is different between the `runelite-plugin.properties` and the `LowestExperiencePlugin.java` file
* Your plugin name is different in every place it appears, you should make them all consistent and ensure that the name which appears in your `runelite-plugin.properties` file is the friendly formatted name and not kebab-case.

This is all super important for when users are trying to find your plugin to understand what it does and also to ensure that everything works and is consistent across every place it is used.

Nightfirecat:
[Do not use `"example"` as your config group.](https://github.com/HoustonBass/lowestSkill4Tears/blob/7a2851a7797f30c9afbf8aeee129a81e09a24b2a/src/main/java/com/lowestLevel/LowestExperienceConfig.java#L8)
[Don't use `loadWhenOutdated = true`](https://github.com/HoustonBass/lowestSkill4Tears/blob/7a2851a7797f30c9afbf8aeee129a81e09a24b2a/src/main/java/com/lowestLevel/LowestExperiencePlugin.java#L33) as that is reserved for plugins which can run without access to `Client`.

LlemonDuck:
Is this plugin's use case covered by [tog-indicator](https://runelite.net/plugin-hub/show/tog-indicator)?

HoustonBass:
> It looks like your repo is missing a lot of the metadata. For example:
> 
> * You didn't fill out the `group` or `version` in your `build.gradle` file
> * Your README has no information about your plugin and is currently incorrect
> * Your `settings.gradle` is unedited
> * Your plugin description is different between the `runelite-plugin.properties` and the `LowestExperiencePlugin.java` file
> * Your plugin name is different in every place it appears, you should make them all consistent and ensure that the name which appears in your `runelite-plugin.properties` file is the friendly formatted name and not kebab-case.
> 
> This is all super important for when users are trying to find your plugin to understand what it does and also to ensure that everything works and is consistent across every place it is used.

fixed

HoustonBass:
> [Do not use `"example"` as your config group.](https://github.com/HoustonBass/lowestSkill4Tears/blob/7a2851a7797f30c9afbf8aeee129a81e09a24b2a/src/main/java/com/lowestLevel/LowestExperienceConfig.java#L8) [Don't use `loadWhenOutdated = true`](https://github.com/HoustonBass/lowestSkill4Tears/blob/7a2851a7797f30c9afbf8aeee129a81e09a24b2a/src/main/java/com/lowestLevel/LowestExperiencePlugin.java#L33) as that is reserved for plugins which can run without access to `Client`.

fixed

HoustonBass:
> Is this plugin's use case covered by [tog-indicator](https://runelite.net/plugin-hub/show/tog-indicator)?

yes and no.
For me, i want to 
know exactly how much xp i need in the skills.
The tog-indicator plugin is great but it only shows my my lowest skill.
I want to see how much xp i need to make it my not lowest skill, or better yet something like this 

how-to

CreativeTechGuy:
@HoustonBass it looks like the metadata is there! Nice. (I'm not a reviewer so I cannot do much more than say that, sorry!)

PR #4986, opened 2023-09-30
Save Evil Creatures

Jinxer42:
A plugin that replaces many gnome models with evil creature models.  Submitting for review to be added to the plugin hub.


YvesW:
Put the file in the plugins folder.

PR #4993, opened 2023-10-01
Costume room highlighter update (change 1)

michaelperino:
Added mining gloves which can now be stored in player houses


Nightfirecat:
This change looks fine. [Can you update your plugin's config group to not be `"example"`?](https://github.com/michaelperino/CostumeRoomChecker/blob/2df11ae618caf5301c84b5a239b64364648f9824/src/main/java/com/CosRoom/CosRoomConfig.java#L8) Be aware that changing the config group will reset any users' configs for this plugin; you can choose to write some code to migrate the old configs to the new group if you'd like.

PR #4997, opened 2023-10-03
Update "GP Per Hour" to 1.1

MosheBenZacharia:
- Fix issue when depositing items via menu option
- Remap granite cannonball price to normal cannonball
- Fix ash sanctifier regex 
- Add option to show session stats on inventory overlay
- Add support for tracking amulet of blood fury charge cost
- Prevent crystal armor from degrading when taking zero damage
- Add option to show inventory overlay when inventory is hidden
- Add option for displaying static gold drops

### DIFF
--- a/plugins/gp-per-hour
+++ b/plugins/gp-per-hour
@@ -1,3 +1,3 @@
 repository=https://github.com/MosheBenZacharia/GP-Per-Hour.git
-commit=dcd197c3a81a6275f2c67bdac7891f6ed47377f3
+commit=57fa4dcdb1a332d5cc1e67c032e8454a2a3a2a7e
 authors=MosheBenZacharia
--- a/plugins/gp-per-hour
+++ b/plugins/gp-per-hour
@@ -1,3 +1,3 @@
 repository=https://github.com/MosheBenZacharia/GP-Per-Hour.git
-commit=ed854db2994b751bcf8cde33c93a1a2525175212
+commit=dcd197c3a81a6275f2c67bdac7891f6ed47377f3
 authors=MosheBenZacharia


raiyni:
Any reason you aren't using `clientThread::invokeLater` here? https://github.com/MosheBenZacharia/GP-Per-Hour/blob/dcd197c3a81a6275f2c67bdac7891f6ed47377f3/src/main/java/com/gpperhour/itemcharges/ChargedItem.java#L783

TicTac7x:
Why are you creqting brand new plugin instead of eoing PR into my plugin? Full copium lmao

CreativeTechGuy:
I wanted to call out that [TicTac7x's comment](https://github.com/runelite/plugin-hub/pull/4997#issuecomment-1747488206) was edited since it was posted.

> The plugin is quite different, but uses lots of same code under the hood, would be nice if this plugin would read the number of charges from the original plugin.

That's a very interesting idea to treat the [Item Charges Improved](https://runelite.net/plugin-hub/show/tictac7x-charges) plugin like a library to avoid the duplication. That would then require that users must have both installed for the GP Per Hour plugin to work.

It seems like a good idea for the long-run, but would require:

* A public Service exposed by the Item Charges Improved plugin which other plugins could access at runtime if it exists.
* A way for RuneLite to let users know that one plugin has a hard dependency on another. - This could also be done within the consuming plugin if RuneLite doesn't want to add this feature at the platform level.

@TicTac7x - Would you be willing to make a service to expose all of the data from your plugin? I think the best (and one of the only) examples would be the [SlayerPluginService](https://github.com/runelite/runelite/blob/master/runelite-client/src/main/java/net/runelite/client/plugins/slayer/SlayerPluginServiceImpl.java).

raiyni:
It's not possible to do cross phub communication like that. It would need to come from a genetic event bus message or something.

CreativeTechGuy:
@raiyni do you know if there's any plans to support cross plugin-hub communications? That seems like it'd be necessary for the plugin ecosystem to grow beyond a certain point. Imagine what would be possible if plugins could have hard dependencies on other plugins. And if plugins could be library-only.

That'd really help avoid the issue of many plugins which have slightly different versions of the same code that keeps getting copied around and slightly better or worse each time. If a new update breaks it, now a ton of plugins need to be updated separately rather than one which they all depend on. That'd also really make the job of plugin reviewing a lot easier as the amount of code in each would go down as everyone would be encouraged to share.

raiyni:
Probably not. It would create unnecessary edge cases and dependencies that we don't want to support. https://github.com/runelite/runelite/pull/16863 Is the closest we would want to get.

The reason there are multiple version of plugins is a complex problem that's rooted around making `better` versions and we have since upped our critique of adding such plugins and opt for adding users as collaborators.

PR #5006, opened 2023-10-05
Update random hopper to v1.2

rugg0064:
Add "beta" and "no save" worlds to filter correctly.
Change getAdjacentWorlds() to show the only world available if there is only one instead of ???.


rugg0064:
Additionally changed the wording of "target" to "bounty".

LlemonDuck:
Remove the println calls

PR #5010, opened 2023-10-06
Update and rename Low Detail Raids v1.0 to Automatic Low Detail v1.1

bepzi:
- Bump version number to 1.1
- Rename plugin from "Low Detail Raids" to "Automatic Low Detail"
- Add support for the Inferno and TzHaar-Ket-Rak's Challenges
- Add support for Hallowed Sepulchre
- Prevent conflicts between Runelite's built-in Low Detail plugin and this plugin


YvesW:
You might want to check out https://discord.com/channels/301497432909414422/419891709883973642/1159642031048642600

I haven't checked if you are migrating the config, but if you are changing the config group, that wipes the config of current users. So you might want to look at that as well.

PR #5012, opened 2023-10-06
Create skull-prevention-icon

i:
Draws an overlay to display whether skull prevention is on or off.


Bram91:
Your repo still contains the code for fuzzybanksearch so you will have to remove that.

LlemonDuck:
* you should check the config group here first and ensure it's your plugin's config group https://github.com/i/rl-plugins/blob/f6ed38f27b74b76daaefd234e3219deebf115c0b/src/main/java/com/github/i/skullprevention/SkullPreventionIconPlugin.java#L68
* you should check the varbit value in `startUp` as well
* the `PluginDescriptor#name` is the one that appears in the RuneLite side panel, so you may want to use a more display-friendly string like `Skull Prevention Reminder`

LlemonDuck:

PR #5018, opened 2023-10-07
Update log-master

ImTedious:
Big refactor to tidy things up a lot, also add an overlay to display current task and persist save data across PCs if on the same RL account


Nightfirecat:
This plugin's overlay [is added during startup but never removed](https://github.com/ImTedious/generate-task/blob/a6de85147e39e268da95551b961efdc8d5c7035d/src/main/java/com/logmaster/LogMasterPlugin.java#L99), please rectify that.
Changes LGTM otherwise

ImTedious:
Added a line in shutdown to remove the overlay.
Should be good to merge now?

Kind regards

Op vr 13 okt 2023 om 04:14 schreef Jordan ***@***.***>:

PR #5025, opened 2023-10-09
Create Spec Overlay

OmbudRov:
Adds an Overlay/Info Box to show your special attack energy when wielding the Soul Reaper Axe


Nightfirecat:
FYI, this is something we're interested in adding to the main client (mentioned in https://github.com/runelite/runelite/issues/16819), if you're interested in working with us to get something incorporated there instead of as a hub plugin.

OmbudRov:
i've already put a PR for that, this is just overlaying your normal special energy when wielding the Axe

PR #5026, opened 2023-10-09
tob-qol: release v1.2.0

damencs:
- fixes: entry mode not resetting room upon death
- fixes: instance timer building on launch (not using config font size and offset)
- adds: salve and spellbook reminder
- adds: verzik green ball wee-woo alert
- updates: discord name reference

### DIFF
--- a/plugins/tobqol
+++ b/plugins/tobqol
@@ -1,2 +1,2 @@
 repository=https://github.com/damencs/tob-qol.git
-commit=60dadc51adbf6c1edaeae874bc41beb251d528e5
+commit=5342777881403c54c6f8e425cf5958066e3b683b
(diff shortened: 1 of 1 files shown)


Nightfirecat:
As best we're aware, the green ball does not have a sound effect, and despite the config's description, this code doesn't do any replacing/muting of any original sound effect. As such, please do not play a sound effect for this on projectile move. You could, however, play the sound when the chat message indicating you're targeted by it is added. (this would end up playing it only for the player targeted)

damencs:
@Nightfirecat - went ahead and converted it to be based on the chat message for both original target and bouncing. It should now comply with what you stated. - Thank you for pointing out that.

PR #5028, opened 2023-10-09
Create vardorvis-health-tracker

AnuVen:
Tracks the health Vardorvis gains during a fight and displays it during, as well as with a chat message after the fight ends.


damencs:
Is this not covered in Monster HP Percentage? - https://runelite.net/plugin-hub/show/monster-hp-percentage

AnuVen:
This plugin tracks how much Vardorvis heals, which happens based on how much damage the player takes. That plugin just displays its health % at all times.

I think I should have reworded it better

Nightfirecat:
Your build is failing because you specify in your plugin properties file that your plugin is located at `com.example.VardorvisHealthPlugin`, but it is actually at `com.vardorvishealth.VardorvisHealTrackerPlugin`

AnuVen:
Fixed the issue, sorry!

PR #5030, opened 2023-10-10
Add steel-bar-drop-sound

Alex-Banna:
Adds the metal bar falling sound whenever you drop a steel bar


Nightfirecat:
What's the source of this sound effect? I'm told it's a meme, but that doesn't mean we can distribute it as it's likely still copyrighted.

Alex-Banna:
From here: https://www.youtube.com/watch?v=3MPXjMLBlQg

Tbh if it's gonna cause any problems I don't see a need to add this - closing.

PR #5031, opened 2023-10-10
Update star-calling-assist

zodaz:
- Changed plugin display name and description.
- Added default endpoint which is maintained by the starminers discord.
- Fixed issue where the miner count of a previous star would be sent when not close enough to the current star to render players.

### DIFF
--- a/plugins/star-calling-assist
+++ b/plugins/star-calling-assist
@@ -1,3 +1,3 @@
 repository=https://github.com/zodaz/star-calling-assist.git
-commit=e4ce3883f3f1fff3c55e8b64539f9c946f01cfd2
-warning=This plugin submits your current world number along with your IP address to a 3rd party website not controlled or verified by the RuneLite Developers. If enabled in the config, your in-game name will also be sent (default off).
+commit=03d0f433c431311d3a4f8c1e97bb1cfa7895c4da
+warning=This plugin submits your current world number along with your IP address to a 3rd party website not controlled or verified by the RuneLite Developers.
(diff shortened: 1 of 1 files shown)


LlemonDuck:
We can't retroactively apply warnings to users, so you should still have a toggle in the config that is default-off which controls whether data is sent.

zodaz:
> We can't retroactively apply warnings to users, so you should still have a toggle in the config that is default-off which controls whether data is sent.

The warning has been in place since day 1 of the plugin, even when no default endpoint was set. So no one has previously installed the plugin without getting the warning.

I adjusted the warning message as the sending of in-game name is fully opt-in and as I’ve come to understand therefore doesn’t need to be included in the warning message.

I apologize if this caused any confusion.

raiyni:
Plugin shouldn't be sending requests with `execute` and instead use `enqueue` with callbacks.

GIMSjoelbak:
enqueue() has been implemented instead of execute()

plugins/star-calling-assist
```diff
@@ -1,3 +1,3 @@
 repository=https://github.com/zodaz/star-calling-assist.git
-commit=e4ce3883f3f1fff3c55e8b64539f9c946f01cfd2
-warning=This plugin submits your current world number along with your IP address to a 3rd party website not controlled or verified by the RuneLite Developers. If enabled in the config, your in-game name will also be sent (default off).
+commit=41cb3d6b53bad244e21dc8ad4e59634cae863d19
+warning=This plugin submits your current world number along with your IP address to a 3rd party website not controlled or verified by the RuneLite Developers.
```
Nightfirecat, line 3:
```suggestion
warning=This plugin submits your current world number along with your IP address to a 3rd party website not controlled or verified by the RuneLite Developers. If enabled in the config, your in-game name will also be sent (default off).
```

Nightfirecat: Actually I'll remove this since this change is fine.

PR #5044, opened 2023-10-11
Update toggle-chat

MoreBuchus:
**Update to v1.0.3**

- Small fix - Hotkey now closes chat regardless of which tab is currently open


YvesW:
Please remove ``enabledByDefault = false``.

PR #5050, opened 2023-10-12
w45-ge-tracker

XyloYoyo:
W45 Ge Tracker records user trades on w45 and adds a sidebar menu to search items and view the last traded prices of items.

### DIFF
--- a/plugins/w45-ge-tracker
+++ b/plugins/w45-ge-tracker
@@ -1,3 +1,3 @@
 repository=https://github.com/XyloYoyo/w45-ge-tracking.git
-commit=7b3d2df20ca52cfe9cabe3c2a64d3475e505bb37
+commit=92605b78e5e218b65b9ec1e65c8b9aa70b206f66
 warning=This plugin sends requests to a server Runelite doesn't own, trust, or have some affiliation with. This plugin sends trade data (id, price, buy/sell) to 3rd party servers for price tracking. Prices reported by this plugin may be inaccurate or false and could be manipulated by plugin maintainers or other users.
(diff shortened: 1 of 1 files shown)


raiyni:
This plugin looks like it could very easily be used for price manipulation, even on a deadman world that seems bad. We've denied/disabled plugins in the past for the same reasons.

It would be better for the core client to use the deadman api the wiki provides probably.

Nightfirecat:
This plugin should have a `warning` applied as its default configuration sends requests to a server we don't own, trust, or have some affiliation with. Ideally it should also help clarify for users that the prices sent by the server are not necessarily indicative of the real price of items. (whether that be because prices change or because, hypothetically, you could serve bogus/fake prices)

Nightfirecat:
[Your nav button is added during plugin startUp](https://github.com/XyloYoyo/w45-ge-tracking/blob/92605b78e5e218b65b9ec1e65c8b9aa70b206f66/src/main/java/com/w45GeTracking/W45GeTracking.java#L760) but is never removed, please ensure it's removed during shutDown.

plugins/w45-ge-tracker
```diff
@@ -0,0 +1,3 @@
+repository=https://github.com/XyloYoyo/w45-ge-tracking.git
+commit=92605b78e5e218b65b9ec1e65c8b9aa70b206f66
+warning=This plugin sends requests to a server Runelite doesn't own, trust, or have some affiliation with. This plugin sends trade data (id, price, buy/sell) to 3rd party servers for price tracking. Prices reported by this plugin may be inaccurate or false and could be manipulated by plugin maintainers or other users.
```
Nightfirecat, line 3:
```suggestion
warning=This plugin submits your Deadman mode Grand Exchange trade data and IP address to a server not controlled or verified by the RuneLite developers. Prices reported by this plugin are sourced from this server, thus could be inaccurate, false, or manipulated by the server owner or other users.
```

LlemonDuck:

PR #5051, opened 2023-10-12
Committing my new external plugin. Loot Broadcast Party uses the party system to broadcasts collection logs, valuable items, level ups, and CAs to each party member.

07Zuko:
For context: if you are in a friends chat with, say, IRL friends, and then in a clan chat with in game friends, it's a bummer that you cannot share drop notifications with the IRL friends. This plugin solves that. If you get in a party with your IRL friends, this plugin will let you broadcast loot to everyone else in your party.


Nightfirecat:
[Your config group, `"party"`](https://github.com/07Zuko/loot-broadcast-party/blob/aca38c58e1755e076fdba427c8abfe6cf8917d1a/src/main/java/com/lootbroadcastparty/LootBroadcastPartyConfig.java#L35), is already used by the core Party plugin. Please use a unique config group, I recommend "loot-broadcast-party".
[Do not use `enabledByDefault = true`](https://github.com/07Zuko/loot-broadcast-party/blob/aca38c58e1755e076fdba427c8abfe6cf8917d1a/src/main/java/com/lootbroadcastparty/LootBroadcastPartyPlugin.java#L69); it's confusing for users to install a plugin which isn't enabled right away.

Nightfirecat:
It doesn't look like you updated this branch

PR #5057, opened 2023-10-13
Update inferno-stats to v2.1.4

InfernoStats:
Fixes a bug where Bow of faerdhinen attacks would count as a 5 tick action (tbow default) instead of 4.


Nightfirecat:
This code does not compile, please look at fixing that

InfernoStats:
> This code does not compile, please look at fixing that

I have on main, but went out of town immediately after submitting. I'll update the commit later

PR #5060, opened 2023-10-15
Camdozaal fishing helper plugin

ConorJS:
Adds conditional highlighting to world objects used by the player during Camdozaal fishing. See the associated README for more information.


YvesW:
Please remove `enabledByDefault = false` since this confuses users.
Also reading your readme, is there any reason why you didn't add a RegionID check? Seems like it'd prevent unnecessary #support traffic and it's simple to implement

ConorJS:
> Please remove `enabledByDefault = false` since this confuses users. 

Will do.

> Also reading your readme, is there any reason why you didn't add a RegionID check? Seems like it'd prevent unnecessary #support traffic and it's simple to implement

I just didn't get around to this (there's a TODO in the plugin file for it), but I see your point - I'll implement a check.

ConorJS:
@YvesW Done

PR #5061, opened 2023-10-15
Add Impolite Slayer plugin

jdkleuver:
Changes the dialog option and following player dialog from "Okay, great!" to "Screw you!" (configurable) when you get a bad task from a slayer master.

The list of bad tasks can be configured by the user.

### DIFF
--- a/plugins/impolite-slayer
+++ b/plugins/impolite-slayer
@@ -1,0 +1,2 @@
+repository=https://github.com/jdkleuver/impolite-slayer.git
+commit=3dc1d738cce50433e1cbefdcf4f6adc124026bdf


YvesW:
Please rename your ConfigGroup from `example` to something unique. You probably also want to check for your group when the ConfigChanged event fires https://static.runelite.net/api/runelite-client/net/runelite/client/events/ConfigChanged.html#getGroup()

Nightfirecat:
For what it's worth, we don't want to add anything which is overly angry or offensive to the plugin hub. We don't mind if you make the angry message configurable, but "Fuck you!" is probably a bit too rude as a default.

jdkleuver:
Sure thing, I'll make the message configurable and make the config changes

PR #5073, opened 2023-10-16
Update loot-broadcast-party -- Fixing issue on startup.

07Zuko:
Fixed an issue where the plugin was trying to pull in a panel.


07Zuko:
Okay I believe I fixed the issue by re-creating a pull request after updating my branch. Is that the correct way to do it? @Nightfirecat

Felanbird:
This was not correct, but since you have been doing your changes on master, it's not unexpected that you ended up with this solution.

In the future you should checkout a separate branch, that way updating your PR would not have merge conflicts, which was likely the reason why you force-pushed instead of cleanly updating, because git did not see a clean resolution.

PR #5078, opened 2023-10-17
Add ZeroDeath's plugin

kingston45:
This plugin uses an API to track boss data and kills and score them based on KC increases. It uses the official OSRS HiScores data. 
This API put's the user's name and score onto a leaderboard.

If the player dies they are marked as dead on the leaderboard, this also happens when they shutdown the plugin to avoid cheating.


raiyni:
You need to inject our okhttpclient and use enqueue

kingston45:
@raiyni thanks, I've changed it to use that let me know if that's ok or if it need's anything further.

plugins/zero-deaths
```diff
@@ -0,0 +1,2 @@
+repository=https://github.com/kingston45/ZeroDeathsPlugin.git
+commit=d1b8d0b7ad3d3480433e17b6d3d7c3a2a3c4a9cf
```
Nightfirecat, line 2:
```suggestion
commit=d1b8d0b7ad3d3480433e17b6d3d7c3a2a3c4a9cf
warning=This plugin submits your username and IP address to a 3rd party server not controlled or verified by the RuneLite Developers.
```

PR #5083, opened 2023-10-18
Update star-calling-assist

zodaz:
Forgot to set auto call and auto update to be enabled by default on fresh installs, been causing confusion.


Adam-:
Note that you cannot retroactively change a `warning` for a plugin, and so you may not be able to use it as a way to show agreement with a privacy policy.

PR #5091, opened 2023-10-20
update "stardust-per-hour-calculator" plugin

kevindarbydev:
reset count and timer variables if stardust is banked


Adam-:
this needs rebased.

PR #5093, opened 2023-10-21
Added Posiedien's Leagues Planner to the pluginhub

JacobNelsonGames:
Hello!

This plugin is to aid people for planning their route in the upcoming trailblazer 2 league. The plugin allows people to put in their own custom tasks or hook into tasks I've manually added into the planning. With this, you can visually see where tasks can be completed and the plugin forks from the shortest path plugin from Skretzo (with his permission) to allow pathing between points.

I plan on updating it in the future with trailblazer 2 tasks once leagues 4 is out.

image

image

image

image

image


raiyni:
Is that first screenshot something you have in the plugin with that overlay?

JacobNelsonGames:
The first screenshot isn't available for most users, but its my editor for the regions.

JacobNelsonGames:
(Need to edit the source to enable the region editor)

raiyni:
Okay, you can clip the overlay with the bounds of the world map so it doesn't overflow the bounds if you really care.

JacobNelsonGames:
Oh yeah old screenshot, I fixed the clipping, hah

PR #5106, opened 2023-10-23
Create escape-crystal-notify

Xylot:
Notifies you when you do not have an active escape crystal

Useful for HCIM

### DIFF
--- a/plugins/escape-crystal-notify
+++ b/plugins/escape-crystal-notify
@@ -1,2 +1,2 @@
 repository=https://github.com/Xylot/escape-crystal-notify.git
-commit=1e2ca241d484f807a1decd2aa07bbb45bad2a5ba
+commit=7352e10be881f5a44a0a8def606a0ee12d3b0097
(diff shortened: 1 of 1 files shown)


YvesW:
Please change your config group to something unique, instead of ``example``.

Xylot:
> Please change your config group to something unique, instead of `example`.

Fixed! Thank you for pointing that out!

Xylot:
Added option to display time left before triggering the inactivity teleport

Felanbird:
The two images you just added are enormous, please scale them down otherwise java will be very upset.

Xylot:
> The two images you just added are enormous, please scale them down otherwise java will be very upset.

They get scaled down by a factor of 15 in memory and cached before they are drawn to the screen. I had them large to improve fidelity when scaled up by the user. Is that enough or should I prioritize a certain size as the base image?

Nightfirecat:
The loaded images will be kept in memory by Java, so it doesn't matter that you scale them down after the fact. You should be able to make them substantially smaller without impacting user experience.

Xylot:
@Felanbird @Nightfirecat I have reduced the size of the two images! Let me know if the new sizes are acceptable or if there are any further modifications required!

PR #5110, opened 2023-10-24
Create touchscreen

ggaviglio:
Add/fix controls for touchscreens.
Left click activates after release. This fixes issue with clicks occurring before mouse position has updated. 
Click hold is rebound to right mouse.
Click drag is rebound to middle mouse, except when an item container is being clicked.


ggaviglio:
The plugin repo was private and has been made public now.

YvesW:
> The plugin repo was private and has been made public now.

You should either close and reopen your PR, or push new code then to force CI to rerun. Since your configgroup is still called ``example`` instead of something unique, I suggest the latter.

raiyni:
After discussion, we don't think we can safely support these plugins unfortunately.

PR #5111, opened 2023-10-24
Add RuneSync

skillingdev:
Adds the plugin companion for [RuneSync](https://www.runesync.com)

Last leagues I was frustrated by the hiscores only being a current snapshot of an account (rather than a time series), and hiscores not tracking collection log items.

For now, this plugin sends loot to the RuneSync API, which allows RuneSync to display users' unique drops. In the future I'd like to display which tasks they have completed, what zones they unlocked, etc. See https://www.runesync.com/users/mid%20packs for an example (except imagine that stats are changing over time, and the loot shown are only collection log items)

### DIFF
--- a/plugins/runesync
+++ b/plugins/runesync
@@ -1,0 +1,2 @@
+repository=https://github.com/skillingdev/runesync-plugin.git
+commit=a7fca820d4e91afab97fea850502f2d621e7afa2
--- a/plugins/runesync
+++ b/plugins/runesync
@@ -1,2 +1,2 @@
 repository=https://github.com/skillingdev/runesync-plugin.git
-commit=a7fca820d4e91afab97fea850502f2d621e7afa2
+commit=c12fff4e9c6af1f512117b1538e4c909c335152d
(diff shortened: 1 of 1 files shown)


skillingdev:
My original commit hash broke a build.gradle lint rule - I've fixed that with the latest commit hash

Adam-:
Your plugin depends on `SessionManager`. For security reasons, we do not allow plugins to depend on or use it. Additionally, you need to set a warning such as `This plugin shares your IP and username with a 3rd party server not controlled or verified by the RuneLite Developers.`

skillingdev:
Thanks for the review :)

I didn't realize about `SessionManager` - luckily (or not luckily, since it was my mistake for leaving it in) this was unused.

I'll remove the `SessionManager`, and add a warning as well.

Nightfirecat:
The `warning` field should be added to the file in this PR, [not in your runelite-plugin.properties file](https://github.com/skillingdev/runesync-plugin/blob/c12fff4e9c6af1f512117b1538e4c909c335152d/runelite-plugin.properties#L6)

skillingdev:
Gah, ok! Thanks for the pointer. Done

PR #5121, opened 2023-10-27
Add Touchscreen Plugin

Notloc:
*Fixes*
- Fixed the dead input bug.
    - (Without this plugin touch inputs will be missed/inaccurate as the mouse position from last frame is used for interacting, but touch inputs teleport the mouse, so that can be a massive difference)

*Features*
* Ability to rotate the camera by dragging on the game view
* Ability to zoom the camera by dragging vertically on the minimap
* Ability to zoom the minimap by dragging horizonally on the minimap
* Ability to scroll various menus by dragging on the menu
* Faster right clicking option
    * Click and hold to open the right click menu as normal on touchscreen, without releasing the click move your finger onto the desired option then release

*Video Examples*
- Before
    - https://www.youtube.com/watch?v=HEus8rMrut0
- After
    - https://www.youtube.com/watch?v=UGWogZZtLlk

### DIFF
--- a/plugins/touchscreen
+++ b/plugins/touchscreen
@@ -1,3 +1,3 @@
 repository=https://github.com/Notloc/Runelite-touchscreen.git
-commit=99bf9962235ae782690448f343d655e54fd2cb78
+commit=8c2ce0a7092d7a3c02979222f75ff851b424763d
 authors=Notloc
--- a/plugins/touchscreen
+++ b/plugins/touchscreen
@@ -1,0 +1,3 @@
+repository=https://github.com/Notloc/Runelite-touchscreen.git
+commit=e79926b16ca6837130c7a1f429bf44181831219a
+authors=Notloc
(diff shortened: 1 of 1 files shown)


Notloc:
Apologies for the commits post PR opening.
I've cleaned up all the noted issues and added tags.

Felanbird:
we don't care about commits, we'd rather you spam commit than spam open/close PRs every change

raiyni:
My feedback for this pr and the other one that is open is the same. Emulating mouse events, trying to time clicks, and `fixing` events sounds not great maybe in regards to Jagex cheat detection.

Notloc:
Unfortunate.
I don't think there is any way I could achieve a good UX without post-processing the events as I've done.
Even if I somehow managed to receive actual touch events, they would still need to be translated into emulated mouse events.

If that is the final verdict, then I think touchscreen support as a concept is just dead in the water and this PR can be closed.

raiyni:
After discussion, we don't think we can safely support these plugins unfortunately.

PR #5122, opened 2023-10-27
Create TickImprover

Chvii:
Improves tick consistency on certain WIFI cards with energy saving measures in place.


LlemonDuck:
I've just merged `tick-fixer-for-mac` (#4830), which was ahead of this plugin in the review queue. Does it sufficiently cover this plugin's use case or do you still want this one reviewed?

Chvii:
I would say it sufficiently covers it yes, thanks

PR #5125, opened 2023-10-28
Create ChatPanel

Yenof:
This plugin displays Public, Private, Clan chats and Game messages to the side bar.
Not sure if I did the custom icon right.


YvesW:
Please remove the exampleConfig and examplePlugin files. I'd also recommend renaming the example folder to something else but that's my personal preference.

PR #5128, opened 2023-10-29
Update zebak-roar-chat

ruttie3:
Fixed incorrect zebak id
Added sound functionality


Adam-:
This is doing a bunch of sound io on the client thread in toa

LlemonDuck:

make sure to load the clip off-thread, ideally only once when the plugin starts instead of every time it is played

PR #5143, opened 2023-11-02
Update default-minimap-zoom

YvesW:
Replaces WidgetInfo with new ComponentID and InterfaceID constants, among others.
Please hold for release


YvesW:
1.10.15 has been released, so it's release ready.

PR #5146, opened 2023-11-03
Archipelago Multiworld Randomizer Plugin

digiholic:
Creates a plugin for interfacing with the Archipelago Multiworld randomizer. Players of dozens of games can all collectively submit seeds to have items placed randomly between all games. Plugin uses the same [Archipelago MultiClient](https://github.com/ArchipelagoMW/Archipelago.MultiClient.Java) used by Minecraft and Slay The Spire.

Players submit a seed on the [Archipelago Website](https://archipelago.gg/), which will give them a URL and port to connect to. Players enter that information in the plugin settings and hit the server tells the player which map chunks and items they are allowed to use, as well as sending checks to the server when tasks are completed.


Felanbird:
Reflection is not permitted in hub plugins

digiholic:
Reflection isn't used in this plugin, did you mean in the library it uses? Would that be a problem?

Felanbird:
I'm not actually sure of the answer to that, my mistake if it does end up being okay :+1:

raiyni:
Our reply is going to be the same as the last time you submitted this plugin.

https://discord.com/channels/301497432909414422/419891709883973642/1130606782973223043

digiholic:
I pulled the last one because it still required a lot of work. Even if it were accepted it would almost immediately need patches. This one should be stable.

Is there anything I can do to make the review process easier to review? I could see if we can get the multiclient plugin added to Gradle or some other tool chain so that it can be vetted somewhere?

raiyni:
Pulling the source files that you need into your plugin is the easiest/fastest.

digiholic:
So, including the plugin files directly in the project, rather than using a jar as a dependency? I can probably do that.

raiyni:
Right. If you can narrow down which sources you need, it is easier for the average reviewer. If it's going to be a huge amount of files, might as well wait on dependency review. I haven't taken a look at the source of Archipelago to know what's up.

digiholic:
So, I tried to untangle the stuff I need from the stuff I don't, and ended up keeping the entire plugin. I could swap out the jar for the source code if that would help? Maybe I can see if I can get them to rework it to not use reflection?

digiholic:
The reflection is for an annotation event listener, I don't think it can be removed, but I don't think it's much of a vulnerability. It lets the _library_ run code from the _plugin_, not vice-versa, so it wouldn't be giving any access to data this plugin doesn't explicitly give it. Any changes that could potentially compromise anything would need to be approved. So, you'd just need to check that the code that's annotated isn't giving away any kind of account information or potential backdooring, and you can see that these event functions don't do any of that.

Nightfirecat:
Sorry for the late review on this one; the build is failing. If this change is just vendoring the entire library, it's probably fine to just leave it as a dependency and we'll review it.

digiholic:
Sorry for the late response to the late review: I had another project that was higher priority than this one I had to dedicate my time to, but that's done and I can get back to working on this one.

I've fixed the build errors, as well as a few UI issues while I was at it. This one's properly compiling, and back to the jar. In the meantime, the version I was using _has_ gotten actual release so it's a practical tagged release instead of me compiling from source to use a new feature that I required. You can see the build here, and check out the repository:
https://github.com/ArchipelagoMW/Archipelago.MultiClient.Java/releases/tag/1.12

Nightfirecat:
Is it at all possible to exclude the duplicate dependencies used in the Archipelago client? It brings along its own copies of gson, apache, hc, slf4j; most/all of which we already provide to plugins at build time. This would substantially help reduce the burden of reviewing it

Felanbird:
Images that are packed with a plugin are blown up to their max size by Java when loaded into memory, meaning your [panel_icon.png](https://github.com/digiholic/osrs-archipelago/blob/master/src/main/resources/gg/archipelago/panel_icon.png) is astronomical in size (width * height * 4)

digiholic:
Turns out the plugin doesn't work without these requirements:

```
implementation 'org.apache.httpcomponents.client5:httpclient5:5.2.1'
implementation 'org.java-websocket:Java-WebSocket:1.5.2'
implementation 'com.google.code.gson:gson:2.8.9'
```

It was previously mentioned that apache and gson were already provided, but without these in the gradle, the plugin doesn't connect to the Archipelago API or parse the data from it, so I've had to add them back in. At least they're in this project instead of a dependency of it.

digiholic:
Hm, doesn't look like the build process liked that. I'll see what I can do tomorrow.

Nightfirecat:
[You add a nav button for your plugin's panel here](https://github.com/digiholic/osrs-archipelago/blob/f55ab7fba99b50116b82578cc3e32f4f89ff37f5/src/main/java/gg/archipelago/ArchipelagoPlugin.java#L97) but it is never removed. Please ensure it is removed on plugin shutDown().

Plugin code LGTM otherwise. I haven't yet reviewed the added dependency or the archipelago lib

digiholic:
Just to make sure, is this waiting on anything on my end or is it just getting the dependencies reviewed?

Nightfirecat:
Correct, I'm hoping to catch up on a bunch of reviewing this week.

Nightfirecat:
Sorry this all took so long to get reviewed and accepted--I'm happy to say all looks great and this is good to go!

PR #5153, opened 2023-11-05
This plugin recolors herb seeds to they match their respective herb

Mike-U5:


raiyni:
Remove your config class

Mike-U5:
@raiyni Done, also added the missing license file.

PR #5159, opened 2023-11-07
Add interface-hider plugin

CreativeTechGuy:
A plugin which enables players to make transparent (or completely hide) almost any interface element.

To comply with Jagex's rules on hiding interfaces, this plugin prohibits any modifications of the spellbook.

### DIFF
--- a/plugins/interface-hider
+++ b/plugins/interface-hider
@@ -1,0 +1,2 @@
+repository=https://github.com/CreativeTechGuy/interface-hider.git
+commit=59a782ce895f77ce6859468b3579ea76cbc3c6c1
--- a/plugins/interface-hider
+++ b/plugins/interface-hider
@@ -1,2 +1,2 @@
 repository=https://github.com/CreativeTechGuy/interface-hider.git
-commit=59a782ce895f77ce6859468b3579ea76cbc3c6c1
+commit=ef325451bb7dd3553fe6b596b88a9f94ceecff29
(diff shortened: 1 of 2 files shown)


LlemonDuck:
while we normally see this kind of issue with id based plugins (npc id / object id / etc), these kinds of arbitrary-input plugins are incredibly difficult to moderate and lead to a game of cat and mouse when issues are found, that we prefer not to accommodate where possible. 

if there are specific interfaces you would like to hide within the confines of the third party client guidelines, feel free to resubmit with a specific, fixed list of interfaces instead.

CreativeTechGuy:
@LlemonDuck is that documented somewhere in the rules for Plugins? If so, I didn't notice it. If not, can it be? I asked on Discord and the general sentiment was that this would be okay as long as certain interfaces were blocked. Other than that, I'm not sure what I should have done to not waste such an incredible amount of time on this.

LlemonDuck:
This is a newer stance we're starting to take after some recent mishaps with plugins being used in unexpected ways. The interface system is too overarching and touches on mechanics all over the game, which makes it almost impossible to accurately moderate such sweeping modifications. It is also not future proof toward new content.

If you feel you have a compelling argument, feel free to come chat in the discord again.

CreativeTechGuy:
Will this be documented as a new rule? Totally fair if when I made this it wasn't a rule yet. I get it. But I hope that going forward there can be better documented guidance so other's don't go down the same path.

Thanks! 😄

PR #5168, opened 2023-11-08
Cheese Coins Plugin

DuhonTheGuy:
Created this simple plugin, only changes the names of coins to be more "cheesy".


Felanbird:
Your config group needs to be something other than `example` 
https://github.com/DuhonTheGuy/Cheeze-Its-Gp/blob/525d8728e27532ecd90eaa1a74800c1763f02314/src/main/java/com/cheesegp/CheeseGpConfig.java#L7

DuhonTheGuy:
Oh hey, ya mean the "example" part? Yeah I forgot that, let me fix it

DuhonTheGuy:
I decided this is too much trouble for a bad meme

PR #5182, opened 2023-11-12
tobqol: overlay prio fix

damencs:
- updates overlay priority to not be on top, but over widgets rather


damencs:
Going to fix a different way.

PR #5195, opened 2023-11-14
Fix tasks-tracker

tylerthardy:
Changes:
- The team identified the most likely issue with client freezing (described as "changed 10,000 frame redraws on client thread to 15" on world hop)
- The team added Leagues 4 metadata to support fetching new task json from our json store


tylerthardy:
Will reopen in a day or two with League 4 updates

PR #5196, opened 2023-11-14
Added nyloer side panel

EIKOOT:
Added nyloer side panel.

Role Swapper
Each role has configuration for
- menu swaps (except for cast and other non-tos swaps)
- menu shift swaps (except for cast and other non-tos swaps)
- nylo tile highlighting

Stats tables:
- room times
- boss times
- boss spawn times
- aggregated stall counts by wave intervals
- stalls
- number of bigs alive during certain wave spawns

### DIFF
--- a/plugins/nyloer
+++ b/plugins/nyloer
@@ -1,2 +1,2 @@
 repository=https://github.com/EIKOOT/nyloer.git
-commit=dca18af1a9fb738094c7bd6affb0380cf7ef03f3
+commit=b9d3ebb6adca2fe694d6110ae75b4769075021e5
--- a/plugins/nyloer
+++ b/plugins/nyloer
@@ -1,2 +1,2 @@
 repository=https://github.com/EIKOOT/nyloer.git
-commit=947a28419efb6ca84566a2de0183d303b1f51ec1
+commit=c80ab438a45624c5e7b10ffecc56148d678405c4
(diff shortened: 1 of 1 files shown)


Nightfirecat:
[You need to remove the sidebar button you've added here on plugin shutdown.](https://github.com/EIKOOT/nyloer/blob/f548cb23383ad6cb0ecc00a39261454497f2c046/src/main/java/com/nyloer/NyloerPlugin.java#L136)
On that note, [can you scale down your sidebar icon?](https://github.com/EIKOOT/nyloer/blob/f548cb23383ad6cb0ecc00a39261454497f2c046/src/main/resources/ico.png) Java's image loading will keep the full-size image in memory even when a scaled-down version is displayed on the button otherwise.

LGTM otherwise

EIKOOT:
> [You need to remove the sidebar button you've added here on plugin shutdown.](https://github.com/EIKOOT/nyloer/blob/f548cb23383ad6cb0ecc00a39261454497f2c046/src/main/java/com/nyloer/NyloerPlugin.java#L136) On that note, [can you scale down your sidebar icon?](https://github.com/EIKOOT/nyloer/blob/f548cb23383ad6cb0ecc00a39261454497f2c046/src/main/resources/ico.png) Java's image loading will keep the full-size image in memory even when a scaled-down version is displayed on the button otherwise.
> 
> LGTM otherwise

pushed changes

PR #5204, opened 2023-11-16
Update 3D Weather to v1.2.2

ScreteMonge:
- bugfix: audio not correctly transitioning to muffled form when stepping indoors


raiyni:
Can you remove sysout calls or replace with `log.debug`?

ScreteMonge:
Sysout calls removed, thanks

PR #5225, opened 2023-11-21
Watchdog v3.1.0

adamk33n3r:
Chat picker and some import fixes
Also, parent toggle fix, non-windows tray notification, request focus notification, message picker on notification fired

### DIFF
--- a/plugins/watchdog
+++ b/plugins/watchdog
@@ -1,2 +1,2 @@
 repository=https://github.com/adamk33n3r/runelite-watchdog.git
-commit=ec5966b3ee6b10c141d9737065d6d7a7f2cd587f
+commit=f239549d55bd7f7c4a4710f2ccb68dc57742501f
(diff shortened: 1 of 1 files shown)


raiyni:
What's copying the notifier class for?

pajlada:
> What's copying the notifier class for?

Hey @raiyni, the previous implementation that Watchdog used to send tray notifications only worked on Windows - for it to work on Linux & macOS the way we wanted, we needed to use an API from the Notifier class that wasn't public.
Instead of making those parts public, which wasn't desired as per [this discussion in #development](https://discord.com/channels/301497432909414422/419891709883973642/1175876740262936606), we copied the necessary functions. Since the functions spawn processes, the functions are kept as close to the original as possible to make it easily diffable.

The PR that implements this can be found here https://github.com/adamk33n3r/runelite-watchdog/pull/112

PR #5228, opened 2023-11-22
Create undying-retribution-timer

DominickCobb-rs:
Create undying-retribution-timer

Simple timer based on chat messages provided during usage of the relic.

Used a good chunk of code from betteroverload plugin, but didn't see any licensing in the top of the source.


Felanbird:
> Used a good chunk of code from betteroverload plugin, but didn't see any licensing in the top of the source.

Hub plugins have a License file exactly like you added to this one, follow the license as it states
https://github.com/tdb48/better-timers/blob/master/LICENSE

PR #5232, opened 2023-11-23
added TombsOfAmascutStats plugin

sreilly64:
This plugin, [TombsOfAmascutStats](https://github.com/sreilly64/TombsOfAmascutStats), has essentially the same purpose/functionality as the [Theatre of Blood Stats Plugin](https://github.com/HSJ-OSRS/theatreofbloodstats) but is designed to support the Tombs of Amascut raid. It provides a detailed breakdown of phase times, damage tracking (per enemy, per encounter total, and percentage of team total), and some encounter specific stats such as boss healing. All of which can be configured to be displayed as icons on the buff bar and/or as messages in the chat box.

### DIFF
--- a/plugins/tombs-of-amascut-stats
+++ b/plugins/tombs-of-amascut-stats
@@ -1,2 +1,2 @@
 repository=https://github.com/sreilly64/TombsOfAmascutStats.git
-commit=becccf32a24102da5e4f3ca205499951554d125f
+commit=a0fc5bd373ba0206e7b4c9e005a552b5c56b5a36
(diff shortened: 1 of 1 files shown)


Felanbird:
Don't use `enabledByDefault = false` it's confusing for users who add a plugin that isn't enabled after installation.
https://github.com/sreilly64/TombsOfAmascutStats/blob/becccf32a24102da5e4f3ca205499951554d125f/src/main/java/rr/raids/tombsofamascutstats/TombsOfAmascutStatsPlugin.java#L59

I assume this line is included due to it being used in the plugin you forked from, but it should be removed.

sreilly64:
Thank you for the feedback. You are correct, that bit of code was present in the repo that I forked from. I have removed that configuration and updated the commit id.

PR #5241, opened 2023-11-25
Create ImprovedInventoryTag

3mot:
Adds a dot option to inventory tagger and adds some new options for sizing. More shapes to be added as well. I really hate outline, underline and fill so added Dot!


Nightfirecat:
[Don't use `enabledByDefault = false`](https://github.com/3mot/InventoryTagSquares/blob/eeb0b98255b05af46b1703a3349eba7d1d7d04fe/src/main/java/com/ImprovedInventoryTag/InventoryTagsPlugin.java#L65), it's confusing for users to install a hub plugin which isn't enabled right away.
Also [use a unique config group](https://github.com/3mot/InventoryTagSquares/blob/eeb0b98255b05af46b1703a3349eba7d1d7d04fe/src/main/java/com/ImprovedInventoryTag/InventoryTagsConfig.java#L36); this would overwrite the core plugin's configs if changed.
Additionally, I believe there are namespace issues for class names as well, please use new names for your plugin's classes instead of using copies of the core plugin's class names.

3mot:
Took me a lot of tries but I am now educated on this process. Hopefully that looks better

3mot:
enabled by default missed that part

Nightfirecat:
Upon second look, it seems like you're trying to add this to aid your AHK scripts. We're not really interested in doing that.

PR #5244, opened 2023-11-26
Create Collection Log Luck plugin

peanubnutter:
A [Runelite](https://github.com/runelite/runelite) plugin that calculates luck for collection log items by pulling data from [collectionlog.net](https://collectionlog.net/).


peanubnutter:
Just verifying that both accounts are me - not sure why my computer is hooked up to a different account than github.com

Nightfirecat:
We don't allow reflection in the plugin hub. Please remove any uses of the `java.lang.reflect.*` APIs.
Also, it looks like this plugin is a fork of the existing Collection Log hub plugin. Have you considered contributing to that plugin instead?

peanubnutter:
Ah, will do. This was originally an improvement to that plugin, but the owner went completely missing for months and ignored messages from me and everyone else. I had to close the pull request and decouple the plugin from that one. 

I'll continue to be completely separate from that plugin as I don't foresee any compatibility issues and only foresee issues updating in the future.

peanubnutter:
Ah, I think I saw that page in the past and was a bit afraid of losing commit history or something. Thanks so much for the review and advice!

I did update the config, classes, README etc. everywhere except for a couple of classes which are identical (deserializing the website response). For some reason, the README was showing up as the old README until I merged the branch to master. Should be fixed now: https://github.com/peanubnutter/collection-log-luck

Also removed the reflection. Not sure why I even did it that way in the first place. Must have been a late night.

peanubnutter:
Any other feedback on this one? I've been wanting to make some updates for a few months but am hesitant to put in any more work if it's not even guaranteed this plugin will get merged.

Nightfirecat:
I haven't yet had time to re-review this PR, particularly since it introduces a new dependency.

Nightfirecat:
[You're still using `java.lang.reflect.Type` here.](https://github.com/peanubnutter/collection-log-luck/blob/c7f05a80b8195ac2633ecf579aaf0c3e8d173f94/src/main/java/com/peanubnutter/collectionlogluck/util/CollectionLogDeserializer.java#L6)
[Please use a distinct user agent so it's not confused for the Collection Log plugin.](https://github.com/peanubnutter/collection-log-luck/blob/c7f05a80b8195ac2633ecf579aaf0c3e8d173f94/src/main/java/com/peanubnutter/collectionlogluck/util/CollectionLogLuckApiClient.java#L21)
Also it looks like your license removed the copyright notice given from the original plugin you've forked. You should preserve that to comply with the license terms.

plugins/collection-log-luck
```diff
@@ -0,0 +1,3 @@
+repository=https://github.com/peanubnutter/collection-log-luck.git
+commit=c7f05a80b8195ac2633ecf579aaf0c3e8d173f94
+warning=This plugin pulls data from collectionlog.net, a 3rd party server not controlled or verified by the RuneLite developers, and which can access your IP address.
```
Nightfirecat, line 3:
```suggestion
warning=This plugin submits your username and IP address to a server not controlled or verified by the RuneLite developers.
```

peanubnutter: The original collection log plugin has that language, but this plugin does not POST any data - only GETs data, so it doesn't really "submit" anything, especially not username. Thoughts?

peanubnutter:
Ah, no problem, thanks for the review!

> [You're still using `java.lang.reflect.Type` here.](https://github.com/peanubnutter/collection-log-luck/blob/c7f05a80b8195ac2633ecf579aaf0c3e8d173f94/src/main/java/com/peanubnutter/collectionlogluck/util/CollectionLogDeserializer.java#L6) [Please use a distinct user agent so it's not confused for the Collection Log plugin.](https://github.com/peanubnutter/collection-log-luck/blob/c7f05a80b8195ac2633ecf579aaf0c3e8d173f94/src/main/java/com/peanubnutter/collectionlogluck/util/CollectionLogLuckApiClient.java#L21) Also it looks like your license removed the copyright notice given from the original plugin you've forked. You should preserve that to comply with the license terms.

1) For the Reflect API, I'm not actually using any reflection - only the java.lang.reflect.Type class definition, which is required to override JsonDeserializer.deserialize(). The collection log plugin already does this [here](https://github.com/evansloan/collection-log/blob/02d2a212a2e177741e9a15207290693d51c47f42/src/main/java/com/evansloan/collectionlog/util/CollectionLogDeserializer.java#L13), and this class is one of the few preserved without changes from the original plugin. 

2) I could be wrong, but I think the server might be using the user agent for some kind of API versioning - not sure. I tested it and it seems to work, though. Will submit a new commit in a sec.

3) Would I have two licenses? Or would I just add both of our github names to the copyright? The license file is identical otherwise.

Nightfirecat:
You should be able to copy their license file, then add a line with your copyright below theirs.

PR #5255, opened 2023-11-28
Update Undying retribution timer

DominickCobb-rs:
Add config setting to show infobox only on cooldown

Prevent timer from continuing on non-league worlds


DominickCobb-rs:
```
if (!e.getGroup().contains(CONFIG_GROUP))
{
    return;
}
```
Like this?

dekvall:
> ```
> if (!e.getGroup().contains(CONFIG_GROUP))
> {
>     return;
> }
> ```
> 
> Like this?

Yes, but with `equals`

DominickCobb-rs:
Done

PR #5263, opened 2023-12-01
Added controller plugin

willthehuman:
Added the controller plugin file.

This plugin allows the use of a controller to play osrs.

You can bind different buttons to different tabs that already have a shortcut (see https://oldschool.runescape.wiki/w/Shortcut_keys).

See demo [here](https://github.com/willthehuman/controller-plugin)


raiyni:
Much like our response to the touch screen plugins, we aren't interested in using robot nor forwarding awt events to cause clicks and actions within the game.

PR #5275, opened 2023-12-07
Updated Bone Runner Plugin

Elrol:
Added host overlay to show who is hosting for runners


Nightfirecat:
This change would prevent your plugin from being unloaded as it is registered with java.sql's DriverManager but never unregistered. Furthermore, it leaks jdbc connections. We're hesitant to support (and add a substantial new dependency for) a plugin which uses plaintext credentials to connect directly to a mysql database, even if it is read-only. Can't this data be served via http?

Elrol:
The data that is being sent isn't anything confidential, but we want the data to not be able to be changed by others.
Its basically to keep track of user stats when running bones so we can have a leaderboard kind of thing. 

Could you point me in the right direction to send this data using http?

Nightfirecat:
By "sending data over HTTP", I meant serving a static file, or something of the sort. (this file could be generated and updated by a server as changes are needed) A server script could also generate and deliver data from the mysql database over HTTP.

PR #5297, opened 2023-12-13
Adds barrows nothing sound plugin

JillevdW:
Adds a sound that plays when the user does not receive an item from the Barrows loot chest. This is in reference to the Swampletics Youtube series, where this sound is played when Settled opens an empty chest.


plugins/barrows-nothing-sound
```diff
@@ -0,0 +1,2 @@
+repository=https://https://github.com/JillevdW/barrows-nothing-sound.git
```
Felanbird, line 1:
```suggestion
repository=https://github.com/JillevdW/barrows-nothing-sound.git
```

Nightfirecat:
What is the source of nothing.wav? It sounds like the one used in Spongebob, which we would not be allowed to distribute due to copyright.

Nightfirecat:
Closing due to inactivity.

PR #5299, opened 2023-12-16
Update herblorerecipes to load cache on LOGGED_IN

climbridecode:
Was seeing intermittent NPEs, maybe due to a race condition between cache load and itemManager? Haven't been able to reproduce after moving cache load to happen on `LOGGED_IN` state.


climbridecode:
Will address build issues when time permits :)

Alexsuperfly:
ci is failing:
> Use ComponentID or InterfaceID instead of WidgetInfo or WidgetID

climbridecode:
@Alexsuperfly resolved the build issues!

PR #5300, opened 2023-12-16
remove GoTR AFK plugin

xenotrace:
**Reason for requested removal**
This plugin serves as a net detriment to legitimate players engaging with the GoTR minigame.

**Background**
It encourages (and facilitates) players entering the minigame and going AFK after absolute minimal interaction to abuse the point system.

This is problematic because:
1. Too many people doing this (note: currently has _over 10k active installs_) leads to games being repeatedly lost as proper engagement is required throughout, not just at the very beginning.
2. Games are capped to 200 players per world - in peak times you must compete with other players to get into the minigame across basically all GoTR worlds. More leechers mean a harder time for those who wish to actually play the minigame as intended to get in.

I am unsure the proper protocol for requesting a plugin to be removed - please do let me know if there is a different procedure to follow.

image

### DIFF
--- a/plugins/zom-afk-gotr
+++ b/plugins/zom-afk-gotr
@@ -1,2 +1,0 @@
-repository=https://github.com/JZomDev/zom-external-plugins.git
-commit=ba5c479d39b8f99c50d73ce95a6066f72905d4d3


JZomDev:
It's probably better for Jagex to fix their minigame from preventing 150 points as a floor being enough to get through. By the time you get 500 points at WT the minigame is almost over in mass where 150 points at GOTR is 30 seconds of play and you'll lose if everybody does only 150 points. 

This plugin was made only because of how absurd the strategy is. 

I wouldn't mind the plugin being deleted but only if Jagex fixes the AFK meta.

raiyni:
We aren't gatekeeping content strategies for plugins if they aren't violating Jagex 3pc guidelines. If the author wants this plugin removed, it can be removed by them.

PR #5307, opened 2023-12-18
Uri Quotes Plugin

rozzan:
This plugin allows users to change what Uri says when you complete an Emote clue step. The data is provided through raw text, or via a URL, both provided via the plugin config.

PR #5314, opened 2023-12-23
Plugin that replaces Uri dialogue with Dril tweets

Mike-U5:
This plugin does not connect to any kind of API, the tweets are randomly picked from a hardcoded list in DrilQuotes.java.

(NpcDialogueReplacementPlugin was the original name for this plugin before it was repurposed)


geheur:
There is another uri quotes plugin #5307 . Is yours better than theirs?

raiyni:
I don't think we can accept this plugin as it's pretty vulgar

Mike-U5:
@geheur If sidesteps the issue mentioned by @raiyni then it seems to be better.

Mike-U5:
The other plugin doesn't seem to be in the hub yet but I will test to see if that one covers this use-case first.

PR #5332, opened 2023-12-31
Add potion-dose-overlay plugin definition

trs:
This plugin adds an overlay to potion-type items showing the number of does left.

icon

I created this as I sometimes have a hardtime determining if a potion is 4 or 3 dose.


geheur:
how is this different from the existing "item charges" plugin?

YvesW:
This already exists as core plugin. https://github.com/runelite/runelite/wiki/Item-Charges#show-potion-doses

trs:
> how is this different from the existing "item charges" plugin?

Ha, I didn't know this was a thing. I looked into it, and yeah it essentially does the same thing.

Oh well, good excuse to learn how to make a plugin at least.

PR #5334, opened 2024-01-02
Add Day/Night Cycle

Natchuuuu:
Adds a configurable day/night cycle based on your system time.

Automatically changes sky-box color, fog and lighting settings to match with your time of day, or select what part of day/night you want to experience. Please view the README for upcoming feature plans and contact information.

### DIFF
--- a/plugins/daynight-cycle
+++ b/plugins/daynight-cycle
@@ -1,0 +1,2 @@
+repository=https://github.com/Natchuuuu/day-night.git
+commit=311786b9da950eb7914602fc95b78de7e491848d


aHooder:
Hi, maintainer of 117 HD here :wave:. I think this approach could cause confusion for users with how it's currently implemented. I'm not opposed to other plugins modifying the behaviour of 117 HD, but I think we should try to find a better way than modifying configs directly. Maybe 117 HD could listen for `JsonObject` instances on the event bus with some kind of identifier, and accept supported commands that way. @Natchuuuu if you want to discuss this further, feel free to drop by the [117 HD Discord](https://discord.gg/U4p6ChjgSE) :slightly_smiling_face:

Natchuuuu:
After discussion with @aHooder on the current difficulty of cross-plugin communication, this pull will be drafted while a better collaborative solution is being worked on. 

If there are any additional concerns please reach out to ナチュ on the RuneLite discord or 117 HD discord.

Nightfirecat:
Hey, wanted to check in on this PR. Is this still planned to be integrated in 117hd? If so, does this need to remain open?

cc: @Natchuuuu @aHooder

aHooder:
Ah, sorry. I still haven't implemented the API which would be needed for this, as we didn't land on the specifics. I would still like to do it at some point, and make Region Locker work with 117 HD through that API too, but I don't know when that might happen, unfortunately. Maybe there's a way we could fast-track this if @Natchuuuu would prefer to not wait for a more proper API. If it'll be through a new API though, this code will likely end up looking quite different.

PR #5342, opened 2024-01-04
Add clicker training plugin

pyonium:
A simple plugin that plays a nice dog clicker audio file when you're leveled up! Classical conditioning at its best.


raiyni:
Don't close and open new PRs to resolve issues, continue pushing changes the same branch.

Since you deleted your branch, you can resolve that problem by clicking the restore branch button

PR #5343, opened 2024-01-04
Add clicker training plugin

pyonium:
Adds a clicker training plugin that plays a dog-clicker sound every time you level up a skill. Classical conditioning!

### DIFF
--- a/plugins/clickertraining
+++ b/plugins/clickertraining
@@ -1,0 +1,2 @@
+repository=https://github.com/pyonium/clicker-training-osrs.git
+commit=6ab3ead1d468e4806a9816e0ec72f5ffa048ff59


Nightfirecat:
[I'm fairly sure Adam doesn't have sole copyright of the code you wrote.](https://github.com/pyonium/clicker-training-osrs/blob/a922393991e7cf91da5e9afeed9c507c561dde9e/LICENSE#L3) You should update this file accordingly. (particularly including the plugins you borrowed code from)
Also I don't know the source of this audio file (given it is so generic), I presume it is licensed in a way which we can distribute it?

pyonium:
Should be fixed now! Updated the license to add the original authors of the borrowed code.

Sound is from https://freesound.org/people/dnewtonjr/sounds/231559/ under Creative Commons 0, modified. Should be good.

PR #5344, opened 2024-01-04
Escape Crystal Teleport Notification

oceijt:
Allows user to get notified if the escape crystal would activate upon taking damage


Nightfirecat:
You must ensure the added infobox is removed when your plugin is turned off; currently no `shutDown()` method exists, so this is not happening.

Nightfirecat:
Closing due to inactivity.

PR #5346, opened 2024-01-04
Add combat-logger plugin

SuperNerdEric:
Add combat-logger plugin.

Logs combat events to a file. Combat logs can then be analyzed by an external tool to calculate DPS, etc.

image

### DIFF
--- a/plugins/combat-locked
+++ b/plugins/combat-locked
@@ -1,2 +1,2 @@
 repository=https://github.com/christianlegge/combat-locked-plugin.git
-commit=e7df05c4ad0f686cf5377019714c324f711011ac
+commit=c8c23059c674c579fd1cff687e916cffc4953378
--- a/plugins/combat-locked
+++ b/plugins/combat-locked
@@ -1,2 +1,2 @@
 repository=https://github.com/christianlegge/combat-locked-plugin.git
-commit=c8c23059c674c579fd1cff687e916cffc4953378
+commit=e7df05c4ad0f686cf5377019714c324f711011ac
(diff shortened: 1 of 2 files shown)


Nightfirecat:
[Reflection is not permitted in hub plugins.](https://github.com/SuperNerdEric/combat-logger/blob/c8c23059c674c579fd1cff687e916cffc4953378/src/main/java/com/combatlogger/HitSplatUtil.java#L5) Copy the code from HitsplatID into your codebase if you need to reference it.

SuperNerdEric:
> [Reflection is not permitted in hub plugins.](https://github.com/SuperNerdEric/combat-logger/blob/c8c23059c674c579fd1cff687e916cffc4953378/src/main/java/com/combatlogger/HitSplatUtil.java#L5) Copy the code from HitsplatID into your codebase if you need to reference it.

I have removed the reflection part and hardcoded the ID conversion. 

Also for reference my hour long log file recording with over 9000 log lines came out to 499 KB. Not sure if there's concern about file size / io.

image

SuperNerdEric:
Is there anything else I need to address?

Felanbird:
The queue is a bit backed up at the minute, the newly added plugins merged yesterday were originally opened in December, so you will have to wait a bit longer to get a thorough review.

Nightfirecat:
As a note, [maybe you shouldn't create a combat_log directory when the plugin is initialized; rather, when it is started (via `startUp()`](https://github.com/SuperNerdEric/combat-logger/blob/eb54a9471cada44d902b96b3ed4dadf93f5b057a/src/main/java/com/combatlogger/CombatLoggerPlugin.java#L44) as it currently will create said dir whether it's enabled or not, but that's a minor issue.

PR #5353, opened 2024-01-07
Create pneck-reminder

AbusiveTuna:
simple plugin that highlights pnecks when nothing is being worn in the amulet slot.


Nightfirecat:
FYI, you can use [`ItemContainer#contains()`](https://static.runelite.net/api/runelite-api/net/runelite/api/ItemContainer.html#contains(int)) rather than [looping the Item array from the inventory itemcontainer](https://github.com/AbusiveTuna/PneckReminder/blob/7bd3925a0d417a3d93fd972abaae51f108eff436/src/main/java/com/pneck/PneckReminderPlugin.java#L51).

PR #5366, opened 2024-01-10
Create disable-wearing

AlbeyAl:
Plugin that will block left clicking to wear an item in inventory. A list is set in the configurations, as well as, default enable/disable of plugin and a hotkey.

### DIFF
--- a/plugins/disable-wearing
+++ b/plugins/disable-wearing
@@ -1,0 +1,2 @@
+repository=https://github.com/AlbeyAl/RuneLite-Disable-Wear.git
+commit=1e31e05ae5ce5f2cb2fbf90e6e55ff4ba14be4b2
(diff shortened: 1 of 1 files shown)


raiyni:
You don't need an  executor to update a variable in your plugin.

AlbeyAl:
Thanks for the information. Removed the executors in favor of calling `reset()` itself.

Nightfirecat:
Please `remove your `log.info()` calls](https://github.com/AlbeyAl/RuneLite-Disable-Wear/blob/da379abcaeeb0d90ca89656929d93a9f63aaabd8/src/main/java/com/disablewearing/DisableWearingPlugin.java#L87) or reduce their visibility to debug, as this will clutter user logs.

Nightfirecat:
Closing due to inactivity.

PR #5374, opened 2024-01-12
Maya's Chat Filter

Madam-Herta:
Get rid of spam bots around Oldschool Runescape worlds.
Plugin uses a weekly updated bank of words mostly used by spam bots around the world at Grand Exchange.
It overrides default Chat Filter plugin configuration, but there's a warning popup before it happens.


YvesW:
https://github.com/Psicoses/ChatFilterUpdater already exists and basically does the same. Please contribute to that one, especially because you literally copied that repo (including even the comments) and changed the absolute bare minimum.

Alexsuperfly:
The entire functionality of this is contained in the currently available [Chat Filter Updater plugin](https://github.com/runelite/plugin-hub/blob/master/plugins/chat-filter-updater), please contact that author if you would like to extend it in the future in any way, you can easily be added as a contributor by them in order to collaborate together.  As such this PR will be closed since it adds nothing to the current landscape of the plugin hub.

PR #5383, opened 2024-01-13
Fix Banked Experience for fatlaf

TheStonedTurtle:
Looks like the only issue was a reference to a remove RL ui component


TheStonedTurtle:
My bad my branch was based off master and not flatlaf 🤦

abextm:
this needs to target master since flatlaf was already merged

PR #5392, opened 2024-01-15
add google-form-submitter

umer-rs:
Submits drop screenshots to a Google Form.

### DIFF
--- a/plugins/google-form-submitter
+++ b/plugins/google-form-submitter
@@ -1,2 +1,2 @@
 repository=https://github.com/umer-rs/google-form-submitter.git
-commit=ac7148f4486296f4395123e9ec4ecc791f716b86
+commit=9596acd0340c291d9239847df3a621e1baf804d9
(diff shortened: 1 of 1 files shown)


Nightfirecat:
This plugin always submits requests to imgbb.com, even if its key is empty. Please either add a guard to prevent outgoing requests with the default (empty) configuration, or add a `warning` to the plugin indicating that it will make outbound requests. ([see example](https://github.com/runelite/plugin-hub/blob/0cfd0e2c09b4bd88cc0f58cb9677c0a4f123b872/plugins/neverscapealone#L3))

umer-rs:
> This plugin always submits requests to imgbb.com, even if its key is empty. Please either add a guard to prevent outgoing requests with the default (empty) configuration, or add a `warning` to the plugin indicating that it will make outbound requests. ([see example](https://github.com/runelite/plugin-hub/blob/0cfd0e2c09b4bd88cc0f58cb9677c0a4f123b872/plugins/neverscapealone#L3))

Hi, I have made both changes that you suggested.

PR #5399, opened 2024-01-16
CombatStyleIcons Plugin v1

EnzinGit:
Used for marking combat styles


raiyni:
Can you you change the name of your plugin to be more relevant to what it does? Iconify has a meaning of minimizing to just an icon for programs and I think it would be confusing.

EnzinGit:
> Can you you change the name of your plugin to be more relevant to what it does? Iconify has a meaning of minimizing to just an icon for programs and I think it would be confusing.

Absolutely, shall do that now.

EnzinGit:
> Can you you change the name of your plugin to be more relevant to what it does? Iconify has a meaning of minimizing to just an icon for programs and I think it would be confusing.

Have update it all, hopefully that is a bit better. Please let me know if there are any other problems, cheers!

Nightfirecat:
[Your uses of `ImageIO.read()` are not safe on some launchers.](https://github.com/EnzinGit/CombatStyleIcons/blob/b4b0db4acfe062006c03b105663dc139e474b055/src/main/java/com/combatstyleicons/NpcOverlay.java#L29-L31) (they must be synchronized on `ImageIO.class`) [Please use `ImageUtil.loadImageResource` instead.](https://static.runelite.net/api/runelite-client/net/runelite/client/util/ImageUtil.html#loadImageResource(java.lang.Class,java.lang.String))
Your plugin [registers an overlay on startup](https://github.com/EnzinGit/CombatStyleIcons/blob/b4b0db4acfe062006c03b105663dc139e474b055/src/main/java/com/combatstyleicons/CombatStyleIconsPlugin.java#L47) which is never removed if you were to turn it off. Please ensure it is removed on plugin `shutDown()`.

LGTM otherwise

EnzinGit:
> [Your uses of `ImageIO.read()` are not safe on some launchers.](https://github.com/EnzinGit/CombatStyleIcons/blob/b4b0db4acfe062006c03b105663dc139e474b055/src/main/java/com/combatstyleicons/NpcOverlay.java#L29-L31) (they must be synchronized on `ImageIO.class`) [Please use `ImageUtil.loadImageResource` instead.](https://static.runelite.net/api/runelite-client/net/runelite/client/util/ImageUtil.html#loadImageResource(java.lang.Class,java.lang.String))
> 
> Your plugin [registers an overlay on startup](https://github.com/EnzinGit/CombatStyleIcons/blob/b4b0db4acfe062006c03b105663dc139e474b055/src/main/java/com/combatstyleicons/CombatStyleIconsPlugin.java#L47) which is never removed if you were to turn it off. Please ensure it is removed on plugin `shutDown()`.
> 
> 
> 
> LGTM otherwise

Will sort that tomorrow, appreciate it!

EnzinGit:
I believe I have sorted the problems regarding the loading of Images and loading/unloading the Overlay. Thank you!

PR #5423, opened 2024-01-23
Create GodlikeHttp

bastianjoergensen:
A HTTP server plugin


bastianjoergensen:
> What is this plugin doing? It looks like it just writes events to files locally? If so, you don't need like 85% of this code.

It starts a HTTP server that you can make requests to, to get the players info etc.

But enlighten me, I want to learn. :)

raiyni:
There's already two different plugins that do something similar that might be what you want?

https://github.com/MW90-NL/httpeventserver.git

https://github.com/Adam-/runelite-plugins/tree/http-plugin

bastianjoergensen:
They aren't really the same, I think. I also think theirs is pretty poorly written (not trying to be rude, sorry), and not easily extendable.

bastianjoergensen:
Added the missing license

Nightfirecat:
Why are you [creating a separate EventBus instance](https://github.com/bastianjoergensen/GodlikeHttp/blob/ffc4e99eaba52567643728296a7929ddc7235225/src/main/java/com/godlikehttp/GodlikeHttpPlugin.java#L61) instead of `@Inject`ing it? You can still manually register and unregister classes to the injected instance.

bastianjoergensen:
> Why are you [creating a separate EventBus instance](https://github.com/bastianjoergensen/GodlikeHttp/blob/ffc4e99eaba52567643728296a7929ddc7235225/src/main/java/com/godlikehttp/GodlikeHttpPlugin.java#L61) instead of `@Inject`ing it? You can still manually register and unregister classes to the injected instance.

I simply didn't know, will definitely try that!

Nightfirecat:
Closing due to inactivity. Feel free to re-open this PR if you get around to making the noted changes.

PR #5444, opened 2024-01-27
Update npc-dialogue

leejt:


Alexsuperfly:
CI is failing : `plugin uses terminally deprecated APIs:
Use ComponentID or InterfaceID instead of WidgetInfo or WidgetID`

PR #5447, opened 2024-01-28
Update toa-puzzle-helper

TimOosterom:
* Fixes for compatibility


Felanbird:
Can you also scale down your [baba_wave_spawns.png](https://github.com/TimOosterom/TombsOfAmascutHelper/blob/master/src/main/resources/com/tim/tombshelper/puzzles/baba_wave_spawns.png) file, it's enormous and being shipped with the plugin.

TimOosterom:
> Can you also scale down your [baba_wave_spawns.png](https://github.com/TimOosterom/TombsOfAmascutHelper/blob/master/src/main/resources/com/tim/tombshelper/puzzles/baba_wave_spawns.png) file, it's enormous and being shipped with the plugin.

Good catch, removed it, thank you.

PR #5468, opened 2024-02-01
added big fish probability plugin

sepcol:
Plugin to show the player the probability of receiving a big fish. (only implemented sharks atm)


YvesW:
Don't forget to unregister your overlay. Also please remove some of the ``log.info`` calls. Otherwise reading a log of someone that's fishing sharks would be terrible.

sepcol:
> Don't forget to unregister your overlay. Also please remove some of the `log.info` calls. Otherwise reading a log of someone that's fishing sharks would be terrible.

Fixed PR comments

PR #5480, opened 2024-02-04
inner-zoom

Haftago:
Allows you to adjust how far you can zoom in while allowing you to set a further away default


Haftago (commented):
I believe I fixed the changes asked.

PR #5484, opened 2024-02-04
package: update actions to node20

Felanbird:
Actions using node16 are [DEPRECATED!!!](https://github.blog/changelog/2023-09-22-github-actions-transitioning-from-node-16-to-node-20/) so I went ahead and bumped them all to their node20 versions

We've been using these in Dink since whenever dependabot thought we should have them, and so far they have not killed us

also should `CACHE_VERSION` be changed to `2.3.2` or `2.4.0`, I didn't change it yet

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -22,14 +22,14 @@ jobs:
     if: github.event_name != 'push' || github.repository_owner == 'runelite'
     runs-on: ubuntu-20.04
     steps:
-    - uses: actions/checkout@v3
+    - uses: actions/checkout@v4
       with:
         fetch-depth: 0
-    - uses: actions/setup-java@v3
+    - uses: actions/setup-java@v4
       with:
         distribution: adopt
         java-version: 11
-    - uses: actions/cache@v3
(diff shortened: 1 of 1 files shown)


Nightfirecat:
[Based on this commit](https://github.com/Felanbird/runelite-plugin-hub/commit/bd0b598987b6c086a23388842a0287863ba5131b), use `CACHE_VERSION=2.4.0`.

Nightfirecat:
Most of these action updates seem fairly innocent. The only ones I see being potentially problematic are upload-artifact and download-artifact because artifacts created on v3 are not necessarily compatible with v4. ([1](https://github.com/actions/upload-artifact/releases/tag/v4.0.0), [2](https://github.com/actions/download-artifact/releases/tag/v4.0.0), [3](https://github.com/actions/upload-artifact/blob/main/docs/MIGRATION.md))

abextm:
the cache version doesn't need to be changed here. I have this staged with some other changes I need to finish validating

abextm:
360b9daa7a134eafc9a0c5c2ec2015e72f1825e9

PR #5485, opened 2024-02-04
Create tickcounter

Pojken-Arn:
Adds a visual customisable tick counter.
Can help keep track of ticks and help you with prayer-flicking.

### DIFF
--- a/plugins/customticks
+++ b/plugins/customticks
@@ -1,2 +1,2 @@
 repository=https://github.com/Pojken-Arn/tick-flick
-commit=217f55c6a6bbfe7a44d6eea458cf8b9919b13581
+commit=7270f848dc247a68315c8aa4010b8035f81f90a6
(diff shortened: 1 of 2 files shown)


YvesW:
Any reason you did not contribute to https://github.com/NathanQuayle/visual-ticks ?

Pojken-Arn:
@YvesW Well I don't mind contributing to https://github.com/NathanQuayle/visual-ticks, I just had some ideas for features and more customisble settings and didn't wanna clutter an already existing plugin, espacially when I hope to implements some more features in the near future aswell. If you think it would be better for me to contribute an already exsisting plugin then I don' t mind it.

YvesW:
If possible, we generally recommend people to contribute to already existing plugins as to not end up with the 10 discord plugins situation that we have right now, and because that plugin already has a userbase. Thus, more people might use your improvements. It's just that you might have to be a bit more conservative with enabling stuff by default if you already got a bigger install base.

Pojken-Arn:
@YvesW Okay sounds good 👍

PR #5498, opened 2024-02-07
Update combat-logger to v0.0.3

SuperNerdEric:
- Send reminder message so players don't forget they are logging
- Log combat stat boosts
- Log equipment


SuperNerdEric:
I took the opportunity to also:

- Log splashes
- Log the hitsplat number when it is unknown

PR #5508, opened 2024-02-10
Update toa-mistake-tracker to v1.5

QuestingPet:
Fix visual glitch with new sidebar panel

### DIFF
--- a/plugins/toa-mistake-tracker
+++ b/plugins/toa-mistake-tracker
@@ -1,2 +1,2 @@
 repository=https://github.com/QuestingPet/ToaMistakeTracker.git
-commit=8a1b1d809f204ea9630f3664b43581c4ff75a518
+commit=f82a273fa40b150d9dd3a36ad55123cac058646b
--- a/plugins/toa-mistake-tracker
+++ b/plugins/toa-mistake-tracker
@@ -1,2 +1,2 @@
 repository=https://github.com/QuestingPet/ToaMistakeTracker.git
-commit=f82a273fa40b150d9dd3a36ad55123cac058646b
+commit=c73766d005af28741be13dee635723f12a41815b


Adam-:
Is `com/toamistaketracker/crondis-example.png` required at its full resolution? It is almost 2mb.

QuestingPet:
> Is com/toamistaketracker/crondis-example.png required at its full resolution? It is almost 2mb.

Where are you seeing that? I'm seeing it as 219 KB both locally and in the GitHub repo. In any case, it's used in the readme as an example/teaser of the plugin. It's not *required* to be of any size, and can be changed if needed, though probably can be separate from this fix I'd think.

Adam-:
Java explodes images to their full resolution when you load it into memory, so it is 1,888,224 bytes (length * width * 4) in memory

QuestingPet:
I can crop it to 690 x 380. Which puts it down to 1MB. Does that work, or do you require a new (lower-res) photo altogether?

CreativeTechGuy:
@QuestingPet I highly recommend running all of your images through [Squoosh](https://squoosh.app/). You can easily compress your images to 30% of the size (or smaller) without any noticeable visual quality loss. You just need to play around with the compression settings and visually compare to find what works best for each image. Just the default settings (MozJPEG, 75% quality) for the `crondis-example.png` took it down to 32kb. If you also resize it you could easily get below 25kb.

Nightfirecat:
@CreativeTechGuy We're not talking about compressed size, we're strictly talking about image dimensions, as that is what causes large in-memory size when loaded by Java.

For the image Adam referenced, it looks like it's only use is to be shown in the README. In that case, you can remove it from your resources directory (you could move it to the project root, or something outside of the `src/` directory) and it would be fine as it would no longer be shipped in the jar.

QuestingPet:
Are we blocking merging this bug-fix on improving the image bloat (which is not modified in this PR)? Does everything else look fine?

Felanbird:
I'm not sure if Adam looked at the actual code you changed when he made his request, but yes we would like you to make the change to the existing file.

Nightfirecat:
Yes, we would like the image size issue addressed. We ask that hub plugin authors update their plugins in line with review guidelines at the time of review.

QuestingPet:
I have moved the README photos out of `src/`

PR #5510, opened 2024-02-10
Update collection-log-luck with visual overlay

peanubnutter:
Check out the updated demo here: https://github.com/peanubnutter/collection-log-luck


peanubnutter:
In case the code is confusing, here's an explanation: calculating luck is potentially expensive (a few milliseconds or more in rare cases), so we can't afford to do this on the game thread every frame. Instead, the first frame a calculation request is issued. On subsequent frames, no overlay is shown until the result is available. Then, the cached result is used from them on.

PR #5520, opened 2024-02-13
Create Ruby counter

hyppis1:
This plugin tracks your attacks, procs and rate of ruby bolt procs.
There is options to separate Armadyl/Zaryte crossbow special attacks from the normal counter.
You must have sounds enabled for the plugin to track ruby bolt procs.

There is 2 commits by accident. I noticed a small typo when i was making 1st commit and had to replace it. So the newer is the correct one


hyppis1:
Okay took few more attempts than just 2 commits, was doing some pointless mistakes during commits and naming

PR #5539, opened 2024-02-17
Create Clan-data-exporter

HudRud:
This plugin exports clan data, like names, join dates and ranks, to .csv format.


cdfisher:
What's the difference between this and existing plugins such as "Clan List Exporter", "Clanmate Export", "Copy clan usernames", etc.?

HudRud:
> What's the difference between this and existing plugins such as "Clan List Exporter", "Clanmate Export", "Copy clan usernames", etc.?
Clan List Exporter won't let you select the save location, same with clanmate export. Copy clan usernames copies clan usernames.
Im aiming to create a plugin that has more functionality over the others by making it easier to select  what data you want.

raiyni:
We'd prefer collaboration with another plugin over another new plugin. Especially if their feature set is the same.

PR #5541, opened 2024-02-17
Add RBW Bitwarden Authenticator Plugin

nea89o:
This plugin adds the ability to pull passwords from bitwarden using the [`rbw`](https://github.com/doy/rbw) CLI tool. That tool has the advantage of having a daemon over the regular bitwarden tool. This means the login into your password manager does not happen via a session id (which would need to be stored in your
    settings or in your .profile).

### DIFF
--- a/plugins/rbw-bitwarden
+++ b/plugins/rbw-bitwarden
@@ -1,0 +1,3 @@
+repository=https://github.com/nea89o/rbw-bitwarden-runelite.git
+commit=c888c5569f8573a93d9e57555d94ddadd909c008
+


YvesW:
1. Does this plugin offer additional benefits over the already existing Bitwarden plugin?
2. If so, is there any reason you did not contribute to that plugin?

nea89o:
> 1. Does this plugin offer additional benefits over the already existing Bitwarden plugin? Is the `regular tool` you are talking about the already existing plugin, or is it something else?

There are two bitwarden clients out there nowadays. The official bitwarden cli [bw](https://github.com/bitwarden/cli) and [rbw](https://github.com/doy/rbw). `bw` is stateless, meaning that in order for any command to work you need to set an environment variable containing your session key. `rbw` has a daemon, meaning that at first startup/first usage you get asked for a password and the daemon remembers your session key for the next (by default) 1 hour.

This also shows in the way the plugins are used. The official `bw` tool needs the user to set the session key in the settings for the tool, while `rbw` will either already have a daemon running or will fork a daemon that then prompts the user for the password. The state is just managed in different places (either the plugin, or the actual bitwarden client).

> 2. If so, is there any reason you did not contribute to that plugin?

Because of the different ways state is handled means the actual design of the plugin is practically orthogonal. `rbw` basically involves invoking a single command, while `bw` requires complex management of state. There are for example routines for asking for the session key on every startup, something that is pointless with `rbw`, since `rbw` does this itself.

In theory I could of course add one master switch to the existing bitwarden plugin, but that would be pointless in terms of code sharing, and it would also involve me meddling with a code base that hasn't had a commit in 3 years.

I also wrote this for myself in hopes of not having to involve myself in open source shenanigans, but sadly runelite does not support sideloading. This is a choice that was made which means that slightly different forks of plugins will exist.

> 3. Is this unofficial bitwarden cli safe to use?

Depends on your threat model, i guess? It is of course unofficial, but also wildly trusted and audited and to my knowledge it doesn't pull in any remote updates that would make a hostile take over easy. If this means anything, it is available in the *official* arch repo: https://archlinux.org/packages/extra/x86_64/rbw/ (not even the AUR).

raiyni:
> it would also involve me meddling with a code base that hasn't had a commit in 3 years.
> I also wrote this for myself in hopes of not having to involve myself in open source shenanigans

Not sure how this justifies anything as the code works and hasn't needed change, hasn't had outside contribution, and I've been active.

With that said, I'm not sure I would want to support this in a plugin that promotes the official cli anyway. I'm struggling to find information about the safety of the library outside of the GitHub page and I'm not going to use this as an excuse to read rust. It looks like the project has been around for a few years, so it has that going for it. If there was a decent user base of need, I'd think about it but this is the first time I've heard about it.

nea89o:
> Not sure how this justifies anything as the code works and hasn't needed change, hasn't had outside contribution, and I've been active.

I'm not saying your codebase is bad. (It probably isn't, in fact I haven't looked into it much at all). But in general, I am much more comfortable contributing to a project with weekly or monthly commits than i am bothering someone who seems to have been happy enough with their product for a few years.

Nightfirecat:
We're concerned both that this would be a plugin which (a) handles user credentials, built by someone not known or trusted by the team, and (b) is not using the official password manager tool. Neither of those points, nor both together, is to say that we suspect this plugin is malicious or insecure, however we do not want to add this plugin to the hub at this time, sorry.

PR #5544, opened 2024-02-18
Create naturalspeech

phyce:
Naturalspeech is a plugin that uses an advanced text-to-speech engine to generate sound with different voices.

### DIFF
--- a/plugins/naturalspeech
+++ b/plugins/naturalspeech
@@ -1,3 +1,3 @@
 repository=https://github.com/phyce/rl-natural-speech.git
 commit=45e124a46608daccc27a3a397a56aa61bc2f22e0
-warning=This plugin downloads voice model files from external services.
+warning=This plugin downloads voice model files from, (and sends your IP address in the process, to) huggingface.co a 3rd party website not controlled or verified by the RuneLite Developers.
--- a/plugins/naturalspeech
+++ b/plugins/naturalspeech
@@ -1,2 +1,2 @@
 repository=https://github.com/phyce/rl-natural-speech.git
-commit=fef12d35a587764289858b6d42649ec7a47d3776
+commit=3098e0db31a860045228f83cfbac740d55a391e3
--- a/plugins/naturalspeech
+++ b/plugins/naturalspeech
@@ -1,2 +1,2 @@
 repository=https://github.com/phyce/rl-natural-speech.git
-commit=18055dfdc6c66723e80924f00fe269d269a0ab76
+commit=fef12d35a587764289858b6d42649ec7a47d3776
--- a/plugins/naturalspeech
+++ b/plugins/naturalspeech
@@ -1,2 +1,3 @@
 repository=https://github.com/phyce/rl-natural-speech.git
 commit=45e124a46608daccc27a3a397a56aa61bc2f22e0
+warning=This plugin downloads voice model files from external services.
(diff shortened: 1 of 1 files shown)


Nightfirecat:
Your plugin needs a `warning` [as it automatically connects to external services with the default configuration](https://github.com/phyce/rl-natural-speech/blob/fef12d35a587764289858b6d42649ec7a47d3776/src/main/java/dev/phyce/naturalspeech/tts/DownloadManager.java#L19).

phyce:
@Nightfirecat warning message added, along with a bunch of other updates. Please let me know if the plugin needs any other amendments.

raiyni:
The warning message goes in the file you created in this pr.

phyce:
@raiyni  Ohhh, my bad. I thought it was in the plugin info. Hopefully I've done it right this time.

plugins/naturalspeech
```diff
@@ -0,0 +1,3 @@
+repository=https://github.com/phyce/rl-natural-speech.git
+commit=45e124a46608daccc27a3a397a56aa61bc2f22e0
+warning=This plugin downloads voice model files from external services.
```
Felanbird, line 3:
```suggestion
warning=This plugin downloads voice model files from huggingface.co, a 3rd party website not controlled or verified by the RuneLite Developers.
```
basically all the warnings follow the same [structure](https://github.com/search?q=repo%3Arunelite%2Fplugin-hub+warning%3D&type=code)

Nightfirecat: As felanbird said, please use the same format used by most/all other plugins with warning fields. Additionally, please do not exclude the warning about IP addresses being sent. osrs users are paranoid, and may be upset to learn this some other way.

phyce: Yeah, that is absolutely fair. I've commited the suggestion.
I am looking at traing custom models eventually, and might end up just hosting those along with these all on a repo on github. If the models are hosted in github will it still need the warning?

Nightfirecat: No, there's a short list of servers/services we trust, and GitHub is one of them.

Nightfirecat:
You're registering several classes to the event bus (at a glance, TopLevelPanel, MenuEventHandler, and SpeechEventHandler) which don't seem to be unregistered on plugin shutdown. This means that turning your plugin off potentially does nothing, as those classes would continue to receive and respond to fired events. Please ensure they are unregistered on plugin shutdown.

plugins/naturalspeech
```diff
@@ -0,0 +1,3 @@
+repository=https://github.com/phyce/rl-natural-speech.git
+commit=7d7371bf03200c5ca1155b6e5e6a266ca6283a4f
+warning=This plugin downloads voice model files from, (and sends your IP address in the process, to) huggingface.co a 3rd party website not controlled or verified by the RuneLite Developers.
```
Nightfirecat, line 3:
```suggestion
warning=This plugin downloads voice model files from, (and sends your IP address in the process, to) huggingface.co, a 3rd party website not controlled or verified by the RuneLite Developers.
```

TheLouisHong:
> You're registering several classes to the event bus (at a glance, TopLevelPanel, MenuEventHandler, and SpeechEventHandler) which don't seem to be unregistered on plugin shutdown. This means that turning your plugin off potentially does nothing, as those classes would continue to receive and respond to fired events. Please ensure they are unregistered on plugin shutdown.

I believe I resolved this with this commit: https://github.com/phyce/rl-natural-speech/commit/cba9dc9b77ec51dfad0c7d41f7c717ae8a9803ff

- TopLevelPanel no longer registers tab panels.
- EventBus::register/unregister calls are restricted to NaturalSpeechPlugin, managing handlers on startup() and shutdown() is clear and easy to spot.

PR #5538, opened 2024-02-17
New Borderless Fullscreen Plugin

TheLouisHong:
<p align="center"></p> 

---

- Borderless Fullscreen
- Multiple Monitors Support
- Hotkey Toggle

# Demo
https://github.com/TheLouisHong/Borderless-Fullscreen/assets/1916460/42e02982-1ac3-4e9c-9182-1a3aa94ba59d

https://github.com/TheLouisHong/Borderless-Fullscreen/assets/1916460/479e03e5-17ab-41f6-8de6-db3a48851442

PR #5555, opened 2024-02-21
Update borderless-fullscreen - Added "custom chrome" warning 

TheLouisHong:
Borderless-fullscreen now warns users when attempting fullscreen with custom chrome.

image


DominickCobb-rs:
https://github.com/runelite/plugin-hub/pull/5538 is open and unmerged. I believe you should be pushing new commits to that branch and not having multiple PRs open for the same plugin.

PR #5560, opened 2024-02-21
Create Party Special Tracker

1Defence:
Customizable view of party members special information

- Color change on chosen special threshold
- Filter specific names to only see visuals of 1 or 2 players as opposed to 20
- Optional timer to know how long it's been since a player has used special
- Location offset and conditional text settings

https://github.com/runelite/plugin-hub/assets/63735241/71c682bb-cadd-4d55-beaa-9e9562ffbf40

### DIFF
--- a/plugins/party-special-tracker
+++ b/plugins/party-special-tracker
@@ -1,2 +1,2 @@
 repository=https://github.com/1Defence/party-special-tracker.git
-commit=1d8942bfd54bd8f996c699411a75ab3f509a0962
+commit=2b36b5148248ea5a89f05cab5898a6d6f08abd6d
--- a/plugins/party-special-tracker
+++ b/plugins/party-special-tracker
@@ -1,2 +1,2 @@
 repository=https://github.com/1Defence/party-special-tracker.git
-commit=b947dd6ef17881fd86629fbd4bc1dcf2acd6df5c
+commit=1d8942bfd54bd8f996c699411a75ab3f509a0962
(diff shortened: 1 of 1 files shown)


1Defence:
Updated this to use standard StatusUpdate packet.

Nightfirecat:
I'm curious what the conditions are where this is needed: https://github.com/1Defence/party-special-tracker/blob/2b36b5148248ea5a89f05cab5898a6d6f08abd6d/src/main/java/com/partyspecialtracker/PartySpecialTrackerFringeConditionUpdate.java. If it's something which should be fixed in core, we'd appreciate being made aware of it.

1Defence:
> I'm curious what the conditions are where this is needed: https://github.com/1Defence/party-special-tracker/blob/2b36b5148248ea5a89f05cab5898a6d6f08abd6d/src/main/java/com/partyspecialtracker/PartySpecialTrackerFringeConditionUpdate.java. If it's something which should be fixed in core, we'd appreciate being made aware of it.

Sure, if you lose special energy on the same tick that you receive an energy transfer you receive 2 special change varbits in the same tick;
So, you can for example go from 50%->100%->50% in the same tick, you also don't need to start with spec for that to occur you can similarly go from 0%-100%-0% in the same tick.

If you're tracking locally this likely isn't a problem as you can set a flag in the varbit but as I'm going off party data party doesn't see a change so no update occurs.

PR #5566, opened 2024-02-22
Move profiles repo from spedwards to jzomdev

JZomDev:
I did try to PR this and contact spedwards but they seem to be inactive. 

image

Add RuneLite's profile switching feature to trigger on selecting this plugin's panel rsns

Not sure how to make it show less of a diff while also being under my repo's branch structure.


Alexsuperfly:
@Spedwards are you available? By April, if this is not responded to by you requesting otherwise, ownership will be transferred.

PR #5568, opened 2024-02-22
Create pest-control-points

UniqueName992:
Tracks and Displays your Pest Control Points on-screen whilst in a Pest Control Game or on the Void Island.

### DIFF
--- a/plugins/pest-control-points
+++ b/plugins/pest-control-points
@@ -1,0 +1,2 @@
+repository=https://github.com/UniqueName992/runelite-pest-points.git
+commit=880816aa75e422617e7e95453a266df42c3940dc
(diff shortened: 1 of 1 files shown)


Nightfirecat:
[Do not use "example" as your config group.](https://github.com/UniqueName992/runelite-pest-points/blob/2427eb5b47c579ec7019330330505406852dc92b/src/main/java/com/pestPoints/pestPointsConfig.java#L7) Use something more specific to your plugin, such as "pest-control-points".
LGTM otherwise.

UniqueName992:
Thanks for the review, new [push amends the config group](https://github.com/UniqueName992/runelite-pest-points/blob/36809c3db7de9afed2b92ddc2907d2714bceb9d7/src/main/java/com/pestPoints/pestPointsConfig.java#L7C1-L7C36)

PR #5579, opened 2024-02-24
Create easylootsell

TarrenG:
Highlights items with more than 0 qty in a specific bank tab for easy identification while selling off a loot tab to prevent time wasted trying to track down the last few items hidden between placeholders.

### DIFF
--- a/plugins/easylootsell
+++ b/plugins/easylootsell
@@ -1,0 +1,2 @@
+repository=https://github.com/TarrenG/easylootsell.git
+commit=85f1d55abc374dd69924bc00546d9542b0db7068
--- a/plugins/easylootsell
+++ b/plugins/easylootsell
@@ -1,2 +1,2 @@
 repository=https://github.com/TarrenG/easylootsell.git
-commit=85f1d55abc374dd69924bc00546d9542b0db7068
+commit=d3004325b8add7086813ff873257f3de24e64ba2


raiyni:
Does searching by  value not cover this?

raiyni:
You can also have config keys with the Color class.

TarrenG:
> Does searching by value not cover this?

searching by value is for all items in the bank, this is specifically for a designated loot tab within the bank, to avoid bringing up items not wanting to be sold when searching for a something like ">500k" for example. I keep a bank tab with all the placeholders for items I deem as "junk to sell" and sometimes its hard to see which items are actually there compared to the slightly faded placeholders, so I wanted to highlight them to stand out.

raiyni:
I believe the bank heatmap plugin would serve this purpose then?

TarrenG:
The bank heatmap plugin appears to be similar however, the two main differences are that it is for the entire bank, not a single bank tab and the color highlights are automatically determined based off value. I would prefer to avoid coloring my entire bank, and its easier to identify straggling items in the loot tab if I know exactly what color to look for, such as a bright yellow or green outline.

TarrenG:
It displays the overlay in every tab you click on unless you turn it back off each time you don't want to use it, instead of only highlighting items in a specific tab, as well as being unable to set a single color as the highlight color.

PR #5584, opened 2024-02-24
Updated Advanced Mining v1.0.2

a-h-designs:
Fixed an error where it wasn't displaying gems collected.


cdfisher:
Don't keep closing PRs and making new ones, just push additional commits to the same branch.

a-h-designs:
Sorry it was only once and it was the only way I could see to do it at the time

PR #5592, opened 2024-02-25
Projectile swapper

SirAnomen:
This plugin swaps projectile visuals.

Similar to the existing Vardorvis Projectiles plugin, but more generalized.


Nightfirecat:
After some discussion, we decided we would not like to add this functionality, and have [added an entry to our Rejected or Rolled Back Features wiki page](https://github.com/runelite/runelite/wiki/Rejected-or-Rolled-Back-Features) covering this, and similar features:

We would welcome a more restricted plugin (like is mentioned) but we've found that open-ended ID user inputs often cause problems for our moderation.

PR #5617, opened 2024-03-02
Wintertodt idle

ryansand20:
A Wintertodt Idle plugin inspired by Wilderness Player Alarm.
Super simple, super visible overlay for playing on a second monitor.

I tried every existing Wintertodt plugin, and none of them are visible enough to avoid losing time.
RuneLite needs this plugin!


Nightfirecat:
Please do not close and open new PRs; you can simply push new commits to your branch instead

ryansand20:
> Please do not close and open new PRs; you can simply push new commits to your branch instead

Apologies. Won't happen again.

PR #5624, opened 2024-03-03
DropTracker re-write

joelhalen:
Re-wrote the entire plugin to remove lots of excess code and messy logic.
Used a lot of code to rewrite from [Adam's Discord Loot Logger](https://runelite.net/plugin-hub/show/discord-loot-logger).

All players' drops are now tracked via our Discord webhook system, and players can enable their own "discord loot logger"-style webhook notifications for drops if they don't have a clan or don't want to use the Discord bot.

Players can sign up for events in our Discord server, which are centered around drops/pvm content.

Implemented a new side panel with enhanced functionality and much better management of thread usage to prevent latency issues that many users were reporting with the API enabled

### DIFF
--- a/plugins/droptracker
+++ b/plugins/droptracker
@@ -1,2 +1,2 @@
 repository=https://github.com/joelhalen/droptracker-plugin.git
-commit=1dafa1742831aa44fbd46d020337674c6da1c8d8
+commit=e8c14fc151ef844e8377cdecf01bd20df83947fc
--- a/plugins/droptracker
+++ b/plugins/droptracker
@@ -1,2 +1,2 @@
 repository=https://github.com/joelhalen/droptracker-plugin.git
-commit=e8c14fc151ef844e8377cdecf01bd20df83947fc
+commit=86157d364618463ec65dadd04015cb37ad20c90b
--- a/plugins/droptracker
+++ b/plugins/droptracker
@@ -1,2 +1,2 @@
 repository=https://github.com/joelhalen/droptracker-plugin.git
-commit=7d1f67fc6c94e3ed83704140f8c1e62e47ef2a74
+commit=8aed64b51aa42933ae5bfb3ca9b10b832c90d471
--- a/plugins/droptracker
+++ b/plugins/droptracker
@@ -1,2 +1,2 @@
 repository=https://github.com/joelhalen/droptracker-plugin.git
-commit=0f05faead8821b3cf57fa5e6cbff3e8a30391f66
+commit=aafdaf9854f3113bdd3607e3ea202ec05085768d
--- a/plugins/droptracker
+++ b/plugins/droptracker
@@ -1,2 +1,2 @@
 repository=https://github.com/joelhalen/droptracker-plugin.git
-commit=7de61b713043e9819db91e4b752912a60edd718f
+commit=6187c7293bf1e729f5de5b2e704be3c9291c4484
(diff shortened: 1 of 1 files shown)


Nightfirecat:
[You can't use null.com as an "invalid" domain](https://github.com/joelhalen/droptracker-plugin/blob/e8c14fc151ef844e8377cdecf01bd20df83947fc/src/main/java/io/droptracker/api/DropTrackerApi.java#L58); <https://null.com> is a real website. Please ensure your plugin is not sending unwanted outbound connections when your config is not set up to do so.
The [chat commands you're registering in your plugin](https://github.com/joelhalen/droptracker-plugin/blob/e8c14fc151ef844e8377cdecf01bd20df83947fc/src/main/java/io/droptracker/DropTrackerPlugin.java#L142) are never unregistered; please do so in your plugin's shutDown method.
Your [plugin panel is registered to the event bus on init](https://github.com/joelhalen/droptracker-plugin/blob/e8c14fc151ef844e8377cdecf01bd20df83947fc/src/main/java/io/droptracker/ui/DropTrackerPanel.java#L183) but is never unregistered. (`deinit()` is never called) Please ensure it is unregistered when it is removed and/or on plugin shutDown.

joelhalen:
> [You can't use null.com as an "invalid" domain](https://github.com/joelhalen/droptracker-plugin/blob/e8c14fc151ef844e8377cdecf01bd20df83947fc/src/main/java/io/droptracker/api/DropTrackerApi.java#L58); https://null.com is a real website. Please ensure your plugin is not sending unwanted outbound connections when your config is not set up to do so. The [chat commands you're registering in your plugin](https://github.com/joelhalen/droptracker-plugin/blob/e8c14fc151ef844e8377cdecf01bd20df83947fc/src/main/java/io/droptracker/DropTrackerPlugin.java#L142) are never unregistered; please do so in your plugin's shutDown method. Your [plugin panel is registered to the event bus on init](https://github.com/joelhalen/droptracker-plugin/blob/e8c14fc151ef844e8377cdecf01bd20df83947fc/src/main/java/io/droptracker/ui/DropTrackerPanel.java#L183) but is never unregistered. (`deinit()` is never called) Please ensure it is unregistered when it is removed and/or on plugin shutDown.

Apologies for that oversight; should be fixed. I've defaulted the API to use an empty url and made sure the commands and side panel are both removed on shutdown.

Thanks for your time

Nightfirecat:
[You were already removing your plugin's nav button on shutDown](https://github.com/joelhalen/droptracker-plugin/blob/86157d364618463ec65dadd04015cb37ad20c90b/src/main/java/io/droptracker/DropTrackerPlugin.java#L211), however you are never calling `panel.deinit()`, so it remains subscribed to the event bus.

joelhalen:
> [You were already removing your plugin's nav button on shutDown](https://github.com/joelhalen/droptracker-plugin/blob/86157d364618463ec65dadd04015cb37ad20c90b/src/main/java/io/droptracker/DropTrackerPlugin.java#L211), however you are never calling `panel.deinit()`, so it remains subscribed to the event bus.

Ah alright; took care of that. Thanks for your patience.

Nightfirecat:
[This will cause an NPE](https://github.com/joelhalen/droptracker-plugin/blob/68a57d5575555c07f3568626186db91308ac1d20/src/main/java/io/droptracker/DropTrackerPlugin.java#L215-L216) (because you nulled it the line before) and not deinit it.

PR #5631, opened 2024-03-04
Create displaynamedisguiser

YaIPartake:
A plugin that disguises a player's display name based on config settings.
Allows you to change the display name of other players too.
This is a modified version of the Fake RSN plugin.


YaIPartake:
The most recent commit should fix the Build issue with WidgetInfo being used instead of ComponentID.

Nightfirecat:
Changing the displayed name of other players (to specific other names) is not something we're interested in adding, as that could (and likely would) be used to aid in scamming or falsely implicating other players of things they didn't do. Obfuscating their names, however, would be fine. (Perhaps a fixed obfuscation cipher shouldn't be used--instead generated on first run, or on each launch?)

YaIPartake:
Would it help if there was a visual indicator like an icon for instances where another player's name is changed? I could make it look like an incognito browser icon or a silly mustache glasses to remove the sense of authenticity.

I'll make the cipher randomized along with enabling the encrypt function for users. I'm just not certain how to prompt the user to execute it yet.

YvesW:
Please remove ``enabledByDefault = false``. It's confusing for users to install a plugin from the hub which isn't enabled right away.

PR #5642, opened 2024-03-06
Create csvexport

fl4xman:
This plugin gives the ability to export key information out of the game via tabular formatted csv files for non technical users to be used for analysis in Excel/Tableau/PowerBI.

Useful for informative data analytics, account progress capturing and generally fun pet projects.

### DIFF
--- a/plugins/csvexport
+++ b/plugins/csvexport
@@ -1,0 +1,2 @@
+repository=https://github.com/fl4xman/csvexport.git
+commit=9844963d90bdd0c6af3beb7e732947327753f391


raiyni:
1. Please stop trying to submit this code as noted in previous PRs
2. The execution of writing all of this data each tick to a flat file is horrendous
3. Using a different user to submit the project isn't going to help

Nightfirecat:
Regarding http plugins, what exactly do you need from one which isn't already implemented in one of the existing plugins? I know not every http plugin author is inactive so I'd definitely recommend reaching out via email/Discord/other means to those authors and seeing if they would be willing to work with you to add whatever missing feature you're interested in. If that all fails, we're willing to consider adding another, but I don't know what could be missing that is so crucial.

PR #5646, opened 2024-03-08
Add party-games plugin

Maurits825:
Add party games plugin. This plugin uses the Party plugin to allow party members to play games. Current games include rock paper scissors and coin flip.

### DIFF
--- a/plugins/party-games
+++ b/plugins/party-games
@@ -1,0 +1,2 @@
+repository=https://github.com/Maurits825/party-games.git
+commit=996a9a8de407c8cf10bc3a4698b9c050e12e30da


Maurits825:
> 1. This plugin is an abuse of our party service that wasn't intended for this kind of usage

Is the idea of the plugin an abuse or the implementation itself? I had a look at some other plugins like the [tag ur it](https://runelite.net/plugin-hub/show/tag-ur-it) that also uses the party service to send messages.
> 2. These games seem like they are nearing gamba territory

Is it the rock paper scissors, the coin flip or both? The initial idea of the plugin was to play rock paper scissors to decide who runs the head at olm.

Nightfirecat:
> Is the idea of the plugin an abuse or the implementation itself? I had a look at some other plugins like the tag ur it that also uses the party service to send messages.

A bit of both. We're not really interested in supporting plugins which have no ingame behavior to use the party service. This plugin is effectively just a Swing UI which reads chat messages, which doesn't show any overlays or otherwise interface with the game itself.

Maurits825:
Thats fair, I might re-work the plugin in the future to not use the party service, ill close the pr for now.

PR #5655, opened 2024-03-09
Update taskjingle

maximonster:
Updated plugin with a command to test custom sounds and error messages when sounds are not correctly setup


Nightfirecat:
You must unregister your chat command when the plugin is disabled.

PR #5656, opened 2024-03-09
Vorkath kicker

K0BIE:
A simple plugin that changes the poke vorkath option to a kick vorkath option. Also changes the animation that plays.


raiyni:
The boss kicker plugin already exists.

Nightfirecat:
Like raiyni said, this seems to behave the same as the Boss Kicker plugin. Given that, I'm closing this PR.

PR #5659, opened 2024-03-10
Updated Advanced Mining to v1.0.5

a-h-designs:
Temporarily removed Dense Essence until fixed properly as is was showing up no matter what you mined.


Nightfirecat:
[FYI, you can add a plugin conflict to your `@PluginDescriptor` instead of instructing users to disable the original plugin.](https://github.com/a-h-designs/Advanced-Mining/blob/c015ca41c69e9329aea2e30e9ceb1044f6da8182/README.md?plain=1#L11) That would certainly be a better user experience.

PR #5667, opened 2024-03-13
Update commit hash of rc-pouch

LameGames:
- Updated the commit hash to the latest master of the rc-pouch-usage plugin.

Updates since last hash inclue:

- Fixed counter not working when Bank window is not open;
- Improved onItemContainerChanged to prevent filling multiple pouches simultaneously from being treated as only filling the last-clicked pouch with the total stored essence;
- Removed ZMI-related stuff since plugin should not be dependent on being near ZMI now


Felanbird:
We don't accept updates from non-authors, have you tried contacting the author to update this plugin?

PR #5675, opened 2024-03-14
Create spec-jingle plugin

Samstiles:
Plays a sound effect at arbitrary configurable spec regen thresholds.

Useful for if you want to be audibly notified that you are able to spec with [your spec weapon of choice] when your spec regens when you may not be looking.


Nightfirecat:
It looks like this plugin just doesn't work at all the way it says it does? This code will play a jingle for any spec energy change which leaves you with above 0 spec energy: https://github.com/blert-io/spec-jingle/blob/d7ae00d0480410484c8dcbb4da6d740edbd61d2d/src/main/java/com/blert/spec_jingle/SpecJinglePlugin.java#L60-L65

PR #5679, opened 2024-03-14
Update Milestone levels to v1.0.2

Antimated:
- Added multi-level support; leveling from level 1 to 20 at once will now trigger notifications for level 10 and 20 (if those are set in your levels list).
- Notifications now check if the notification interface is not opened before displaying it.
- General cleanup of code


Adam-:
We do not allow reflection in pluginhub plugins. Update your plugin to remove reflection.

Antimated:
> We do not allow reflection in pluginhub plugins. Update your plugin to remove reflection.

Fixed and pushed

PR #5684, opened 2024-03-15
Update inferno-stats to v2.1.5

InfernoStats:
Add support for the blazing blowpipe animation to the Tick Loss handler


Felanbird:
Is [blob-square.png](https://github.com/InfernoStats/InfernoStats/blob/master/src/main/resources/blob-square.png) still needed/used anywhere? It's very large and Java blows it up to it's max size when it loads it into memory. (width * height * 4)

InfernoStats:
> Is [blob-square.png](https://github.com/InfernoStats/InfernoStats/blob/master/src/main/resources/blob-square.png) still needed/used anywhere? It's very large and Java blows it up to it's max size when it loads it into memory. (width * height * 4)

It is, but I can resize it with this PR

PR #5685, opened 2024-03-15
Update milestone levels to v1.0.4

Antimated:
- Removed setting to enable multi-levels, this is now default behaviour.
- Added support for virtual levels


Antimated:
Is there a reason why v1.0.3 doesnt have a checkmark behind it?

Felanbird:
> Is there a reason why v1.0.3 doesnt have a checkmark behind it?

Yes because Github doesn't run actions on every commit, unless they are all pushed separately. You pushed that commit along with the commit updating the branch with master changes, the exact same thing happened with your first branch update.

image


image

Felanbird:
As long as the most recent commit is ✔️ then yes there's nothing from the workflow check that would prevent merge

PR #5688, opened 2024-03-16
New plugin - Tostky Tracker. 

example-git:
Tostky Tracker is a plugin that adds an overlay that keeps track of all the ammo in your inventory and a panel that tracks total usage of ammo overall- Including cannonballs over time with the ability to reset the counter on demand.

### DIFF
--- a/plugins/tostky-tracker
+++ b/plugins/tostky-tracker
@@ -1,0 +1,2 @@
+repository=https://github.com/example-git/tostky-plugin.git
+commit=89860777ba9adbd0b8cb2ed18d4ba5520ec38003
--- a/plugins/tostky-tracker
+++ b/plugins/tostky-tracker
@@ -1,2 +1,2 @@
 repository=https://github.com/example-git/tostky-plugin.git
-commit=89860777ba9adbd0b8cb2ed18d4ba5520ec38003
+commit=7b2e6f988d057fdc8f491a6eda0045112bf8d143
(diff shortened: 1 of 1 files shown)


example-git:
Anyone get a chance to look at this?

cdfisher:
> Anyone get a chance to look at this?

Typically reviews for hub PRs, and especially for new plugins, can take several weeks, depending on when the people who do those reviews have time. Right now there's a lot of focus on getting updates together for the core client for Varlamore, so it may be a bit longer.

raiyni:
Does this plugin provide something the current item charges plugin doesn't?

example-git:
This is more focused on range equipment for grinding as well as tracking overall usage, not just remaining charge for cost tracking 
image
image


While the use case isn't as extensive as the itemlist covered by item charges I don't believe they allow you to track overall usage in this way. I was requested to make this for a friend who was trying to keep track of cannon ball usage over their grinding session and covered other ranged ammunition for the sake of it.

Nightfirecat:
You should [use a real author field](https://github.com/example-git/tostky-plugin/blob/7b2e6f988d057fdc8f491a6eda0045112bf8d143/runelite-plugin.properties#L2) for your plugin properties.
You should [scale your panel image down](https://github.com/example-git/tostky-plugin/blob/7b2e6f988d057fdc8f491a6eda0045112bf8d143/src/main/resources/icon.png) as Java will keep the full-size image in memory even when it is scaled down to be shown on the nav button.

LGTM otherwise.

example-git:
> You should [use a real author field](https://github.com/example-git/tostky-plugin/blob/7b2e6f988d057fdc8f491a6eda0045112bf8d143/runelite-plugin.properties#L2) for your plugin properties. You should [scale your panel image down](https://github.com/example-git/tostky-plugin/blob/7b2e6f988d057fdc8f491a6eda0045112bf8d143/src/main/resources/icon.png) as Java will keep the full-size image in memory even when it is scaled down to be shown on the nav button.
> 
> LGTM otherwise.

Done

PR #5694, opened 2024-03-16
Update wikisync to v3 and add jayden/lezed as authors

andmcadams:
@lezed1 could you give a quick summary of the changes here?

### DIFF
--- a/plugins/wikisync
+++ b/plugins/wikisync
@@ -1,4 +1,4 @@
 repository=https://github.com/andmcadams/WikiSync.git
-commit=8246ef597be7554010737e8c47559b23d689410f
+commit=cb98273687f697bb4ff8d9c11127c16dd71ac045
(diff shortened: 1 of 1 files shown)


lezed1:
WikiSync now hosts a WebSocket server to allow for a web-based tool pull data directly from the client! This will be used by the [OSRS Wiki DPS Calculator](https://tools.runescape.wiki/osrs-dps) to import your current gear. Here is an example of how the DPS Calculator will be able to use the new feature (I'm logged in to two instances of RL):

image

I tried to enforce that arbitrary websites are not authorized to talk to the WebSocket server. I'm happy to open it up to any site that has a good use case, but it felt wrong to open a server on localhost without an allowlist.

lezed1:
This also includes some copy/pasted code from https://github.com/runelite/runelite/pull/17524. Once that PR is merged into RL core, we can delete our copy. All the copied code is clearly marked, so it will be easy to remove!

raiyni:
Can the `log.info` calls be made `log.debug` ?

lezed1:
@raiyni I pushed a new commit switching the logs from info to debug!

PR #5704, opened 2024-03-18
Update advanced-raid-tracker

capslock13:
Long term fix for live chart performance issues. Only draws/allocates the image for the size of the window and manually implements scrolling


capslock13:
Forgot the other PR was still pending when I hit discard commit, will not do that in the future

PR #5705, opened 2024-03-18
Update Inventory Setups to v1.18 and Bank Tag Layouts

dillydill123:
Inventory setups v1.18

- V3 data (separate key per setup)
- search by item or notes in setup
- blazing blowpipe -> toxic blowpipe
- Fix patch notes panel not displaying correctly

That last point made me want to quit Java Swing

Also update Bank Tag Layouts to use the new V3 data. @geheur please confirm the commit is correct. These need to be merged at the same time to avoid a headache


geheur:
LGTM!

PR #5711, opened 2024-03-19
Create No Steven plugin

two-zee:
A RuneLite plugin that hides everything related to the "SteVen" bots that plague F2P.

It can hide their:
- Entities (e.g. models)
- 2D entities (e.g. hitsplats)
- Overhead chats
- Chat interface chats
- The models of who they are interacting with

### DIFF
--- a/plugins/no-steven
+++ b/plugins/no-steven
@@ -1,0 +1,2 @@
+repository=https://github.com/two-zee/no-steven.git
+commit=b60a773ef6957270e9f9c0b5918a898117166522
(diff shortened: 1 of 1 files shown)


raiyni:
Personally, I'm not keep on entity hiding players like this. Other maintainers can chime in but this is moderately unfair if someone happens to be named `steven`.

two-zee:
I do have a config toggle that enables only checking for case-sensitive "SteVen", and I can gladly enable that by default. 

Additionally, their names tend to follow a certain pattern ("SteVen" + 6 characters). Would it make a difference to you if I changed my strict mode to look for a regex that more closely matches their names?

Of course, I also don't want any innocent Stevens to be hit by this plugin, so let me know if there's anything I can do to make this more compliant.

two-zee:
I just changed strict mode to be on by default, and made it match against name length as well.

Nightfirecat:
No, we would close the PR if we came to that decision. I expect your changes are probably sufficient, but I haven't reviewed this plugin and ron probably has not re-reviewed this since those have been implemented.

PR #5712, opened 2024-03-19
Added new plugin: streamer redeems

molteber:
This plugin allows you to connect to a Server-sent event endpoint hosted by yourself or someone else which is connected to a streamer service (ie. twitch) 

Taking twitch into account, streamer can redeem rewards. A streamer can install this plugin and allow him/her/it to be pranked when chatters redeem specific rewards.

This plugin does for now only support triggering a fake collection log popup (which includes a chat message stating that the person was pranked).

### DIFF
--- a/plugins/streamer-redeems
+++ b/plugins/streamer-redeems
@@ -1,0 +1,2 @@
+repository=https://github.com/molteber/streamer-redeems.git
+commit=a490304c745b844692939adf10df90f94673b5e6
--- a/plugins/streamer-redeems
+++ b/plugins/streamer-redeems
@@ -1,2 +1,2 @@
 repository=https://github.com/molteber/streamer-redeems.git
-commit=a490304c745b844692939adf10df90f94673b5e6
+commit=abc4050b16fb0bdf68f67e88d499ab2406404e47


YvesW:
Does this plugin do anything that https://runelite.net/plugin-hub/show/twitch-live-loadout doesn't do? Would you not like to contribute to that plugin?

molteber:
@YvesW thanks for the quick reply

The main difference is the use of channel points instead of bits (when we're speaking of twitch), which that plugin does not support.

The plugin was made to goof in a particular stream, but made generic to allow further use of others wants to setup something similar

raiyni:
If you can work bits into that plugin, that would be the preferred avenue of this feature.

molteber:
> If you can work bits into that plugin, that would be the preferred avenue of this feature.

Not sure what you're trying to convey. Just to be clear:
This plugin: supports whatever, it just listens for events from an endpoint.
Twitch load out: can only use bits to start random events.

My use of this plugin will use a private web application which receives webhooks from twitch and pushes events to all listeners of a specific endpoint.
If anyone else wants to use this plugin, they will have to contact me or setup their own application to output what the plugin expects (mentioned in the readme)

I'm sure you would like to have similar plugins combined, but can they live side-by-side until twitch load out can support channel points?

raiyni:
I might have misunderstood on what your previous comment meant for feature parity. 

> I'm sure you would like to have similar plugins combined

This is more what I meant. If it's a similar `feature`, it's much preferred  you try and collaborate with an existing plugin's author first. In the event they are awol, or refuse, we would be happy to assist getting it on the hub in some other fashion.

molteber:
Im collaborating with the creator of twitch load out instead. Closing this

PR #5777, opened 2024-03-30
Update animation-transmog to V2.1.2

BarelyFunctionalCode:
Very small update that just removes the chat message that appears when the client starts asking the player to enable the multiplayer option if they don't already have it enabled.


BarelyFunctionalCode:
I've just now realized that after I changed my Github username that I did not update the `repository` variable in that file with the new username, so that is probably why it has flagged this as a "non-author plugin change".

Alexsuperfly:
correct, you can update that and it will be fine

BarelyFunctionalCode:
Should be good now.

PR #5781, opened 2024-03-31
Create anchovy-pizza-helper

vpavanv:
This is a plugin that helps players combo-eat in F2P with anchovy pizzas. It only highlights full anchovy pizzas as only eating the first halves is what allows quick healing vs eating the second halves.


cdfisher:
Does this do anything that can't already be done with Inventory Tags in core RuneLite?

vpavanv:
not at all, just wanted to make something more lightweight and hyper-specific to my scenario. in large f2p clan wars (100v100), it gets quite hectic so we tend to shut off as many plugins as possible to reduce lag and client overhead to improve rendering and thought this would help. will close this for now and gather performance specs during our next event

PR #5783, opened 2024-03-31
Item Charges Improved - v0.5.5

TicTac7x:


TicTac7x:
I have not seen any such config migration examples anywhere, link me please to some and I'll take a look on how easy it would be to do it.

Previous configs implementation with multi-select was really sluggish and i doubt it was even widely used. I don't think its worth the effort to spend more time on such migration script, but i am more than happy to let users know to reselect their infoboxes/overlays settings with the ingame plugin message that is displayed when new version is installed, hopefully that would be enough.

YvesW:
Multiple plugins have migrated configs. I think inventory setups recently went from config implementation v2 to v3 or something like that. You can just check in StartUp if the old key exists, and if it does, read its value and set the appropriate new config keys using ConfigManager. Then you can unset the old key or add a new key to mark that you've already migrated settings for that key (if the data is important like inventory setups, but you clearly don't care about it, so unsetting will probably be more applicable to you).

TicTac7x:
It makes sense to write config migration for inventory setups, because otherwise you would lose so much time spent on super custom settings.

I feel like this is not the case for this plugin, it's only a matter of checking few checkboxes for very specific items that you want to hide. Can we proceed without the config migration?

YvesW:
It prevents potential #support traffic and improves the user experience; the only reason you haven't done it is... because you're too lazy? Very briefly looking at your code, you can even remove the underscores in your enum + camelcase them, and add ``Overlay`` or ``Infobox`` to it to receive the name of the new config var? It takes 5-10 minutes to write that code.

TicTac7x:
If you think it only takes 10 minutes to write that code, then please do and help me out? Because for me it would take more than 10 minutes.

TicTac7x:
Here you go, updated it, hope this is enough. This took me 1 hour and 3 minutes. Blatant lies about 10 minutes, you are perfect candidate to be SCRUM master.

PR #5791, opened 2024-04-01
japanese plugin

YS-jack:
Can translates text of most if not all widgets into japanese, can also chat in japanese. can use DeepL api to translate thing more accurately.
This plugin is targeted for Japanese players who can't read English but want to play this game.

Currently some widgets are set not to be translated because I havent provided a translation for that part, and bad translation can cause confusion.

Can selectively choose not to translate if desired in config section.


raiyni:
Please don't close and open new prs. Push commits to the same branch.

YS-jack:
sorry about that, i wont

YS-jack:
oh wait, i closed it somehow

PR #5793, opened 2024-04-01
Update taskjingle

maximonster:
Added functionality to play level up music

### DIFF
--- a/plugins/taskjingle
+++ b/plugins/taskjingle
@@ -1,2 +1,2 @@
 repository=https://github.com/maximonster/TaskJingle.git
-commit=9d00dbc0ebfe462a93007ceef6c62f0888aeb580
+commit=e2a55f6130d16ed2dd9b1d440c45cc01c3c51880
--- a/plugins/taskjingle
+++ b/plugins/taskjingle
@@ -1,2 +1,2 @@
 repository=https://github.com/maximonster/TaskJingle.git
-commit=2663256a4160913552556d0937c7014070b1e7a9
+commit=9d00dbc0ebfe462a93007ceef6c62f0888aeb580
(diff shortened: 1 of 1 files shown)


Nightfirecat:
I'm not sure if you meant to add more changes here... this PR only updates the project's README. Does the plugin already support level up jingles, but didn't have info indicating that in its README?

maximonster:
No the level up jingle functionality is new as per commit 823789838ec1fa2c937abd2fc24aaca5b47a119f

I updated the readme afterwards making that the most recent commit

Felanbird:
I'm unsure why you would say "no" after being told what you've done does was incorrect, you almost certainly forgot to update your master branch with these changes last week and had to push them today.

If you type `git checkout 9d00dbc0ebfe462a93007ceef6c62f0888aeb580`, which is what the update to your plugin will do, you will get a branch without these changes just as NFC said. If you want to update the plugin with the changes, you'll need to update this pr to `e2a55f6130d16ed2dd9b1d440c45cc01c3c51880`, the merge commit that you pushed today that fixed up your issue.

maximonster:
Thanks for the help!

Nightfirecat:
[We're not going to be able to distribute the Final Fantasy 7 victory theme](https://github.com/maximonster/TaskJingle/blob/e2a55f6130d16ed2dd9b1d440c45cc01c3c51880/src/main/resources/lvlup.mp3), and for that matter, [the existing task-jingle.mp3 file is the GTA San Andreas mission passed theme](https://github.com/maximonster/TaskJingle/blob/e2a55f6130d16ed2dd9b1d440c45cc01c3c51880/src/main/resources/task-jingle.mp3) which we also should not be distributing. Please change or remove those files from distribution. (if you want, you can host the files separately and have the plugin download them on startup, like many other plugins such as the C Engineer plugin do)

This change LGTM otherwise.

maximonster:
I've replaced the problematic mp3s with original jingles inspired by the previous ones. 
I've made lvlup.mp3 myself and task-jingle.mp3 is made by a friend specifically for the plugin. 
Please let me know if this is a suitable solution.

Nightfirecat:
Thanks, I'll double check on that today

PR #5799, opened 2024-04-02
Add "No Player Use" plugin

oohwooh:
Simple plugin that removes menu options related to using items on other players. This is especially helpful for UIMs looking to unnote/note items in crowded areas.

### DIFF
--- a/plugins/no-player-use
+++ b/plugins/no-player-use
@@ -1,0 +1,2 @@
+repository=https://github.com/oohwooh/no-use-players.git
+commit=dff362d01fa7c562729f3b859cc25dce984696df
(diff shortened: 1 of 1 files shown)


cdfisher:
This seems like it would possibly go against the third-party client guidelines, specifically

> Reorders or removes player-based options, such as 'Trade with'

oohwooh:
Custom Menu Swaps can be used to accomplish this, yes. The "Easy Unnote" plugin also provides similar functionality, however only to a whitelist of items and targets.

The same functionality this plugin provides can also be gained by turning entity hider on.

Is anyone aware of a statement from Jagex clarifying this ruling?

PR #5806, opened 2024-04-04
Update gauntlet-map

StickySerum:
Refactor code & add new features:
-New overlay map option
-Show fishing spots option
-Show grym leaves option
-Highlight demi boss room node option

### DIFF
--- a/plugins/gauntlet-map
+++ b/plugins/gauntlet-map
@@ -1,2 +1,2 @@
 repository=https://github.com/StickySerum/gauntlet-map.git
-commit=f5fbc30784885f3bf237490adad00871bf94fa86
+commit=71e2be20ce0cdd304d921b026222f5ca4cee73d3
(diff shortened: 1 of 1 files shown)


StickySerum:
Good to know, thanks! First fancy README file I've ever done lol.

I have the images in tags with a set width. Is that still going to be an issue? Also where do you recommend I store the images for the readme?

Felanbird:
Just need to put them anywhere outside of the `src/` folder, they can be in their own folder like [tombs-of-amascut](https://github.com/LlemonDuck/tombs-of-amascut/tree/master/docs) or just in the root like [dink](https://github.com/pajlads/DinkPlugin/blob/master/img.png) whatever you prefer.

Nightfirecat:
A couple things:

* You've committed your `build/` directory. This doesn't do anything good or bad, but it does clutter up the diff view; I'd recommend adding it to your `.gitignore` to reduce the size of your repo (and make our lives easier by not needing to collapse/skip so many irrelevant files during review)
* You need to remove the map and demiboss overlays on plugin shutDown; presently they are not removed if you turn off your plugin

PR #5813, opened 2024-04-06
Add Unbalanced Trade Prevention plugin

TheStonedTurtle:


YvesW:
Not going to comment on if the plugin should be accepted or not, but could you add a chat message that makes it clear that this plugin is blocking the trade to prevent support traffic?

TheStonedTurtle:
image
PR has been updated so this will now show in the chatbox.

PR #5862, opened 2024-04-13
Create Venom Timer Plugin

SimonFlapse:
## Description
This plugin shows venom related information to help players identify envenomed actors. 

Information available includes:
- Next venom splat hitting
- Next venom splat damange
- Total venom damage applied
- Expected time before target will die from venom damage

## Screenshot
Screenshot 2024-04-13 210219

### DIFF
--- a/plugins/venom-timer
+++ b/plugins/venom-timer
@@ -1,0 +1,2 @@
+repository=https://github.com/SimonFlapse/venom-timer.git
+commit=9f3ad66fa66262939d385a13f2a2ef86274e6236


YvesW:
Does this significantly differ from <https://runelite.net/plugin-hub/show/poisoned-npcs>? Could it be added to that plugin?

SimonFlapse:
> Does this differ significantly from https://runelite.net/plugin-hub/show/poisoned-npcs? Could it be added to that plugin?

I did consider it, but I found that it was very specific to the more advanced mechanics of poison.

Nightfirecat:
While [you are removing your overlays from being rendered on plugin shutdown](https://github.com/SimonFlapse/venom-timer/blob/289c2e7926382096a9b36f9455e25f828689f06b/src/main/java/com/simonflapse/osrs/venom/ui/OverlayOrchestrator.java#L54), they are not cleared from `activeOverlays`, so they are effectively leaked memory (and end up not being rendered) if you toggle the plugin off and back on. I recommend adding an `activeOverlays.clear()` call in `OverlayOrchestrator#shutDown()`.

PR #5863, opened 2024-04-13
Create dodgy-necklace-reminder

LiamStachiw:
This plugin will remind the user to equip a dodgy necklace with an overlay whenever the user has a dodgy necklace in their inventory while also not having one already equipped.


cdfisher:
I believe this is already done by both "Pickpocket Info" and "Pickpocket Helper".

PR #5866, opened 2024-04-14
Cry-Alchemy Plugin

romeshdev:
### Description
Replaces the default high and low level alchemy animations with the cry animation, keeps the alch gfx but raises it a little to meet with head in hands.
First attempt at a simple plugin inspired by dance party, testing the waters a little here.

### Screencap
cryalching


raiyni:
Can this be added to animationtransmog?

PR #5877, opened 2024-04-16
Adds lf-skill plugin file

FaceTanked:
The "LF Skill" plugin simply replaces the overhead text displayed during the special attack of the dragon axe, harpoon, and pickaxe to "Let's fucking chop!", "Let's fucking fish!", and "Let's fucking mine!", respectively.

Additionally, there is a "no profanity" toggle for those that want to get festive without that specific language.

### DIFF
--- a/plugins/lf-skill
+++ b/plugins/lf-skill
@@ -1,0 +1,2 @@
+repository=https://github.com/FaceTanked/runelite-lf-skill.git
+commit=e95d50705fc4b218c14313a9d3e774c534e99225


cdfisher:
I believe "Spec Dialogue Plugin" allows for custom special attack overhead text, so this appears to be a use case already covered by that.

FaceTanked:
> We have generally stayed away from memey plugins that could be covered already by another plugin.

Fair enough, I appreciate the reviews 🤝

PR #5911, opened 2024-04-21
add hunter pouch overlay plugin

trs:
This plugin adds an overlay to the hunter pouches showing the number of meat/furs inside.

icon

### DIFF
--- a/plugins/hunter-pouch-overlay
+++ b/plugins/hunter-pouch-overlay
@@ -1,2 +1,2 @@
 repository=https://github.com/trs/runelite-hunter-pouch-overlay.git
-commit=e127bb56c32f4a879b3c7f35fe65bec472404d1e
+commit=dcc682ba14bb869cdbc2ac7f80aacfb466aa0da3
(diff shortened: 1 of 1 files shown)


trs:
I've found a few issues I'd like to sort out before adding to the hub

Felanbird:
Just switch this PR to a draft until you've made the changes you feel necessary, and then update this PR to the new commit hash.

trs:
Okay, after some fixes and testing this should be working good now. It previously wasn't taking into account if multiple actions were occurring during a single tick (like filling both meat and fur pouches quickly).

Nightfirecat:
I see your plugin doesn't currently have any config options, [but "pouch" isn't a sufficiently distinct config group](https://github.com/trs/runelite-hunter-pouch-overlay/blob/dcc682ba14bb869cdbc2ac7f80aacfb466aa0da3/src/main/java/com/pouch/HunterPouchConfig.java#L7); please either remove the config interface or change the config group to be more unique. (either would help avoid possible conflicts in the future should you decide to add configs later on)
Code LGTM otherwise

LlemonDuck:
Convert this off of draft when you've made the requested changes

MESLewis:
Closed for inactivity, you can reopen this if you would like to make the requested changes.

PR #5912, opened 2024-04-21
Update Last Man Standing to 1.7.1

pwatts6060:
Add Lobby Summary
Add option to mark players as sweats for future reference and display in lobby summary/ingame


Nightfirecat:
[You're adding an infobox in your plugin which is never removed.](https://github.com/pwatts6060/runelite-plugins/blob/b1a74c9b8a0e211566f9ad17d1821fbf96841127/src/main/java/com/lmsnotifier/LMSPlugin.java#L181) Please ensure turning the plugin off will remove it.

pwatts6060:
> [You're adding an infobox in your plugin which is never removed.](https://github.com/pwatts6060/runelite-plugins/blob/b1a74c9b8a0e211566f9ad17d1821fbf96841127/src/main/java/com/lmsnotifier/LMSPlugin.java#L181) Please ensure turning the plugin off will remove it.
> 
> 

itemUpgrade.reset() should remove the infoboxes already in ItemUpgrade.java when it is called in shutdown event

PR #5915, opened 2024-04-22
Update Item Charges Improved to v0.5.7

TicTac7x:


raiyni:
You need to unregister the key/mouse listener on shutdown.

TicTac7x:
Thanks, updated, made it bit clearer as well, the handlers are registered/unregistered now in the plugin class

PR #5918, opened 2024-04-22
NPC Animator Plugin

SomewhatUsefulSoftware:
Lets users set and run animations on NPCs and their own character.

This plugin is intended to be used to make the NPCs feel more alive and to give them a little extra personality. It might also be useful for players who want to make role-playing videos.

Features:
- Save any animation in the game with your own customizable name for it.
- Run any animation on your own player.
- Run any animation on any NPC.
- Save animations for individual NPCs as 'Greetings'
    * Example: Save the bow animation for King Roald so when you interact with him, he bows to you.
- Run all saved NPC animations at the same time.
    * Use this to run all the saved greeting animations or just 1 specific animation on all saved NPCs.
- Run an animation on all spawned NPCs at the same time.

### DIFF
--- a/plugins/npc-animator
+++ b/plugins/npc-animator
@@ -1,0 +1,2 @@
+repository=https://github.com/SomewhatUsefulSoftware/npc-animator.git
+commit=98debd79289e46b6b507e915c5214c1afd5f190f
--- a/plugins/npc-animator
+++ b/plugins/npc-animator
@@ -1,2 +1,2 @@
 repository=https://github.com/SomewhatUsefulSoftware/npc-animator.git
-commit=98debd79289e46b6b507e915c5214c1afd5f190f
+commit=862451ee9d11fb9ce613fbb95477d389df52de28


Nightfirecat:
You're [adding a nav button here](https://github.com/SomewhatUsefulSoftware/npc-animator/blob/98debd79289e46b6b507e915c5214c1afd5f190f/src/main/java/com/npcanimator/NPCAnimatorPlugin.java#L105) which is never removed, even on plugin shutdown, so disabling your plugin would not remove this button and toggling it would add multiple buttons. Please ensure it is removed on `shutDown()`.

SomewhatUsefulSoftware:
> You're [adding a nav button here](https://github.com/SomewhatUsefulSoftware/npc-animator/blob/98debd79289e46b6b507e915c5214c1afd5f190f/src/main/java/com/npcanimator/NPCAnimatorPlugin.java#L105) which is never removed, even on plugin shutdown, so disabling your plugin would not remove this button and toggling it would add multiple buttons. Please ensure it is removed on `shutDown()`.

I fixed this as requested and also fixed a UI bug. However, I cannot verify the workflow checks pass as it's awaiting approval from a maintainer.

PR #5924, opened 2024-04-24
Update rumour-reminder - Remove Kebbity tuft related highlighting

DrJam:
Removing equivalent turn-in options per update today
- Kebbity Tufts from Hunter rumours can now only be handed in if they were obtained from the same kebbit that you've been assigned.


DrJam:
Closing for a bit as expecting item ids for the different items so will need to add that as well

YvesW:
For future reference, we recommend marking the PR as draft instead of closing it in such cases.

DrJam:
Makes sense, also doesn't seem like there's new item IDs so reopening this, sorry for not drafting 😅

PR #5934, opened 2024-04-26
Create click-to-minimize

WaylonJ:
This plugin minimizes the game client when certain actions are pressed.

For example, with the config set to:
Pick-Fruit: Sq'irk tree, Mine: Crashed Star

Anytime the player tries to pick a sq'irk fruit, or mine a crashed star, the game will minimize immediately.

This helps when you want to minimize the distraction the game causes while still afk'ing to get those gains.


Nightfirecat:
[Do not use "example" as your config group.](https://github.com/WaylonJ/OSRS_ClickToMinimize/blob/80e1a91f0caa66d8929603e46556951c6e09dabb/src/main/java/com/clicktominimize/ClickToMinimizeConfig.java#L7)
Code LGTM otherwise

WaylonJ:
Updated to not use "example" as config group.

PR #5940, opened 2024-04-26
Add osrs-soapstones plugin file.

KevinJDurant:
If you're familiar with the Dark Souls series, then you know players can use a Soaptone to leave messages for other players.

This basically (in a very simple way) does the same thing but then with tiles. It's in a very barebones state.

There's no way for people to actually use / test this plug-in unless they write their own backend to support the HTTP-calls this plug-in needs. Which are documented in the README.md. 

I didn't add a master service yet because I lack the resources to provide the infrastructure at this moment. Also, to provide this unfiltered, uncached, etc. doesn't seem like a good idea right now.

My main goal is to share this plug-in with my friends without sharing shaded builds.

Example (the white boxes are the player usernames):

image

The code lacks tests.

### DIFF
--- a/plugins/osrs-soapstones
+++ b/plugins/osrs-soapstones
@@ -1,0 +1,2 @@
+repository=https://github.com/KevinJDurant/osrs-soapstones.git
+commit=87e663b04762669b767fa0168dd67fffe4fb4bd4


cdfisher:
> There's no way for people to actually use / test this plug-in unless they write their own backend to support the HTTP-calls this plug-in needs.

I suspect that you would probably be better off waiting to submit this plugin until either you or someone else has actually written and started hosting the backend required for the plugin to actually function.

It doesn't make much sense for maintainers to add a plugin to the hub if they know for a fact that it's non-functional in its current state.

KevinJDurant:
> I suspect that you would probably be better off waiting to submit this plugin until either you or someone else has actually written and started hosting the backend required for the plugin to actually function.
> 
> It doesn't make much sense for maintainers to add a plugin to the hub if they know for a fact that it's non-functional in its current state.

I fully understand your point. 

For context, I have written my own backend for it, for me and my friends. I just cannot expose this backend for the masses. A public backend would require decent caching, some sort of authentication, profanity and gold seller filtering and more.

I thought we could release plug-ins "for friends" on the market, so we wouldn't have to create our own builds. But if that's not realistic, I don't mind closing the PR.

raiyni:
If it's just a plugin for friends, we would prefer you pass around a shaded build unless you are going to provide a public use option--or self hosted server option.

KevinJDurant:
Ok, fair enough :)

PR #5943, opened 2024-04-27
Add ironclad clan goals

dextermb:
Ironclad are an ironman focused OldSchool RuneScape clan and we're looking for new and inventive ways to include clan members in community activities.

We currently want to create OldSchool skill xp goals – the best way to do this, we thought, would be to create a RuneLite plugin that can keep track of clan member xp gains.

This plugin allows for a RuneLite user to add an API key to the "Ironclad Clan Goals" configuration panel.

If the API key is valid (checked on plugin start up and configuration change) then we log:
- Which account is using the API key
- Which skills they are currently training

Outside of the plugin the backend service (https://progress.quest) will keep track of the information above along with any contributions towards an ongoing goal.

The backend service also outputs contributions to a Discord channel where people can see how everyone is doing.

### DIFF
--- a/plugins/ironclad-clan-goals
+++ b/plugins/ironclad-clan-goals
@@ -1,4 +1,4 @@
 repository=https://github.com/ironclad-osrs/runelite-plugin.git
-commit=601ef36263f6a15bab48bf3357e789936dd9ec8d
+commit=80d39c0d9e3f0dffeea5395962434419493f50ee
 authors=dextermb
 warning=This plugin sends player data (including IP address) to a 3rd party server not controlled by RuneLite or Jagex.
--- a/plugins/ironclad-clan-goals
+++ b/plugins/ironclad-clan-goals
@@ -1,2 +1,2 @@
 repository=https://github.com/ironclad-osrs/runelite-plugin.git
-commit=adf6e5a87946c01ad38520a6f9a547a20eef321a
+commit=ef47a6d6f8912bc497306834682a1ef3db8118ee
--- a/plugins/ironclad-clan-goals
+++ b/plugins/ironclad-clan-goals
@@ -1,2 +1,4 @@
 repository=https://github.com/ironclad-osrs/runelite-plugin.git
 commit=601ef36263f6a15bab48bf3357e789936dd9ec8d
+authors=dextermb
+warning=This plugin sends player data (including IP address) to a 3rd party server not controlled by RuneLite or Jagex.
--- a/plugins/ironclad-clan-goals
+++ b/plugins/ironclad-clan-goals
@@ -1,2 +1,2 @@
 repository=https://github.com/ironclad-osrs/runelite-plugin.git
-commit=ef47a6d6f8912bc497306834682a1ef3db8118ee
+commit=601ef36263f6a15bab48bf3357e789936dd9ec8d
--- a/plugins/ironclad-clan-goals
+++ b/plugins/ironclad-clan-goals
@@ -1,2 +1,2 @@
 repository=https://github.com/ironclad-osrs/runelite-plugin.git
-commit=e8acb56c078cba0193c5508c2ea424977f6a13f6
+commit=adf6e5a87946c01ad38520a6f9a547a20eef321a
--- a/plugins/ironclad-clan-goals
+++ b/plugins/ironclad-clan-goals
@@ -1,0 +1,2 @@
+repository=https://github.com/ironclad-osrs/runelite-plugin.git
+commit=e8acb56c078cba0193c5508c2ea424977f6a13f6


raiyni:
Is there a reason you can't use Wise Old Man for this?

dextermb:
> Is there a reason you can't use Wise Old Man for this?

We're looking at a custom solution so we can track specifics (kills, drops, et al) in the future and hope to gain the most up to date information as fast as possible. 

WiseOldMan quite potentially would solve our problems but then we'd be interfacing with something we don't own or control. 

By using something we control it allows our clan members to agree to usage and see direct results of that usage.

raiyni:
I don't understand your concerns about agreeing to usage as all account data is public on the high scores and groups are managed by a leader. There is a utility plugin they can *choose* to install.  https://runelite.net/plugin-hub/show/wom-utils

dextermb:
> I don't understand your concerns about agreeing to usage as all account data is public on the high scores and groups are managed by a leader. There is a utility plugin they can *choose* to install.  https://runelite.net/plugin-hub/show/wom-utils

Not concerned about members agreeing to usage. But, again, I am looking for something we control. 

WSO is a great service for those who don't want to deal with setting up their own stuff whereas I don't mind.

Thankfully tracking RuneScape stats is not as vital as hosting a website, using serverless functions, or storing files/data safely and securely so "re-inventing" the wheel here for some extra personalisation and control is something I'd consider to be fine.

dextermb:
While waiting for @Adam- or another core maintainer to respond it's worth noting that [another clan has their own plugin](https://runelite.net/plugin-hub/show/embargo-clan-plugin) as well.

Alexsuperfly:
I don't understand the point there, they perform a unique task, you don't.  As long as what you're doing is already being done by something else why are you pushing to have a duplicate.

dextermb:
> I don't understand the point there, they perform a unique task, you don't. As long as what you're doing is already being done by something else why are you pushing to have a duplicate.

Because this is the MVP. This gives us something to iterate on

cdfisher:
> > I don't understand the point there, they perform a unique task, you don't. As long as what you're doing is already being done by something else why are you pushing to have a duplicate.
> 
> Because this is the MVP. This gives us something to iterate on. There's no point for us to put in tons of effort into something and then it be rejected.

I think you're missing that in its current state this is pretty likely to be rejected because it doesn't add really anything that can't be done with other plugins.

raiyni:
> While waiting for @Adam- or another core maintainer to respond it's worth noting that [another clan has their own plugin](https://runelite.net/plugin-hub/show/embargo-clan-plugin) as well.

My reply was to coerce you into using WOM's groups features as that's what you are doing. You can even pull data to your server and do with it what you want there. https://docs.wiseoldman.net/

dextermb:
> > > I don't understand the point there, they perform a unique task, you don't. As long as what you're doing is already being done by something else why are you pushing to have a duplicate.
> 
> > 
> 
> > Because this is the MVP. This gives us something to iterate on. There's no point for us to put in tons of effort into something and then it be rejected.
> 
> 
> 
> I think you're missing that in its current state this is pretty likely to be rejected because it doesn't add really anything that can't be done with other plugins. 

I'm more surprised the "review" process isn't "does it break Jagex rules, no? Okay it's in" but more a whole backstory and reasoning to why a plugin is created. 

You're right that in its current state it just tracks XP which can be done elsewhere but we have plans to include more as time goes on. Hence all my previous statements:

We want to create the plugin, so that we have something we control. This version is the minimum viable product for us to get started and test the waters of how RuneLite reviews work. Once approved and we use the current features we can expand its use cases (as described, and more). 

You guys are trying to solutionize the use cases for IronClad of which none of you have spoken to the clan leadership to understand the end goal. 

It's a shame.

dextermb:
> > While waiting for @Adam- or another core maintainer to respond it's worth noting that [another clan has their own plugin](https://runelite.net/plugin-hub/show/embargo-clan-plugin) as well.
> 
> 
> 
> My reply was to coerce you into using WOM's groups features as that's what you are doing. You can even pull data to your server and do with it what you want there. https://docs.wiseoldman.net/

That's perfectly fair. But the second we want to do more then we need to submit a plugin anyway as we don't control the source code for WOM. 

We have plans and niche use cases.

plugins/ironclad-clan-goals
```diff
@@ -0,0 +1,2 @@
+repository=https://github.com/ironclad-osrs/runelite-plugin.git
+commit=601ef36263f6a15bab48bf3357e789936dd9ec8d
```
raiyni, line 2:
```suggestion
commit=601ef36263f6a15bab48bf3357e789936dd9ec8d
warning=This plugin sends player data (including IP address) to a 3rd party server not controlled by RuneLite or Jagex.
```

raiyni (changes requested):

raiyni:
1. Inject our `okHttpClient` and create a new builder from that. Enqueue should also be used from it.
2. Logs should be debug level if they aren't useful to the average user

dextermb:
> 1. Inject our `okHttpClient` and create a new builder from that. Enqueue should also be used from it.
> 2. Logs should be debug level if they aren't useful to the average user

Hey, I believe I've addressed these points. Let me know if you have any other feedback

PR #5949, opened 2024-04-28
Create PvMRandomize

JDSCHMITZ20:
Randomize PvM drops, if you get a whip as a drop it may unlock ability to use bandos tassets but you cannot use the whip until you get the item that unlocks it.


cdfisher:
This appears to just be the example plugin with names changed in a handful of places. Please don't submit things to the hub before you've actually written them.

geheur:
I think this plugin uses your imagination to function. It's fully functional as is.

PR #5950, opened 2024-04-29
Create graardor-kc

Gavin-TC:
Small plugin that adds counter boxes to the screen to see Bandos KC. Mostly useful with groups where you're not always getting the kill and would like to see how many kills you're getting per trip.

### DIFF
--- a/plugins/graardor-kc
+++ b/plugins/graardor-kc
@@ -1,2 +1,2 @@
 repository=https://github.com/Gavin-TC/GraardorKC.git
-commit=2442922dc8d059ed414325c48a4654204f7ee395
+commit=d72f953b06843b789084fed50fba8cad7e87b48e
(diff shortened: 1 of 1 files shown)


DominickCobb-rs:
Could this not be achieved by looking at "Current Session Data" in the Loot Logger plugin or at the individual data in the Loot Tracker? It would require a client restart between trips if you're more worried about the separation of multiple trips in one go.

Gavin-TC:
> Could this not be achieved by looking at "Current Session Data" in the Loot Logger plugin or at the individual data in the Loot Tracker? It would require a client restart between trips if you're more worried about the separation of multiple trips in one go.

I suppose it could if it shows info from kills made by people in your group. I haven't used the plugin before, so I'm not sure.

umer-rs:
I believe `boost-performance` would do what you want. It tracks kills (at whitelisted bosses, bandos is included), regardless if you were the one getting kill credit.

Gavin-TC:
> I believe `boost-performance` would do what you want. It tracks kills (at whitelisted bosses, bandos is included), regardless if you were the one getting kill credit.

Ahh, alright. Should I close the PR then since my plugin doesn't really add anything?

PR #5952, opened 2024-04-29
Item Processing Warner plugin

ConorJS:
Adds alert specific to item processing tasks (e.g. smithing, crafting, fletching). There are two, separately configurable alerts: one which fires when the processing is _almost_ complete (i.e. an advance warning), and another when it completes. The former is the main reason for the plugin, and from where the plugin name is derived. The latter is more an afterthought, to allow a (potentially) more strong, or at least distinct alert when the user needs to more urgently respond. 

The functionality of the advance warning works by keeping a history of 'changed' items in the user's inventory, identifying which items are changing into what, and calculates when the number of the input item will reach zero in the user's inventory, based on how recently there was a conversion, the current number of existing input items, and the number of ticks per conversion action (configured by the user - this is not calculated).

The alerts use a breathing effect (a little more pleasant than the RuneLite default, IMO), similar to my [Camdozaal fishing plugin](https://github.com/runelite/plugin-hub/pull/5060).


ConorJS:
@raiyni in case this fell off your radar

raiyni:
It's been 2 days.

PR #5953, opened 2024-04-29
new bank-black-list plugin

spektrum1:
This plugin allows you to blacklist certain items, causing a warning every time you leave them in your bank. Helpful for group ironmen forgetting to put their items back in the group storage.

### DIFF
--- a/plugins/bank-black-list
+++ b/plugins/bank-black-list
@@ -1,2 +1,2 @@
 repository=https://github.com/spektrum1/bank-black-list.git
-commit=f58f164499d73a9bc2f8351caa751142e676ddba
+commit=58716bb438eba86b40b7174840574f32ff557983
--- a/plugins/bank-black-list
+++ b/plugins/bank-black-list
@@ -1,2 +1,2 @@
 repository=https://github.com/spektrum1/bank-black-list.git
-commit=b7cf1027841feea632e84ecd3d4ba383d09b30e8
+commit=671114571363e3fac7d5916de330076a5ca9d715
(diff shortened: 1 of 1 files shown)


raiyni:
Don't use reflection

spektrum1:
> Don't use reflection

Fair. Is there a preferred way to retrieve an ItemID from a name string? Looking around the existing plugins I can't seem to manage to find an example

geheur:
Why do you need an itemid? why not get the item name and see if it matches the input?

spektrum1:
> Why do you need an itemid? Why not get the item name and see if it matches the input?
> 
> ItemManager is probably what you want to use. It can get you the name from the id.

I was hoping to avoid doing a lookup for every bank item instead of just the items on the black list, but I made the change now. Thanks for the input

geheur:
I think you are too worried about performance. Also, in the old version the reflective lookup was inside the double loop instead of in the config reading... so this new version is definitely faster XD.

spektrum1:
> I think you are too worried about performance. Also, in the old version the reflective lookup was inside the double loop instead of in the config reading... so this new version is definitely faster XD.

True woops. Talking about it might as well optimise a little more

Nightfirecat:
Can you [change your config group?](https://github.com/spektrum1/bank-black-list/blob/58716bb438eba86b40b7174840574f32ff557983/src/main/java/com/bankblacklist/Config.java#L31) "Block List" on its own is not as specific as "Bank Block List", or something of the like--I'd rather this be more specific than less to avoid possible conflicts down the line.
Code LGTM otherwise

spektrum1:
Done, thanks

PR #5955, opened 2024-04-30
add umer-rs to list of maintainers of dey0's plugins

umer-rs:


dey0:
Confirming @umer-rs as a new maintainer for my plugins.

PR #5964, opened 2024-05-02
Create death-inventory

UniqueName992:
Initial
A plugin which captures and shows your inventory when you died so you can replace used supplies and continue.


raiyni:
Is what's in this plugin different than `Dude wheres my stuff?`

UniqueName992:
I'd say so
That one monitors overall contents of IRS / Gravestone for monitoring where items are 
This one shows what you had when you died last as a snapshot, for easy re-gearing

PR #5971, opened 2024-05-04
Spider Hider

straightupgangsta:
Very simple plugin based on the code from the entity hider. All it does is remove all spider models from the game so that my friend who is arachnophobic can play the game.


YvesW:
Please don't close and open new PRs. Just push to the same branch. You can mark it as draft if you really want to let people know to not merge it yet, but it won't be merged anyway if checks are failing.

PR #5988, opened 2024-05-10
item-stats-banking: add warning

Alexsuperfly:
users constantly complain in #support that they install this and then are confused why item stats are now disabled


plugins/item-stats-banking
```diff
@@ -1,2 +1,3 @@
 repository=https://github.com/Boredska/bank-item-stats-toggle.git
 commit=b9d65c7472e59224895ff13b05bd459c81661a32
+warning=This plugin disables item stats, dont install it if you want to see item stats.
```
Nightfirecat, line 3:
```suggestion
warning=This plugin can disable the Item Stats plugin, don't install it if you want to see item stats.
```

PR #6001, opened 2024-05-12
Create blast-mine-dynamite-restriction

Fabletownn:
If the player runs out of dynamite in the Blast Mine, this plugin will alert the player (with a sound effect and chat message) and deprioritize chiseling Hard Rock (set it to right-click only) until they have it again. This way, they don't waste time chiseling the next rock if they can't load it, and helps with increasing efficiency.


YvesW:
Please use a unique config group, not ``example``.

PR #6012, opened 2024-05-15
Create recommended-equipment

t8or:
Added Recommended Equipment plugin. This awesome plugin takes your OSRS gameplay to the next level by seamlessly integrating equipment recommendations from the wiki, via the approved wiki scraper, directly into your RuneLite client and then filtering your bank, similar to Quest Helper.

It helps you gear up quickly for bossing and mini-games by using wiki data without ever leaving RuneLite.

### DIFF
--- a/plugins/recommended-equipment
+++ b/plugins/recommended-equipment
@@ -1,2 +1,2 @@
 repository=https://github.com/t8or/runelite-recommended-equipment.git
-commit=648398d386948d6aef39975d91a8f1a2c4f2c632
+commit=da919e4bb4394c4d2de186561017a979a00ff541
--- a/plugins/recommended-equipment
+++ b/plugins/recommended-equipment
@@ -1,2 +1,2 @@
 repository=https://github.com/t8or/runelite-recommended-equipment.git
-commit=1fb3ea49fcdc572cf089e814085ab40532898147
+commit=a41ed77fe623873159332087e76af9bdf2af49eb
--- a/plugins/recommended-equipment
+++ b/plugins/recommended-equipment
@@ -1,0 +1,2 @@
+repository=https://github.com/t8or/runelite-recommended-equipment.git
+commit=1fb3ea49fcdc572cf089e814085ab40532898147
--- a/plugins/recommended-equipment
+++ b/plugins/recommended-equipment
@@ -1,2 +1,2 @@
 repository=https://github.com/t8or/runelite-recommended-equipment.git
-commit=a41ed77fe623873159332087e76af9bdf2af49eb
+commit=648398d386948d6aef39975d91a8f1a2c4f2c632


adamk33n3r:
image
I'm not sure why it's saying it needs to wrap at 120 when it's at 114 🤔

raiyni:
https://github.com/runelite/plugin-hub/blob/d40f8d47274a84dbe0a33a5b4db8d218db5082be/package/package/src/main/java/net/runelite/pluginhub/packager/Plugin.java#L400

adamk33n3r:
oh fun.....lol why EIGHT for tabs :lul:

Nightfirecat:
that's the default size of tabs in github's ui

adamk33n3r:
alright, i converted the file to spaces so hopefully that works. but i dont have permissions for @t8or's fork at the moment to update the hash, but he can do it in the morning. https://github.com/t8or/runelite-recommended-equipment/commit/a41ed77fe623873159332087e76af9bdf2af49eb

Nightfirecat:
[You should absolutely not be mutating event objects, that would negatively impact every other plugin which would be listening for it.](https://github.com/t8or/runelite-recommended-equipment/blob/648398d386948d6aef39975d91a8f1a2c4f2c632/src/main/java/com/adamk33n3r/runelite/recommendedequipment/banktab/BankTab.java#L123)
What's going on [with RecEquipLAF?](https://github.com/t8or/runelite-recommended-equipment/blob/648398d386948d6aef39975d91a8f1a2c4f2c632/src/main/java/com/adamk33n3r/runelite/recommendedequipment/RecEquipLAF.java) Why does it require reflection? How is that needed here? I'm generally a little confused about all this LAF code (both for this file, and the use of RuneLiteLAF) you have going on.
You [add your nav button here](https://github.com/t8or/runelite-recommended-equipment/blob/648398d386948d6aef39975d91a8f1a2c4f2c632/src/main/java/com/adamk33n3r/runelite/recommendedequipment/RecommendedEquipmentPlugin.java#L62) which is never removed, even on plugin shutdown. Please ensure it is removed when disabling your plugin.

adamk33n3r:
Hey nfc, I'm the dev on this project.
> [You should absolutely not be mutating event objects, that would negatively impact every other plugin which would be listening for it.](https://github.com/t8or/runelite-recommended-equipment/blob/648398d386948d6aef39975d91a8f1a2c4f2c632/src/main/java/com/adamk33n3r/runelite/recommendedequipment/banktab/BankTab.java#L123) 

The code in the `banktab` package is more or less copied from QuestHelper. That line specifically is also in quest helper here: https://github.com/Zoinkwiz/quest-helper/blob/master/src/main/java/com/questhelper/bank/banktab/QuestBankTab.java#L227 I don't know exactly what it's doing, or if it's needed for our case. I just tried to keep everything there.

> What's going on [with RecEquipLAF?](https://github.com/t8or/runelite-recommended-equipment/blob/648398d386948d6aef39975d91a8f1a2c4f2c632/src/main/java/com/adamk33n3r/runelite/recommendedequipment/RecEquipLAF.java) Why does it require reflection? How is that needed here? I'm generally a little confused about all this LAF code (both for this file, and the use of RuneLiteLAF) you have going on. 

The reflection specifically is copy/pasted from the RuneLiteLAF to get the colors from the colorscheme into the style file. As for what `RecEquipLAF` is for, it's to customize the style for our own panel. Whenever our panel is open, we switch to `RecEquipLAF` and when our panel is closed we switch back to `RuneLiteLAF`. You can see that [here](https://github.com/t8or/runelite-recommended-equipment/blob/648398d386948d6aef39975d91a8f1a2c4f2c632/src/main/java/com/adamk33n3r/runelite/recommendedequipment/PanelWrapper.java#L28). Unfortunately, swing doesn't really support multiple LAF's so this was the cleanest solution I could come up with. You can see our stylesheet [here](https://github.com/t8or/runelite-recommended-equipment/blob/648398d386948d6aef39975d91a8f1a2c4f2c632/src/main/resources/com/adamk33n3r/runelite/recommendedequipment/RecEquipLAF.properties)

> You [add your nav button here](https://github.com/t8or/runelite-recommended-equipment/blob/648398d386948d6aef39975d91a8f1a2c4f2c632/src/main/java/com/adamk33n3r/runelite/recommendedequipment/RecommendedEquipmentPlugin.java#L62) which is never removed, even on plugin shutdown. Please ensure it is removed when disabling your plugin.

Yep, that's totally my bad. Speaking of shutdown, should we be removing the sprite override as well? (if possible)

raiyni:
1. You should be doing what bank tag layouts does

2. You should be able to set whatever properties on the object. Swapping in a new LAF is not good.

t8or:
Hey Rayini,

Can you explain why we should be doing what bank tags does instead of quests helper?

Can you point us in the direction to set visual design properties on the objects instead of using a LAF? We couldn't find an easy way to adjust objects that repeat many many times; hence creating the LAF.

adamk33n3r:
More specifically, rounded buttons. Can easily make a rounded border, but the background would still be a rect. flatlaf has the capability to do full rounded buttons. Maybe some manual `paintComponent` override or something? 🤮

abextm:
the laf is global, you are not allowed to set it

abextm:
and really, you should not be using rounded buttons because nothing else in the client uses rounded buttons; your plugin integrates into the client, not the other way around.

t8or:
This is how the plug-in should look. Would love direction on how to edit each object so it inherits this styling. 

image

abextm:
Swing doesn't support multiple styles in a single application, you should inherit the styles provided by the LAF so it looks like the rest of the application.

t8or:
Appreciate the direction Abex!

Intecpsp:
I'm not a developer/maintainer of this particular repo/application, nor do I have any say in the approval of this PR. I'm just a software engineer who follows chatter on repos and PRs I find interesting (like RL and this plugin in particular).

Another way to put it is that you want your plugin's panel to feel like it's part of the core RL app (even though we all know it isn't). And if you stray too far from the core with your own LAF it can feel jarring to the end user and feel like it's own app inside of RL. 

That is all, back to lurking. Good luck with your plugin, I'm looking forward to seeing it on the hub!

adamk33n3r:
Alrighty, removed the custom LAF and removed that event mutation 👍🏻

t8or:
It's been a little while. Just wanted to check in and see if there is anything else to fix before merging into the plug-in hub. Thank you!

raiyni:
Core is adding bank tag layouts and will be exposing an API to register your own layouts, recommend looking at that.

t8or:
@raiyni do you know when that API will be available? 

I'm anxious to get this plugin released as I know it will be a great QoL update for all the new equipment and combat changes that just came out.

YvesW:
It's already on master, so next release.
https://github.com/runelite/runelite/commit/ee0c998621dde080c80a968bdf35e773783271d1

t8or:
Just so I can make sure I understand, this PR won't be merged until the plugin is changed to use the API that will be out soon?

adamk33n3r:
Does this new API let us create sections and dividers like we have in this plugin?
 image

I looked through the commit you linked @YvesW but I'm not seeing anything that would let us do that, the sections especially are important.

geheur:
That is not currently supported.

t8or:
With the announcement of Bank Tags in the new official client, is that using the same API as the one above? I hope that this plugin can be integrated before the great plugin migration.

PR #6023, opened 2024-05-15
Added Entity Masker

Droei:
Plugin made to mask out specific entities in Runescape while still keeping them clickable.
Having the ability to mask out specific entities is something I've read a few requests for.
Currently it has a default list for epilepsy and a few common phobia's but you can also make your own.

I'm no expert in the field so feedback on the concept as well as the actual code is always highly appreciated!


YvesW:
Don't forget to unregister your overlay using overlaymanager and npcoverlayservice.
Also probably resize your jad image to be smaller and re-encode your secretImage.

PR #6038, opened 2024-05-19
Update tog-crowdsourcing: Add backoff retries, US E, W icons

jcarbelbide:
… icons


jcarbelbide:
I'll fix the build when i have time in the next day or so

PR #6040, opened 2024-05-19
Create camera keyframe plugin

mlgudi:
Adds a camera keyframing plugin. Users can create sequences of camera positions using the side-panel, and then play them back with customisable durations/easing between each point.

### DIFF
--- a/plugins/keyframecamera
+++ b/plugins/keyframecamera
@@ -1,2 +1,2 @@
 repository=https://github.com/mlgudi/rl-plugins.git
-commit=d90a67bcc431078cb633850287e72a96e657da44
+commit=eda3b712785ed482719c6bca67186e1349bb0f0b
--- a/plugins/keyframecamera
+++ b/plugins/keyframecamera
@@ -1,2 +1,2 @@
 repository=https://github.com/mlgudi/rl-plugins.git
-commit=eda3b712785ed482719c6bca67186e1349bb0f0b
+commit=a6455af947b927650f251cfc62cdbc6b962da9cd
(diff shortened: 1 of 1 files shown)


adamk33n3r:
Surprised this hasn't existed yet!

cdfisher:
This may be better as an enhancement to the existing hub plugin "Camera Points". Have you tried reaching out to the author of that about possibly contributions to that plugin?

mlgudi:
> Surprised this hasn't existed yet!

The recent API addition of setting of camera targets made it possible! A beautiful addition to the API which I am sure will have many benefits.

> This may be better as an enhancement to the existing hub plugin "Camera Points". Have you tried reaching out to the author of that about possibly contributions to that plugin?

I was not aware this plugin existed. I would note the new camera target APIs are only available in detached/oculus mode, though I am happy to rewrite it as an extension of Camera Points if it is deemed more appropriate.

raiyni:
I think it's fine as its own plugin as it's more of a camera orchestrator instead of spots to switch back and forth between.

raiyni:
If you can add an indicator the chat message came from your plugin, will be good to go. eg prepending the message with a highlighted [Keyframecamera] or something similar.

mlgudi:
> If you can add an indicator the chat message came from your plugin, will be good to go. eg prepending the message with a highlighted [Keyframecamera] or something similar.

Thanks! I've now added the message prefix.

mlgudi:
Unsure of the etiquette on bumping PRs, but it has been a while so... Bump! Any chance this could get merged?

Felanbird:
The queue for new plugins is very behind right now, you almost never would need to bump a PR on this repo

But it still needs a full proper review before it can be merged

PR #6047, opened 2024-05-21
Create combat-details plugin

ErikOverflow:
This plugin adds 2 info panels with details about combat performance. It is similar to the PVP Performance tracker, but functions for all NPCs, and even tracks multi-combat performance.

Here is a screenshot containing both info panels and their details in their default configuration.
image


ErikOverflow:
I've removed the ability to display time in combat, as I saw some other concerns around the Watchdog timer and abuse for some bosses. 

This plugin tracks and displays total attacks/hits similarly to the pvp performance tracker plugin, but since the total number of attacks can be equally abused to track boss mechanics in an unfair manner, I've also removed that and only display overall accuracy as a percentage.

The above changes should remove any risk of abuse for counting/timing boss attacks.

PR #6059, opened 2024-05-23
Create DialogueOptionRebinder Plugin

Devdeve:
This is a plugin to allow different keybindings for dialogue with NPC's for example, using Q for option 1.


YvesW:
Please use a unique config group, not ``example``.

Devdeve:
🤦🏻‍♂️

PR #6068, opened 2024-05-24
Update Path marker 1.05

GeChallengeM:
Bugfix, update loc blocking data


YvesW:
You don't need to close your PR; it won't be merged if it doesn't pass CI. You can mark it as draft if you want to though.

PR #6076, opened 2024-05-24
Add httpx plugin

AirDancerX:
Add httpx plugin


Bram91:
This appears to be a 1:1 copy of <https://github.com/MorgApps/morghttpclient> with 3 lines of code removed. We'd prefer you try contributing to that project.

PR #6078, opened 2024-05-24
Create TonzoAPI

Garretonzo:
Hosts a simple local HTTP server that has endpoints for player stats, inventory, equipment


raiyni:
Unless you are making completely incompatible changes, we aren't keen to add it. We'd rather you try and contribute to an existing plugin.

PR #6079, opened 2024-05-24
Update click-to-minimize

WaylonJ:
This changes the plugin's config from:
key: value 
to
key: list\<value\> 

so that players can have multiple of the same action leading to other actions.

e.g. 
search: chest,
search: wardrobe

previously it would overwrite the first value causing it to fail.

### DIFF
--- a/plugins/click-to-minimize
+++ b/plugins/click-to-minimize
@@ -1,2 +1,2 @@
 repository=https://github.com/WaylonJ/OSRS_ClickToMinimize.git
-commit=a74fda91d879df931e97fe1278b5496485b6807f
+commit=40105daa568c34b000cb6fdffa67d03befaeaa15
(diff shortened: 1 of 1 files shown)


Alexsuperfly:
https://github.com/runelite/runelite/blob/master/runelite-client/src/main/java/net/runelite/client/util/Text.java#L82 ?

WaylonJ:
Removing jsoup and using that, thanks! 🙌
Ill have an update shortly

WaylonJ:
Updated the original issue to remove jsoup and fix duplicate action issue,

Included new changes in recent commit. Updated config to include 
- 'ignore case' toggle for actions, 
- 'ignoreFullInventory' toggle to not do anything if the inventory is full,
- Keybind for minimizing the window (useful when cutting gems or other menu context items).

WaylonJ:
Added a new config to allow functionality to work with menu options (e.g. grand exchange confirm menu)

Comparison of all changes to the current commit for this pr can be found here:
https://github.com/WaylonJ/OSRS_ClickToMinimize/compare/536c4473442d20c79772df7f5e64d792500142b9...ac730fa35a3751f746db234980bf9d33b2eb0692

PR #6080, opened 2024-05-24
Added Discord Xp and Kc Notifier plugin

Ischy-roos:


Felanbird:
Does [Dink](https://github.com/pajlads/DinkPlugin) not already fulfill both these features?

PR #6081, opened 2024-05-24
Updated afk canafis marks

powerus117:
Made some changes to the afk canafis marks plugin:
- No longer resets on game state changes. This was triggered when "LOADING" was triggered and caused the plugin to reset timings. Removed the game state subscription.
- Made leeway seconds configurable for unaccurate PC timings.
- Added feature request to swap menu entry when plugin suggests to wait to prevent jumping down the course unintentionally. Default disabled

Thanks for checking 🙏

### DIFF
--- a/plugins/afk-marks-canafis
+++ b/plugins/afk-marks-canafis
@@ -1,2 +1,2 @@
 repository=https://github.com/powerus117/RLAfkMarksCanafis.git
-commit=ecb775029561dce0013cd08dc5e7614a27f11794
+commit=222884ffde8884f57dbaa78746c81723d53f6f4a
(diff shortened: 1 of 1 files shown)


raiyni:
Don't do this https://github.com/powerus117/RLAfkMarksCanafis/blob/609d38d3be413562e66a801659ea21c13cb2aea7/src/main/java/com/afkmarkscanafis/AfkMarksCanafisPlugin.java#L167-L177 your plugin will no longer receive updates for events after you leave the area

TomBMNA:
But that's the idea no? To prevent unnecessary callbacks when not needed, to save performance. The events are only used for menu entry changes

Nightfirecat:
If your plugin is no longer receiving game tick events, how would it ever re-run that check to re-register to the eventbus?

TomBMNA:
Ah I didn't realize the game update was also tied to this eventbus. I'll remove it thanks

PR #6083, opened 2024-05-24
Create examine-plugin

sethks:
Adds a Right-Click option to NPCs that allows the user to pull up the relevant combat information about the NPC. It displays the combat information in the side panel.


ErikOverflow:
This appears to do something similar to what DPS Calculator already does. 
image

Additionally, I see your code leverages a custom lambda function in Azure, but there is no documentation on its owner or its function. If your code does something different than DPS Calculator, Can you provide some detail on its maintenance/support?

sethks:
I see it is similar to the DPS Calculator, thank you.

PR #6086, opened 2024-05-25
Update bank-tab-names - Minor Fixes, Removed Deprecated APIs, Added LICENSE

Psyda:
Minor fixes & Removed deprecated APIs

### DIFF
--- a/plugins/bank-tab-names
+++ b/plugins/bank-tab-names
@@ -1,2 +1,2 @@
 repository=https://github.com/Psyda/bank-tab-names.git
-commit=524316bc3b36abca7ca0ecbf84636f79f54b376a
+commit=92e26e049bb96acac210b507684ec2760497f65d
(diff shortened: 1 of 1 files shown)


YvesW:
IIRC your plugin will cause the tab names to be blank on install, right? We sometimes get support traffic because people don't understand that the plugin will cause their tab names to be empty with the default config. Could you fix that as well? E.g. show a chat message that points to the plugin when all tabs are blank.

Beagon:
> IIRC your plugin will cause the tab names to be blank on install, right? We sometimes get support traffic because people don't understand that the plugin will cause their tab names to be empty with the default config. Could you fix that as well? E.g. show a chat message that points to the plugin when all tabs are blank.



> IIRC your plugin will cause the tab names to be blank on install, right? We sometimes get support traffic because people don't understand that the plugin will cause their tab names to be empty with the default config. Could you fix that as well? E.g. show a chat message that points to the plugin when all tabs are blank.

I've created an issue for this https://github.com/Psyda/bank-tab-names/issues/8

PR #6096, opened 2024-05-26
Update `odablock` add config for emotes

DapperMickie:
After getting some complaints in Oda's Discord and in PMs I'm adding a config for the emotes I've released in the previous update.


DapperMickie:
The compare isn't 100% correct, I only added the config options, the rest was already in there...

PR #6109, opened 2024-05-30
Update Pet info

microtavor5:
Sped up the pets under cursor check using aabb


microtavor5:
I'll be honest, I have next to no idea how gradle works, but clearly something got messed up

PR #6114, opened 2024-05-31
Update clan-member-list-sort

erwinvc:
FPS improvement by @AbusiveTuna


plugins/clan-member-list-sort
```diff
@@ -1,2 +1,2 @@
 repository=https://github.com/jodelahithit/runelite-plugins.git
```
iProdigy, line 1:
```suggestion
repository=https://github.com/erwinvc/runelite-plugins.git
```

PR #6116, opened 2024-06-01
Update party-defence-tracker

MoreBuchus:
Adds elder maul and updated CoX defence scaling

Thanks to Umer and Jaayjee


jaayjee:
Can this be run again? The change depended on a modification to SpecialWeapon from https://github.com/runelite/runelite/pull/17895 . I just tested locally with the Runelite 1.10.31.3 and it's working as intended. The build should no longer fail.

PR #6121, opened 2024-06-02
Create limpwurt-login

Gavin-TC:
Simple plugin that plays a sound clip from YouTuber Limpwurt when the player logs in. Basically a modified "C Engineer: Completed" plugin which only plays a single sound.

### DIFF
--- a/plugins/limpwurt-login
+++ b/plugins/limpwurt-login
@@ -1,2 +1,2 @@
-repository=git@github.com:Gavin-TC/limpwurt-login.git
+repository=https://github.com/Gavin-TC/limpwurt-login.git
 commit=de4d6bb9ec280705f499316f49bfabc8149f2c78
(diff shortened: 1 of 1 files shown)


Nightfirecat:
I've not reviewed the code whatsoever, but do you have permission from limpwurt to use his voice clip in this way?

Gavin-TC:
> I've not reviewed the code whatsoever, but do you have permission from limpwurt to use his voice clip in this way?

For some reason I never thought about this, but I did go ahead and DM him on Discord and he said go for it.

Gavin-TC:
> Can you include said conversation? We can't take your word that you spoke to someone

Gotcha. Here's a screenshot:image

Nightfirecat:
You should not [use the C Engineer plugin's icon](https://github.com/Gavin-TC/limpwurt-login/blob/3906a2c888e94479c359bf8b41f5623c46be1b54/icon.png) for this plugin.
Code LGTM otherwise

Gavin-TC:
> You should not [use the C Engineer plugin's icon](https://github.com/Gavin-TC/limpwurt-login/blob/3906a2c888e94479c359bf8b41f5623c46be1b54/icon.png) for this plugin. Code LGTM otherwise

Good point, thanks for mentioning that!

PR #6122, opened 2024-06-03
readme: better clarify steps for running the test

Felanbird:
changed the wording a little bit and also moved the second sentence it's own number, both changes are for making sure you don't read past the full stop, and skip the running the test part of the step

also host the imgur link on github just because i'm already here


Felanbird:
this was more-or-less added in 207ee9c5b8518287ac1de967eb15cb2deb0e234a

PR #6124, opened 2024-06-03
Create loot-lens

Zzeetaa:
a simple plugin which adds a sound effect from the content creator lens when opening chests in the gauntlet

### DIFF
--- a/plugins/loot-lens
+++ b/plugins/loot-lens
@@ -1,2 +1,2 @@
 repository=https://github.com/Zzeetaa/loot.git
-commit=50795043d95b5f2b50747c57dea657ebd76e84aa
+commit=f050d97ef33e682abd533d25a46dfa9098eb7598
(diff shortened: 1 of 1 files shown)


Zzeetaa:
consent

proof of permission

iProdigy:
`LootConfig#greeting` is unused and you should not use `example` as the config group: https://github.com/Zzeetaa/loot/blob/f050d97ef33e682abd533d25a46dfa9098eb7598/src/main/java/com/loot/LootConfig.java#L7

<sup>perhaps the clip volume should be configurable via config</sup>

to read the wav resource, you should use `getResourceAsStream` because `new File("src/main/resources/sound/loot.wav")` will not work as explained here: https://github.com/runelite/plugin-hub?tab=readme-ov-file#plugin-resources

LlemonDuck:
Marking as draft for now. Please make the requested changes, then mark the plugin as ready for review.

MESLewis:
Closed at author request.

PR #6126, opened 2024-06-04
Create closed-door-highlighter

cabrilliant:
Closed Door Highlighter is a simple plugin that will highlight closed doors within a specified radius. This helps unattentive people like me remember to open doors before going into buildings.

### DIFF
--- a/plugins/closed-door-highlighter
+++ b/plugins/closed-door-highlighter
@@ -1,0 +1,2 @@
+repository=https://github.com/cabrilliant/DoorHighlighter.git
+commit=1ad2157bd4a20adaac241698b76b9b0b8026c5f4


raiyni:
I don't see the justification for this plugin other than using it for a color bot.

cabrilliant:
> I don't see the justification for this plugin other than using it for a color bot.

two things:

1) if i wanted to use it for a bot there'd be no reason to merge it into the offical client lol, why would i not just use a custom runelite build?
2) by that logic not sure why "object highlighter" is an allowed plugin as you can quite literally do what this plugin does with "object highlighter" but this just saves you from having to manually right click -> mark object on every single closed door in the game, which is obviously an annoying thing to have to do (esp since closed and open doors have different object ids)

This plugin is for unattentive people who go to click into a buliding and then look away for a second, look back and realize their character has  pathed somewhere irrelvant because a closed door was in the way. Honestly, im sure anyone who smokes the "devils lettuce" understands why theres a need for this plugin, its annoying and IMO it should be a base feature(really your guy should open doors on thier own FLAMEX). Some doors blend into walls really well. Id KIND OF undertsand your logic if it let you mark anything like "object highlighter"(but you guys already allowed that in lmao)  but how is something so niche gonna be some big botting tool lmao. Sorry not trying to be rude but like, i just dont see how this is even 1/50th as risky to be "only used for color bots" as some pre existing plugins that do WAY MORE.

PR #6129, opened 2024-06-04
Create clue-scroll-notifier

SoaresPT:
This plugin sends a notification or plays a sound when a clue scroll is obtained either by pickpocketing, fishing, woodcutting, or as a drop.


Nightfirecat:
Is there any reason in particular you're downloading the sound on plugin startup instead of including it as a resource? I'm aware some other plugins use this approach, however it is usually because they have a dynamic set of sounds, or periodically add/update sounds.

SoaresPT:
> Is there any reason in particular you're downloading the sound on plugin startup instead of including it as a resource? I'm aware some other plugins use this approach, however it is usually because they have a dynamic set of sounds, or periodically add/update sounds.

That makes sense. I will fix on the next PR

PR #6133, opened 2024-06-05
cg-recolor

denaelc:
A plugin that recolors the corrupted gauntlet.


raiyni:
This plugin won't work when packaged as a jar for live. You need to load your files through resource streams.

PR #6140, opened 2024-06-07
Add RsPaint plugin

geheur:
an initial version. none of gamma's crazy ideas. hopefully later.


raiyni:
What's the difference with this and `Brush Markers` ?

PR #6142, opened 2024-06-07
Update partial sets

Fiskmans:
Closes: https://github.com/fiskmans/rl_partial_sets/issues/1
Organize settings into tabs with content sorted alphabetically
Add tracking for the crystal sets used within the gauntlet (and corrupted gauntlet)
Add info to tooltip about which items from the set the player is or isn't wearing


Fiskmans:
image
New tooltip

PR #6143, opened 2024-06-08
Add auto-tags plugin

i:
Automatically assigns inventory tags to gear based on item stats


Nightfirecat:
There's nothing strictly against it, but usually plugin names are Title Cased, not lowercased and hyphen-separated: https://github.com/i/rl-plugins/blob/b86ca7acc5dd2224c6ff6736e3f9e2d3cd8bcace/src/main/java/com/github/i/autotags/AutoTagsPlugin.java#L35

Code LGTM so it can be merged as-is, though you may want to update the linked line.

PR #6147, opened 2024-06-09
chore: update region-chat commit hash

WoutDev:
@Zoinkwiz hopefully we can do this with your blessing 🙏 

I want to use this plugin at Zalcano 😅


Felanbird:
I'm fairly sure Zoinkwiz knows how to open update PRs here when he wants to update his plugins
image

PR #6161, opened 2024-06-13
 Update Companion Pet Plugin- Update for compatibility with WorldView.

Mrnice98:
Update Companion Pet Plugin- Update for compatibility with WorldView. 
(Very small change, Plugin currently not working without fix)


Felanbird:
Can you please scale down your [nav_icon](https://github.com/Mrnice98/Companion-Pets-Plugin/blob/master/src/main/resources/nav_icon.png) image, it's very large due to being blown up to it's max size when Java loads it into memory (width * height * 4)

PR #6171, opened 2024-06-15
[Fix] Updated bank-tag-generation version

steveacalabro:
The version of this plugin was out of date and the original develop seems to have gone MIA. So I decided to step in

I tested commit 491d7054b3dfb34f66e7c30ed66712a973086128 locally on my machine which is marked as the latest commit on the [original repo of the plugin](https://github.com/MitchBarnett/wiki-bank-tag-integration). This commit appears to work as expected with no errors.

I updated the commit to match this in this PR

### DIFF
--- a/plugins/bank-tag-generation
+++ b/plugins/bank-tag-generation
@@ -1,2 +1,2 @@
 repository=https://github.com/MitchBarnett/wiki-bank-tag-integration.git
-commit=233f143f9994558bb44696411eccc6fd74d92d88
+commit=491d7054b3dfb34f66e7c30ed66712a973086128


Felanbird:
Please don't close PRs to resolve issues, you can just push the new commit to this existing PR

steveacalabro:
@Felanbird Apologies, didn't know that was the practice. I just re-opened it now. Thanks :)

Felanbird:
> I do have to wait for my fix in the plugin repo to get merged before this PR can be reviewed

If you want to take over this plugin you should attempt to contact the author, they have their email listed on their Github profile. 
After that we can go ahead and merge a PR updating the plugin to a new commit & your repo depending on the response.

Otherwise we aren't going to accept a change submitted by somebody else, which was committed to the repo 2 years ago, if the current author wanted this change to be live they would have submitted it to us.

steveacalabro:
@Felanbird Ok in that case, I think I'm going to just move these features into a similar plugin that is built with up-to-date standards. To be honest, I don't like that this plugin currently uses chat commands. I would rather see input's being done in the side panel. I think I am going to close this indefinitely in about 48hrs. I am only waiting because I did just email the original dev to see if they would respond

YvesW:
We prefer that you contribute to existing plugins rather than create a new one, to prevent pluginhub bloat and so users can benefit from your updates and fixes. Please contact the original author via any way possible so they can add you as contributor.

You should:
- Create a PR with in their repo and ask to be added as contributor to the repo.
- Contact the author via any way possible:
  - @ them in the RL discord if you know their discord name and ask to be added as contributor
  - Check their GH for contact info and use that to ask them if you can be added as contributor
  - Check their commits for their email so you can ask them via email if you can be added as contributor
  - Can create an issue as well asking to be added to the repo as contributor

Nightfirecat:
Failing all of the above, we would be willing to transfer ownership of the plugin to you so existing users can have their installs updated automatically rather than needing to install a different, new plugin.

Alexsuperfly:
Since this PR doesn't even build, and it looks like you never circled back after claiming to make an attempt to contact the original author I'm just going to close this.

PR #6191, opened 2024-06-18
Create local-consciousness

YeahImRose:
Local Consciousness is a plugin that lets you make an item bounce around on your screen.

Hugely inspired by Global Consciousness, which was removed from the Plugin Hub due to no longer working and not being maintained.

### DIFF
--- a/plugins/local-consciousness
+++ b/plugins/local-consciousness
@@ -1,2 +1,2 @@
 repository=https://github.com/YeahImRose/local-consciousness.git
-commit=e3cee9dddd3aadfcdb31344af53f4a1bc2f390f3
+commit=aeb912d33236507bd96ec58ac9cba2d44c822192
--- a/plugins/local-consciousness
+++ b/plugins/local-consciousness
@@ -1,2 +1,2 @@
 repository=https://github.com/YeahImRose/local-consciousness.git
-commit=7f13fbf6675080df28c6eeccae253a21844f5800
+commit=e3cee9dddd3aadfcdb31344af53f4a1bc2f390f3
--- a/plugins/local-consciousness
+++ b/plugins/local-consciousness
@@ -1,2 +1,2 @@
 repository=https://github.com/YeahImRose/local-consciousness.git
-commit=aeb912d33236507bd96ec58ac9cba2d44c822192
+commit=f033ae3c82015349f25f9d1efb9a67a380efd448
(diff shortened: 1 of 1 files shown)


YvesW:
Does this plugin do anything that's significantly different from Global Consciousness?

YeahImRose:
> Does this plugin do anything that's significantly different from Global Consciousness?

Not at the moment, but I couldn't locate the Global Consciousness plugin's repo, so I wrote this from scratch from my friend's memory of its functionality

YeahImRose:
I looked a bit further and found the original repo, I'll try to reach out to the old dev and see if they plan on fixing the problem or if I should PR them/if they're okay with my replacement plugin.

YeahImRose:
Couldn't find a contact email so bringing the discussion to this thread.
@JacobLindelof Hello, I know you developed [the original plugin this is based on ](https://github.com/JacobLindelof/global-consciousness-plugin), but do not seem to be interested in maintaining, would it be alright by you if I maintain my iteration of the concept and credit you for the original idea? Thanks!

Nightfirecat:
FYI you can find their email from their commits, eg. <https://github.com/JacobLindelof/global-consciousness-plugin/commit/977f4c5b63d6deb89cd6fa590d522ddfa161b293.patch>

YeahImRose:
> FYI you can find their email from their commits, eg. https://github.com/JacobLindelof/global-consciousness-plugin/commit/977f4c5b63d6deb89cd6fa590d522ddfa161b293.patch

Oh true, didn't even cross my mind. I'll send the same message to them on email and link them here. Thanks!

YvesW:
You are adding the overlay twice.

YeahImRose:
> You are adding the overlay twice.

Oops, thanks for the heads up!

YeahImRose:
Done with features, just waiting on permission since I still haven't back from the creator of global consciousness. If that gets the okay then it should just need a once over review and then be good for publishing.

YeahImRose:
Well, it's been over a month since I reached out to the creator of the original plug-in, and still no response. They're clearly still around since they've made commits to a private repo since I've tried to contact them, just not responding.

Is there a way to check what was said to them before their plug-in originally got removed? I assume they were contacted to update it once it broke but clearly they didn't bother, so can we consider it abandoned yet?

PR #6192, opened 2024-06-18
Readd config for emotes `Odablock`

DapperMickie:
After my last update, I forgot to add these emotes to my main branch so the last update removed these config items, just trying to add them back because I'm being flooded with dms about this.


DapperMickie:
Build has been fixed :)

PR #6195, opened 2024-06-19
Update Stardust Per Hour Tracker

kevindarbydev:
impl resetConfiguration(), text color switching.
also added comments, updated title in PluginDescriptor to 'Stardust Per Hour'


raiyni:
needs rebased

PR #6197, opened 2024-06-20
Create pest-control-notifications

Cryslacks:
This plugin adds the ability to get notifications about different things regarding the Pest Control minigame, such as:
- Round start
- Round end
- Shield down for each of the portals
- Activity going below a limit


raiyni:
If you don't plan on using your overlay for anything, remove that please

PR #6200, opened 2024-06-21
Create httpeventserver-reloaded

NathanDunahoo:
This Plugin provides game and character insights on your localhost


Adam-:
We aren't accepting new plugins which expose character information over http because they are only used by color bots and aren't worth our time looking at.

PR #6205, opened 2024-06-22
Added gauntlet performance tracker plugin

powerus117:
As described on the plugin page, this plugin adds an overlay that shows statistics about your fight with Hunllef in the (corrupted) gauntlet. It will not show anything until Hunllef is fought and will show a number of statistics mentioned on the plugin page.

Thanks for checking in advance 🙏


YvesW:
Please use a unique config group, not [example](https://github.com/powerus117/RLCGPerformanceTracker/blob/a92408c6ad6b7c7beffaced8dd59e407dfe49956/src/main/java/com/gauntletperformancetracker/GauntletPerformanceTrackerConfig.java#L7).

PR #6210, opened 2024-06-23
Create nature-runecraft-utilities

alex-macm:
This plugin swaps the achivement diary cape left click option to teleport to Jarr when standing at the nature altar.

### DIFF
--- a/plugins/nature-runecraft-utilities
+++ b/plugins/nature-runecraft-utilities
@@ -1,0 +1,2 @@
+repository= https://github.com/alex-macm/Nature-Runecrafting-Util.git
+commit=820856abbc7491120431e4a6607e609dc9097e38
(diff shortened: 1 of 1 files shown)


YvesW:
What if you just left click to TP to Jarr, and shift-click to teleport to Kaleb, using MES?

alex-macm:
> What if you just left click to TP to Jarr, and shift-click to teleport to Kaleb, using MES?

MES left click to kaleb, and then this overrides the left click at the altar to Jarr. no shift click required, for maximum reclined gaming.

LlemonDuck:
Use a more specific config name that is less likely to conflict with other plugins (e.g. `naturerunecraftutilities`)

https://github.com/alex-macm/Nature-Runecrafting-Util/blob/820856abbc7491120431e4a6607e609dc9097e38/src/main/java/com/NatureRCUtil/NatureRCUtilConfig.java#L34

alex-macm:
> Use a more specific config name that is less likely to conflict with other plugins (e.g. `naturerunecraftutilities`)
> 
> https://github.com/alex-macm/Nature-Runecrafting-Util/blob/820856abbc7491120431e4a6607e609dc9097e38/src/main/java/com/NatureRCUtil/NatureRCUtilConfig.java#L34

changed

geheur:
Did you try getting this added to runecrafting utilities instead of as a standalone plugin? It feels like it would fit there perfectly.

PR #6212, opened 2024-06-24
Adding Quest Voiceover Plugin 

KevinEdry:
# Quest Voiceover Plugin
This plugin aims to provide voice over acting to the brilliant quest writing in Old School Runescape as a Runelite plugin.

### Example Video
https://github.com/KevinEdry/runelite-quest-voiceover/assets/42734162/3ecd069a-e26a-4a0f-9d7e-36af3b157a88

### Technical Notes
Im adding here some technical notes explaining how the plugin works to streamline the review process (We doing this at work, theres no reason for people not to include those here).
- The plugin streams mp3 audio files from the plugin's Github repo, specifically from a branch called `sounds` (see `SoundEngine` for how it runs - [permalink](https://github.com/KevinEdry/runelite-quest-voiceover/blob/dd895fecb5c2df839548a19266895d7f0ee721b7/src/main/java/com/quest/voiceover/SoundEngine.java#L30).
- We use a standard Jaco player to play the sounds, specifically the jar was taken from an already [published plugin repository](https://github.com/techgaud/TTS/tree/main/libs).
- Sounds were generated with Elevenlabs offline, and then uploaded to the Github branch.
    - I was looking through some of the other plugins pull request and saw that you put an emphasis on sound copyrights, so for the records, all of the generated sounds are from the default Elevenlabs library, no streamer voice was used.
- In the repo there is a python tool to generate those sounds, this might be a point of friction because it is not purely just the plugin, but the tool is minuscule and shouldn't contribute to the plugin size in any meaningful way.
    - I'd rather keep it in the repo as this is will enable the community help with the quest converge efforts.
    - You can refer to the [README](https://github.com/KevinEdry/runelite-quest-voiceover/blob/main/README.md#usage) to see the tool's functionality.
- Because the OSRS wiki transcripts are not written exactly like the text is served in the game and because there is no feasible way (that I could think of) to distinguish between those dialog lines, I've set up a versionable SQLite server that is being downloaded ones someone installs the plugin or onces the db version becomes outdates, and then fuzzy seaches the text to find the most similar one by the same NPC.
    - This is the best way I found to reliably check if a quest dialog is  available (has been generated) and to assign the "in-game" text to the wiki transcript text.

- The plugin config includes two option at this moment at `QuestVoiceoverConfig` [permalink](https://github.com/KevinEdry/runelite-quest-voiceover/blob/main/src/main/java/com/quest/voiceover/QuestVoiceoverConfig.java):
    - Volume 0-100
    - Mute
- Laslty, I am adding a "Mute" button on the top right of the dialog widget, invoked [here](https://github.com/KevinEdry/runelite-quest-voiceover/blob/dd895fecb5c2df839548a19266895d7f0ee721b7/src/main/java/com/quest/voiceover/QuestVoiceoverPlugin.java#L109), and implemented [here](https://github.com/KevinEdry/runelite-quest-voiceover/blob/dd895fecb5c2df839548a19266895d7f0ee721b7/src/main/java/com/quest/voiceover/DialogEngine.java#L44).

This plugin code isn't perfect, but its a good starting point.

<br>
<b>Thanks for your time and effort reviewing my plugin.</b>


KevinEdry:
@Adam- Hey Adam, we talked a bit in the discord about incorporating SQLite in a plugin.
The `xerial/sqlite-jdbc` bumps it 3mb~ over the limit.

Adam- (changes requested):
pull sqlite in via build.gradle as a normal dependency instead of vendoring it.

KevinEdry:
> pull sqlite in via build.gradle as a normal dependency instead of vendoring it.

Changed it so it will now pull `build.gradle` as a normal dependency.

Adam-:
> > pull sqlite in via build.gradle as a normal dependency instead of vendoring it.
> 
> Changed it so it will now pull `build.gradle` as a normal dependency.

Your plugin has sqlite pulled in as `testImplementation`, but it will need to be either implementation or runtime for it to work outside of the test.

KevinEdry:
> > > pull sqlite in via build.gradle as a normal dependency instead of vendoring it.
> > 
> > 
> > Changed it so it will now pull `build.gradle` as a normal dependency.
> 
> Your plugin has sqlite pulled in as `testImplementation`, but it will need to be either implementation or runtime for it to work outside of the test.

Oh I didn't know that was required, I've changed it to `implementation`  and tested the functionality to confirm that everything is working.

iProdigy:
fyi the verification-metadata.xml change should happen in this repo rather than your plugin repo

KevinEdry:
> fyi the verification-metadata.xml change should happen in this repo rather than your plugin repo

Thanks for pointing it out!, it took me a while to figure it out.
I've also added it to the `build.gradle` (its not documented, but I've seen another PR doing the same).
Hopefully this would now be approved.

Adam-:
Your plugin leaks prepared statements, which I think leaks native resources in the jdbc driver, don't do this.

KevinEdry:
> Your plugin leaks prepared statements, which I think leaks native resources in the jdbc driver, don't do this.

Yup, I've added them to a try-with-resource statements, that way they will close automatically once we are done with them.

PR #6218, opened 2024-06-26
Tile Packs Update

TrevorMDev:
- Refactor - pull panel code out into small separate panel pieces
- Only load data from config once, rather than on every update
- UI Update, new layout and icons. This is to make space for more buttons.

before

after


raiyni:
This page doesn't exist https://www.flaticon.com/free-icons

Unmoon:
[Attribution requirements](https://support.flaticon.com/s/article/ka03V0000008Xs3QAE ) for the icons seem to be not met.

PR #6221, opened 2024-06-26
Update wasted-bank-space

mcgeer:
Updated for plugin break after id change in https://secure.runescape.com/m=news/colosseum-npc-clickboxes--more?oldschool=1


mcgeer:
Speedy!

PR #6228, opened 2024-06-29
Create arduino-idle-notifier

HeyImJ0hn:
This plugin allows the user to run a command on a connected arduino when the player goes into idle, like turning on an LED light or triggering a servo


iProdigy:
re: `jSerialComm` - https://github.com/runelite/plugin-hub/pull/3596#issuecomment-2052552984

PR #6230, opened 2024-06-30
Create zoom-keybinding

R4lfE:
A plugin to set custom zoom keybings other than the scroll wheel.
My scroll wheel broke and as far as I know there is no option or plugin to do this yet.


YvesW:
FYI, you could also just rebind the key using AHK if it's just because your scrollwheel broke (or get a new mouse; good mice are really cheap nowadays).

PR #6244, opened 2024-07-04
Update first-person to 1.0.2

Zoinkwiz:
Adds the GpuPlugin's content, to allow for the scene to be rendered through that as an alternative to the detached camera mode.

I'd appreciate if the following Jagex guidelines can be considered when reviewing this PR, as I think this could count as conflicting with one of the following. I can't imagine the plugin itself providing any benefit to players, given it's just effectively a more cleaner version of zooming in a lot, but it might still fall under these:

* Any movement or resizing of click zones for 3d components

Because the camera in the GPU mode is different from what is it actually is in the game, I could imagine this being in conflict.

* Offers world interaction in any detached camera mode 

This doesn't provide interactions in detached mode, but having the camera seem like it's somewhere its not is kinda similar in effect.

### DIFF
--- a/plugins/first-person
+++ b/plugins/first-person
@@ -1,2 +1,2 @@
 repository=https://github.com/Zoinkwiz/FirstPerson.git
-commit=b7f25373188cb89666c69b73839318659fe09c76
+commit=464a774299b06146d65dd9932160d0f9db5bea21


LlemonDuck:
What's the point of consuming all these keycodes? https://github.com/Zoinkwiz/FirstPerson/compare/b7f25373188cb89666c69b73839318659fe09c76...Zoinkwiz:464a774299b06146d65dd9932160d0f9db5bea21#diff-137be197120b01c69561b724e6f5ffbd89d9a4df8706432bc583471bfecd7f3aR166-R183

Zoinkwiz:
> What's the point of consuming all these keycodes? [Zoinkwiz/FirstPerson@b7f2537...Zoinkwiz:464a774299b06146d65dd9932160d0f9db5bea21#diff-137be197120b01c69561b724e6f5ffbd89d9a4df8706432bc583471bfecd7f3aR166-R183](https://github.com/Zoinkwiz/FirstPerson/compare/b7f25373188cb89666c69b73839318659fe09c76...Zoinkwiz:464a774299b06146d65dd9932160d0f9db5bea21#diff-137be197120b01c69561b724e6f5ffbd89d9a4df8706432bc583471bfecd7f3aR166-R183)

Those are used in the detached camera mode for things such as moving the camera up and down, rotating it, and so on. If they're not disabled you can potentially get weird behaviours with the camera jittering from the position it's moved to by A, and then back to the forced perspective from the plugin.

PR #6245, opened 2024-07-04
Update Tick Fixer for Mac to v1.1.0

jonathangarelick:
configurable ip and ping interval
do not ping when logged out
resolve plugin restart issues with executor (thanks @Molorius)


Nightfirecat:
It's more likely that github is having issues, not the bot. See https://www.githubstatus.com/

PR #6250, opened 2024-07-06
Update dt2-mistake-tracker

DominickCobb-rs:
Add Vardorvis pillar hider (thank you @JZomDev)
Add Whisperer projectile swap -- Took code from @InfernoStats and their Vardorvis Projectiles. Licensing should be correct where applicable, will correct if needed. Per their request I have not integrated their plugin into mine and added a note in the config to use their plugin.

E: Rebranded to DT2 Boss Utilities as well.


aHooder (approved):

iProdigy (approved):
LGTM but you may want to call `vardorvisPillarHider.getPillarTiles().clear()` on `dt2pbfPlugin#shutDown` (within the `clientThread.invoke` task) for the edge case where the plugin is disabled mid-fight

PR #6251, opened 2024-07-06
Added casket-saver

Notespeon:
Added casket-saver plugin to plugin list.
It changes open to use on caskets if you already have a master clue.


YvesW:
Please use a unique config group, not example.

Notespeon:
oops

PR #6252, opened 2024-07-06
Update "Dude, Where's My Stuff?" to v2.7.0

Thource:
Release notes: https://github.com/Thource/dude-wheres-my-stuff/releases/tag/v2.7.0


Thource:
Sorry, trying to create a PR for hotfix and accidentally pushed to this branch

Felanbird:
@Thource as a reminder you pushed this branch to have no changes, also I pushed this plugin back 1 commit to before the cat_h changes, just so you are aware when/if you resolve this PR.

Thource:
Thanks @Felanbird, I believe this should be ready now

LlemonDuck:
Is there a reason you need the newer gsheets dependency? Does `v4-rev581-1.25.0`, which is already approved, no longer work?

Thource:
We thought it’d be best to use an updated version, rather than the 2019 version. See here:   https://discord.com/channels/301497432909414422/419891709883973642/1259902701165351053
  
  
  
  
  
  
  

  
  
>

Thource:
We tried to downgrade our `google-api-services-sheets` to the already accepted version, but doing this would have meant that the plugin users would need to create their own Google API keys and paste them into the plugin config, rather than just using Google's OAuth flow, which does not seem like a good user experience.

abextm:
this is not supposed to be public https://github.com/Thource/dude-wheres-my-stuff/blob/19dc452292300d48d96ce0fa14a905ee6409c875/src/main/resources/credentials.json#L8

abextm:
Hmm, OK. I would expect Google to use a pkce flow without a client secret, but the docs seem to indicate otherwise, so that is okay I guess. Obfuscating it wouldn't be useful or add anything, so this seems fine.

Do you know its excluding grpc is possible? It looks almost entirely unused.

abextm:
not blocking, but you should use int[] here. https://github.com/Thource/dude-wheres-my-stuff/compare/749e63ef6871ae27d45b5ac401e1b716e1c95fae...Thource:19dc452292300d48d96ce0fa14a905ee6409c875#diff-6e5d858f33efa0e0391d79d78318fb1423603c06103c4486f44d96851f83b364R14

PR #6258, opened 2024-07-08
Add Prayer or Restore plugin.

Gnuder:
Plugin adds a side panel which, when clicked on, displays a sorted list of the current wiki price per does of all prayer pots (1-4) and super restores (1-4).

This plugin can help decide if it is worth buying  prayer potions or super restores.

### DIFF
--- a/plugins/prayer-or-restore
+++ b/plugins/prayer-or-restore
@@ -1,0 +1,2 @@
+repository=https://github.com/Gnuder/prayer-or-restore.git
+commit=6b8966bb7f76628a1b7f5a1c7d840e5ce89160e9


raiyni:
I don't see a reason to add this plugin when the GE search panel exists.

raiyni:
I'm not sure if this is for merching, or not, but there is already flipping plugins that can be used for that.

Gnuder:
It is simply to help determine at a quick glance if someone wants to buy prayer potions or super restores. There are times when super restores end up being priced lower than prayer potions, even though super restores restore an additional +1 prayer point. Is simply meant to help quickly inform that decision if a player thinks it worth to spend for the extra prayer point of a super restore.

PR #6275, opened 2024-07-12
Create tictac7x-deposit-worn-items

TicTac7x:


raiyni:
The bank plugin already does this?

TicTac7x:
Yes, but it still opens the menu on left click, which i don't want to see. My plugin also adds lock icon to clearly see that left click is disabled, which the bank plugin does not do.

tylerwgrass:
Your desired changes would make more sense as proposed to the bank plugin instead of creating a new plugin

TicTac7x:
Maybe, but last time i opened PR for core plugin, the PR stayed idle for 2 years until it got rejected. Somif possible, I would like to go forward with this plugin.

PR #6277, opened 2024-07-12
Update tile packs

TrevorMDev:
- Added custom configs storable per tile pack. This will allow users to customize each tile pack more.
- Added first per pack config, a visibility toggle. This allows users to hide the packs from the list.
image

- Added global filter that allows the user to control if they the list shows the visible, hidden, or all tile packs.
image
- Update README with useful information.

### DIFF
--- a/plugins/tile-packs
+++ b/plugins/tile-packs
@@ -1,2 +1,2 @@
 repository=https://github.com/TrevorMDev/tile-packs.git
-commit=704e10891abb4b40cd4c035d5917cf15aa642ac4
+commit=6940f003dd3092358a88fd4444d37e138677d614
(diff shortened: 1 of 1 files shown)


aHooder (approved):
All looks good here, but I have one minor suggestion. If you disable HTML for [your `TilePackPanel`'s `JLabel` here](https://github.com/TrevorMDev/tile-packs/blob/6940f003dd3092358a88fd4444d37e138677d614/src/main/java/com/tilepacks/ui/panel/TilePackPanel.java#L97) like so:
```java
packName = new JLabel();
packName.putClientProperty("html.disable", Boolean.TRUE);
packName.setText(tilePack.packName);
```
it won't be possible for your users to accidentally leak their IP by pasting malicious HTML into a custom pack name. Setting the text *after* the hint to disable HTML is important btw.

TrevorMDev:
Thanks for the suggestion, I've added it in. I also fixed a bug I found while testing it where newly added packs were enabled, but the point overlays were not being rendering.

aHooder (approved):

iProdigy (approved):

PR #6284, opened 2024-07-13
Update combat-logger to v1.1.0

SuperNerdEric:
- Add a built-in damage meter
- Log Damage messages from Party
- All logs now go in a queue with a 2 tick delay before they are written so that previously unknown hitsplat sources can now be matched up with a Party members hitsplat
- Create a death log for end of Verzik P1


iProdigy:
on shutdown, you are missing `EventBus#unregister` for `LogQueueManager`

SuperNerdEric:
> on shutdown, you are missing `EventBus#unregister` for `LogQueueManager`

I just pushed a fix for this. Thanks!

PR #6293, opened 2024-07-14
Update Multi lines

tsbreuer:
Performance improvements (arround 65% frame time usage reduction, from 24% to 8%)

Compare between commits for ease of review:

https://github.com/tsbreuer/Multi-Lines/compare/6eca6fe3031ec62a08d1a5ef9e073a286d8c99db...9949ec4a0683456f08da293d06e2aa0c50962143


YvesW:
It's probably a slight bit easier for reviewing purposes/cleaner to put the json file in a separate branch; that way it doesn't show up in the diff.

tsbreuer:
> It's probably a slight bit easier for reviewing purposes/cleaner to put the json file in a separate branch; that way it doesn't show up in the diff.

I dont think it matters too much if you just ignore the file, and having it on a separate branch could turn out to be confusing for people just looking for it for the geojson. Just doesnt make that much sense to have it on a separate branch for the sake of it not being on the diff comparison.

Felanbird:
Github's UI can be very terrible and if the reviewers browser freezes because there's x amount of JSON changes, it's very likely the review may get thrown further back into the queue.

I don't think what Yves has suggested is required, but it would benefit your updates and our review process.

Nightfirecat:
(for what it's worth, it does not matter to me personally--I will generally ignore json file changes as they are non-code changes)

raiyni:
Honestly doesn't matter as long as it doesn't grow too huge.  But it *should* be in a resources folder if you are wanting to be pedantic about java.

iProdigy:
You have numerous `System.out.println` calls that are not commented out

tsbreuer:
Did all the changes mentioned in tihs PR.

PR #6294, opened 2024-07-14
Create pet-to-npc-transmog

Tony905x:
Customize your pets appearance to be any NPC/Object

### DIFF
--- a/plugins/pet-to-npc-transmog
+++ b/plugins/pet-to-npc-transmog
@@ -1,2 +1,2 @@
 repository=https://github.com/Tony905x/pet-to-npc-transmog.git
-commit=dd74e50d1e0d1df1e953a945d523f7a804896435
+commit=872998f7cef6fedbc436934353be41db2e0245d0
--- a/plugins/pet-to-npc-transmog
+++ b/plugins/pet-to-npc-transmog
@@ -1,2 +1,2 @@
 repository=https://github.com/Tony905x/pet-to-npc-transmog.git
-commit=68598bdf38c4e8542cf498055fc51bfb2447810c
+commit=50666c8921ec405773fd7267531eb6f3072cb018
--- a/plugins/pet-to-npc-transmog
+++ b/plugins/pet-to-npc-transmog
@@ -1,2 +1,2 @@
 repository=https://github.com/Tony905x/pet-to-npc-transmog.git
-commit=872998f7cef6fedbc436934353be41db2e0245d0
+commit=68598bdf38c4e8542cf498055fc51bfb2447810c
--- a/plugins/pet-to-npc-transmog
+++ b/plugins/pet-to-npc-transmog
@@ -1,0 +1,2 @@
+repository=https://github.com/Tony905x/pet-to-npc-transmog.git
+commit=dd74e50d1e0d1df1e953a945d523f7a804896435


raiyni:
This seems like a better addition to companion pets.

Tony905x:
I believe companion pets is to transmog a pet to any other pet.  This one is to transmog into any NPC.  wise old man, gnome child, konar ect, and even massive ones like leviathan.

raiyni:
This would be a better addition to companion pets because it's nearly the same plugin and would provide more options for the already existing userbase.

Tony905x:
I downloaded the plugin check but it only appears to work when you summon a thrall, not for actual pet. unless I am missing something

aHooder:
> I downloaded the plugin check but it only appears to work when you summon a thrall, not for actual pet. unless I am missing something

It currently only does that, but I believe what raiyni is suggesting is that you should instead try to get your features added to the companion pets plugin, i.e. by submitting a pull request to their plugin over here: https://github.com/Mrnice98/Companion-Pets-Plugin. If the author of that plugin is happy to oblige, it's a win for the plugin users in the end. That said, if there are disagreements between you and that plugin author, I would think a standalone plugin could be reconsidered :+1:

Tony905x:
I think the goal of the plugins as well as the code is a bit too different.  pet transmog to NPC vs thrall transmog to pet.  One is changing away from a pet while other is changing to become a pet.  My plugins current functionality is having a permenant pet transmog of any NPC by way way of entering ModelID's and Animation ID's that.  The end goal with later changes being the NPC's animations and overhead changing based on what the player is doing.  In order to add these functionalities to the other plugins would require too many changes and it probably won't align with what the other plugins owner wants with his.  If you think it will confuse people having 2 plugins for pets I could maybe change the name of mine to make it less confusing?  maybe not even mention pets at all

raiyni:
The rational is the hub is bloated and we would rather provide enhancements to existing plugins where possible. Encouraging collaboration with developers is the first option.

Tony905x:
I understand. In this scenerio though I don't think it would be possible as it would likely either restrict allowing any NPC to be used or it would require drastic code change on other plugins owner to allow custom entry rather then selecting from a list of available pets.  In this post I put on reddit showing the plugin I mentioned that any NPC can be used and I'd worry that wouldnt be possible if combined into the other plugin.   https://www.reddit.com/r/2007scape/comments/1dyfuvg/i_created_a_plugin_that_lets_you_change_your_pet/

raiyni:
I don't see how it wouldn't be possible. Maybe contact them and ask?

Tony905x:
I get it that there is a ton of plugins. However, I’m planning to add more things like giving the NPC custom overhead texts based on the NPC’s actions like combat, leveling, etc. This wouldn’t fit with a plugin that focuses on temporary thralls appearing as pets.  Merging the two plugins would require changing the idea of the existing plugin, which I don't think would be fair to the person that created it.  It would also need a lot changes to my code to align with the other.  Keeping both separate would mean both plugins can do what they are intended to. I hope this makes sense, I'm open to changes that could allow the two to function seperate though

aHooder:
I can empathize with that, and personally I don't think there's much harm in allowing this as a separate plugin.

It wouldn't hurt to hear what @Mrnice98 (author of Companion Pet) thinks about this though. The TL;DR is @Tony905x wants to allow replacing your own pet with one of a few NPCs from a list, and wants to add contextual overhead text to these NPC companions further down the line. Would this be something suitable for addition to the Companion Pet plugin, as some sort of collab?

raiyni:
The description of the plug-in implies it either works with pets or you can summon a pet if you don't have one. I'd think this follows under the same premise.

Tony905x:
ya I see what you mean "Allows you to spawn any pet in the game (including transmogs) which will follow you around and act nearly identical to the real thing or replace your thralls."  but I'm trying it and it seems to be for thralls only, if you want to give it a try yourself in case I am missing something.  Also it references changing into a pet where as I am changing into an NPC or even an object.  If it is intended to allow change pet though it would still be different in my opinion, his would actually be possible in game if you got the pet where as making it a NPC is kind of a goofier and gimmicky experience especially once I add my planned functionality

Mrnice98:
@Tony905x @aHooder 
So I have no problem if he wants to make his own plugin that transmogs existing real pets into NPC's. 

My plugin does allow people to spawn "fake pets" which follow the player (click the whistle with no pet following you), along with replacing thralls with pets of your choice. In the future the plugin will allow people to use custom models, such as player models or Npc's to use as "fake pets" or to replace thralls. I have held off expanding in that direction as internal RL changes need to be made to allow for proper application of animation on scaled RL objects. 

In short, if his plugin is just replacing the existing pet models for pets people already have with NPC models I don't have any major issue, I'm not here to gatekeep. 

(I would however urge you to ensure that you have tested out all the models your after work properly, as for many models once scaled down the animations will be not applied correctly until the RL internal changes are made. )

raiyni:
The question isn't really about competition but if you would be open to accepting contribution for it.

Mrnice98:
Sure id be open to it, would depend on what exactly he would be looking to do. If he wants to just kind of allow the existing pets to be overridden with only NPC's. It could be a relatively easy addition, but I'm hesitant to make changes like that before it can even support all the pets in game in its core functionality as currently limited by the animation scaling issue for RL objects.

Mrnice98:
After testing out the plugin now, it seems like it would be a fine addition if it wanted contribute to the companion pet plugin as its very similar functionality and its something id likely be adding in the future regardless post animation scaling changes. 

Some things need cleaning up though. When you have the plugin on you cant pick up your pet and therefore are stuck with the NPC following you,  when you shut the plugin off the RLObject still remains in place, and a few other things.

Tony905x:
Hey @MrNice, @aHooder, @raiyni I appreciate the offer to contribute and can understand the reason behind it.  My concern is losing control of this as it has been a means for me to learn JAVA, I'd probably lose motivation to add onto it as I wouldn't be able to put changes I had planned without approval and making sure it aligned with another plugin.  My reasoning for doing this exact plugin was because there wasn't a plugin that currently allowed for changing a pet into a NPC and I thought it would be hilarious to be able to alter their interactions with the game based on whats happening in the game.  I've been having a ton of fun making this and I had a lot of plans.  @Mrnice98 I understand that you may have had future plans for NPC's but this isn't something I could have known about.   I am hoping it will be ok for this to go in seperately because it's been such a fun way to learn JAVA and would love to keep adding onto it with the plans I had.

Mrnice98:
of course man, I have no issue with making the plugin. That was more to say, if in the future the internal RL changes are made to fix animation scaling, don't be surprised if somewhat similar features are added into the companion pet plugin. (Wouldn't be the exact same) 

for what its worth I used plugins as a way to learn Java so I understand your desire.

Tony905x:
Thanks, I appreciate it!

PR #6297, opened 2024-07-14
Update ruby-counter

hyppis1:
Added quiver ammo slot to work for tracking. Also added "Attacks that hit" tracking, and overall accuracy with that. Added "attacks that hit" tracking to saved data also


YvesW:
FYI the quiver varps were added to core in https://github.com/runelite/runelite/commit/17e3793947d188964fabeabd6ee096fcfcc458c4

hyppis1:
Oh thanks for that info, updated the plugin to read "public static final int" for the quiver ammo slot now

PR #6298, opened 2024-07-14
Update tilemanmode

sram1337:
https://github.com/ConorLeckey/Tileman-Mode/pull/47

> Skipped tiles when running instead of walking are currently incorrectly interpolated when doing L-shaped movement around an object. This fix allows the filled in tiles to be correct for all cardinal directions (both negative and positive delta x- and y-movement).


YvesW:
FYI, currently only ConorLeckey is marked as author. He should add you or other people, as author to the plugin file and then submit a PR here, if he wants to allow you to update the plugin. After that's merged, you can create PRs yourself.

PR #6300, opened 2024-07-14
Update attacktimer

ngraves95:
* Eliminate false positives when alching

The 0-ticking Patch (1.1) (#56)
* The 0-ticking Patch (1.1)

* New support for getting the correct weapon stats when switching weapons on the same tick that a player attacks.
* Adds Salamander Support 🎉.
* Adds Harmonized 4t Support (only triggers when using standards).
* Basic Red Keris special attack support.
* Further improvements to powered staff detection.
* Fixes many false positives

The quick swapping support (known as 0-ticking in the PvP community), is a helpful feature I wanted for rooms like Nylos and Baba-puzzle where many frequent and fast swaps happen. I noticed that the underlying cause of the problem was that we are essentially caching the results of `getWeaponStats()` and related functions when the `onGameTick()` subscription was triggered. Unfortunately this subscription happens at the start of a tick so you could very easily create a bug like so: https://youtu.be/t_2h5uJM5kk. This change tackles the root of the problem by using a new a subscriber to listen a client int which from testing will always monotonically go up as the client switches any worn item. This means we can use a fresh non-stale weapon id for the equipped weapons stats.

This required some more finicky details, including a new state for the first tick of attack delay, this is so the plugin can only do a "weapon" swap. If it's early enough to not have already been rendered.

------------------------------------------------------------------------

Next up is Salamander support, which turned out to just need a spot anim. As it turns out the ranged and magic attacks have no animation at all and simply trip up the `isPlayerAttacking()` code to think no attack was happening.

------------------------------------------------------------------------

Finally the even more improved staff support, as I was fixing the 0-ticking I was noticing quite a few problems with d971729 when staffs were involved. Especially when quickly swapping and manual casting. So I tackled it from a bunch of angles trying to reduce the false positives and false negatives that come from the current solution.

This starts with the new PoweredStaves.java file (which renamed from WeaponIds.java but git has lost the rename), the first step of this is too explicitly link every staff to it's corresponding animation and projectile. With this new information we can then already be more accurate by only switching to 4t when the animation and weapon Id match, whereas the old version just looked at id and animation separately.

Next there's some new code Spellbook.java and CastingSoundData.java which allows the plugin to correlate which spell book the player is on and if that animation can be manually cast. Giving a us a really big win in 99% of cases. This is because most things people do they're on thralls/ancients/lunars and all the powered staves tend to use the standard spellbook animations. We also use the sound to help detect manual casts, but a lot is still left on the table here as we should also be detecting the sound of the staff being used too, but I haven't got around to implementing that.

The other reason I haven't worked on the sound side of it too much is that it doesn't work when the client is muted so it feels half-baked in premise.

Edit: No longer do we use projectiles in the plugin, the data is still there incase the API becomes un-deprecated again.

------------------------------------------------------------------------

Fixes many false positives: This includes fishing spots, flinching, summoning thralls etc. We now use a ban list for animations which can occur while in combat don't actually indicate an attack.

This also includes the better flags to detect attackable NPCs (like the ToA oblisk) and a hard-coded exception for the target dummy.

------------------------------------------------------------------------

Misc details:
 * Version bump the whole plugin to 1.1 since this a big change. New subscribers, different state management, etc.
 * Fixed a small bug with blowpipe tailing animation being picked up as a second attack.
 * Added @ngraves95 to more copyright headers. (I've just assumed it's all copyright circa 2022 because checking the history was a touch too tedious for me).
 * Added some very basic sanity check unit tests which while adding I discovered that I couldn't run them locally. This is because Runelite has moved to lombok:1.18.30 instead of lombok:1.18.20, this is a breaking change with the V7.4 of gradle (Fails with: `java.lang.IllegalArgumentException: Unsupported class file major version 65`). Therefore to match the lombok version we just have to bump the gradle version to 8.X.

* Fix staves when on matching spell book

### DIFF
--- a/plugins/attacktimer
+++ b/plugins/attacktimer
@@ -1,2 +1,2 @@
 repository=https://github.com/ngraves95/attacktimer.git
-commit=d97172936b4e8f82ecdbb1666e9cd2634aa50e30
+commit=a01456308844de3e766ea5fc4d55c6a097bfd92a


aHooder (approved):

iProdigy (approved):
LGTM but technically you should do `11` instead of `1.11` at https://github.com/ngraves95/attacktimer/compare/d97172936b4e8f82ecdbb1666e9cd2634aa50e30...ngraves95:a01456308844de3e766ea5fc4d55c6a097bfd92a#diff-49a96e7eea8a94af862798a45174e6ac43eb4f8b4bd40759b5da63ba31ec3ef7R28 (or use `JavaVersion.VERSION_11.toString()`)

PR #6303, opened 2024-07-15
Create apollo-pet

ivieedev:


ivieedev:
> We prefer to not have meme plugins that are inside jokes. I think the Companion Pets plugin also covers this.

It doesn't cover it, only cover thrall, but you can just say you don't want it. It's okay to be honest.

PR #6314, opened 2024-07-17

adhansen:
## Summary
The release of notification override integration into this plugin with PR #6239 was met with overwhelming backlash on Reddit and in my github repo. I was excited to release this at the time because it would be a simplification to relinquish owning the logic of how flashing is handled and instead let the implementations in core handle that. However, it has been brought to my attention that this is a pretty significant usability regression- so this PR rolls that back. Notification overrides are instead implemented in place of the old checkbox option to send a tray notification, and the homebrewed flashing is back. I've also snuck in an enum dropdown control to manage the flashing speed. This was an accessibility feature request that has been open for a long time.

## Detail
* Notification overrides rollback:
    * Rather self-explanatory. The old flashing functionality is back, and notification overrides replace the old "Desktop Notification" checkbox. So now the feature set is a superset of what it used to be and nobody can complain. It does mean there are two places in the interface now where flashing can be configured which I'm sure will be somewhat confusing. But for the people that are happy to have the new notification overrides features (such as playing a sound and forcing the window to be focused) I want to keep that functionality in the UI.
* AlarmOverlay rewrite:
    * The original implementation of the AlarmOverlay was abysmal, which is why I was so happy to get rid of it. The screen flash is implemented as a hundred line breaks in a `PanelComponent` trenchcoat. It still is with the rewrite, but now that is produced one time in the constructor whereas previously it re-added 100 empty LineComponents every time the `render` method executed.
* Flash control:
    * Old issue that I never got around to doing: https://github.com/adhansen/plugin-repo/issues/22. Since I was rewriting the AlarmOverlay anyway, I worked in this feature too.

### DIFF
--- a/plugins/wilderness-player-alarm
+++ b/plugins/wilderness-player-alarm
@@ -1,2 +1,2 @@
 repository=https://github.com/adhansen/plugin-repo.git
-commit=acf721f8d9372b0524b05c438c8ad0c4ee4121b0
+commit=c234e6d20492e483bf22893d0861ed25d5d53829


YvesW:
It's def not the best implementation (check core!), but you can do something like this https://github.com/YvesW/resource-area-notifier/blob/master/src%2Fmain%2Fjava%2Fcom%2Fywcode%2Fresourceareanotifier%2FScreenNotificationOverlay.java

adhansen:
@YvesW yeah that's pretty clean. Definitely preferable to what I've done, but I don't mind mine either so long as it can clear code review. The important thing is that it's not creating 100 new objects and pushing them into the component children every 20ms anymore.

aHooder (approved):
I would also recommend implementing this using `Graphics2D#fillRect` as YvesW suggested, but lgtm otherwise.

TheStonedTurtle (approved):

PR #6315, opened 2024-07-17
Create tts-alerts

Morris1927:
Adds TTS alerts for hp and superiors spawning


YvesW:
Does this plugin do something that's not already covered by Watchdog or the TTS plugin?

Morris1927:
Ahh sorry I didn't realise watchdog already existed, I tried looking for something similar already but it didn't come up under "tts" so I assumed it wasn't a thing yet. It is slightly different as the exact hp is called out instead of set intervals (unless I'm missing something), but I suppose this isn't really needed. Oh well, was a fun practice project anyway! :P

PR #6323, opened 2024-07-18
Update empty-forestry-basket

DominickCobb-rs:
Resolve potential NPEs, slight refactoring
Add rough state tracking for whether the option in the bank will be empty or fill.
Add Forestry Kit as it also can be filled/emptied (I think)


geheur:
This should probably use `ConfigManager.setrsprofileconfiguration` instead of `set/getConfiguration`.

DominickCobb-rs:
Done in 3010a1388290957902b39e378a27fb0d8a6ab5d1

geheur:
You missed one. Also parseBoolean returns false for null. so you could simplify the `startUp` code if you want to.

geheur (approved):

TheStonedTurtle (approved):

PR #6337, opened 2024-07-20
Update csvexport

fl4xman:
Significantly improved write performance and created a new file for equipment as opposed to throwing it in with inventory data.


fl4xman:
Good catch, thank you. Fixed 👍

PR #6343, opened 2024-07-21
add privatege

AcidTree:
Runelite grand exchange plugin with GrandExchangeClient.java code removed


Adam-:
We need the GrandExchangeClient to provide accurate pricing data, and so I don't want to merge something which undercuts that.

PR #6346, opened 2024-07-22
Create Equipment Lock

Matimatician:
Plugin Name: Equipment Lock

Description: This plugin allows players to create a group ID within which only one player can equip a piece of gear. Once a player equips an item, any other player in the group attempting to equip the same item will be unable to do so.

How it functions: When a new item is equipped, the group ID, item name, and a hash of the player's username are sent to a server that tracks ownership. Relevant data for a player's group is retrieved when they open the client and cached locally. The plugin only requests/writes information from/to the server if they attempt to equip an item not in their cache.

Repository URL: https://github.com/Matimatcian/Equipment-Lock


iProdigy:
Please remove the awssdk dependency as it's not being used - you'll want to change your lambda calls to utilize gson and okhttp (you should `@Inject` both)

iProdigy:
```
Equipment Lock: invalid plugin file name "Equipment Lock"
in file plugins/Equipment Lock
plugin file names must be lowercase alphanumeric + dashes. try: "equipment-lock"
```

also gson is already provided by runelite so you can remove the dependence on a specific version

iProdigy:
the file in *this* repo should be renamed, not your repo's plugin class name

iProdigy:
You need a `LICENSE` file in the root of your repo

plugins/equipment-lock
```diff
@@ -0,0 +1,2 @@
+repository = https://github.com/Matimatician/Equipment-Lock.git
+commit = a5259f4dbd00f5673ec58617c05695b67cafbe39
```
YvesW, line 2:
```suggestion
repository=https://github.com/Matimatician/Equipment-Lock.git
commit=a5259f4dbd00f5673ec58617c05695b67cafbe39
```

Additionally, I have not reviewed this plugin at all, but if you are consuming clicks/causing people to not be able to equip armour, please make it known to your users in some way that the cause is this plugin. That way you can prevent prevent RL support traffic. You could for example send a game message ``Equipment Lock plugin: unable to equip item because...`` with a cooldown between said messages.

Matimatician: Hi, I know you said you haven't reviewed the code yet, but I just wanted you to know that this is already a feature. here's the bit of code from when the event is consumed:

client.addChatMessage(ChatMessageType.GAMEMESSAGE, "", "Someone else in your group has claimed the right to this item, so you cannot equip it.", null);
				event.consume();

pajlada: To YvesW's point, that message doesn't clarify what plugin it comes from. Somehow mentioning "Equipment Lock" will give users looking to figuring it out something to search for. maybe adding `[Equipment Lock] ` in front of messages from the plugin too?

Matimatician: understood. I've updated the code to start the game message with "[Equipment Lock]:" to clarify the cause of the consumed action to users.

Unmoon:
Please add a warning since you are doing requests, see https://github.com/runelite/plugin-hub/blob/master/plugins/tts for example.

Additionally you should use `log.debug` instead of `log.info`, since the latter will spam users' logs.

Matimatician:
> Additionally you should use `log.debug` instead of `log.info`, since the latter will spam users' logs.

Where is the most appropriate place to put the warning?

plugins/equipment-lock
```diff
@@ -0,0 +1,2 @@
+repository=https://github.com/Matimatician/Equipment-Lock.git
+commit=fda812ee461ee6d8aee63730a9352a40dcdb9cc4
```
iProdigy, line 2: The warning goes in this file, something like:

```suggestion
commit=fda812ee461ee6d8aee63730a9352a40dcdb9cc4
warning=This plugin submits your account hash, IP address, group ID, and items equipped to a 3rd party server not controlled or verified by the RuneLite Developers. Also, the plugin may momentarily freeze the client when you equip items.
```

iProdigy:
The current `warning` is not comprehensive in terms of information transmitted to your backend. I'd recommend using/adapting the suggested text above

Matimatician:
> The current `warning` is not comprehensive in terms of information transmitted to your backend. I'd recommend using/adapting the suggested text above

Ok, I've updated the warning to reflect what you included - though the lag only happens on equipping a previously unequipped item as that requires it to use the server instead of data stored locally.

iProdigy:
In `AWSLambdaClient`,  can you switch your `System.err` calls to utilize slf4j instead?

iProdigy:
To fix the build failure, I'd revert the build.gradle change and simply use `@Slf4j` on `AWSLambdaClient` (like you have in `Equipment_Lock`)

Matimatician:
> To fix the build failure, I'd revert the build.gradle change and simply use `@Slf4j` on `AWSLambdaClient` (like you have in `Equipment_Lock`)

I'm not sure why it's still failing - I reverted build.gradle to how it was and the only changes in AWSLambdaClient I made were switching the AWSLambdaClient file with a single import line and using log instead of logger for slf4j

iProdigy:
Your `build.gradle` is missing a final `}` for the `tasks.withType(JavaCompile)`

also in `AWSLambdaClient`, you need to annotate the class with `@Slf4j`, not just import the annotation

Matimatician:
> Your `build.gradle` is missing a final `}` for the `tasks.withType(JavaCompile)`
> 
> also in `AWSLambdaClient`, you need to annotate the class with `@Slf4j`, not just import the annotation

Thank you so much! I really appreciate all your help throughout this entire process.

LlemonDuck:
You should use OkHttpClient instead of your own custom input/output stream stuff (it will also greatly simplify your lambda caller). Inject the base instance with `@Inject OkHttpClient httpClient`, and make sure to update the user agent as well.

LlemonDuck:
Marking as draft for now. Please make the requested changes, then mark the plugin as ready for review.

iProdigy:
@Matimatician could you remove the latest `System.out.println` calls you added in favor of slf4j (e.g., `log.debug`)

LlemonDuck:
Ah, good catch. Thanks @iProdigy

Matimatician:
> @Matimatician could you remove the latest `System.out.println` calls you added in favor of slf4j (e.g., `log.debug`)

I've done this now, but I don't seem to be able to edit the file anymore where I'd paste the "commit=" - is this because it was merged?

iProdigy:
Correct, you'll need to open a new PR to update the hash

Matimatician:
Ok, I'll do that shortly.

PR #6362, opened 2024-07-25
Add new plugin: Walk here

Thource:
Adds a configurable keybind to make "walk here" left click.

### DIFF
--- a/plugins/walk-here
+++ b/plugins/walk-here
@@ -1,0 +1,2 @@
+repository=https://github.com/Thource/walk-here.git
+commit=4bd4510c081c252baad7ffa98266cd72bd5c2636


YvesW:
What is the use case for this? Is there anything that's not already covered by MES's shift click walk here settings?

Thource:
> What is the use case for this? Is there anything that's not already covered by MES's shift click walk here settings?

This allows you to bind a separate key to walk through things, so that you don’t lose your swapped menu entries when you hold the walk here button.

I’ve been made aware that it’s possible to do this with custom entry swaps plugin in the config, but this is a zero setup required option.

raiyni:
I don't really see the point in adding new plugins that can be a config option

MESLewis:
As mentioned above this plugin is very similar to core functionality and also other popular hub plugins. If this functionality is something you would still like I suggest PRing to <https://runelite.net/plugin-hub/show/hotkeyable-menu-swaps>, since it fits with that plugins existing functionality.

PR #6363, opened 2024-07-26
Create tormented-demon-helper

sreilly64:
This plugin adds a timer overlay that indicates when the player's attacks become boosted to 100% accuracy against Tormented Demons.


raiyni:
I think this would probably fall under "Indicates the time where a boss mechanic may start or end" or maybe even attack counters.

pwatts6060:
Jagex is patching this anyway and removing the 100% accuracy so this plugin is no longer needed.

YvesW:
> Jagex has patched this anyway and removed the 100% accuracy so this plugin is no longer needed. https://old.reddit.com/r/2007scape/comments/1e4rtbi/tormented_demons_stealth_nerfed/ldh837f/

That's probably not the case, but it's irrelevant what happens to the mechanic if the plugin is against Jagex's guidelines.

> First of all, a return to an idea we pitched in [last week's newspost](https://secure.runescape.com/m=news/deadman-armageddon--while-guthix-sleeps-updates?oldschool=1). We discovered an error that gave players 100% Accuracy against tormented demons, which negatively affected weapon variety and power. To offset this effect, we suggested lowering the demons’ stats while increasing their Magic Defence bonus to compensate. We also planned to have Demonbane weapons reduce their Defence by at least 5%. However, after taking your feedback on board, we’ve decided not to make these changes. We understand that the community has already developed strategies for the existing mechanics, and that you love the unique element it adds to the combat. We’ll still look for small tweaks we can make, but for now, the tormented demons will keep their stats and we won't be fixing the accuracy bug.
<https://secure.runescape.com/m=news/a=13/deadman-armageddon--while-guthix-sleeps-tweaks?oldschool=1>

sreilly64:
Thank you for bringing up these points. As YvesW quoted, they have decided to leave this bug in the game. I can see how this plugin is arguably against the aforementioned rule. But to play devil's advocate, it is less a mechanic and more a bug (with zero indicators in game, mind you) that they decided not to fix. Even if you consider it a mechanic, Tormented Demons are not a "boss" as per the wording of the rule. 

I can close this PR if you still think that the plugin violates the rules. Please let me know.

PR #6367, opened 2024-07-27
Create bank-searcher

SentryTheBroken:
This plugin allows players to see and search the items in their bank while they are away from a bank.


raiyni:
I think the bank memory plugin does this?

SentryTheBroken:
@raiyni Yeah, you're right. My plugin is very similar to this one. I'll go ahead and close the PR.

PR #6374, opened 2024-07-29
Update Smart Chat Input Color to 1.5.1

Ririshi:
In this update:
* At least a partial fix for https://github.com/MarbleTurtle/Color-Coordinator/issues/15
* Reformat code to 120 characters per line 
* Improved Slash Swapper compatibility
* Simplify parsing of chat channel color VarPlayers
* Remove `contains` calls before using `split`
* Use `@Varp` on some ChatChannel field getters

### DIFF
--- a/plugins/input-recolor
+++ b/plugins/input-recolor
@@ -1,3 +1,3 @@
 repository=https://github.com/MarbleTurtle/Color-Coordinator.git
-commit=1de5ab055335c634d24637c22d9f76e55eb40345
+commit=0cb5eaab0434ee9a928b96aa771e17863edfdd8a
 authors=Ririshi,MarbleTurtle


iProdigy (approved):
LGTM

did you mean `input.split(":", 2)` at https://github.com/MarbleTurtle/Color-Coordinator/blob/0cb5eaab0434ee9a928b96aa771e17863edfdd8a/src/main/java/com/smartchatinputcolor/SmartChatInputColorPlugin.java#L104

TheStonedTurtle (approved):
You were offline when I reviewed this so if you want to fix the issue @iProdigy brought up go ahead and create another PR.

Ririshi:
Thanks for the review and the quick merge. I don't think it should be split with limit 2 in this case, because I wanted it to split once, giving back at most 2 strings. If it turns out to be wrong somehow, there might be a 1.5.2 release soon anyway once I get some feedback on issue #15.

PR #6376, opened 2024-07-30
Create dimmer

kingLordsy:
Dim the client based on current region.

### DIFF
--- a/plugins/dimmer
+++ b/plugins/dimmer
@@ -1,2 +1,2 @@
 repository=https://github.com/kingLordsy/lordsys-plugins.git
-commit=3cb63836a7bbce2824e545c07fd98c56cfeaa211
+commit=2f84519ccf9d61000b47b50a86577c5f0e33b49a
--- a/plugins/dimmer
+++ b/plugins/dimmer
@@ -1,0 +1,2 @@
+repository=https://github.com/kingLordsy/lordsys-plugins.git
+commit=52d82f8885852eda464b6383080f3760546fbe23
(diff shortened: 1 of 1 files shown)


iProdigy:
please utilize slf4j (e.g., `log.debug`) instead of the `System.out.println` calls

geheur:
what is clickedRegionId? it seems to not do anything.
You also need a unique config group.

raiyni:
You can de/serialize your list with our Text utility class. `Text:: toCSV` `Text::fromCSV` or you can use GSON to go to a typed list/set.

iProdigy (approved):
LGTM - a couple bugs you may want to fix:

* if a user sets the dimmerRegions config to an empty string, the DimmerPlugin.regions list is not cleared (or nulled): https://github.com/kingLordsy/lordsys-plugins/blob/fe0084c0e98e6e7c145420f4d4746116ae28d7ad/src/main/java/com/dimmer/DimmerPlugin.java#L85
* the region calculations don't account for local instances. here's an example from a core plugin on how to deal with this: https://github.com/runelite/runelite/blob/cc4d83a47d386ee750fb3fb7310d8eee6f85f32c/runelite-client/src/main/java/net/runelite/client/plugins/groundmarkers/GroundMarkerPlugin.java#L250

iProdigy (approved):
one more place you'd want to account for instances: https://github.com/kingLordsy/lordsys-plugins/blob/3cb63836a7bbce2824e545c07fd98c56cfeaa211/src/main/java/com/dimmer/DimmerPlugin.java#L149
core example: https://github.com/runelite/runelite/blob/cc4d83a47d386ee750fb3fb7310d8eee6f85f32c/runelite-client/src/main/java/net/runelite/client/plugins/discord/DiscordPlugin.java#L310

iProdigy (approved):

aHooder (approved):

PR #6378, opened 2024-07-30
Update Chunk Tasks plugin

nathanreidok:
- Track farming tasks
- QOL UI enhancements
- Refresh tasks from Chunk Picker


nathanreidok:
@LlemonDuck - You were one of the reviewers for this plugin, and I just wanted to make sure that I'm not missing anything on my PR. Completely understand that it takes time to get to. Just don't want it to sit here forever if I'm missing something dumb. Thanks in advance!

MESLewis:
For http requests to 3rd party websites we require a warning to the user that their ip address will potentially be available to the plugin author. If you can change your requests to use a github.io domain then the warning will not be needed, otherwise you must add a config option that defaults to off with a warning such as this:
```
@ConfigItem(
...
warning = "This plugin submits your IP address to a 3rd party website not controlled or verified by the RuneLite Developers."
)
```

nathanreidok:
@MESLewis - thanks for the review. Added the IP address warning as requested

PR #6380, opened 2024-07-31
Add Obelisk Reminder plugin

AcolyteOfFire:
Shows a warning when near wilderness obelisks if the destination is set to the current obelisk.


geheur:
Have you considered using varbit 4966 to test what the obelisk destination is? This would simplify your code massively.

Also, you never remove your overlay. Also keep in mind that LOGGED_IN can happen multiple times, e.g. when you teleport you will get LOGGED_IN again.

Why are you using client.getCanvas().addKeyListener() instead of using `@Inject KeyManager keyManager` and registering with that? That is the standard way to do this.

AcolyteOfFire:
I'll get back to you, I'm out of time rn @geheur wish I'd known about varbit 4966

AcolyteOfFire:
@geheur thanks for the advice, I was able to massively simplify the code and improve functionality using that varbit, also changed it so the overlay is added and removed at startup/shutdown respectively

aHooder (approved):
I would suggest exiting early in `onGameTick` if the player isn't in the wilderness, since looping through all tiles takes a little bit of time. Also, [here you're calling Lombok getters](https://github.com/AcolyteOfFire/Obelisk_Reminder/blob/fa69baa32dba0cc252e3b2a4510e2d63893f53d8/src/main/java/com/obeliskReminder/obeliskReminderOverlay.java#L23-L25), which are only for reading values. Those don't make the overlay movable/resizable. Looks good to me otherwise :+1:

AcolyteOfFire:
with advice from some guys on the discord I changed it so the plugin watches objectSpawned events to keep a list of loaded obelisks instead of looping through all the tiles every gametick

AcolyteOfFire:
@raiyni done

raiyni (approved):

iProdigy (approved):
LGTM - as a performance optimization, instead of doing `(Arrays.stream(allObelisks).anyMatch(x -> x == id)` you could just do `(id >= WILDERNESS_OBELISK_ACTIVE_OBJECT_ID && id <= WILDERNESS_OBELISK_OBJECT_ID_LVL_50) || id == POH_OBELISK`

PR #6384, opened 2024-08-01
Fix autotags config group name

i:
Config name was mismatching the plugin name, so tag overrides were not doing anything.


Nightfirecat:
I would recommend writing a config migration function to ensure this does not effectively "erase" your plugin users' existing settings, as they would not be associated with the new config group.

aHooder (approved):

PR #6386, opened 2024-08-02
Create attack-ranges

tylerwgrass:
Displays the player's attack range for their currently equipped weapon.

This is similar to some of the other line of sight plugins/settings but automatically renders on weapon or attack style changes. I did a best-effort attempt to catalog all weapons but I am sure I missed some and I have not done extensive testing.

I tested:
- Bowfa
- Rune cbow
- magic shortbow (i)
- Red chin
- Dragon knife/thrownaxe
- Air battlestaff
- Dorgeshuun cbow

The general idea was given an OK, but will call out anyways that I could see this potentially violating the following Jagex guideline:

I plan to extend this to NPCs at some point, but decided to prune out what I had in progress for now, so there might be some remnants of that that work in the current case, but might not exactly look right in the current state (e.g. the tile visibility check).

### DIFF
--- a/plugins/attack-ranges
+++ b/plugins/attack-ranges
@@ -1,0 +1,2 @@
+repository=https://github.com/tylerwgrass/attack-ranges.git
+commit=3aa4f6e9b088276479fc8ed5dc799065e89adf8b
--- a/plugins/attack-ranges
+++ b/plugins/attack-ranges
@@ -1,2 +1,2 @@
 repository=https://github.com/tylerwgrass/attack-ranges.git
-commit=3aa4f6e9b088276479fc8ed5dc799065e89adf8b
+commit=263896b7efef4268f7b14c3814859f696961fe57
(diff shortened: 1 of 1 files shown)


Felanbird:
You can still host your images on your repo, just don't put them in `src/`

aHooder (approved):
Not a blocker, but I would call `getVisiblePoints` once per game tick instead of every frame where the overlay is drawn. I also think your current tile marking method will break in POH and the Gauntlet, since those are assembled from non-contiguous chunks. Maybe also in Chambers of Xeric.

tylerwgrass:
I'll update `getVisiblePoints`.

Gauntlet seems fine, unless there's something I'm missing:


Also CoX seem okay from the first few rooms (I have yet to do the raid), and PoH works but will probably just disable it there

aHooder:
Hmm, it actually works better than I had expected :eyes: I see now that you're using world coords without mapping from instances to template chunks, so this should actually work quite nicely even in assembled instances.

aHooder:
Incidentally, I just ran into an NPE which you might want to fix. It's possible for `Perspective.getCanvasTilePoly(client, lp)` to return null, so it requires a null check.

tylerwgrass:
I've fixed the NPE and moved the point calculations to `onGameTick`.

Also, here's a screenshot in the POH checking across rooms of the same type (it was brought up in discord):

aHooder (approved):

iProdigy (approved):

PR #6388, opened 2024-08-02
Update Pneck reminder to add necklace of faith

AbusiveTuna:


pajlada:
Your config group right now is `didICompost`, which I assume is not what you want

TheStonedTurtle (changes requested):
You must update the config group from `didICompost` to a value unique to this plugin. Config group names can not be shared between plugins.

You should probably also do the change @pajlada mentioned but that is not required.

AbusiveTuna:
> You must update the config group from `didICompost` to a value unique to this plugin (and matches or is at least very closely related to the plugin name). Config group names can not be shared between plugins.
> 
> You should probably also do the change @pajlada mentioned but that is not required.

Oops that was dumb of me, updated config group and changed to use RL's ItemIds

iProdigy (approved):

TheStonedTurtle (approved):

PR #6390, opened 2024-08-02
Create `Odablock DMM Finale` plugin

DapperMickie:


Nightfirecat:
After looking into this a bit more, I think this runs afoul [our own rules against generic player highlighting](https://github.com/runelite/runelite/wiki/Rejected-or-Rolled-Back-Features):

I understand that isn't the intent here, but I don't know how this plugin could work without also allowing a malicious or vindictive player to use the plugin to target individuals.

DapperMickie:
If I were to hardcode the list, could it get accepted? And I completely get that, I am just looking for a way to get this sort of functionality for groups larger than the 500 cap on friendschats.

PR #6395, opened 2024-08-03
Disable `Odablock DMM Finale` plugin

DapperMickie:


DapperMickie:
As this plugin was only used for the B2 finale, the plugin can be disabled.

PR #6403, opened 2024-08-06
Create ruby-bolt-spec-counter

yototheg:


cdfisher:
Does this do anything that the existing plugin "Bolt proc counter" doesn't already do?

yototheg:
> Does this do anything that the existing plugin "Bolt proc counter" doesn't already do?
Didn't know that there was one, but this only specifically does Ruby bolts

raiyni:
That doesn't really answer the question.

MESLewis:
<https://runelite.net/plugin-hub/show/ruby-counter> seems to already cover this functionality. If there is functionality missing then that should be PR'ed to the Bolt Proc Counter plugin repo.

PR #6411, opened 2024-08-07
npc-timer

sethks:
Adds a timer/tracker display to the screen so the player can track any NPC in RuneScape. Helps keep slayer tasks more engaging with personal bests, or helps the player see kill times with different gear setups.


geheur (approved):

LlemonDuck:
We have previously rejected plugins / requested changes on plugins which showed the current "alive" time of NPCs as it can reveal some mechanic timings. Can you please remove that ability?

LlemonDuck:
fwiw you may wish to update the readme

you can also display "last kill time" after the NPC has been killed, to still show times on a per-kill basis

PR #6413, opened 2024-08-07
Create HideAndSeekTracker

jeromkiller:
This plugin is to be used by the hiders during the Hide and Seek events hosted on the official osrs discord server.

Hiders setup their tracking area around them, enter the list of participants and hide somewhere.

when one of the participants finds the hider and enters the tracking area their placement will be logged.
The placement of players and their amount of hints used can be coppied to the clipboard for easy sharing with the host

afbeelding

afbeelding


jeromkiller:
sorry, i should also commit this on a different branch

PR #6415, opened 2024-08-08
Deforestation Plugin

Ocknon:
Makes it so that when you cut down a tree it never respawns, either leaving a stump behind or a blank tile. Saves a file to make sure that data is saved between sessions, can be deleted through the config.


raiyni:
The dataset looks like it could get quite large, yes?  Some things to keep in mind:

1. There is a max size for config entries
2. We'd prefer large static sets like this be saved to a file on disk instead of in config
3. You can use serializable properties to decrease the save size

LlemonDuck:

PR #6418, opened 2024-08-09
add wiki-bank-tags plugin

karearl:
This plugin adds the ability to generate bank tags by querying https://oldschool.runescape.wiki/

### DIFF
--- a/plugins/wiki-bank-tags
+++ b/plugins/wiki-bank-tags
@@ -1,0 +1,2 @@
+repository=https://github.com/karearl/wiki-bank-tags.git
+commit=27227d72a00e514715d26106a350c7bdbf45acb2
--- a/plugins/wiki-bank-tags
+++ b/plugins/wiki-bank-tags
@@ -1,2 +1,2 @@
 repository=https://github.com/karearl/wiki-bank-tags.git
-commit=27227d72a00e514715d26106a350c7bdbf45acb2
+commit=5c82620c62584adeb6e4e2273682fc57e6b83471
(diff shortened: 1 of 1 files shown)


raiyni:
Does this do anything that `bank-tag-generation` doesn't? Can it be contributed to that?

karearl:
`bank-tag-generation` seems to just be broken and abandoned so i decided i'd create a new project. if you prefer i update `bank-tag-generation` instead i can try that as well.

cdfisher:
Generally the preference would be to have you contribute to the existing plugin. If the creator is unresponsive, ownership of the plugin could potentially be given to someone else.

karearl:
if a plugin breaks on the plugin hub you should probably remove it after a set amount of time so it doesn't prevent people from making an alternative if the original contributors stop maintaining their plugin.. 
it seems like Mitch and Caleb are not maintaining their repo anymore so creating a new pull request on it seems a bit pointless..
also there is some pull request already trying to fix some build issue from a couple of months ago that is still open.
though i dont really understand why `bank-tag-integration` is not compatible anymore.. when i build it locally it works just fine.. (after i change `ICON_SEARCH` to `TAG_ICON_PREFIX`)

raiyni:
> if a plugin breaks on the plugin hub you should probably remove it after a set amount of time so it doesn't prevent people from making an alternative if the original contributors stop maintaining their plugin.. it seems like Mitch and Caleb are not maintaining their repo anymore so creating a new pull request on it seems a bit pointless.. also there is some pull request already trying to fix some build issue from a couple of months ago that is still open. though i dont really understand why `bank-tag-integration` is not compatible anymore.. when i build it locally it works just fine.. (after i change `ICON_SEARCH` to `TAG_ICON_PREFIX`)

We do disable plugins that are broken but they need to be reported to us as broken.

karearl:
so what should happen now? do i just close this pull request?

YvesW:
You should communicate with the author of the other PR to ask if they've received responses to their communication and what their future plans are. They can take over the plugin if they can show the proper receipts, but we generally prefer people be added as contributors instead of changing authors. You can then contribute to that repo.

PR #6423, opened 2024-08-11
Create chat-message-tracker

PhyrWall:


MESLewis:
What is the purpose of this plugin? From looking at the code it seems all functionality can be accomplished by using <https://runelite.net/plugin-hub/show/chat-logger> and having an independent program that ingests those logs into your database. It seems unlikely this plugin has a widespread use case suitable for being distributed through the plugin hub. You can contact the chat-logger people if you would like to add any extra functionality to their existing plugin.

PR #6430, opened 2024-08-12
Updates OSRS Archipelago Plugin to v2.0.0

digiholic:
- ~~Archipelago Client Plugin is now on Maven! No more direct vendoring of plugin code!~~ Nevermind it needed to be vendored anyway!
- Adds new checkbox for toggling AP messages going to OSRS chatbox
- AP Server Password is now an actual password field, using placeholders instead
- Replaces hard-coded tasks with data parsed from github, using version information from Archipelago slot data. This should allow for logic changes that don't require a whole new plugin version to use.
- Adds in a local DataPackage storage, which will store the Slot's connected player and last received item index. This will make reconnection smoother as well as allowing you to see popups for items received while offline.


hex-agon:
It seems like you forgot to include your plugin changes

digiholic:
I did indeed. Updated now.

iProdigy:
you haven't added `org.java-websocket:Java-WebSocket:1.5.2` to verification-metadata.xml - we've already approved version 1.5.6 for the wikisync plugin, so I'd recommend updating your version to that

package/gradle/verification-metadata.xml
```diff
@@ -210,5 +210,13 @@
             <sha256 value="b51f8867c92b6a722499557fc3a1fdea77bdf9ef574722fe90ce436a29559454" origin="Generated by Gradle"/>
          </artifact>
       </component>
+      <component group="org.junit" name="junit-bom" version="5.9.3">
```
iProdigy, line 213: this should be in `package/verification-template/gradle/verification-metadata.xml` instead of this file

package/verification-template/gradle/verification-metadata.xml
```diff
@@ -534,6 +560,22 @@
...
          </artifact>
       </component>
+      <component group="dev.koifysh" name="archipelago-client" version="0.1.18">
+         <artifact name="archipelago-client-0.1.18.jar">
+            <sha256 value="8ebb3dbaa85dd6edebd348151ead99b781f27fddacaef33a8d5f516c21db5ade" origin="Generated by Gradle"/>
+         </artifact>
+         <artifact name="archipelago-client-0.1.18.pom">
+            <sha256 value="3354696605f14bc952380651786ab3bc97a4bbd5a63d74a5973f177c18c17725" origin="Generated by Gradle"/>
+         </artifact>
+      </component>
```
iProdigy, line 570: is this necessary given https://github.com/digiholic/osrs-archipelago/blob/34a7bacdda632ae961998562a3982647de17fd5d/build.gradle#L28

package/verification-template/gradle/verification-metadata.xml
```diff
@@ -245,6 +245,14 @@
...
          </artifact>
       </component>
+      <component group="com.google.code.gson" name="gson" version="2.11.0">
+         <artifact name="gson-2.11.0.jar">
+            <sha256 value="57928d6e5a6edeb2abd3770a8f95ba44dce45f3b23b7a9dc2b309c581552a78b" origin="Generated by Gradle"/>
+         </artifact>
+         <artifact name="gson-2.11.0.pom">
+            <sha256 value="c0e547bea998888e6e25c5886a90e762272bc88b5275343dd2c05ded6ca2e360" origin="Generated by Gradle"/>
+         </artifact>
+      </component>
```
iProdigy, line 255: consider adding a gson exclusion [here](https://github.com/digiholic/osrs-archipelago/blob/34a7bacdda632ae961998562a3982647de17fd5d/build.gradle#L28-L32) so 2.11.0 doesn't need to be added to the verification metadata

package/verification-template/gradle/verification-metadata.xml
```diff
@@ -897,6 +952,14 @@
...
          </artifact>
       </component>
+      <component group="org.java-websocket" name="Java-WebSocket" version="1.5.7">
+         <artifact name="Java-WebSocket-1.5.7.jar">
+            <sha256 value="28d8392a882660e636603c0a72d3e1ed5da1993210214645754e93655c0d806a" origin="Generated by Gradle"/>
+         </artifact>
+         <artifact name="Java-WebSocket-1.5.7.pom">
+            <sha256 value="16b0757548ed6e4a834293d6202a896470a145a7a4083250b4d64c960cdeb411" origin="Generated by Gradle"/>
+         </artifact>
+      </component>
```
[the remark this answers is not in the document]
digiholic: Well, I removed it and now it says it's an error, so I guess I gotta put it back.

digiholic:
Since the automated tests are passing the dependency checks, is there anything else remaining on this before it can get reviewed?

package/verification-template/gradle/verification-metadata.xml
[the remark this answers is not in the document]
abextm: it failed because you didn't remove it from the thirdParty deps list

abextm:
You are writing files outside of .runelite, which is disallowed. You are also using java's built-in object serialization, which we also disallow.

digiholic:
> You are writing files outside of .runelite, which is disallowed. You are also using java's built-in object serialization, which we also disallow.

There are no files being written to outside of `.runelite`. The generated `.save` file is in `.runelite/APData/`.

I will remove the Java object serialization and write my own serializer and parser.

digiholic:
Serialization has been removed. If I could get clarification on where I'm writing a file outside of `.runelite` I could fix that

iProdigy:
You should use `@Inject` to populate the gson instance instead of doing `new GsonBuilder()`

digiholic:
Donezo. Build successful this time, thanks.

abextm:
You are writing to the cwd in `dev.koifysh.archipelago.Client::loadDataPackage` which you call (and still uses java serialization). You also are pinning a bunch of deps that I think you don't use, so don't do that.

digiholic:
That's in an external dependency, would it be sufficient to just override the functions that call those so they aren't used? The pinned dependencies have been found through trial-and-error to be used. The plugin won't build without them.

abextm:
One of the methods is private, so you can't just override it. I would prefer if they were gone entirely or had to be called manually. Even if you could override it, verifying that you always use the overridden one puts too much burden on reviewers handling your plugin in the future.

Wrt your deps, I deleted half of them and it still built. The error you were seeing earlier is because you didn't have the same excludes in verification-template's build as your plugin's, so it was failing to build because you are pulling more deps there than in your plugin.

digiholic:
So, at this point I'd need to just embed the source of the plugin instead of using the one on maven, so I could rewrite these methods? Or ask the maintainer of that plugin to add in some way to disable them?

abextm:
Yes

digiholic:
Okay, it's been added to the project, and modified to remove all Gsons, Serializables, and writing to improper directories. While I was in there, since the source is now embedded, I was able to excise the dependency on Apache httpcomponents, which means that now the only dependency change is _removing_ one that the original build of the plugin already had.

Hopefully that'll make the process much simpler for review.

digiholic:
Updating Plugin to v2.0.0

Runelite Plugin Changes:
- Updates Archipelago Java Client version so items are no longer displayed as "Unknown Item" or "Unknown Location"
- Plugin now pulls data from [logic repository](https://github.com/digiholic/osrs-archipelago-logic) based on versions baked into AP Seed, allows for minor logic changes without requiring updates.
- AP slot data stored in Runelite directory, for more consistent detection of player login and more safeguards against connecting with the wrong OSRS account
- More consistent connectivity, no longer loses connection to AP server on world hops or loading, properly disconnects on logout
- "Tiered" items no longer mismatch between what is displayed and what is actually unlocked
- Adds new item categories for Duds, Care Packs, and Claimed Care Packs to item panel
- _Every_ task should now be consistently auto-detected by Runelite. New methods of detecting completion have been devised that should be lag-proof.
- Plugin now displays an overlay when you are in a locked region, regardless of whether region-locker is installed or enabled.
- Many new types of Tasks, to auto-detect completion of new task types.

PR #6433, opened 2024-08-13
added multi-notes plugin

sama305:
Multi Notes adds a simple and powerful note editor that has note organization baked in. Unlike the default note plugin, this has the ability for separate notes. Those who prefer the simplicity of the default note editor can still use it, but those who want a little more organization can opt for this!


raiyni:
Notes would be better stored as individual config keys or files as there is a limit for data length that could erroneously impact this with multiple notes.

LlemonDuck:
Marking as draft for now. Please make the requested changes (at least the ones by raiyni), then mark the plugin as ready for review.

LlemonDuck:

PR #6437, opened 2024-08-14
Applying a plugin

sanderaigar:
A notification plugin with TTS. 
Currently only available for HP detection, will add more in the future.


cdfisher:
I believe this is already something that Watchdog does.

sanderaigar:
> I believe this is already something that Watchdog does.

Oh. I was not aware. I’ll give it a look.

sanderaigar:
> I believe this is already something that Watchdog does.

It's close, but not what I intended. At least I could not figure out how to setup it to do what I want it to do :)

YvesW:
What do you want to do?
How does your plugin significantly differ from Watchdog?

LlemonDuck:
Closing as stale. Reopen if you are still interested in submitting this plugin, although we'd prefer not to duplicate Watchdog here.

PR #6448, opened 2024-08-14
Added Town Crier's Cries

JREWeav:
A little plugin that allows users to change what the Town Crier says, just for fun.


cdfisher:
Does this do anything that "NPC Overhead Dialogue" can't already do?

JREWeav:
> Does this do anything that "NPC Overhead Dialogue" can't already do?

Hey! I just looked into it now. As far as I can see, NPC Overhead Dialogue displays all NPC's already set dialogue above their head, as well as some custom ones set by the developer. This plugin gives the Town Crier custom messages that are set by the user.

PR #6461, opened 2024-08-15
Clan member list sort

erwinvc:
- Updated usage of ClanMember api
- Added sort by chat sorting type
- Added sorting delay


iProdigy:
`lastChatTimestamps` is written to, but never cleared

consider clearing it upon:
* plugin shutdown
* `ClanChannelChanged`

also would be ideal to remove entry upon `ClanMemberLeft`

erwinvc:
Thanks for the heads-up. I now clear the map when the plugin shuts down.
I've decided not to remove entries on clan member leave since that might be confusing when players only log out for a second/hop worlds.

iProdigy (approved):

TheStonedTurtle (approved):

PR #6466, opened 2024-08-16
Wilderness Boss Scouter - Discord Webhook Integration

Darjow:
After scouting an initial world, and the world is active, a discord webhook will be send to your channel with information about the scouted world.

This way your PK team is automatically notified that they should hop to this world. 

[Wilderness Boss Scouter Plugin](https://github.com/Darjow/WildyBossScouter_Discord-Webhook)


raiyni:
We do not want to distribute scouting plugins for the wilderness as it's against 3pc guidelines.

PR #6469, opened 2024-08-16
Add Perfect Balance plugin

Thudcrackers:
Tiny plugin to change the skill icon to be perfectly balanced, created for ThisIsLandon on YouTube (unbeknownst to him). It's probably not ideal.

### DIFF
--- a/plugins/perfect-balance
+++ b/plugins/perfect-balance
@@ -1,0 +1,2 @@
+repository=https://github.com/Thudcrackers/perfect-balance.git
+commit=89499ec0a51da49eb0bac4cc9f06024b651a9e10


cdfisher:
Your config group needs to be something other than "Example".

Also, remove `enabledByDefault = false`. It confuses users when they install a plugin and then it's not enabled.

Thudcrackers:
> The repository link should go to the plugin repo, not your fork of the plugin hub.
> 
> Your config group needs to be something other than "Example".
> 
> Also, remove `enabledByDefault = false`. It confuses users when they install a plugin and then it's not enabled.

That's what I get for trying to rush through before leaving. Should all be fixed now.
Is there a way to remove the Config file? I am not using it but I didn't see anything about removing unused configs in the docs.

Thudcrackers:
> The repository link should go to the plugin repo, not your fork of the plugin hub.
> 
> Your config group needs to be something other than "Example".
> 
> Also, remove `enabledByDefault = false`. It confuses users when they install a plugin and then it's not enabled.

That's what I get for trying to rush through before leaving. Should all be fixed now.

PR #6472, opened 2024-08-17
Price graph opener

Raitab:
Version 1.1.0 update - add irreversible items as display options.


plugins/price-graph-opener
```diff
@@ -1,2 +1,2 @@
 repository=https://github.com/Raitab/plugins.git
-commit=ad901e62f3e7530d7c96439c798ad9b713047bdb
+commit=b9f04b47e2b14c66446015041cba490156954b24
```
iProdigy, line 2: `ad901e62f3e7530d7c96439c798ad9b713047bdb` is on the `price-graph-opener` branch
while `b9f04b47e2b14c66446015041cba490156954b24` is on the `price-graph-opener-fix` branch

did you mean to update the hash to [`398577a17500606467b5a615435d434a5fff05f1`](https://github.com/Raitab/plugins/commit/398577a17500606467b5a615435d434a5fff05f1) instead?

Raitab: No, this was intentional. I used filter commits to update my authorship on my previous commits, but only on this branch, hence the diff being more than just the recent commits. I wasn't sure if completely deleting the previous commit that this file pointed to would result in a broken plugin, so it's in a separate branch. I'll overwrite the main branch after.

Raitab: Wondering if this can get another view here in the near future. It's been almost two months since it was last viewed. As stated, the branch and change are intentional, I have nothing to change on my end.

pajlada: The diff you've created is for the entire plugin - it would be easier if you keep a linear git history

Raitab: I'm was aware of this from the get-go. I would generally agree, but I have no intention on changing in this instance. Apologies for the burden.

PR #6476, opened 2024-08-17
submission of custom-vital-bars plugin

qt31415926535-femboy:
Love the Status Bars but just wish they could be repositioned and/or customised?
You can recreate a sleek and classic style with just the bars, change the direction the bar fills, or modify the border to track related progress such as Hitpoint regeneration, or more!


qt31415926535-femboy:
not sure what's going on, I've supplied the correct key into the commit= field but no dice?

qt31415926535-femboy:
looks like it's failing now due to being pointed the incorrect package name but not sure how I would fix that

qt31415926535-femboy:
I think I've fixed it by changing the value in runelite-plugin.properties

raiyni:
The test class should be in a test resources folder like the example plugin 

https://github.com/runelite/example-plugin

Unmoon:
Your gradle files etc don't diff cleanly to `example-plugin`. Please make sure they do to make the review easier.

qt31415926535-femboy:
> Your gradle files etc don't diff cleanly to `example-plugin`. Please make sure they do to make the review easier.

sorry, I don't quite understand which files exactly to modify. could you please point me in the right direction by telling me which files and/or directories to copy from the example-plugin repo?

MESLewis:
At least some of these features seem suitable as improvements to the core plugin, you should consider contributing to that plugin instead of starting a brand new one.

PR #6478, opened 2024-08-17
Update attack-ranges to 1.2.0

tylerwgrass:
- Add hotkey support for rendering on toggle or hold
- Add NPC highlighting for targets in range
- Add drag protection
  - Hides NPC attack option if target is out of attack range 

If there is an API somewhere that returns the NPCs on a given tile I could use that'd be great (I couldn't find one).


iProdigy:
On plugin shutdown, please clear `AttackRangesUtils.targetableNpcs` and `AttackRangesUtils.npcPointMap`

also, your `ConfigChanged` handler should check whether `AttackRangesConfig.ATTACK_RANGES_GROUP.equals(event.getGroup())`

iProdigy (approved):

PR #6485, opened 2024-08-19
Update recommended-equipment

t8or:
Updated auto-incrementer as per Abex request
Added plugin icon
Updated tags and description
Updated Readme page to fix broken link

### DIFF
--- a/plugins/recommended-equipment
+++ b/plugins/recommended-equipment
@@ -1,2 +1,2 @@
 repository=https://github.com/t8or/runelite-recommended-equipment.git
-commit=da919e4bb4394c4d2de186561017a979a00ff541
+commit=dd45953b8bfd1f67a609eb37acd7177797480550


raiyni:
Can you provide a reference for this requested change?

t8or:
@raiyni Here is the reference from RuneLite dev channel: https://discord.com/channels/301497432909414422/419891709883973642/1273977480956477440

`@adamk33n3r Please delete this from watchdog and reccomended-equipment. It writes the date&time into the jar and makes it non reproducable, which makes it hard for me https://github.com/adamk33n3r/runelite-watchdog/blob/master/build.gradle#L55`

t8or:
Ugh im sorry - I don't know how I closed this PR. I made a minor update to the readme (spelling), and then pushed it to the master branch of my fork. And I guess it auto-closed this PR?

iProdigy:
Is writing the builds count on compilation all that important? if not, you could remove all of this from the build script https://github.com/t8or/runelite-recommended-equipment/blob/dd45953b8bfd1f67a609eb37acd7177797480550/build.gradle#L43-L55

t8or:
@iProdigy We included it as it is helpful for development and testing. We will be keeping it in please.

raiyni:
You can generate this information in your shadowJar task instead?

adamk33n3r:
> You can generate this information in your shadowJar task instead?

Then the build number would only increment when building a shadowjar, which isn't always done. Would be weird to have the same build number for different release versions. This is already being restricted to only the build task (and explicitly ignoring the plugin hub task) so we're good here.

pajlada:
for plugin hub builds, you have versions as the identifier
for local development, you could check an environment variable you set on your machines that increments the build numbers, meaning you'd have version number + build number
would that solve your development/testing needs?

adamk33n3r:
I suppose, but that's extra overhead for any developer to have to do. I'm not sure why the discussed and implemented solution isn't sufficient. The issue was the version changing during plugin hub deploy. This fixes that issue.

raiyni:
What you are doing isn't easily discernible by multiple reviewers and the easiest path forward is to not do it, as originally asked in the discord link. 

I'm honestly still not sure *why* you need it for development and testing in its current state. Are you maintaining a live master jar somewhere that you are letting people use instead of release runelite? If that's the case, why does the increment count matter and not the commit hash?

adamk33n3r:
> I'm honestly still not sure _why_ you need it for development and testing in its current state. Are you maintaining a live master jar somewhere that you are letting people use instead of release runelite? If that's the case, why does the increment count matter and not the commit hash?

A commit hash is still not granular enough when attempting to fix issues, at least in my experience. I can't speak directly for this plugin since I just copied it here, but for watchdog I have often sent multiple jars with fixes to users who reported issues to test them out. If using just the hash, since I haven't made commits, the file name would be the exact same and overwrite itself with every change. Having an incrementing build number let's each change I make be distinguishable no matter what without me having to think about it. 

I hope that explains the process better and its usefulness. It's been like this in watchdog for years and is part of my development process when interacting with my users. In this case, since @t8or is the owner of this plugin, I'll leave the decision up to him for this PR. But I'd like more cooperation with my own.

pajlada:
> for watchdog I have often sent multiple jars with fixes to users who reported issues to test them out

how are you tracking which changes are in which build? this sounds like the perfect use-case for commit, build, send shadowJar with commit hash in it to track exactly which commit someone is on

raiyni:
> But I'd like more cooperation with my own.

The feedback is the same, as you are doing the exact same thing.

t8or:
The nuance of this is way beyond me, so I asked Chat GPT to translate this into something I can understand lol

> **The Core of the Disagreement:**
> Developers like raiyni and pajlada believe that using commit hashes (a standard approach in development) should be sufficient to track changes.
> 
> Adamk33n3r argues that an incrementing build number is easier and more practical for his workflow, especially in development scenarios where uncommitted changes are tested. This method avoids confusion with file overwrites and makes each build distinct. 

Please forgive my ignorance here. Essentially what this means is that Adamk33n3r prefers to track his changes via a value in the code that is updated rather than commits because he believes that you shouldn't commit testing changes until they are validated? While Raiyni and Pajlada you both prefer to push code and commit it to track your changes. Do I have this all figured correctly? Essentially, there are two different ways to track changes to the code - one is in the public way, and the other is inside the code?

pajlada:
My intent wasn't to sidetrack the PR, this was probably a question better asked in Discord - I was trying to understand the value of the build number if it cannot be tracked to code, not necessarily arguing against your PR. It'll be up to the maintainers to review this PR.

LlemonDuck:
Merging as this is still better than the current behaviour

PR #6487, opened 2024-08-19
Update nature-runecraft-utilities

alex-macm:
fixed + added Sir Rebral option

### DIFF
--- a/plugins/nature-runecraft-utilities
+++ b/plugins/nature-runecraft-utilities
@@ -1,2 +1,2 @@
 repository= https://github.com/alex-macm/Nature-Runecrafting-Util.git
-commit=04b6c282e44a270d3613c9db2a2ba802c94ba333
+commit=97000c0b548af6de03b69dfa91fcf3f325afdf28
--- a/plugins/nature-runecraft-utilities
+++ b/plugins/nature-runecraft-utilities
@@ -1,2 +1,2 @@
 repository= https://github.com/alex-macm/Nature-Runecrafting-Util.git
-commit=97000c0b548af6de03b69dfa91fcf3f325afdf28
+commit=965363d4f735554a2809e44d2548cda953970237


iProdigy:
you probably want to move the `teleportIdx == -1` check before `menuEntries[teleportIdx]` to avoid a `IndexOutOfBoundsException`: https://github.com/alex-macm/Nature-Runecrafting-Util/blob/97000c0b548af6de03b69dfa91fcf3f325afdf28/src/main/java/com/NatureRCUtil/NatureRCUtilPlugin.java#L87-L93

similarly, you don't have a bounds check on `jarrIdx`: https://github.com/alex-macm/Nature-Runecrafting-Util/blob/97000c0b548af6de03b69dfa91fcf3f325afdf28/src/main/java/com/NatureRCUtil/NatureRCUtilPlugin.java#L95

lastly you may want to `break;` here as a performance optimization: https://github.com/alex-macm/Nature-Runecrafting-Util/blob/97000c0b548af6de03b69dfa91fcf3f325afdf28/src/main/java/com/NatureRCUtil/NatureRCUtilPlugin.java#L112

iProdigy (approved):

geheur (approved):

geheur:
Does this work with menu entry swapper? You are accessing menuentries.length - 2 which is static and may be different when you are in postmenusort, as you are now using an event that is after menu entry swapper has swapped menus.

FredP1:
It doesn't work when swapping teleport to the top using MES. It's designed to function when the top value is set to Kaleb in the sub-menu teleport.

halo145:
Not fixed yet as of 22 AUG 2024 at 18:12 hrs. EST.

alex-macm:
> Not fixed yet as of 22 AUG 2024 at 18:12 hrs. EST.

I've just tested it in game and it is fixed. Open a ticket on the plugin repo with more info and i will investigate when i get the chance.

PR #6488, opened 2024-08-19
Update delayed-healing

DapperMickie:
Fixes:
https://github.com/DapperMickie/delayed-healing/issues/1 
https://github.com/DapperMickie/delayed-healing/issues/2

Better reliability and works in bank now.

Thanks to https://github.com/CreativeTechGuy for the PR

### DIFF
--- a/plugins/delayed-healing
+++ b/plugins/delayed-healing
@@ -1,2 +1,2 @@
 repository=https://github.com/DapperMickie/delayed-healing.git
-commit=fe699f342beda037c57766606e6fe4a5b3fbe415
+commit=f8863f5343869375d58f73dcc0b6f6c593d7567e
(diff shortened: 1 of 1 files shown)


iProdigy (approved):
LGTM

iProdigy:
some optimizations you can make if desired:

`Widget#isHidden` can be expensive, so I'd move this check into the previous conditional block: https://github.com/DapperMickie/delayed-healing/compare/15922240fcf794905806783a2bac4d440d8052fb...DapperMickie:fe699f342beda037c57766606e6fe4a5b3fbe415#diff-69603ddaecdb9381dabadfcf7585204dcb3ad6009965bee1e350f6db80c4135cR47-R49

`map.put(key, map.getOrDefault(key, 0) + newValue)` could be replaced with `map.merge(key, newValue, Integer::sum)`

DapperMickie:
@iProdigy so for when INVENTORY_CONTAINER is not hidden, we don't check it twice is what you're saying? I'll fix that.
And the 2nd suggestion lgtm, I'll do that as well.

PR #6505, opened 2024-08-24
Create monster-monitor 

ChunkyAtlas:
Tracks NPC kills regardless of item drops and allows setting kill limits with optional notifications

### DIFF
--- a/plugins/monster-monitor
+++ b/plugins/monster-monitor
@@ -1,2 +1,2 @@
 repository=https://github.com/ChunkyAtlas/monster-monitor.git
-commit=87dd221ac40a24f1ce64a712dcd9983cd12af714
+commit=b18e48c7619c822853abf7083e7b9367732a3b7f
--- a/plugins/monster-monitor
+++ b/plugins/monster-monitor
@@ -1,2 +1,2 @@
 repository=https://github.com/ChunkyAtlas/monster-monitor.git
-commit=5634410d8ec8f4b24db5efab1db8070d6ed7697d
+commit=87dd221ac40a24f1ce64a712dcd9983cd12af714
--- a/plugins/monster-monitor
+++ b/plugins/monster-monitor
@@ -1,2 +1,2 @@
 repository=https://github.com/ChunkyAtlas/monster-monitor.git
-commit=b18e48c7619c822853abf7083e7b9367732a3b7f
+commit=e4fb16a569e7a2e347b77a24010ec439836ff49a
--- a/plugins/monster-monitor
+++ b/plugins/monster-monitor
@@ -1,2 +1,2 @@
 repository=https://github.com/ChunkyAtlas/monster-monitor.git
-commit=7fa696a50df39f01527ed1abcb493d7822209c60
+commit=5634410d8ec8f4b24db5efab1db8070d6ed7697d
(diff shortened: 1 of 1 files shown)


ChunkyAtlas:
Sorry for all the spam on commits, this should be the last major update (until others send me deathanimationids to fully implement), given the changes it should be fully functional and ready :)

ChunkyAtlas:
Sending back to draft, revamping and reassessing

ChunkyAtlas:
Ready for review.

ChunkyAtlas:
Hate be a bother, but if any maintainers have a minute or two, could you take a look at this please! Thank you!

raiyni:
This plugin looks like it logging every animation from an npc and always saving it?

ChunkyAtlas:
> This plugin looks like it logging every animation from an npc and always saving it?

If buffers animations when interacting with an NPC only buffering unique animations that are not -1. Upon despawn it logs the last animation as the Death Animation. then does a clean up of the buffer.

ChunkyAtlas:
Any other feedback or questions you have of this PR? :)

ChunkyAtlas:
> Can't you use something like onActorDeath instead of logging and tracking all this animation stuff? https://static.runelite.net/api/runelite-api/net/runelite/api/events/ActorDeath.html
> 
> This just sounds like asking for a nightmare every time a new piece of content comes out.

I could’ve used onActorDeath instead of tracking NPC animations with an array, but I chose this approach to get more accuracy and detail when logging NPC deaths. By capturing the death animations, I’m also hoping to help fill in some of the missing animation IDs within RuneLite. As the plugin grows and we identify more death animation IDs, I plan to eventually move away from tracking animations altogether and rely on the DeathAnimationIDs class for logging.

raiyni:
That doesn't really make sense as llemonduck mentioned.

ChunkyAtlas:
> That doesn't really make sense as llemonduck mentioned.

I understand! I'll make the corrections right now and utilize onActorDeath instead of the mess I made for tracking animations.

ChunkyAtlas:
@raiyni @LlemonDuck Updated per your suggestions

iProdigy:
DeathTracker is not being unregistered from the EventBus on plugin shutDown

ChunkyAtlas:
> DeathTracker is not being unregistered from the EventBus on plugin shutDown

@raiyni @Felanbird @LlemonDuck @iProdigy Changes made

raiyni:
You don't need to ping everyone every time you update it.

iProdigy:
This isn't a blocker for merging but: if someone stops and starts the plugin, the DeathTracker won't be re-registered with the eventbus. Personally I'd just do the `register`/`unregister` calls within the plugin `startUp`/`shutDown` methods (and then you don't need to inject EventBus into DeathTracker)

iProdigy:
Also you create this executor but never use it - could you remove this line: https://github.com/ChunkyAtlas/monster-monitor/blob/5634410d8ec8f4b24db5efab1db8070d6ed7697d/src/main/java/com/monstermonitor/DeathTracker.java#L84

LlemonDuck:
this should be on swing thread not client thread https://github.com/ChunkyAtlas/monster-monitor/blob/87dd221ac40a24f1ce64a712dcd9983cd12af714/src/main/java/com/monstermonitor/MonsterMonitorPlugin.java#L93-L103

LlemonDuck:
Sorry to bounce you back and forth a bit, but my last request is that you should remove uses of `Exception#printStackTrace`

ChunkyAtlas:
> Sorry to bounce you back and forth a bit, but my last request is that you should remove uses of `Exception#printStackTrace`

no worries! I appreciate you going through it all!

PR #6514, opened 2024-08-26
Special Attack Notifier plugin

Foxori:
Displays info box alert at specified special attack threshold.

   - Configurable alert expiry time (in ticks)
   - Configurable color and flash between two colors

Displays info box alert before threshold is reached

  -  Configurable expiry time (in ticks)
   - Configurable delay before threshold (in ticks)
   - Configurable color and flash between two colors


Unmoon:
You are adding `specialAttackSoonOverlay` in `notifySpecialAttack`, I assume you meant to add `specialAttackOverlay` since you are removing that in `clearSpecOverlay`.

Foxori:
I have now corrected the overlay adding and overlays are now removed on shutdown.

PR #6517, opened 2024-08-27
Custom Sound Notifications Plugin

launchd:
This plugin is meant to provide the user with the ability to get sound notifications on a game message. The user can add a list of keywords to a text box in a comma separated list. The plugin will check game messages and play a sound if it matches on of the keywords.


cdfisher:
This is already something that can be done with Watchdog.

launchd:
Thank you, I can stop development on this now.

PR #6518, opened 2024-08-27
Rewrite of Pneck Reminder

AbusiveTuna:
Rewrites Pneck reminder to "Equip Alert" Which is basically the same code just adds a lot more items. 
Makes this niche plugin more widely usable.


YvesW:
Please note that you are wiping your users' config. I did not see a migration method using ctrl+f, so I'd recommend migrating the config in startUp and ProfileChanged. Something like this should work (untested, just written on my phone):
```java
private void migrate()
{
	boolean oldConfigValue = configManager.getConfiguration(configGroup, oldKeyName, Boolean.class);
	if (oldConfigValue != null)
	{
		configManager.setConfiguration(configGroup, newKeyName, oldConfigValue);
		configManager.unsetConfiguration(configGroup, oldConfigKey);
	}
}
```

AbusiveTuna:
> Please note that you are wiping your users' config. I did not see a migration method using ctrl+f, so I'd recommend migrating the config in startUp and ProfileChanged. Something like this should work (untested, just written on my phone):

As in the config is completely reset upon exit? Or the new config settings wipe out pre-existing ones from the old version of the plugin? If it's wiping the config on exit everytime I'll have to add something in, otherwise the old config was only a singular option for something almost no one used so I'm not worried about wiping it out.

DapperMickie:
Your previous configuration in the `pneckReminder` group called `neckOfFaith` is not migrated to the new `faith` in the `equipAlert` group.

Something like this:

```java
private void migrate()
{
	String oldConfigGroup = "pneckReminder";
	String oldKeyName = "neckOfFaith";
	String newConfigGroup = "equipAlert";
	String newKeyName = "faith";

	boolean oldConfigValue = configManager.getConfiguration(oldConfigGroup, oldKeyName, Boolean.class);
	if (oldConfigValue != null)
	{
		configManager.setConfiguration(newConfigGroup, newKeyName, oldConfigValue);
		configManager.unsetConfiguration(oldConfigGroup, oldConfigKey);
	}
}```

Which is then ran at startup of your plugin, would migrate the old config to the new config.

YvesW:
Keep in mind that it also needs to be called in ProfileChanged, not just startUp.

AbusiveTuna:
Thanks both for the help. Updated with the config merge.

iProdigy (approved):

PR #6521, opened 2024-08-28
Update attack-ranges to 1.2.1

tylerwgrass:
- Add Noxious halberds
- Fixes to drag protection
  - Fix issues with menu entries being swapped incorrectly in melee range
  - Fix attackable targets without an attack level having their menu entries swapped incorrectly
  - Decouple the drag checks from the overlay rendering checks

### DIFF
--- a/plugins/attack-ranges
+++ b/plugins/attack-ranges
@@ -1,2 +1,2 @@
 repository=https://github.com/tylerwgrass/attack-ranges.git
-commit=2e6266b30f254030a59e7640619c16a38295a373
+commit=0ace05f4196242dbfc3e59c24fd92a7f1fcc72c2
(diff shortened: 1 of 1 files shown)


iProdigy (dismissed):
LGTM but you may want a bounds check [here](https://github.com/tylerwgrass/attack-ranges/blob/2e6266b30f254030a59e7640619c16a38295a373/src/main/java/com/attackranges/AttackRangesUtils.java#L39)

tylerwgrass:
Seems like the array is statically sized but I added just in case, as well as a check that it exists (there's some that don't have it after browsing the cache)

iProdigy (approved):

geheur (approved):

PR #6544, opened 2024-09-01
Address NPE causing plugin to fail

climbridecode:


Nightfirecat:
This seems like a bad way to solve this problem. Why not just initialize the cache as `private Map<Integer, Tooltip> tooltipCache = new HashMap<>();`?

climbridecode:
@Nightfirecat agreed, thanks

iProdigy:
Since `tooltipCache` is never null now, you can remove the try-catch you had added

iProdigy (approved):

TheStonedTurtle (approved):

PR #6552, opened 2024-09-04
Create cox-tool-rack-sound

pollyrobin:
new plugin request, which plays a sound when you pick up a tool from the cox old tools tool rack


cdfisher:
You should remove `enabledByDefault = false`. It confuses users when they install a plugin and it's not enabled.

MESLewis:
We try and limit extremely niche plugins since every plugin requires reviewer resources which are limited. If you are still interested in having this functionality, I suggest contacting the authors of some of the other CoX plugins and seeing if this functionality fits within their existing plugin.

PR #6556, opened 2024-09-07
Create NPC Animation Notifier

usa-usa-usa-usa:
Users define a list of NPCs and animations. When the animations are performed, a notification in the chatbox is triggered.

### DIFF
--- a/plugins/npc-animation-notifier
+++ b/plugins/npc-animation-notifier
@@ -1,0 +1,2 @@
+repository=https://github.com/usa-usa-usa-usa/usas-plugins.git
+commit=88dd1a32bf1166f0a5099f1087a78afe0c86b2c7


Felanbird:
> Reviewing

usa-usa-usa-usa:
> > Reviewing
> > We will review your plugin to ensure it isn't malicious, doesn't [break Jagex's rules](https://secure.runescape.com/m=news/third-party-client-guidelines?oldschool=1), or isn't one of our previously [Rejected/Rolledback features](https://github.com/runelite/runelite/wiki/Rejected-or-Rolled-Back-Features).
> 
> This plugin would fall under our Rejected feature for any new `ID based plugins`
> 
> > Plugins that use player provided IDs for the entirety of their functionality can cause moderation issues and outcomes that break Jagex's plugin rules. Due to this fact we will not be accepting any new ID based plugins. Plugins that use a specific set of IDs but do not allow user input will still be accepted. (e.g. plugins like Vardorvis Projectiles which only allows you to change a specific projectile to a set list of projectiles.)

I was just looking for a dumb little use case to start a plugin project, I was doing hydra at the time, and thought this would be helpful since I am so rusty.

If I make this specifically a hydra notification. Ex. notify specifically when hydra performs 9111, Would it comply?

cdfisher:
> If I make this specifically a hydra notification. Ex. notify specifically when hydra performs 9111, Would it comply?

That sounds like it would fall afoul of 

I would encourage you to read the pages felanbird linked to before you start any other projects.

PR #6559, opened 2024-09-07
[New Plugin] - Max-Hit Sounds - v1.0.0 - Sololegends

sololegends:
# Max-Hit Sounds
Add a special sound for when you hit max hit

- Set sound effects per range, mage, and or melee hits. 
- Whitelist only certain items to trigger the sounds (Looking at p2 wardens dds deafening you)

### DIFF
--- a/plugins/max-hit-sounds
+++ b/plugins/max-hit-sounds
@@ -1,0 +1,3 @@
+repository=https://github.com/sololegends/Runelite-Max-Hit-Sounds.git
+commit=8ada88c3efa0d7e58fb4713ea842ea1070b8d286
+authors=Sololegends


DominickCobb-rs:
This may sound silly, but I have a similar plugin that is half baked. Have you looked at Attack Sound Notifications? A poorly named plugin that also can achieve this

sololegends:
> This may sound silly, but I have a similar plugin that is half baked. Have you looked at Attack Sound Notifications? A poorly named plugin that also can achieve this

I didn't see that one in my searching for something that could do this. The main thing I wanted to have that I did find, and yours also doesn't seem to do this either, is to use in-game sounds and not sounds the user would have to load in themselves. 

There are a bunch of plugins that allow people to load in sounds from files, but not use in-game sounds which is where this ones comes in.

DominickCobb-rs:
I could definitely make a small change to the panel to accommodate this. A toggle between an external sound and a sound ID is not a significant change

sololegends:
That sounds fine to me. 

Would you also make it handle when you manual cast magic with the weapon and the hitsplat isn't directly tied to the weapon being held? That is something I'm working on today as as addition to this proposed plugin as I ran into that issue this morning. 

Additionally a categorical setting for all mage weapons / all melee / all mage would be nice.

DominickCobb-rs:
I'll be honest, I don't have the time or will to expand the scope beyond the initially mentioned panel changes on mine. Apologies for bringing it up. :)

sololegends:
> I'll be honest, I don't have the time or will to expand the scope beyond the initially mentioned panel changes on mine. Apologies for bringing it up. :)

No worries, it is always good to make sure there isn't a bunch of plugins doing the exact same thing on there

sololegends:
This getting looked at maybe?

iProdigy:
To avoid plugin proliferation:
@sololegends would you be willing to contribute your changes to the existing plugin?
and @DominickCobb-rs would you be willing to add them as a collaborator on the repo?

DominickCobb-rs:
I'm willing to give full access/ownership of my existing plugin, but the way it sounded before is that this would supercede mine and cover the use case.

LlemonDuck:
Even still, it would be better to add it to the existing plugin so that current users are maintained instead of needing to find a new plugin, and new users aren't required to determine which one supercedes the other. 

@sololegends Can you move your code into @DominickCobb-rs's repo?

LlemonDuck:

PR #6567, opened 2024-09-08
Create nex-timers

imKwok:
This plugin displays splits of the Nex boss fight, between Nex herself and the minions, and can print results to game chat as well.

### DIFF
--- a/plugins/nex-timers
+++ b/plugins/nex-timers
@@ -1,0 +1,2 @@
+repository=https://github.com/imKwok/nex-timers.git
+commit=351d435ea070f36ad637301d6977063e9312436d


cdfisher:
Does this do anything either of the two existing Nex splits hub plugins don't already do?

imKwok:
> Does this do anything either of the two existing Nex splits hub plugins don't already do?

Yes, I believe one of the splits plugins is for loot splits, and the other one doesn't function as a split timer with an overlay.

cdfisher:
> > Does this do anything either of the two existing Nex splits hub plugins don't already do?
> 
> Yes, I believe one of the splits plugins is for loot splits, and the other one doesn't function as a split timer with an overlay.

In that case we would recommend that you try to collaborate with the author of the existing Nex Splits plugin to add additional functionality to that rather than adding a whole new plugin to the hub.

imKwok:
> > > Does this do anything either of the two existing Nex splits hub plugins don't already do?
> > 
> > 
> > Yes, I believe one of the splits plugins is for loot splits, and the other one doesn't function as a split timer with an overlay.
> 
> In that case we would recommend that you try to collaborate with the author of the existing Nex Splits plugin to add additional functionality to that rather than adding a whole new plugin to the hub.

I see. I will try to contact the author.

PR #6572, opened 2024-09-09
Update Itemdespawnnotifier

maximonster:
Configuration is now setup properly and is no longer opening grounditems config instead.


iProdigy (approved):
non blocking nits:
* on game tick, do not call `toArray` on `values()`, just iterate over the values without creating an intermediate array
* would recommend checking the config group in your `ConfigChanged` event subscriber
* you can delete `highlightedItemsList` since it's no longer used
* on plugin shutdown, you could set `NotifyItems` to null

PR #6575, opened 2024-09-10
Update health-notifications

TommyAndy:
Added an option to set re-notification timer

Allows users to choose how often a notification will fire for breaching a threshold

### DIFF
--- a/plugins/health-notifications
+++ b/plugins/health-notifications
@@ -1,2 +1,2 @@
 repository=https://github.com/TommyAndy/health-notifications.git
-commit=f6142082d93c3a24946909946baa963d4ac9b0d7
+commit=f1230c7bc817b50eeb807800fa7b438cdfbd0912


Alexsuperfly:
can you set the default to not spam, so that we don't have to get thousands of people asking why they suddenly have notifications going off every second.  The common behavior is 0 is equal to disabled for these style of config options

wolfsilver00:
> can you set the default to not spam, so that we don't have to get thousands of people asking why they suddenly have notifications going off every second. The common behavior is 0 is equal to disabled for these style of config options

This change comes after the plugin was "fixed" and we asked the dev to make it spam again (so I think the default behaviour on spam is ok, as it was how it worked before the last commit and how we've been using it.. This is important for HCs that may expect the spam to be there if they come back to the game)

raiyni:
I have used this plugin for a very long time and don't remember it spamming notifications by default. Regardless, we believe that would be a very poor UX to include by default.

wolfsilver00:
> I have used this plugin for a very long time and don't remember it spamming notifications by default. Regardless, we believe that would be a very poor UX to include by default.

That's so weird.. It always spammed in prayer for me, although not in hp (in fact, I thought HP was bugged because it didnt spam) and Ive confirmation most HCs in my clan used it this way

Alexsuperfly:
the original behavior is to notify when you drop below the threshold and to only enable notifications again when once again above, there was a bug where prayer reset when above the health threshold and not prayer which was recently fixed

TommyAndy:
Alexsuperfly is correct on the history, my only concern with not having spamming be the default now was that it might unintentionally cause some folks to miss notifications if they were relying on the broken behavior.

The concern around thousands of folks suddenly wondering why they're getting spammed is fair, that probably outweighs the reliance on the unintended behavior.

Curious if you think there's a middle ground on repeat notifications to meet in the middle of the different uses to balance the two cases.

cdfisher:
Default behavior won't affect people who already have the plugin installed, just fresh installs, as far as I'm aware.

Alexsuperfly:
The broken behavior is already fixed, they can no longer be relying on it.  Just set the config to be disabled by default so that the current intended default behavior does not change, users can adjust it to whatever they want from there

wolfsilver00:
Yeah, the plugin was broken, the point is that the broken behavior has persisted for at least a year, in which many people have used it considering normal behavior (as in an alarm that won't stop until you fix the "thing").

I myself almost died because I missed the first notification and it didn't keep spamming. While the broken behavior was fixed, I would like to postulate that the benefit of a single notification that can be achieved through other means, is not much... If you are on your pc or paying attention or whatever, a single notification (even more in a setting with many) wouldn't do much good besides what is done by other plugins. 

This plugin specifically tints your screen permanently until you are above the threshold (which lends into the "hey, look, you are below x" and I think its usefulness as an alarm is paramount. 

I don't think because the behavior has been fixed that they can no longer rely on it, it was "fixed" just for a little while, people do other things besides OSRS and they may come after a 2 week break and die because the expected behavior has changed.

Due to it being broken only with prayer, I would put only prayer on alarm by default. Anyone who had been using it is expecting that behavior, anyone who just installs it will see the config with spam set up, and can disable it. 

I insist, even if it was a bug, it has become a feature and is what many are expecting of this, changing its default behavior will mess with a lot of people for no reason, and there is no "thousand" people who will come complaining about the spam, due to it being the behavior it had 2 weeks ago, from which point im pretty sure there wasn't a thousand users installing the plugin

Felanbird:
> and there is no "thousand" people who will come complaining about the spam, due to it being the behavior it had 2 weeks ago, from which point im pretty sure there wasn't a thousand users installing the plugin

We've already had to deal with this plugin adding on-by-default features, when the prayer notifier was added there was a sea of people in the discord confused why their screen was suddenly blue.

TommyAndy:
> > and there is no "thousand" people who will come complaining about the spam, due to it being the behavior it had 2 weeks ago, from which point im pretty sure there wasn't a thousand users installing the plugin
> 
> We've already had to deal with this plugin adding on-by-default features, when the prayer notifier was added there was a sea of people in the discord confused why their screen was suddenly blue.

Apologies, I wasn't aware of this. I'll be more mindful when updating that I don't change the existing behavior on folks.

I'm going to close this out and cut a new pull with changes to align with the expectations here.

PR #6576, opened 2024-09-10
Create IdleSit

LockTheMage:
A plugin that lets you sit down if you idle for x amount of seconds.


JonneSaloranta:
There's already a plugin that does it. See:[Resting](https://runelite.net/plugin-hub/show/resting)

MESLewis:
Already covered by an existing plugin.

PR #6578, opened 2024-09-11
Update target-true-tile to 1.3.0

Notloc:
This update does the following:

- Replaces all deprecated function calls with alternatives
- Enemy NPC status is now identified by combat level instead of HP
- Replace checking the worldview for an NPC with handling their despawn in onNPCDespawned
- Add tile border style options


tylerwgrass:
FYI some attackable NPCs (Hespori flowers, for example) don't have an attack level so your combat level check will miss some NPCs.

I fixed this for mine by checking the second option of the NPC actions exists:
```java
String[] actions = npc.getComposition().getActions();
return actions != null && actions.length > 1 && actions[1] != null;
```

Notloc:
Thank you very much, that's a great way to handle that check!
I've added it to the PR.

geheur (approved):

iProdigy (approved):

PR #6581, opened 2024-09-12
Updated afk marks canifis to marks cooldown

powerus117:
After some feedback on the AFK canifis marks plugin, I have added other courses so it no longer is a canifis specific plugin but rather a general mark of grace cooldown tracker. And that's exactly what it does, while running laps, after finding your first mark, you should see a cooldown for when the next mark could spawn.

There were also previously some issues with computer with their time out of sync, which are now resolved using NTP.

Let me know if I can improve anything or if you have any questions, thanks in advance 🙏

### DIFF
--- a/plugins/afk-marks-canafis
+++ b/plugins/afk-marks-canafis
@@ -1,2 +1,2 @@
 repository=https://github.com/powerus117/RLAfkMarksCanafis.git
-commit=afbd393a1a027fe7145b1d7a4bd30bab1e336e4b
+commit=d31418239cc5039cfba36c0a06e27dfa09c387c5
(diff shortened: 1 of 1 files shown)


YvesW:
Not an extensive review but...
- By changing the ConfigGroup, you are wiping people's config. I strongly recommend not changing the config group or migrating config values on startUp and ProfileChanged. The former is the easier option.

raiyni:
We also recommend not changing the plugin file name

raiyni:
Yes, value for configName would be `AfkMarksCanafisPlugin`

powerus117:
Alright thanks, updated 🙏 But when I boot up, I now do see my plugin twice, if I enable 1, the other automatically also enables and vice versa. Is that to be expected?
image

YvesW:
FYI the support link gets ignored nowadays AFAIK. So you can just remove that field from the file.

PR #6583, opened 2024-09-13
Update health-notifications

TommyAndy:
Follow up after the discussion in https://github.com/runelite/plugin-hub/pull/6575

I'm setting the default behavior back to what it had been for over a year, reverting the recent "fix" in https://github.com/TommyAndy/health-notifications/commit/f1230c7bc817b50eeb807800fa7b438cdfbd0912

Prayer notifications are disabled by default, so this default behavior change is unlikely to disrupt any new users in the last few weeks this particular change has been out.

For existing users this will align with their existing expectations of behavior as well as the ability to spam notifications for both hitpoints and prayer.

Frankly, the recent change in default behavior was a mistake on my part. I should have been more aware of how the plugin was being used instead of hastily making a change based on the original intention of use.

### DIFF
--- a/plugins/health-notifications
+++ b/plugins/health-notifications
@@ -1,2 +1,2 @@
 repository=https://github.com/TommyAndy/health-notifications.git
-commit=f6142082d93c3a24946909946baa963d4ac9b0d7
+commit=990a1970cf155d57289edefdd14e4ba41f1a1a6b


Alexsuperfly:
i don't understand the purpose of specifically adding back in the bug where a user can only be notified that their prayer is low if they lose hp, just fix the bug and send the correct notification??? I don't want to have to review another PR of you realizing this was stupid in a week.

TommyAndy:
What you described was not the bug, I'm not sure I'd classify it as a bug either since it was just how it worked since the feature was created. With the original behavior the prayer notification just keeps firing if their HP is above the threshold. People were using this behavior to alert on prayer being low with a repeated notification. 

It was an unintentional feature that people started to rely on that I hastily changed in the update a couple weeks back. 

What I don't want to have happen, which might have already, is to kill a HC because of the behavior change a few weeks ago. Which yes, shouldn't have happened in the first place.

raiyni:
If you are really that worried about the change in functionality, you can add a one time chat message on login telling users to set a notification frequency if they want previous functionality back.

wolfsilver00:
> If you are really that worried about the change in functionality, you can add a one time chat message on login telling users to set a notification frequency if they want previous functionality back.

I think this is a very good compromise

PR #6585, opened 2024-09-13
Added Better Custom Cursor

archiru:
This plugin allows users to specify where they want their cursor to point, for images like crosshairs ect.


cdfisher:
Before a reviewer even really looks through this, can you clean up all the example plugin stuff from the repo? You're basically bundling the whole example plugin along with what you've written.

Also, is there a reason your plugin's license headers are dated 2018 and seem to only name someone who appears to mostly write WoW add-ons?

archiru:
Thank you, I have now removed the example plugin from the repo, I mistakenly merged.

YvesW:
What does this plugin do that the core plugin, or other cursor plugins on the hub don't already do?
Also, could you come up with a more descriptive name? (I know this can be difficult)

archiru:
This plugin allows the user to change where they want the cursor to click. By default, the custom cursor plugin assumes that cursor starts at 0,0, which is not the case for many custom images, such as crosshairs or arrows.

I will change the name of the plugin, Thank you

LlemonDuck:
Your build is failing because you are using `enabledByDefault = false`. You can check the results under the details of the RuneLite Plugin Hub Checks check suite.

LlemonDuck:

PR #6609, opened 2024-09-18
Update examine-log

CarlOmega:
### Summary
One like fix since I used the wrong constant value as a typo. Also had a once over and found another small mistake. Have updated the plugin hopefully not long til review.


CarlOmega:
@Unmoon Apologies for tagging so soon but found a small bug introduced when I made the move to using the disk. I implemented on my lunch break but had a better look and cleaned up the bugs.

PR #6615, opened 2024-09-18
Create counter plugin

pwatts6060:
A plugin that keeps track of a counter that you can manually increment with a hotkey.

### DIFF
--- a/plugins/counter
+++ b/plugins/counter
@@ -1,0 +1,2 @@
+repository=https://github.com/pwatts6060/runelite-plugins.git
+commit=24652d3cda56b6bbefe6f833439dec33c43908a1


YvesW:
Just for my interests, what use cases would this plugin have?

pwatts6060:
> Just for my interests, what use cases would this plugin have?

Counting attacks at hydra/demonic gorillas/cg for people with goldfish memory span. Not too sure in all honesty. I use it at Hydra but it's almost as much effort as keeping count, still helps a bit. The animation for attack style switch doesn't play if it's at the same time as phase transitions which is why counting is still needed. I'm expecting other people to find use cases for it beyond this.

raiyni:
This plugin doesn't seem moderately useful when you can use AHK or any other counter program

PR #6617, opened 2024-09-19
CookingBuddy - Fix/cooking-level-check.

MapleBytes:
Apologize for this update,

I was dumb and never tested after removing my hard-coded cooking level (as I am not level 99)


MapleBytes:
URGENT: The plugin is currently non-functional without this fix.

Nightfirecat:
You need to rebase your plugin-hub branch

PR #6620, opened 2024-09-19
Urgent Fix - Cooking Buddy

MapleBytes:
**Apologies for this update.**

I failed to test after removing my hard-coded cooking level and forgot to fetch the player’s cooking level, which has made the plugin non-functional. This fix is _urgent_ as it restores the core functionality.


MapleBytes:
Thank you Unmoon.

PR #6623, opened 2024-09-20
Update examine-log

CarlOmega:
Migrate from player name to rsprofile and added failsafe loading before writing.
Uses `onGameStateChanged` -> `state == GameState.LOGGED_IN` then uses `configManager.getRSProfileKey()` and stores data inside that folder.

Also fixes for fixed classic mode not showing popups.

### DIFF
--- a/plugins/examine-log
+++ b/plugins/examine-log
@@ -1,2 +1,2 @@
 repository=https://github.com/CarlOmega/examine-log.git
-commit=97174960ca7e6dc129812f648e047692ca3fd356
+commit=118b9af7c03607242ec51858e1a318244e487be3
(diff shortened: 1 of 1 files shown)


YvesW:
Have not checked your code, but keep in mind:
- When loading new areas, your gamestate will be GameState.LOGGED_IN -> GameState.LOADING -> GameState.LOGGED_IN. Thus, it procs semi-often while running around.
- To migrate the data from display name to RSProfileKey. RSProfile even has e.g. a getDisplayName if you want to use that <https://static.runelite.net/api/runelite-client/net/runelite/client/config/RuneScapeProfile.html> (and ConfigManager getRSProfiles).

CarlOmega:
@YvesW thanks for the callout have updated it to only load logs from log in if they haven't loaded. Additionally on world hop or login screen states it clears the loaded account (hopping to temp world etc).

PR #6625, opened 2024-09-20
Create ToA Potion Deposit

ealahd:
Prevents players from using items on other players in ToA nexus room since clicking potions on other players often causes annoyance.


geheur:
Does this actually work?
the menu is rebuilt 30 times per game tick. this sounds like it has a 1/30 chance of working.

ealahd:
Not really too familiar with the API and just salvaged the code from an already approved plugin. Worked in testing but the testing included going into the nexus room with alt and main a few times. I can test further to see if you have a point.

MESLewis:
There is already a [comprehensive toa plugin](https://runelite.net/plugin-hub/show/tombs-of-amascut) that this feature could be added to. I suggest contacting @LlemonDuck to get this feature added to the existing plugin.

LlemonDuck:
I would gladly welcome this as a feature in the Tombs of Amascut plugin, and may add it regardless. Would you be interested in PRing it instead of creating a whole new plugin for users to find and install?

ealahd:
Yeah absolutely would be better not to have it as a separate plugin. I can PR or you can implement it yourself whichever you prefer. I don't really know what I am doing so it might be better you just add it yourself.

LlemonDuck:
Thanks, will do. I'll ping you in the corresponding ToA plugin PR soon as an FYI.

PR #6632, opened 2024-09-22
Update DropTracker to v3.0.1

joelhalen:
Added a variety of new functionality, including a side panel with user lookups, tracking personal bests, combat achievements and collection logs, and more.

Fixed tracking of Araxxor and Whisperer kills, since their mechanics are slightly different than normal npc loot.


YvesW:
I've not reviewed this, but it has conflicts that need to be resolved.
Also, we target java 11 nowadays on the hub, not 9. You can check out build.gradle from the example plugin.

Additionally, without having to check a 3k diff: can the warning indeed be removed? Is this because data was only submitted when a config option was enabled, which is false by default? Or are you also submitting data in other way?

joelhalen:
> I've not reviewed this, but it has conflicts that need to be resolved. Also, we target java 11 nowadays on the hub, not 9. You can check out build.gradle from the example plugin.
> 
> Additionally, without having to check a 3k diff: can the warning indeed be removed? Is this because data was only submitted when a config option was enabled, which is false by default? Or are you also submitting data in other way?

That is correct-external connections are nulled if they're not enabled in the plugin config first. It's odd that I needed to change that this time around, because I previously updated the plugin to remove that--on the plugin hub currently, there is no warning for install.

Do you know what the conflicts it's referring to are? Never received one of these; I've not done anything different than any other time I've updated from what I can recall

YvesW:
Normally I'd recommend to just sync your fork, then rebase your branch onto master, and force push... but you PRed from your master branch (generally, you should never do this). I'd just reset to upstream/master in this case.

joelhalen:
Closed this and opened a new one that appears to fix the conflicts @ https://github.com/runelite/plugin-hub/pull/6635
Thanks for your help

PR #6635, opened 2024-09-23
Update DropTracker to v3.0

joelhalen:
Added a variety of new functionality, including a side panel with user lookups, tracking personal bests, combat achievements and collection logs, and more.

Fixed tracking of Araxxor and Whisperer kills, since their mechanics are slightly different than normal npc loot.

### DIFF
--- a/plugins/droptracker
+++ b/plugins/droptracker
@@ -1,2 +1,2 @@
 repository=https://github.com/joelhalen/droptracker-plugin.git
-commit=ec4282d433ecf2813aa862b577a52e3bad06ef3f
+commit=0416943d653810efb771c49297a16c15369a8fa7
--- a/plugins/droptracker
+++ b/plugins/droptracker
@@ -1,2 +1,2 @@
 repository=https://github.com/joelhalen/droptracker-plugin.git
-commit=0416943d653810efb771c49297a16c15369a8fa7
+commit=13aa8c19f6949aff7ec00d6c368c748c071adfec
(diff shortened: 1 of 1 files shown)


Zoinkwiz (changes requested):
You seem to have removed the useApi check from a few paths, notably the onChatMessage, so it will now send events off to the webhooks you've defined without it set to true. That'll need to be active for that to happen otherwise the player won't see the warning that plugins require for sending data.

joelhalen:
> You seem to have removed the useApi check from a few paths, notably the onChatMessage, so it will now send events off to the webhooks you've defined without it set to true. That'll need to be active for that to happen otherwise the player won't see the warning that plugins require for sending data.

Hey there, am I mistaken in my understanding that Discord links and GitHub Sites pages do not require this warning?
The only external connections made throughout the program are in the side panel's player search feature, elsewhere everything is sent through webhooks using discord URLs--there is a check in place as well to ensure that it is actually indeed a discord-based url

Zoinkwiz:
> Hey there, am I mistaken in my understanding that Discord links and GitHub Sites pages do not require this warning?
> The only external connections made throughout the plugin are in the side panel's player search feature--which does check for the API setting--elsewhere everything is sent through webhooks using discord URLs--there is a check in place as well to ensure that it is actually indeed a discord-based url

At present the list of Discord webhook endpoints are stored in an editable location (the GitHub endpoint), which means that after this initial review the endpoints can be changed without any insight to the RL team to ensure it remains a Discord endpoint. You've got a basic check for the URL containing the word 'discord', but there's multiple ways the url could contain 'discord' and not be an actual discord endpoint.

If you're wanting to keep roughly the same structure as you currently do, you could try hard-coding the `https://discord.com/api/webhooks` into the code, and retrieve from GitHub the webhook ID part.

PR #6637, opened 2024-09-25
Update max-hit-calculator to v2.0.0

j-cob44:
- Added Sidebar Plugin Panel
- Moved Dart Type Setting and Monster Size Setting to Sidebar
- Added Magic Type Weaknesses to calculation
- Added ability to automatically retrieve NPC magic type weakness
- Automatically retrieved NPC's size

### DIFF
--- a/plugins/max-hit-calculator
+++ b/plugins/max-hit-calculator
@@ -1,2 +1,2 @@
 repository=https://github.com/j-cob44/max-hit-calc.git
-commit=3a6835744cd599c04c84fbfd54f4b3370a671b8b
+commit=eb05ebf20f162b118fd9e708c4bb369804d7a34b
--- a/plugins/max-hit-calculator
+++ b/plugins/max-hit-calculator
@@ -1,2 +1,2 @@
 repository=https://github.com/j-cob44/max-hit-calc.git
-commit=eb05ebf20f162b118fd9e708c4bb369804d7a34b
+commit=bb2df38aea06b896b540e7873c39b4dcfac9e2a6
--- a/plugins/max-hit-calculator
+++ b/plugins/max-hit-calculator
@@ -1,2 +1,2 @@
 repository=https://github.com/j-cob44/max-hit-calc.git
-commit=bb2df38aea06b896b540e7873c39b4dcfac9e2a6
+commit=683020e7839409e7ef1b2ef0e7e434c694040d67


MosheBenZacharia:
is the dart type something you still want to persist? you can still store the data using the config classes while allowing it to be controlled from the sidebar

j-cob44:
> is the dart type something you still want to persist? you can still store the data using the config classes while allowing it to be controlled from the sidebar

Absolutely would love to store that if possible, but I didn't have the chance to figure that out yet.

> why do you seemingly have so much duplicated code between PredictNextMax.java and the other ___MaxHit classes?

This is necessary because it's taking the normal version and increasing specific stats by 1 until it finds the next max hit value. Those stats aren't exposed to be changed by parameters of max hit. I don't necessarily want to be able to change those stats via parameters either.

j-cob44:
The changes in PredictNextMax.java are normal for whenever an update occurs to the original Maxhit.java function.

The big change for this 2.0 update was moving everything away from static functions and into their own objects. I have no clue why I originally decided to do it that way, but slowly overtime I was feeling those growing pains of having everything be static and decided to change it to allow for more functionalities in the future.

MosheBenZacharia:
> > is the dart type something you still want to persist? you can still store the data using the config classes while allowing it to be controlled from the sidebar
> 
> Absolutely would love to store that if possible, but I didn't have the chance to figure that out yet.

It's pretty straightforward you just need to call the config manager's setConfiguration and getConfiguration methods:

https://static.runelite.net/api/runelite-client/net/runelite/client/config/ConfigManager.html

j-cob44:
That is legitimately exactly what I was looking for and I gave up looking for it. I was so sure there would be one too. Thank you.

MosheBenZacharia (dismissed):

YvesW:
FYI, ``secret = true`` is used to make a String config option show up as ``*******``

secret

MosheBenZacharia (approved):

Simpuhl:
When can we expect this to go live? Its very much needed

pajlada (approved):
Works mostly as expected for me, nothing blocking

I have this feedback for you if you want to take a polish pass later:
Selected NPC sizes aren't fully "calculated" - when starting the panel and clicking some NPCs, the icons will slightly resize. You can test this by clicking the Hellhound & the Giant spider. Guessing this would be solved if you take the 1 px border into account even if no border is active.

Surge spells in the normal spellbook are missing element type

The dart type is not loaded correctly, so saving it only changes the default dart type selected in the panel, but not the one actually used for the max hit calculation

"Dart Type" should probably be "Blowpipe Dart", or add a tooltip explaining it's for the blowpipe

The "selected NPC" changes when talking to an NPC - maybe you can't differentiate attacking to talking though

j-cob44:
Thank you @pajlada, the feedback is awesome and extremely helpful. 

The surge spell issue is incredibly important so thank you for spotting that. 

With GUI work, I definitely want to take a polish again in the future. It's waaaay better than what I originally started with, but I'm not fully happy with it. I'm not the best with Swing so that was definitely a learning experience.

And the dart type setting saving definitely is a little finicky. I'll have to check that out more when I have some time.

Thank you once again.

PR #6638, opened 2024-09-25
Create HideChatButtonText

salemwatkins:
A plugin that hides text on chat filter buttons.

### DIFF
--- a/plugins/HideChatButtonText
+++ b/plugins/HideChatButtonText
@@ -1,0 +1,3 @@
+repository=https://github.com/salemwatkins/HideChatButtonText.git
+commit=1913a3b39099fc195a7a532f502b95482f366a9d
+authors=salemwatkins,transceiverfreq


YvesW:
Not done an in-depth review, but...
- What could be the use case of this plugin? I can't think of anything
- Why would you try to create a new thread and sleep? There's so many better ways to do this.
- Why are you randomly checking 10 times?
- Have you checked how often GameState == GameState.LOGGED_IN when running through the world, or teleporting around?
- Please remove your log.info calls or replace them with log.debug

salemwatkins:
1. My use for this plugin is making the chat box less "cluttered" looking. (With the help of other plugins)
Example:
example

2-6. To be honest, this is the first code I've ever written in my life and I encountered a lot of technical issues/errors so I resorted to pure insanity to get it work.

PR #6646, opened 2024-09-25
Disable GE Calc plugin

cman8396:
Replicated the issue but no luck on a fix.
Best to disable due to multiple reports on github and RuneLite discord of the plugin interfering with all quantity input fields.


raiyni:
disabled is the field name

PR #6652, opened 2024-09-26
Create mixologyxp

Lep255:
Mixology plugin similar to the giant's foundry. Will show the combination on a info box and outline which tables to use during the minigame.


hex-agon:
Hey, let's collaborate over https://github.com/hex-agon/mastering-mixology, i'll be making sure it has feature parity with yours

PR #6678, opened 2024-09-29
Pluh damage sound

DanttePantte:
[Hey! This plugin changes the sound that plays when your character takes damage to the "Pluh/Plug" meme sound.


MESLewis:
The repository has been deleted. Update this PR with the new repository if you would still like to submit this plugin. Also note that any copyrighted material we will need permission from the copyright holder to include it in the plugin.

PR #6684, opened 2024-09-30
Update target true tile

Notloc:
Just lowering the overlay priority a bit. 
Previously, high priority overlays such as the prayer bar were being obscured, or worse, completely hidden by the actor cutout.


Notloc:
The build failure seems to be on GitHub's side?

iProdigy (approved):

MosheBenZacharia (approved):

PR #6696, opened 2024-10-02
Create boostnotifier

rikbreukers:
Boost Notifier is a handy RuneLite plugin designed to keep track of your skill boosts. You can set up customizable level thresholds for different skills, and the plugin will send you alerts — even email notifications

### DIFF
--- a/plugins/boostnotifier
+++ b/plugins/boostnotifier
@@ -1,0 +1,2 @@
+repository=https://github.com/rikbreukers/boostnotifier.git
+commit=f4472c82c77e42c4778dba7156b2b5051ec027ea


cdfisher:
Please don't close PRs and open new ones if there's an issue. Just push commits to the same branch.

Additionally, if I'm reading it correctly, this would store the user's email password in plaintext in their config, which would be a not-insignificant security concern.

YvesW:
Additionally, could you remove your log.info calls, or replace them with log.debug calls? That way we won't have to read logs with a ton of spam in it.

LlemonDuck:
What possible use case is there for email? It's not real-time, and I don't feel comfortable introducing that form of external communication.

LlemonDuck:

PR #6703, opened 2024-10-03
Update Wasted Bank Space Plugin

mcgeer:
Varlamore pt2 catch-up and delayed runnable removed


MosheBenZacharia (approved):
LGTM

iProdigy (approved):
On plugin shutdown you could also cancel `updateFilterFuture` (but delay is only 250ms so not a blocker here)

PR #6705, opened 2024-10-03
Create skilling-boost-reminder

JonneSaloranta:
This plugin reminds you to use your special attack to boost stats when skilling with appropriate tool in hand.
Incl. woodcutting, fishing and mining with e.g dragon axes which boosts woodcutting when special attack is used.
Didn't find any plugin that does this.

### DIFF
--- a/plugins/skilling-boost-reminder
+++ b/plugins/skilling-boost-reminder
@@ -1,0 +1,2 @@
+repository=https://github.com/JonneSaloranta/skilling-boost-reminder.git
+commit=9563a414f75dc1ac5d094dcc809a3f09813f1d11


YvesW:
Doesn't idle notifier already cover this? Alternatively https://runelite.net/plugin-hub/show/special-attack-notifier if you want a warning box

JonneSaloranta:
The idle notifier doesn't send notifications when in Runelite's notification settings 'Send notifications when focused'.
kuva

and there is only option to use Runelite's notification and it's notifying regardless what the player is doing.
kuva

Seems like the [special-attack-notifier](https://runelite.net/plugin-hub/show/special-attack-notifier) also notifies regardless what the player is doing.

This plugin would only notify when you are doing specific task with specific item(with ability to boost with special attack) in hand and 100% spec. e.g woodcutting with dragon axe/harpoon/pickaxe, otherwise it's completly hidden

Notifies you with:
- per Runelite's notifications settings. (toggleable)
- in player chat (toggleable)
- small arrow blinking near special attack (color settings with alpha)

PR #6706, opened 2024-10-03
Update quest-helper to 4.7.0.1

Zoinkwiz:
Tiny fix removing 4 lines for shortest path to ensure it gets cleared correctly in quest shutdown.

### DIFF
--- a/plugins/quest-helper
+++ b/plugins/quest-helper
@@ -1,2 +1,2 @@
 repository=https://github.com/Zoinkwiz/quest-helper.git
-commit=092724258cf4ea58b5a464d02abf98a960d59a05
+commit=8d1e4693ac61368fdabd44360ecd6c98ecd0bb42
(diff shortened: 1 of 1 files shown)


iProdigy:
Could you retain a `getQuestHelper().getConfig().useShortestPath()` check within `DetailedQuestStep#disableShortestPath` to reduce the chance of clearing a pathfinding target set by a different plugin?

Aside: would be cool if pathfinding supported conditional clear based on the target location (might open a pr later)

Zoinkwiz:
> Could you retain a `getQuestHelper().getConfig().useShortestPath()` check within `DetailedQuestStep#disableShortestPath` to reduce the chance of clearing a pathfinding target set by a different plugin?
> 
> Aside: would be cool if pathfinding supported conditional clear based on the target location (might open a pr later)

I've added a new function for the quest step closing down, which will do a check for the config state. For the existing disableShortestPath() function it won't as it should be being called when the config has been turned off (and thus why it's trying to turn it off).

iProdigy:
ah that explains why the conditional was initially negated, thanks!

iProdigy (approved):

pajlada (approved):

PR #6708, opened 2024-10-03
Update mastering-mixology

whiitehead:
These changes have the updated xp values.


YvesW:
Only the original author or authors that have been added to the plugin file are allowed to update the plugin. That's Hex in this case.

PR #6711, opened 2024-10-03
Create lite-stat-bars

Varietyz:
A plugin to generate skill stat bars that display accurately on the RuneLITE theme, includes extra customization options.

image

from
image

to
image

### DIFF
--- a/plugins/lite-stat-bars
+++ b/plugins/lite-stat-bars
@@ -1,2 +1,2 @@
 repository=https://github.com/Varietyz/litestatbars.git
-commit=a2ba63258c157dfab49c92bb4479d858a733744c
+commit=52b317a2bea4c2f3e0615d02f367dcf18ef0ad6f
--- a/plugins/lite-stat-bars
+++ b/plugins/lite-stat-bars
@@ -1,0 +1,2 @@
+repository=https://github.com/Varietyz/litestatbars.git
+commit=a2ba63258c157dfab49c92bb4479d858a733744c


cdfisher:
Is this something you could integrate with your existing spec regen plugin? I'm not sure if having multiple separate plugins intended for use with a single resource packs is something that would be wanted, and it seems like combining them into a single plugin with the various interface tweaks would make sense.

Varietyz:
> Is this something you could integrate with your existing spec regen plugin? I'm not sure if having multiple separate plugins intended for use with a single resource packs is something that would be wanted, and it seems like combining them into a single plugin with the various interface tweaks would make sense.

The lite meters are universally appliable on the theme's and vanilla.
The status orbs are not universally, so i thought to not cluster the plugins config for those wanting only the lite regen meters and not have the space occupied of the status bars, since the bars plugin is bigger then the meters. 

i do see how it would make more sence if the meters were still pack unique, but after some updates, their universally appliable.

Integrating options to turn these off in the config itself, i feel like would cause more code to have to be written then what they are now. so i terms of total KB's togheter and both seprate. the seprate ones will put less resources on lower end pcs.

YvesW:
I don't think integrating it into one plugin would make it significantly harder to run. Your other plugin also works without the resource pack.
Please add this to your previous plugin as a config option.

I'd also like to take this opportunity to point out that the plugin file for your other plugin is not completely properly licensed. You used large parts of code of both the RegenMeterPlugin and PrayerPlugin, but did not credit the authors of the PrayerPlugin. Neither are their names found in for example the ``LICENSE`` file.

Varietyz:
> I don't think integrating it into one plugin would make it significantly harder to run. Your other plugin also works without the resource pack. Please add this to your previous plugin as a config option.
> 
> I'd also like to take this opportunity to point out that the plugin file for your other plugin is not completely properly licensed. You used large parts of code of both the RegenMeterPlugin and PrayerPlugin, but did not credit the authors of the PrayerPlugin. Neither are their names found in for example the `LICENSE` file.

dont mind me asking, but where exactly could i find the authors details? 
i'm currently having a hard time locating them :l 

i'm so sorry, i'm quiet new to this...

raiyni:
They are referring to the license comment at the top of the runelite files

Varietyz:
> They are referring to the license comment at the top of the runelite files

i have an extremely hard time locating the authors...
a screenshot or path would help me out as to where to look.
sorry to be such a pain :/ i feel stupid.

Varietyz:
i think i have done what you asked of me. although i am not quite sure if i've done it correctly.
Mind checking it for me, to see if i'm within the guidelines now?
thank you!

Varietyz:
Done Merging, Will open another PR to update the LITE meter with these changes.

PR #6712, opened 2024-10-04
Create toa-swarmer plugin

FiveNine:
This plugin displays numbers on the swarms in kephri boss room as they spawn.
In a similar way to how the `nyloer` plugin displays wave numbers on the nylos as they spawn.

Edit: The reason I didn't consider adding this functionality to the tombs-of-amascut plugin is because this plugin is going to eventually have another similar feature to nyloer, where it displays leaked swarms after room completion in a side-panel.


LlemonDuck:
> Edit: The reason I didn't consider adding this functionality to the tombs-of-amascut plugin is because this plugin is going to eventually have another similar feature to nyloer, where it displays leaked swarms after room completion in a side-panel.

Hi, I maintain that plugin. I'd gladly welcome both changes if you're willing.

LlemonDuck:
If not, I may still steal them and add them to that plugin regardless.

FiveNine:
Awesome!
I shall add this to the existing plugin then

PR #6718, opened 2024-10-04
Update marks-of-grace-counter to v1.3.2

Cyborger1:
Should be the last one for now, unless the user who requested this feature reports issues with it.

Essentially just adds an experimental settings section with configs to display an extra line on the overlay which counts the time passed since the beginning of the minute when a mark last spawned. Done by simply truncating the time to the minute with an option to offset the "start" of the minute in case the in-game ticks aren't lining up quite right.

The idea revolves around the marks' spawning mechanics, where the cooldown timer is not based on the exact moment of the mark spawning, but rather the first tick for the minute it spawned in.


iProdigy (approved):

MosheBenZacharia (approved):
very cool. was just thinking yesterday how this would be a nice feature. I can't really figure out if this logic is correct but seems like a good start

PR #6719, opened 2024-10-04
Create crabman-group-bronzeman-plugin

mvdicarlo:
This plugin introduces a group version of the bronzeman plugins.

Originally based off of [another-bronzeman-mode](https://github.com/CodePanter/another-bronzeman-mode) but with heavy modification and changes.

Group collaboration is enabled by users providing an Azure Storage Account SAS Token. I understand this is somewhat limiting, but trying to configure for multiple databases is currently not within scope. As far as setup and costs go are relatively cheap.

Apologies if the setup is a little messy as I forked, and just figured things out from there. Any further guidance is welcome.

### DIFF
--- a/plugins/crabman-group-bronzeman-plugin
+++ b/plugins/crabman-group-bronzeman-plugin
@@ -1,2 +1,2 @@
 repository=https://github.com/mvdicarlo/osrs-crabman-mode.git
-commit=b23e857a02cd0bec5c7a35d2ec631252ca9cd652
+commit=f6e23c2a6e6cfd2dcd52bfa61d77383ae4f732d3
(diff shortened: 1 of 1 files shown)


LlemonDuck:
* A nitpick, but you can just add a second "Copyright" line to the license file instead of duplicating the entire license https://github.com/mvdicarlo/osrs-crabman-mode/blob/9333442a3a2139e6f39242a06f641ae16e3c6d4b/LICENSE
* Replace println calls with log.debug https://github.com/mvdicarlo/osrs-crabman-mode/blob/9333442a3a2139e6f39242a06f641ae16e3c6d4b/src/main/java/mvdicarlo/crabmanmode/CrabmanModeStorageTableRepo.java#L102
* Inject the existing OkHttpClient instead of creating your own, and make sure to update the user agent string https://github.com/mvdicarlo/osrs-crabman-mode/blob/9333442a3a2139e6f39242a06f641ae16e3c6d4b/src/main/java/mvdicarlo/crabmanmode/AzureTableApi.java#L21

mvdicarlo:
Thanks for the feedback.
I have updated the requested parts.

iProdigy:
Please unregister the `GBM_RECENT_STRING` command on plugin shutDown

PR #6731, opened 2024-10-07
Create vorkath-attack-counter

NoExtraSauce:
Vorkath attack counter, that resets after it does one of its two speical attacks. Also flashes with a red box when the attack counter is at 6. This is useful for timing his special attacks without having to manually count each of his attacks.


harbingerofme:
I'm pretty sure this goes against [the third party guidelines](https://secure.runescape.com/m=news/third-party-client-guidelines?oldschool=1).
Namely, "Indicates the time where a boss mechanic may start or end"

PR #6734, opened 2024-10-08
Add sceneryremover

rdutta:
This plugin removes objects from the scene, with the intention of reducing visual clutter when doing any activities.

I have read the [wiki](https://github.com/runelite/runelite/wiki/Rejected-or-Rolled-Back-Features) regarding "ID based plugins", but I am not sure what rules could potentially be broken with this, so I didn't add any kind of constraints or id blacklists etc.

This is essentially the unrestricted version. I will make changes if its possible to have this added.


hex-agon:
As per wiki, we do not want new by-id plugins as moderating them is troublesome.

PR #6737, opened 2024-10-08
VM Recolor Added

rtaylor4444:


YvesW:
Don't use enabledByDefault = false in the plugin descriptor

DominickCobb-rs:
Hey y'all. The VMRecolor is disabled by default so as not to confuse users who just want to download Ultimate Volcanic Mine. Bundled the recolor with this though to reduce clutter and encourage people to use the preferred VM plugin.

raiyni:
Is there a specific reason these are bundled together. It's preferable they are separate plugins.

DominickCobb-rs:
If it's preferred these not be shipped together then I'll submit a PR later this afternoon to add it as its own.

I don't remember where or who told me to ship it with this to reduce plugin hub bloat.

Alexsuperfly:
added in #6751

PR #6741, opened 2024-10-09
Create ladydragontracker

SonderAu:
First stable build of account tracker for clan. 
Using HTTPS, sends players xp values of skills to backend server which gets stored in values. 
Intended for semi-private (within osrs clan) use until backend moved to public repositories.

### DIFF
--- a/plugins/ladydragontracker
+++ b/plugins/ladydragontracker
@@ -1,0 +1,2 @@
+repository=https://github.com/SonderAu/ladydragon.git
+commit=65539805ed7516412806989aee2f5cfa9d5c0230
--- a/plugins/ladydragontracker
+++ b/plugins/ladydragontracker
@@ -1,2 +1,2 @@
 repository=https://github.com/SonderAu/ladydragon.git
-commit=65539805ed7516412806989aee2f5cfa9d5c0230
+commit=602a3fac74fbe5c37cb09d1afb46420d2ae215f0


raiyni:
Would wise old man groups and API usage be better for your clan?

SonderAu:
The consensus from a few of the more savvy leaders is that they want the data housed locally. The React frontend gives them more functionality than tools like CML and WOM. Previous "workaround" was running GroupIronman.Men with a couple different groups, however this has pushed past the scope of that.

cdfisher:
Why not just grab it from the highscores with a script then? I don't really see a need for this to be an entirely separate plugin, when an external Python script or something if the sort could achieve the same thing.

SonderAu:
Understandable, in its current form its quite raw and doing the absolute basics. Our implementation theory was to get the basics in to the community-hub and continue expanding from there, leveraging the review/build process runelite does to mitigate any major design issues we'd run into. Ultimately we're simply doing a fork of https://github.com/christoabrown/group-ironmen-tracker but due to it deviating from not using his backend/frontend deployment understood it to mean we couldn't "piggyback" off that commit without the clan having to run local builds. 
The primary difference in external scripting vs within RL, was to us just real-time vs scheduled scripting to scrape someone elses endpoint (be it hiscores, WOM, or CML). Forgive me for the block text, just trying to add context.

LlemonDuck:
* Use a more specific config group, like "ladydragon" https://github.com/SonderAu/ladydragon/blob/602a3fac74fbe5c37cb09d1afb46420d2ae215f0/src/main/java/com/ladydragon/LadyDragonConfig.java#L7
* Remove the default url, even though it's localhost, and do nothing if it is left blank https://github.com/SonderAu/ladydragon/blob/602a3fac74fbe5c37cb09d1afb46420d2ae215f0/src/main/java/com/ladydragon/LadyDragonConfig.java#L18
* Replace println calls with log.debug https://github.com/SonderAu/ladydragon/blob/602a3fac74fbe5c37cb09d1afb46420d2ae215f0/src/main/java/com/ladydragon/LadyDragonPlugin.java#L73
* Not strictly necessary, but you can craft json using JsonObject or `gson.toJson(someValueClass)` instead of string interpolation, to handle proper escaping of e.g. quotation marks https://github.com/SonderAu/ladydragon/blob/602a3fac74fbe5c37cb09d1afb46420d2ae215f0/src/main/java/com/ladydragon/LadyDragonPlugin.java#L141
* don't use printStackTrace, the log.error already will https://github.com/SonderAu/ladydragon/blob/602a3fac74fbe5c37cb09d1afb46420d2ae215f0/src/main/java/com/ladydragon/LadyDragonPlugin.java#L151

LlemonDuck:

PR #6744, opened 2024-10-09
Update clear-unsent-messages

lightningboltemoji:
* Fixes compatibility with latest RuneLite version
  * I was injecting a `ScheduledExecutorService` and expecting it to run on the client thread which was probably a bad idea


iProdigy:
You should retain the client thread fix, but why are you creating a new single threaded executor now?

lightningboltemoji:
Oh, hmm... I see your point... my gut reaction was 'oh, maybe I shouldn't have used this', but using it with the assumption that it's not going to run on the client thread makes sense. Will update!

lightningboltemoji:
Updated to inject/reuse `ScheduledExecutorService`. Thanks for the nudge.

iProdigy (approved):

geheur (approved):

PR #6750, opened 2024-10-09
Rainbow Rave 1.3.5 - make plugin compile again.

geheur:


larkable:
It appears that this is just awaiting review from a plugin hub maintainer.

Thanks @geheur for the quick patch to get our rainbow fix. <3

PR #6755, opened 2024-10-11
Update quest-splits

jonas-fo:


iProdigy:
Could you also call `overlayManager.remove(overlay);` on plugin shutDown

jonas-fo:
Yes, I have added that now

iProdigy (approved):

PR #6760, opened 2024-10-12
Create cox-scouting

aghasemi78:
Makes it easier to scout for specific raid layout. Makes default option on the steps leaving the cox raid "reload." Takes options from players about specific rooms wanted in the raid. When all criteria are met, it will make the default option on the steps "walk here" so that it can't be accidentally clicked.

### DIFF
--- a/plugins/cox-scouting
+++ b/plugins/cox-scouting
@@ -1,0 +1,2 @@
+repository=https://github.com/aghasemi78/cox-scouting-plugin.git
+commit=337d8fee9da7e7c4529a90607abecca0a6889b07


cdfisher:
Does this do anything that "CoX Scouting QoL" doesn't already do?

aghasemi78:
> Does this do anything that "CoX Scouting QoL" doesn't already do?
> 
> Additionally, your build is failing because your repo doesn't follow the [plugin template](https://github.com/runelite/example-plugin). Information on how to develop for the plugin hub can be found [here](https://github.com/runelite/plugin-hub/blob/master/README.md).

Technically they achieve the same result but I think this is a little more straightforward for finding a specific set of rooms as opposed relying on the knowledge of specific rotations and whatnot. I'm very new to coding so I will take any feedback you can provide on how to get it to work.

PR #6764, opened 2024-10-13
cox scouting

aghasemi78:
work similar to cox scouting qol but a little simpler


Felanbird:
Please stop spam opening pull requests

PR #6768, opened 2024-10-13
cox scouting

aghasemi78:
This is plugin works very similar to cox scouting QoL however it is much simpler to operate. Simply select all of the raid requirements you want (number of rooms, specific rooms, and/or raid start layout) and the plugin will deprioritize the "reload" option on the steps leading out of cox so that it can't be misclicked once a raid meeting all of your selected criteria is scouted.

### DIFF
--- a/plugins/cox-scouting
+++ b/plugins/cox-scouting
@@ -1,2 +1,2 @@
 repository=https://github.com/aghasemi78/cox-scouting-plugin.git
-commit=392cf4f9ee927c6bc0e2fe5124ad1fa7b19491fb
+commit=3445bdc909395c92eec61da7dc14abe6d0ff761e
--- a/plugins/cox-scouting
+++ b/plugins/cox-scouting
@@ -1,2 +1,2 @@
 repository=https://github.com/aghasemi78/cox-scouting-plugin.git
-commit=032a98a31a6cb96e0ee07e0b0bead9cffe15aadc
+commit=ce574c7e03b6e1c3e35e7c8c1c9a0e6524ec2260
--- a/plugins/cox-scouting
+++ b/plugins/cox-scouting
@@ -1,2 +1,2 @@
 repository=https://github.com/aghasemi78/cox-scouting-plugin.git
-commit=3766dd4e7b2426c5152e2c61fe78424d17b0f2b4
+commit=f3a0da866d5815ed3650e7de1e83d575afe3b13e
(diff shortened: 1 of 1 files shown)


pajlada:
@aghasemi78 Please stop opening new pull requests, and instead update your already open pull request

cdfisher:
To continue the discussion from #6760 we would prefer that you contribute to the existing plugin rather than adding a new plugin that does basically the exact same thing as a plugin that's already on the hub.

aghasemi78:
> To continue the discussion from #6760 we would prefer that you contribute to the existing plugin rather than adding a new plugin that does basically the exact same thing as a plugin that's already on the hub.

Although the plugins achieve the same result they work in different ways. Adding the functionality/settings of this plugin to the other would create a configuration panel that would be confusing to the user. It would require that the user disable all the settings in one section of the plugin before enabling any in the other section. I figured creating a new plugin would be much simpler (both for production and user experience)

aghasemi78:
> That would amaze me since the cox-scouting-qol plugin already does this exact thing AFAIK. At worst it's slightly bugged and might need a small fix like adding a priority for the specific event.
> 
> Also, opening a new PR is basically NEVER the right answer to a problem. If you close/open 8 PRs while getting pinged to not do it, you should first ask yourself if you are doing something wrong.

I didn't create all of the new PRs with the intention of going against the standard practice. I'm still new to how github and plugin development works. I was pinged after the final one I made so I'll keep that in mind for the future.

As for the scouting QoL plugin, many of the people I've interacted with find the plugin confusing and difficult to configure to achieve their desired result. I'm sure it has it's uses but it appears to be not very intuitive to use for many people. I was also not aware of there being any restrictions on similarities between plugins before I finished the development on this. I had seen plenty of plugins on the plugin hub achieve the exact same results as other plugins and assumed there was no issue with me going ahead and developing this plugin. If it's a requirement, I can try to incorporate this plugin into the already existing one but I'd much rather just go ahead with this as a separate plugin seeing as it's ready pending review and approval

YvesW:
https://github.com/runelite/plugin-hub?tab=readme-ov-file#contribute-to-existing-plugins

I think some slight changes/fixes (if needed) to the other plugin would be a perfect fit; you previously also described this plugin as "simpler version of cox scouting qol", so I assume you can understand why.

LlemonDuck:
Yeah, this is straight up a duplicate feature set of cox-scouting-qol. Please try contributing there first, and we can revisit this if the changes are rejected / left unreviewed.

PR #6769, opened 2024-10-13
Update acoldone-clogging

OmbudRov:
Tiny Change of Audio Files from 24 bit to 16 bit cause java didn't like 24 bit files


pajlada:
@OmbudRov You'll need to resolve the branch conflict

PR #6770, opened 2024-10-13
Update wintertodt-scouter

Nouser:
The new commit hash fixes pre-existing users' uplink and downlink settings. It also fixes the config listener.

Previously, if you had the plugin installed before it changed ownership, you would have the old endpoints populated.

Now, you will have the new endpoints populated because the config values don't clash with pre-existing ones.

### DIFF
--- a/plugins/wintertodt-scouter
+++ b/plugins/wintertodt-scouter
@@ -1,3 +1,3 @@
 repository=https://github.com/Nouser/wintertodt-scouter.git
-commit=41003ee2a8953819d9179acc247f581d30496452
+commit=45b3195d2561e2d9b71251480965837cb1529d13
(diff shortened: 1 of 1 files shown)


YvesW:
Hard for me to check this on mobile, but unlike your commit message, you are not changing the config group, right? Just the 2 keys, and the group check in the ConfigChanged event?
(Changing the config group would be incorrect)

Nouser:
I was also changing the config group, because there was an internal mismatch of the string name in the codebase already (`scouter` vs `wintertodtscouter`). I can swap the config group back to its old value and make the other variable in code have that value instead.

YvesW:
That's what I gathered on mobile indeed. Perfect, then I don't have to suggest doing it the proper way or migrating the rest of the config.

PR #6771, opened 2024-10-13
Add music-favourites plugin

Syhlex:
Adds a plugin that allows users to favourite music tracks and view them in a separate list in the Music tab.

Please take a look at the [README](https://github.com/Syhlex/music-favourites?tab=readme-ov-file#music-favourites) in the plugin repo for more details.

### DIFF
--- a/plugins/music-favourites
+++ b/plugins/music-favourites
@@ -1,0 +1,2 @@
+repository=https://github.com/Syhlex/music-favourites.git
+commit=cb1b16151fe51fc2f76821faa98d0ebc68173123


YvesW:
Not an extensive review, but does this plugin really conflict with the core music plugin? I assume you can still show the songs while the core music options work?

FYI, you can also make a config key hidden, and/or just get/set (arbitrary) config values using ConfigManager (getting a value for key that doesn't exist yet returns null though).

Syhlex:
Unfortunately it does conflict with the core music plugin, as that plugin overwrites the music header container (where the button is located) and the entire music list.

YvesW:
Isn't there a way to work around this? As the music plugin also has a lot of other functions, like muting specific area sounds, granular sliders etc.

Syhlex:
The sliders and other options of the core music plugin don't cause any conflicts, it's mainly how it overwrites the music list when you search or press the toggles that makes it incompatible with plugins like mine that also update the music list.

Looking at how the core music plugin works internally, I wasn't able to come up with a workaround other than marking it as a conflicting plugin. Hopefully the original author can chime in or refactor the core music plugin to make it more compatible with other plugins that want to modify the music list view.

YvesW:
Usually core plugins aren't changed to work with hub plugins, mostly the other way around. I've not looked at the code for the core music plugin, but examples include setting subscriber priorities (to run before or after core plugins), detecting when certain actions are performed to then e.g. reload the info you want to display, or changing hub settings based on settings of the core plugin.
Conflicts are usually most appropriate in case the hub plugin basically does the same as the core plugin (e.g. 117HD and GPU).

Syhlex:
I don't think any of those ideas would work in this case unfortunately. 😕
Are there any examples of existing plugins that modify the music list that are also compatible with the core music plugin?

Syhlex:
During development I looked at the core music plugin and tried to make mine work with it but to no avail. Feel free to checkout my branch and remove conflicts line in the config file to see how the Music plugin conflicts with mine.

Is having the music plugin as a conflict a blocker though? I can update the description to make users aware that enabling my plugin will disable the core music one automatically.

LlemonDuck:
It's not a direct blocker, but it's not preferred. What about reordering the widgets to just place the favorites at the top? Then it's not needed to switch tabs in the first place.

If you'd really rather keep it as is, let me know and I'll give it a review as it stands.

Syhlex:
I would prefer to keep it separate so there's a clearer distinction as to what's been favourited and hide irrelevant tracks.

PR #6773, opened 2024-10-14
Update random-hopper to v1.3

rugg0064:
Update random-hopper to v1.3
* Block hop commands during loading and hopping game states
* Add persistent config


YvesW:
Not an extensive review, but could you remove the prints, or at least make them log.debug?

FYI you don't need to unset a configuration before setting it. Also, when a config key doesn't exist yet, getConfiguration will return null, but I assume you accounted for that.

rugg0064:
> Not an extensive review, but could you remove the prints, or at least make them log.debug?
I've removed the prints

> FYI you don't need to unset a configuration before setting it. Also, when a config key doesn't exist yet, getConfiguration will return null, but I assume you accounted for that.
Addressed the configuration unset.
getConfiguration should be handled by the generic try catch, I think at least.

iProdigy:
```
random-hopper: plugin uses terminally deprecated APIs:
```

YvesW:
You're adding a binary file now.

rugg0064:
> You're adding a binary file now.
😓 Apologies. I have no idea where that came from...

PR #6776, opened 2024-10-14
Update ge-filters to v1.1.2

salverrs:
- Fixed outdated api
- Fixed Inventory Setups fetching logic

Sorry about the delay on fixing this.


LlemonDuck:
It seems like a bad idea to me to be migrating inventory setups' config between versions yourself. At most you should be reading the config values to add support, but please don't go writing into other plugins' configs unless absolutely necessary.

If the user has inventory setups installed, that plugin will perform the migration already. You can listen for onConfigChanged events under the correct group to determine when to reload entries.

PR #6783, opened 2024-10-15
Create damage-sounds

Worley03:
A simple plugin that plays custom sounds when the player takes set amounts of damage.

### DIFF
--- a/plugins/damage-sounds
+++ b/plugins/damage-sounds
@@ -1,0 +1,2 @@
+repository=https://github.com/Worley03/rlplugin-damage.git
+commit=c2134a112c4a1b27440d3ffd708d794669b43715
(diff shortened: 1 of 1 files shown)


tylerwgrass:
You can update your PR instead of closing and opening a new one.
Also your file should be in the `plugins/` directory

Worley03:
> You can update your PR instead of closing and opening a new one. Also your file should be in the `plugins/` directory

Thanks.

LlemonDuck:
Please use `log.debug` statements instead of `System.out.println`

Also you'll need to make sure you're closing the Clip resources to not leak memory. One common way to do so without cutting them off is to store a reference to the last played clip, and closing it when trying to play the next one (and on plugin shutdown).

The included sounds are very cute 💟

PR #6788, opened 2024-10-15
Disable slayer addtions

JZomDev:
It breaks plugins that try to highlight things.
NPC indicators (core)
the gauntlet
hunter rumors

to name a few


LlemonDuck:
cc @SkyBouncer 

Once the fix is merged you can resubmit the plugin with the new version and removing the disabled line.

@iProdigy has written you a fix already in SkyBouncer/sky_slayer#5

PR #6791, opened 2024-10-16
Update bank-tag-generation

MitchBarnett:
Updated to the wiki bank tag generation plugin with fixes so it is back to working order.

The plugin is currently marked as unavailable and incompatible on the plugin hub, I'm not sure if that is an automatic thing or if someone will need to also update that.


MitchBarnett:
Wrong commit

Felanbird:
You can just push a second commit with the correct hash, it doesn't require a new PR

iProdigy (approved):
I would recommend using `okhttp3.Call#enqueue` to execute the http requests in a separate thread pool that doesn't block the client thread (and once you deserialize the response, you can schedule task invocation back on the client thread as necessary) - but that doesn't need to be done for this PR

PR #6797, opened 2024-10-16
Update Equipment_Lock

Matimatician:
Added toggles to allow to exclude clue scroll emote items


iProdigy (approved):
could refactor to use `ItemID` instead of name for item allowlists, but not necessary

PR #6804, opened 2024-10-17
Update for Skilling Boost Reminder

JonneSaloranta:
Update contains:

- Fixes for detecting fishing and woodcutting correctly.
- Update values in config to make them more coherent
- Fixing notifier arrow positioning

### DIFF
--- a/plugins/skilling-boost-reminder
+++ b/plugins/skilling-boost-reminder
@@ -1,2 +1,2 @@
 repository=https://github.com/JonneSaloranta/skilling-boost-reminder.git
-commit=9563a414f75dc1ac5d094dcc809a3f09813f1d11
+commit=f56ad7a7f92d9f071ddf1781301b2dc35fa4a184


pajlada (approved):

iProdigy (approved):
Heads-up: `@Range` is only enforced when the user changes the config value, so changing the allowed range will not immediately impact saved values that are now outside the accepted range. For `offsetY` this isn't a huge deal so I'll merge anyways & you can submit an additional patch if desired

JonneSaloranta:
@iProdigy For the upcoming patches. What is/are the best way(s) to endorse this in the future?

iProdigy:
> What is/are the best way(s) to enforce this in the future?

on plugin startUp: if offsetY is negative, then set offsetY to zero (you'll want to inject ConfigManager to achieve this, or modify your config file to have a setter for offsetY)

PR #6828, opened 2024-10-20
Skilling boost reminder update

JonneSaloranta:
ADD:
- X-Axis position setting for notification arrow
- More fishing situations where boost can be used

FIX:
- Added check for values to keep them in allowed ranges even after updates

CHORE:
- Added Javadoc comments for animation IDs
- Update README


YvesW:
Needs rebase

PR #6842, opened 2024-10-21
Create object-highlighter

Car-Role:
Object Highlighter Plugin: Initial implementation and README

Implemented a new RuneLite plugin for highlighting game objects in OSRS. Features include:

1. Highlight objects by ID or name
2. Three highlighting modes: tile, hull, and clickbox
3. Customizable colors and border widths
4. Partial name matching support
5. Added comprehensive README with usage instructions and configuration options.

### DIFF
--- a/plugins/object-highlighter
+++ b/plugins/object-highlighter
@@ -1,2 +1,2 @@
 repository=https://github.com/Car-Role/Object-highlighter.git
-commit=1202cf3bc495e66c12d74ae0657bc9fabf34869c
+commit=a975daf989c495bb85a0239ad3fe88b896a4409c
--- a/plugins/object-highlighter
+++ b/plugins/object-highlighter
@@ -1,0 +1,2 @@
+repository=https://github.com/Car-Role/Object-highlighter.git
+commit=868ccba2e1349259830105de7bf50587d2786773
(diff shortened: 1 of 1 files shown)


Felanbird:
> Reviewing

Car-Role:
> > Reviewing
> > We will review your plugin to ensure it isn't malicious, doesn't [break Jagex's rules](https://secure.runescape.com/m=news/third-party-client-guidelines?oldschool=1), or isn't one of our previously [Rejected/Rolledback features](https://github.com/runelite/runelite/wiki/Rejected-or-Rolled-Back-Features).
> 
> This plugin would fall under our Rejected feature for any new `ID based plugins`
> 
> > Plugins that use player provided IDs for the entirety of their functionality can cause moderation issues and outcomes that break Jagex's plugin rules. Due to this fact we will not be accepting any new ID based plugins. Plugins that use a specific set of IDs but do not allow user input will still be accepted. (e.g. plugins like Vardorvis Projectiles which only allows you to change a specific projectile to a set list of projectiles.)

How is this any different than Object markers, the output is the same just a different way of managing the marked objects? this plugin introduces no new functionality

Felanbird:
> How is this any different than Object markers, the output is the same just a different way of managing the marked objects? this plugin introduces no new functionality

You have to be able to right-click an object to mark it, if you do it by ID it's possible you could highlight something without a right-click menu, which would make it harder for us to moderate. For example the clearly problematic behavior that suddenly appeared when Duke's vents were implemented, via the existing ID hub plugins.

LlemonDuck:
Closing, relevant above comment on why.

PR #6845, opened 2024-10-22
Update ban-list-export

P2GR:
Export to Sheet / Import to Sheet. this allows for an external banlist which allows clans to exceed the 100 bans limitter. This makes it possible to clear out old banned usernames, but still keep track of them using the external list.

### DIFF
--- a/plugins/ban-list-export
+++ b/plugins/ban-list-export
@@ -1,2 +1,2 @@
 repository=https://github.com/P2GR/ban-list-export.git
-commit=f0d0ee0a73b3558ceb2399fd349a17962644c568
+commit=0bda2879e95fa997ed3db18c2f47e5713fda8758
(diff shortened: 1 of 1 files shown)


iProdigy:
There's no need to do `eventBus.register(this)` (runelite already takes care of this for the plugin class)

also instead of creating a `ScheduledExecutorService`, you could utilize `@Schedule` from `net.runelite.client.task` (so a new thread doesn't need to be created for your plugin)

lastly within `sendWarning` could you include a prefix that makes it clear which plugin the message is coming from

P.S. not a blocker for merging, but you only need to run `Text.standardize(username)` once (rather than on every loop iteration): https://github.com/P2GR/ban-list-export/blob/0bda2879e95fa997ed3db18c2f47e5713fda8758/src/main/java/com/clan_ban_list_export/ClanBanListExportPlugin.java#L252

P2GR:
@iProdigy I've implemented the feedback you provided, let me know if there's anything else :)

PR #6849, opened 2024-10-23
Embargo - Add minigame completion tracking

Sharpienero:
New:
- Update readme to include new feature information, including a disclaimer that we do not store IP address
- Bump version to 1.3.0
- Add minigame completion tracking

Fix:
- Only send STANDARD profile raid/minigame completions
- Abstract logic to keep raid/minigame completion logic in a single function, `handleActivityCompletion`
- Rename `getRaidMembers()` to `getSurroundingPlayers()` to better describe what the function does


iProdigy:
Could you remove the `printStackTrace` calls within `DataManager`; just combine this with your existing slf4j calls

iProdigy (approved):

PR #6861, opened 2024-10-24
Create goading-helper

NCG-RS:
This plugin contains the following features for use with the new goading potions.

1. Radius overlay to show aggression range.
2. Reminder overlay to remind user to sip potion when they have it in inventory and aren't under effect
3. Notification on potion expiry
4. Tick timer to display when the aggression check happens


NCG-RS:
PR closed after noticing missing info in gradle files, recreating

PR #6862, opened 2024-10-25
Create goading-helper

NCG-RS:
This plugin contains the following features for use with the new goading potions.

Radius overlay to show aggression range.
Reminder overlay to remind user to sip potion when they have it in inventory and aren't under effect
Notification on potion expiry
Tick timer to display when the aggression check happens


cdfisher:
If you need to make additional changes, don't open new PRs, just push to the existing branch.

PR #6863, opened 2024-10-25
Create aerial-fishing-pearl-luck

H4waiianPunch:
This plugin tracks the number of fish that it takes to obtain a Molch Pearl. It will also track your biggest dry streak with persistence via the profile. Lastly, it will give the rough odds of a user getting the Golden Tench.

### DIFF
--- a/plugins/aerial-fishing-pearl-luck
+++ b/plugins/aerial-fishing-pearl-luck
@@ -1,2 +1,2 @@
 repository=https://github.com/H4waiianPunch/Pearl-Luck-Tracker.git
-commit=63f08e8f47870cc82bf3cc0c494f11063e6259bc
+commit=baaf25bb769a4087ea19193a570cb3dea18e209b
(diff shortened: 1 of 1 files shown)


LlemonDuck:
Downgrade log.info calls to log.debug https://github.com/H4waiianPunch/Pearl-Luck-Tracker/blob/9dd68e27417cde0a37ec72a9332e66ed073f3faa/src/main/java/com/AerialFishingPearlLuck/AerialFishingPlugin.java#L78

LlemonDuck:
don't do this, you're duplicating the config instance https://github.com/H4waiianPunch/Pearl-Luck-Tracker/blob/63f08e8f47870cc82bf3cc0c494f11063e6259bc/src/main/java/com/AerialFishingPearlLuck/AerialFishingPlugin.java#L50

`@Provides` registers the object to the injector module, so you can just use 
```
@Inject
AerialFishingConfig config;
```

PR #6867, opened 2024-10-25
Update ban-list-export

P2GR:
- added highlighting in cc panel
- added tooltip in cc panel
- added warning in game chat
- added support for bannedby, reason, date through import.


iProdigy:
```
ban-list-export: plugin uses terminally deprecated APIs:
Do not create fresh OkHttpClients - always @Inject the client's OkHttpClient
```

P2GR:
@iProdigy
Can you point me in a direction about the Terminally Deprecated API calls?

iProdigy:
You've fixed the okhttp issue but if you're curious the full list is here: https://github.com/runelite/plugin-hub/blob/master/package/package/src/main/resources/net/runelite/pluginhub/packager/disallowed-apis.txt

PR #6877, opened 2024-10-26
Create valuable-drop-prices

Anderzenn:
Simple plugin that is meant to replace the default drop notifications while adding high alch prices to the chat notification.
User can change whether to see high alch or grand exchange prices, or to just show them both. Other than this they can also decide if ha or ge is is deemed valuable, or if it should use both (highest valued item).
Colour of chat notification can also be modified.

### DIFF
--- a/plugins/valuable-drop-prices
+++ b/plugins/valuable-drop-prices
@@ -1,0 +1,2 @@
+repository=https://github.com/Anderzenn/valueable-drop-prices.git
+commit=24ccad83606ae612a12fde14650ab1be1d87f61e


LlemonDuck:
Can this be made a lot simpler _and_ durable by just piggybacking off the in-game chat message, instead of disabling it? You can edit the existing message node to append the GE/HA values, which would also remove the need for doing your own drop tracking.

The whole plugin could be ~15 lines of code, excluding boilerplate.

Disabling the native drop notifications will conflict with a lot of other popular plugins.

Anderzenn:
> Can this be made a lot simpler _and_ durable by just piggybacking off the in-game chat message, instead of disabling it? You can edit the existing message node to append the GE/HA values, which would also remove the need for doing your own drop tracking.
> 
> The whole plugin could be ~15 lines of code, excluding boilerplate.
> 
> Disabling the native drop notifications will conflict with a lot of other popular plugins.

That's a good idea, I don't know why I didn't think of that. I'll modify the code to piggybank off the default message instead of the current way I'm doing it :) Thank you!

PR #6888, opened 2024-10-27
Add valuable-drop-prices

Anderzenn:
Simple plugin that replaces the default "coins" value with HA and GE prices to the valuable drop notifications.

Now completely rewritten with the changes recommended in pull request #6877, without having to disable the default notifications! :)


Anderzenn:
I will say I didn't mean to close the previous pull request, but something messed up on my end and it ended up being the easiest solution.

LlemonDuck:
Merging since this all looks fine, but fyi you can `@Inject @Named("developerMode") boolean developerMode;` instead of having a debug config option.

PR #6889, opened 2024-10-27
Update shortest-path to v1.16.3

Skretzo:
- The draw transports option should now work when you don't have an active shortest path, correctly respect your config options, and no longer include teleports
- The wilderness ditch has been extended with 11 tiles to reflect the Defender of Varrock update

### DIFF
--- a/plugins/shortest-path
+++ b/plugins/shortest-path
@@ -1,2 +1,2 @@
 repository=https://github.com/Skretzo/shortest-path.git
-commit=bcb844211869270db1f87f01f158c5464a858fdd
+commit=42b7900e330b20dd9f1811c6289855c59a3e672f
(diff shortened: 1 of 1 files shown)


iProdigy:
technically doesn't matter for the review but: this always evaluates to true because your hashCode uses the latest tickcount rather than storing a constant one (also overriding hashCode like this is hacky but won't impact merging): https://github.com/Skretzo/shortest-path/blob/bcb844211869270db1f87f01f158c5464a858fdd/src/main/java/shortestpath/ShortestPathPlugin.java#L322

iProdigy (dismissed):

Skretzo:
> another fyi: `PathfinderConfig#refresh` will be regularly called now on LOADING -> LOGGED_IN game state change (but also not a blocker for merging)

Knew there was something I had forgotten.

iProdigy (approved):
on plugin startUp, you might want to add: `if gameState == LOGGED_IN: invokeLater(pathfinderConfig::refresh)`

LlemonDuck:
Feel free to submit another PR if you want to do the above ^

PR #6890, opened 2024-10-28
Fixed performance issue, added option to count full potions only

mrchooch:


iProdigy:
Is changing the config group intentional? You will effectively be setting everyone on the default config

mrchooch:
Oops, good catch

iProdigy:
Could you also remove this unused field https://github.com/mrchooch/potion-bars/blob/dd2197a0b904c95ee46be67715288886865365d4/src/main/java/com/potionbar/PotionBarPlugin.java#L30 (also unnecessary given `@Slf4j`)

iProdigy (dismissed):
final nit: could do `potionPanels.clear();` on plugin shutDown

iProdigy (approved):

tylerwgrass (approved):

PR #6896, opened 2024-10-28
add antifire-checker plugin

sariyamelody:
I've written a plugin to warn you when you're almost out of antifire juice. Great for people who walk away from their desk habitually.

### DIFF
--- a/plugins/antifire-checker
+++ b/plugins/antifire-checker
@@ -1,2 +1,2 @@
 repository=https://github.com/sariyamelody/antifire-checker.git
-commit=32992addaedd1947265e4e55dba770a42b9e0038
+commit=a4d1c91d0e5f78ae57a1111ec5aa005329ac7d70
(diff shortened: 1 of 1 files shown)


sariyamelody:
image
Here's a screenshot of it in action!

sariyamelody:
bumping the version to address the gradle lint failure

cdfisher:
Can this not already be accomplished with something like Watchdog?

sariyamelody:
I went looking extensively and Watchdog never came up, that might be a route that could address this need.

Investigating Watchdog now, it appears that it does not have any way to grab arbitrary varbits (i.e. Varbits.ANTIFIRE / Varbits.SUPER_ANTIFIRE) or their related status effects, which means that Watchdog is limited to alerting on "has the 'antifire close to expiry' message fired?" or similar -- the point of this is allowing configurable thresholds for users to manage their own AFK desires :)

sariyamelody:
(Another update, to fix the lint again, hopefully!)

adamk33n3r:
Is that an...UNDERLAY? :O but reading this, there's nothing saying I wouldn't add a varbit option ;) I actually have had pouches requested which I think also use varbits etc so it'll probably happen at some point.

PR #6904, opened 2024-10-29
Create party-status-bars-overlay

lou0815:
plugin extends runelites partyservice to also show a fully configurable overlay of all the partymembers inside the game window


lou0815:
did some updates:
-reversed the threshold for special attack energy since its more useful to see when a spec can be used
-noticed the run energy threshold had the wrong value so fixed that
-cleared up the config a bit to be more readable
-updated readme

PR #6905, opened 2024-10-30
Update Aerial Fishing Pearl Luck to version 1.1

H4waiianPunch:
Fixes:
- Tench Chance now resets when plugin is turned off/on.
- Overlay will no longer show without the Cormoront's Glove
- Tench chance percentage is calculating properly


iProdigy (approved):
on plugin startUp you might want to check if the current game state is LOGGED_IN

on plugin shutDown, you might want to add `overlayAdded = false;`

TheStonedTurtle (approved):

PR #6908, opened 2024-10-30
Create potion-man

Noctiverse:
Plays a "potion man" jingle when Dr. Jekyll spawns


Nightfirecat:
This will not work if it is built and deployed, because the compiled plugin will not include your sound file. See https://github.com/runelite/plugin-hub?tab=readme-ov-file#plugin-resources

Noctiverse:
Thank you for the quick response (and documentation as I haven't worked with Java before this)!

I updated the commit ID in the PR file, do I need to do anything else to help your review processes?

iProdigy:
You should use `getResourceAsStream` instead of `new File`, as explained in the link shared by NFC above

Also you open resources on npc spawn but never close them

Lastly please replace your `Exception#printStackTrace` call with slf4j (e.g., `log.debug`)

PR #6909, opened 2024-10-30
Update shortest-path to v1.16.4

Skretzo:
- Fixes number format exception in the last unsuccessful update, and adds assertions and exceptions to handle errors
----
- Added boat between Burgh de Rott, Meiyerditch, Icyene Graveyard and Slepe
- Added row boat between Port Phasmatys and Slepe
- Added Port Phasmatys energy barrier gates
- Added pendant of ates
- Added Necropolis rubble


iProdigy:
Could you replace your `printStackTrace` calls with slf4j?

also fyi: assertions are not enabled in the production client

iProdigy (approved):

tylerwgrass (approved):

PR #6910, opened 2024-10-30
Bump fixed-mode-hide-chat

deathbeam:


pajlada:
This updates the "Hotkey used to toggle hiding the chat" config key name, which will make people lose their configs. You might want to revert that key change back to `hideChatHotkey`

YvesW:
Is renaming the config key worth it, since it'll be wiping the value?
Additionally, will this consume everyone's escape key by default? Maybe that will cause some support traffic since it's often used to close the bank or to swap tabs.

Finally, it's replacing the hide chat option with a toggle. Maybe people will complain that the old behaviour is gone, but that's a decision you have to make yourself.

deathbeam:
The config rename doesnt matter and it doesnt consume escape key (as it would also need to consume it in the other handlers i guess). But the current behaviour is kinda annoying as it both closes a bank and the chat at once so /shrug

deathbeam:
Yea I think the new behaviour is too annoying so just gonna revert it.

PR #6911, opened 2024-10-30
Re-enable cg recolor

denaelc:
Fixed the bug, very minor changes. Somehow forgot a try-catch block which sometimes resulted in a nullpointer exception

### DIFF
--- a/plugins/cg-recolor
+++ b/plugins/cg-recolor
@@ -1,2 +1,3 @@
 repository=https://github.com/denaelc/gauntlet-recolor.git
 commit=bebcd9ca906e44932fb560754825b3be454a90e8
+disabled=Requires bugfix
--- a/plugins/cg-recolor
+++ b/plugins/cg-recolor
@@ -1,3 +1,2 @@
 repository=https://github.com/denaelc/gauntlet-recolor.git
-commit=bebcd9ca906e44932fb560754825b3be454a90e8
-disabled=Requires bugfix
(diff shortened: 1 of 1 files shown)


Adam-:
Can you make this use the `PreMapLoad` event instead?
 I'd prefer if this plugin wasn't readded as is; the way it works is by invalidating the gpu scene cache causing everything to be streamed each frame, which just gives very poor performance, and it isn't clear to users the reason why. Additionally, if you are getting npes from getModel(), it means you are holding onto and calling getModel() on despawned objects, which is just very wrong.

denaelc:
Yeah, that’s why I took it down in the first place instead of just pushing a quick fix. I may be tied up for the next couple of days, and since some users requested that this be uploaded again soon, I decided to re-upload it in a working state with only these minor changes. I’ll make the "bigger" changes in the near future if that's okay with you. If not, that's totally fine 👍

denaelc:
I'm now using the new preMapLoad event and removed all instances of "setting the scene ID on objects". I'm also no longer trying to call getModel() on despawned Objects. The performance should now be improved substantially - thanks again for the new event!

dan-drews:
Is this one ready to be reviewed? I've been missing this plugin, hope it'll be available soon :)

alexgermann:
Really missing this plugin, eagerly awaiting its return!

Xilador:
I appreciate all the work you do Adam, can't wait to get this plugin back

PR #6912, opened 2024-10-30
Update Fairy-Ring-Butterflies to fix issues which were caused by improper use of models on unloaded rings

Jmmuir:


iProdigy:
Could you replace your System.out/System.err calls with slf4j?

iProdigy (approved):

tylerwgrass (approved):

PR #6925, opened 2024-11-03
Update Essence Pouch Tracker to v1.5.9 & Update repo URL

Infinitay:
(Hopefully) Fixed a few issues since releasing Essence Pouch Tracker initially onto the plugin hub (https://github.com/Infinitay/essence-pouch-tracker/tree/bbc4c83fd2dee9935dd6d460185d345ebe686277 v1.4.0) as well as add a few new features

# Fixes
- Fixed the plugin being broken/uninitialized when using using it for the first time when the player is already logged in with essence already in their inventory
	- Refactored resetting config and plugin start up/initialization 
- Only manipulate pouches that are currently within the player's inventory and not every pouch regardless if it's in the inventory or not
- Mitigate wrong pouch states by changing the behavior to not update the pouch's respective states if a property (stored/decay count) is unknown
- Fix game stutters that would be caused when withdrawing large quantities of stackable items
	- Also optimized EssencePouches type retrieval
- Fixed some cases in which pouches weren't being reflected as repaired due to missing dialog options
	- Fixed not being able to repair colossal pouches, and added some of Cordelia's wrong/missing repair options

# Features
- Add support for specialized equipment preventing pouch decay such as Runecrafting/Max capes and the Abyssal Lantern
- Add a chat message when a user has a pouch with any unknown state in their inventory on a new login session (similar to Daily Tasks Plugin IIRC)
- Add support for pouch state resetting when a user upgrades their pouch from a regular pouch to a colossal pouch or downgrades from a colossal pouch to a regular pouch

# Other Changes

- Renamed the plugin (Plugin Description) from `Essence Pouch Tracking` to `Essence Pouch Tracker`
- Renamed the GH repo and updated the URL to reflect the changes: `https://github.com/Infinitay/essence-pouch-tracker.git`
- Removed left-over debug statements that occurred on every MenuOptionClicked

### DIFF
--- a/plugins/essence-pouch-tracking
+++ b/plugins/essence-pouch-tracking
@@ -1,2 +1,2 @@
-repository=https://github.com/Infinitay/essence-pouch-tracking.git
-commit=bbc4c83fd2dee9935dd6d460185d345ebe686277
+repository=https://github.com/Infinitay/essence-pouch-tracker.git
+commit=000c2f0d461cf9e18a9908d0e6680a7a94219647
--- a/plugins/essence-pouch-tracking
+++ b/plugins/essence-pouch-tracking
@@ -1,2 +1,2 @@
-repository=https://github.com/Infinitay/essence-pouch-tracker.git
-commit=000c2f0d461cf9e18a9908d0e6680a7a94219647
+repository=https://github.com/Infinitay/essence-pouch-tracking.git
+commit=f8aec844f6e952089bae6d6028f0c8f7dbcc4db2


pajlada (commented):
Nothing necessarily blocking, but a few things you might want to sneak in

nit: line 19 of readme, remove [ before the "Install the plugin" text
nit: line 29 of readme: update repo URL
nit: EssencePouch.java line 49: the debug message will post the current current/new stored essence as the "previously" because you log it after you've set `this.storedEssence`
nit: EssencePouch.java line 61: same as above
code style: EssencePouchTrackingOverlay.java line 34: The if-statement is needlessly complicated. If I understand it correctly, it's `if (a == null || !(b || c))` while it could be `if (a == null || !b || !c)`
code style: For the RepairDialog.java messages, you could use [Text.removeTags](https://static.runelite.net/api/runelite-client/net/runelite/client/util/Text.html#removeTags(java.lang.String)) to clean up the tags, that way your list of dialog options would look a bit cleaner
nit: EssencePouchTrackingPlugin.java line 181: `a == 1 ? true : false` is equivalent to `a == 1`
nit: EssencePouchTrackingPlugin.java line 218: same as above

question: EssencePouchTrackingPlugin.java line 1401: Can `pouch` be null there? If yes, it's probably something you should check

Infinitay:
- [x] nit: line 19 of readme, remove [ before the "Install the plugin" text
- [x] nit: line 29 of readme: update repo URL
- [x] nit: EssencePouch.java line 49: the debug message will post the current current/new stored essence as the "previously" because you log it after you've set this.storedEssence
- [x] nit: EssencePouch.java line 61: same as above
- [x] code style: EssencePouchTrackingOverlay.java line 34: The if-statement is needlessly complicated. If I understand it correctly, it's if (a == null || !(b || c)) while it could be if (a == null || !b || !c)
- [ ] code style: For the RepairDialog.java messages, you could use [Text.removeTags](https://static.runelite.net/api/runelite-client/net/runelite/client/util/Text.html#removeTags(java.lang.String)) to clean up the tags, that way your list of dialog options would look a bit cleaner
		---
 I'm not using Text#removeTags to keep it simple and avoid doing more calls with text formatting when a simple comparison would work. Furthermore, I tested it out and noticed that it would have made the code or rather the text look worse (imo). For example, `Fine. A simple transfiguration spell should resolve things<br>for you.` => `Fine. A simple transfiguration spell should resolve thingsfor you.`

 - [x] nit: EssencePouchTrackingPlugin.java line 181: a == 1 ? true : false is equivalent to a == 1
 - [x] nit: EssencePouchTrackingPlugin.java line 218: same as above

pajlada (approved):

tylerwgrass (approved):

PR #6936, opened 2024-11-06
Update Inventory Setups to v1.20.0 and update Bank Tag Layouts to 1.4.33 (Core BTL Migration)

dillydill123:
Core BTL migration for inventory setups.

Disable inventory setups integration in hub BTL


geheur:
I approve

PR #6945, opened 2024-11-08
added spec jingle

Servilabs:
The spec jingle plugin plays a jingle sound whenever the special attack energy reaches a value on the user-specified list. Main usecase is for remembering to equip dps rings on time.


cdfisher:
Is there a reason you can't just use Idle Notifier in core RuneLite? It has an option for spec regen notifications.

Servilabs:
The idle notifier would be great if I could add more than one threshold

PR #6946, opened 2024-11-09
Create Soulreaper Axe QoL

seacelery:
A plugin to add a few quality of life improvements for using the Soulreaper Axe including:
* An additional non-interactable minimap orb to display special attack & Soulreaper Axe stacks at the same time
* Tracks the time left until a Soulreaper Axe stack is lost similar to the spec regen timer
* Ability to set a colour for the spec orb and the Soulreaper orb to keep things distinct

I have some example images in the README.


tylerwgrass:
You'll need to move your readme assets outside of `src/`.

PR #6947, opened 2024-11-09
Update Teleport Maps Plugin (spirit-tree-map) to version 1.0.07

MJHylkema:
- Skills Necklace added - thanks @Blubmin!

image


tylerwgrass (approved):
I recommend updating your  `onConfigChanged` handler to check that `configChanged.getGroup()` matches your plugin config group

iProdigy (approved):
less important than the config group check: you have `Pattern.compile(LABEL_NAME_PATTERN)` multiple times in the plugin, I'd just do this once and save it to a static variable

PR #6955, opened 2024-11-10
Updating Did I compost to version 1.1.1

AbusiveTuna:
Adds new feature to show which patches need compost by @Car-Role


iProdigy:
* you could load these images once rather than multiple times every frame render: https://github.com/AbusiveTuna/DidICompost/blob/0e5a7f6f43a781969fc67e0f487d70e72e01d1b2/src/main/java/com/compost/PatchOverlay.java#L113-L125
* this approach to removing from a list can miss some elements: https://github.com/AbusiveTuna/DidICompost/blob/0e5a7f6f43a781969fc67e0f487d70e72e01d1b2/src/main/java/com/compost/DidICompostPlugin.java#L198-L204 - an easy fix: `currentTiles.removeIf(tile -> tile == oldPatch.tile)` (or you can iterate in reverse order)
* `savedPatches` is written to, but not actually used to do anything
* the `setWorldPoints` and `setNeedsCompostPoints` calls don't really do anything (the list reference never changes; did you mean to yield list *copies* from the getters to avoid ConcurrentModificationException?)
* consider clearing your plugin state on shutdown (worldPoints, needsCompostPoints, savedPatches)

AbusiveTuna:
> * you could load these images once rather than multiple times every frame render: https://github.com/AbusiveTuna/DidICompost/blob/0e5a7f6f43a781969fc67e0f487d70e72e01d1b2/src/main/java/com/compost/PatchOverlay.java#L113-L125
> 
>     * this approach to removing from a list can miss some elements: https://github.com/AbusiveTuna/DidICompost/blob/0e5a7f6f43a781969fc67e0f487d70e72e01d1b2/src/main/java/com/compost/DidICompostPlugin.java#L198-L204 - an easy fix: `currentTiles.removeIf(tile -> tile == oldPatch.tile)` (or you can iterate in reverse order)
> 
>     * `savedPatches` is written to, but not actually used to do anything
> 
>     * the `setWorldPoints` and `setNeedsCompostPoints` calls don't really do anything (the list reference never changes; did you mean to yield list _copies_ from the getters to avoid ConcurrentModificationException?)
> 
>     * consider clearing your plugin state on shutdown (worldPoints, needsCompostPoints, savedPatches)

Updated to latest commit after your pull request, and a quick fix for checking plots

iProdigy (approved):

tylerwgrass (approved):

PR #6958, opened 2024-11-11
Aerial Fishing Pearl Luck Update to Version 1.2

H4waiianPunch:
Looking to Merge in the changes I've made to my plugin.
- More stats
- More control for users
- Hopefully I did everything right this time..!


tylerwgrass (commented):
- Saving multiple config options every fish attempt seems a bit much. I would consider saving maybe once a user ends a session instead? (leaves the island, logs off, shuts down plugin, etc)
- It doesn't really make sense to use the config checkboxes to reset stats
- You are calling `pearlRateWikiCalc` even when the user is doing fishing/hunter activities that aren't aerial fishing
- Less of a blocker maybe, but in my opinion I don't think it's necessary to provide a config option to show/hide every field in the overlay.

H4waiianPunch:
> * Saving multiple config options every fish attempt seems a bit much. I would consider saving maybe once a user ends a session instead? (leaves the island, logs off, shuts down plugin, etc)

Ah, that's a very valid point. Didn't even think about it making constant calls. That's fixed.

>     * It doesn't really make sense to use the config checkboxes to reset stats

I wasn't sure how else to do this as I couldn't find a way to just put a new "Reset" button in or anything else. What do you suggest for this?

>     * You are calling `pearlRateWikiCalc` even when the user is doing fishing/hunter activities that aren't aerial fishing

This will take some more looking into, but I'll fix this.

>     * Less of a blocker maybe, but in my opinion I don't think it's necessary to provide a config option to show/hide every field in the overlay.

I had a user that wished it to be fully customizable so they could choose what they see, I think it's a good option for people these days. Everyone always asks for a toggle after all!

tylerwgrass:
> I wasn't sure how else to do this as I couldn't find a way to just put a new "Reset" button in or anything else. What do you suggest for this?

You can add menu entries to the overlay, which will appear when you shift + right click. 
image
[example](https://github.com/runelite/runelite/blob/b71211d6457dc6f57ece1c863fd1cc713d075cd8/runelite-client/src/main/java/net/runelite/client/plugins/agility/LapCounterOverlay.java#L54-L55)

> > * You are calling `pearlRateWikiCalc` even when the user is doing fishing/hunter activities that aren't aerial fishing

> This will take some more looking into, but I'll fix this.

You could probably add a regionId check for the island

H4waiianPunch:
Hi Tyler, 
I think this new version should meet your requirements now.

I decided to not go with the region ID option(as I couldn't quickly figure it out). Instead, I made sure that the hunter and fishing calcs done in pearlRateWikiCalc were only being done if the overlay was added.

Let me know if there's anything else!

tylerwgrass (commented):
If you would really like to keep the fully customizable panel, then I would recommend storing the config values for use outside of the overlay render method as constantly fetching them from the config in `render` could have performance impacts.

tylerwgrass (approved):

H4waiianPunch:
First off, thank you for the approval.

> If you would really like to keep the fully customizable panel, then I would recommend storing the config values for use outside of the overlay render method as constantly fetching them from the config in `render` could have performance impacts.

I'll have to try and figure out how this works, I don't want to cause performance issues for users.

> Less importantly it looks like [src/main/java/com/AerialFishingPearlLuck/PearlCalcType.java](https://github.com/H4waiianPunch/Pearl-Luck-Tracker/compare/f4f1fdec763962391b41a1fbad2eba7d351745f5...H4waiianPunch:108cad4a02129582ec8297c37c0f174e48ae3ab9#diff-6db4233f3400001f1e09f9b0efee5846d0354e9b3817e8b71d82c0097af6c3a0) is unused

You're right, I was using that and then changed how things worked. Thank you for bringing it up, I'll be sure to remove it.

LlemonDuck:
> I'll have to try and figure out how this works, I don't want to cause performance issues for users.

in onStartUp, and onConfigChanged, read the config entries and store them in variables. Then use those variables during the render call

H4waiianPunch:
I think the changes I've just made should work for what was mentioned. Let me know if there's anything else!

LlemonDuck:
just use something like this instead of parsing all the getNewValue fields

```java
public void onStartup() {
  // ...
  readConfig();
}

@Subscribe
public void onConfigChanged(ConfigChanged configChanged)
{
  if (configChanged.getGroup().equals("pearlluck"))
  {
    readConfig();
  }
}

private void readConfig()
{
  totalFishCaughtEnabled = config.totalFishCaught();
  sessionFishCaughtEnabled = config.sessionFishCaught();
  fishCaughtEnabled = config.fishCaught();
  lastPearlEnabled = config.lastPearl();
  sessionPearlsEnabled = config.sessionPearls();
  totalPearlsEnabled = config.totalPearls();
  actualPearlRateEnabled = config.actualPearlRate();
  wikiPearlRateEnabled = config.wikiPearlRate();
  dryStreakEnabled = config.dryStreak();
  bestStreakEnabled = config.bestStreak();
  totalTenchEnabled = config.totalTench();
  showTenchChanceEnabled = config.showTenchChance();
}
```

also just a stylistic thing but you don't need getters for private fields, that's the point of a private field

if you really want them, they can be autogenerated with lombok's `@Getter` e.g. 
```java
@Getter 
private boolean totalPearls;

// somewhere else ...
isTotalPearls() // defined implicitly by lombok
```

PR #6961, opened 2024-11-11
OSRS DataLog

DeadAlready:
Lets you send the log of state changes in varbits and varplayers to either a file or a custom server. 

It is basically WikiSync plugin, but with controls over manifest and output target.

### DIFF
--- a/plugins/osrs-data-log
+++ b/plugins/osrs-data-log
@@ -1,0 +1,2 @@
+repository=https://github.com/DeadAlready/OSRSDataLog.git
+commit=4f824e7f0309d17a983629c9309bfbd3bf445ff2


raiyni:
I believe action logger plugin already does this

DeadAlready:
Not quite - action logger tracks varbit changes and dialogue and writes it to disc. Gives no control over what is tracked and where the output is going with the purpose of debugging quest helper.
DataLog is meant to be custom WikiSync - so you can track your characters progress (levels, qp and whatnot you deem you want) and either use it for something like custom leaderboards or progress visualisation. That means sending the data to a server in chunks.

They are relatively similar - with enough configuration options ActionLogger could do the same thing. That said if we are going down the route of extending existing to minimize amount of plugins, then most similar would be WikiSync itself as DataLog is WikiSync with configuration options - however I felt that it would be very unlikely that wikisync would merge these options as they would be useless for 99% of users and only provide confusion.

Let me know your opinons
Thank you

iProdigy:
> DataLog is meant to be custom WikiSync - so you can track your characters progress (levels, qp and whatnot you deem you want) and either use it for something like custom leaderboards or progress visualisation. That means sending the data to a server in chunks.

have you considered using [dink](https://github.com/pajlads/DinkPlugin/blob/master/docs/json-examples.md) to create those leaderboards? or is there some varbit you care about that is not tracked by dink?

DeadAlready:
Thank you for the recommendation - after some investigation I think I should be able to get most of my wants covered by the dink plugin with some configuration. Thus I shall close this PR for now.

Also extra thank you for being responsive and helpful :)

PR #6966, opened 2024-11-12
Distraction reducer

Car-Role:
Changed config option of bake pie to prevent false positives of contact NPC


tylerwgrass (approved):
FYI you removed `STRING_JEWELRY_ANIMATION_IDS`, not sure if that was intentional

Car-Role:
It should be removed and I'll address a better way to handle it in the future

@tylerwgrass

iProdigy (approved):

PR #6976, opened 2024-11-15
WildernessAlert

Duhstin:
This plugin is similar to an existing one, except this one will look at the combat levels of the other player within distance and will only flash if they can attack you.


Felanbird:
> Reviewing

This plugin would fall under our Rejected feature for any new `Level-based PvP player indicators`

cdfisher:
This would likely be considered to go against the [third-party client guidelines](https://secure.runescape.com/m=news/third-party-client-guidelines?oldschool=1)

While it may not be your intended use case, this sounds like it could be used for such purposes.

PR #6978, opened 2024-11-15
wilderness-course-ticket-reminder plugin

neafey:
Plugin that reminds the player to claim a ticket after completing a lap in the Wilderness Agility Course.
Highlights the dispenser after completing a lap to help player remember to claim your ticket.
Also changes the default left-click action on the pipe to "Walk here" if no ticket has been claimed.

### DIFF
--- a/plugins/wilderness-course-ticket-reminder
+++ b/plugins/wilderness-course-ticket-reminder
@@ -1,0 +1,2 @@
+repository=https://github.com/neafey/Wilderness-course-ticket-reminder.git
+commit=46514096866e017fb3018971ba8d4810ca603989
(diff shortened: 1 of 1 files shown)


LlemonDuck:
* You should remove the unused TicketReminderConfig. You can re-add it later if you need to introduce config options.
* Objects#requireNonNull is an assertion, not a guarantee, and does not prevent NPEs. Uou're leaking these onto the client thread. Since it _can_ actually be null, you should check for such yourself.

neafey:
Thank you for the feedback! This is my first plugin, so I was a bit unsure about whether to remove the config if it wasn’t used. I’ve removed it now.

You’re absolutely right about Objects#requireNonNull. I’ve replaced it with a proper null check instead.

I hope this addresses everything, but I’m happy to make more changes if needed to get the PR accepted.

PR #6981, opened 2024-11-16
Party map pin

DoinkOink:
- Added the ability to have a custom map pin
- Added the ability to change the map pin size


LlemonDuck:
Your branch needs rebased

https://github.com/runelite/plugin-hub?tab=readme-ov-file#updating-a-plugin

PR #6983, opened 2024-11-16
Update party-map-pin

DoinkOink:
- Added the ability to have a custom map pin
- Added the ability to change the map pin size


DoinkOink:
I'm a little confused why it's saying I'm missing a license file.

PR #6984, opened 2024-11-16
Update spec-jingle

Servilabs:
update spec jingle to new version, which adds 8 new sounds and a new threshold section for values that only play when the spec reaches that exact value.

### DIFF
--- a/plugins/spec-jingle
+++ b/plugins/spec-jingle
@@ -1,2 +1,2 @@
 repository=https://github.com/Servilabs/spec-jingle.git
-commit=2045fb5173674d4ae03af109fb37096b482da726
+commit=ff26195ee22105cfae5d493855cc9ba520c9570e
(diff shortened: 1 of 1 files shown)


tylerwgrass:
How is someone supposed to know how many sounds there are, what they are, and how to use them? 
I don't think someone reading the config is going to understand how to use this.

Servilabs:
Thank you, noted, I have an instruction section to the config panel.

iProdigy:
it's still unclear to users what sound id's 1-9 correspond to

Servilabs:
I just looked for a bunch of different id's that might be good for notifications and somewhat different.

Servilabs:
sound description has been added to the configuration

iProdigy (approved):
the game sound IDs (e.g., 3924) probably don't need to be in the instructions (more likely to confuse users)

regardless, LGTM (code could be optimized, but not a blocker)

tylerwgrass (approved):
some things to consider:
- The "how to" stuff would be good to also maybe have in the README
- if a user has a different sound set for both the threshold configs at the same threshold, the first sound is ignored, which might confuse people (no real suggestions here just fyi)

PR #6988, opened 2024-11-17
Update line-markers to v1.3.6 and npc-id to v1.3.4

Skretzo:
- No longer show line markers on the wrong floor on the minimap
----
- Added options to show npc id and index in the right-click menu


LlemonDuck:
For the future, I would prefer each plugin to be in its own PR for separate review

PR #6995, opened 2024-11-18
Update fun reminders

hybridbrizz:
- switch repository path to updated username
- add idle npcs condition

### DIFF
--- a/plugins/fun-reminders
+++ b/plugins/fun-reminders
@@ -1,2 +1,2 @@
-repository=https://github.com/erversteeg/Proximity-Prompts.git
-commit=3c683879686582caa57d9f75abe406eab9e54480
+repository=https://github.com/hybridbrizz/Proximity-Prompts.git
+commit=4a2efb404e79c63c245b0e346a29a6ee2b320ce9


tylerwgrass:
Can you clear `idleNpcs` and `npcLastPositions` in `shutDown`? I would also recommend this for your other lists, maps, etc. 

It also looks like nothing is ever removed from `npcLastPositions`, which is going to likely cause memory leaks. Maybe they should be removed on despawn?

hybridbrizz:
Sure thing got those items added. I also renamed the plugin and made a few other changes.

tylerwgrass (dismissed):
lgtm, though some others like your original plugin name more (this won't block the merge if you would like to keep the updated name)

hybridbrizz:
I'm good with the updated name. Going to see if it makes more sense to users. Fun reminders is a cool name but hopefully the name will make it more clear what the plugin does.

raiyni:
What's the point of adding idle NPCs? We've been denying ID based plugins that requires user input recently.

hybridbrizz:
No worries guys. If it's no good then do what you need to do. I'm adding it because a user requested it.

hybridbrizz:
Going to go ahead and close the PR. Like I said it's something someone wanted but it's not worth all the trouble. I know you guys are busy and I don't want to be a burden.

PR #6997, opened 2024-11-18
Create: Obelisk protection

Car-Role:
RuneLite plugin that protects players from accidentally using 
their POH Wilderness Obelisk when carrying valuable items. Key features include:

- Wealth-based protection system that calculates risk excluding 3 most valuable items
- Configurable wealth threshold (default 1M GP)
- Ground markers indicating when protection is active
- Customizable marker colors
- Removes dangerous menu options when protection is active

Core files:
- ObeliskProtectionPlugin.java: Main plugin logic
- ObeliskProtectionConfig.java: Configuration interface
- ObeliskProtectionGroundOverlay.java: Ground marker rendering
- README.md: Plugin documentation


tylerwgrass:
Some thoughts/curiosities:
- Does this work as expected for untradeables? I am not familiar with the specifics of how those prices are considered regarding risk/if the GE value is somewhat simulated accordingly
- If you are removing all the teleport options from the menu, I would consider an additional indication somewhere of what is doing that as right now the user could disable the label and forget the plugin is active and not know why they can't teleport.    
- Have you checked this works as expected with swapped menu entries?

Lastly, while not directly related, I could see this functionality potentially making sense as a feature addition to the existing [Obelisk Reminder](https://github.com/AcolyteOfFire/Obelisk_Reminder/tree/master) plugin instead

Car-Role:
Thank you for the feedback

> * Does this work as expected for untradeables? I am not familiar with the specifics of how those prices are considered regarding risk/if the GE value is somewhat simulated accordingly

Yes in my testing

> * If you are removing all the teleport options from the menu, I would consider an additional indication somewhere of what is doing that as right now the user could disable the label and forget the plugin is active and not know why they can't teleport.

This is a very niche situation -- The user has to have the plugin on, and the user has to have 4+ items in their inventory that exceeds the user-defined threshold (1m by default) and the user has to have manually turned off the warning indicator, AND the user has to actually want to use the obelisk to teleport

It might happen but I think it should be monitored for now and if issues come up address it when that becomes a problem users run into

> * Have you checked this works as expected with swapped menu entries?

works as intended


> Lastly, while not directly related, I could see this functionality potentially making sense as a feature addition to the existing [Obelisk Reminder](https://github.com/AcolyteOfFire/Obelisk_Reminder/tree/master) plugin instead

These both interact with obelisks but the use case has no overlap beyond that. There is almost no shared code

PR #6998, opened 2024-11-18
Submit TileMarkerMetronome

RenaudBernon:
Tile Marker Metronome is a adaptation of the built in `GroundMarkers` plugin.
Added functionalities include: 
- Creating groups of Tiles
- Changing tile color every x ticks
Tile groups have separate settings, and can be hidden/shown separately as well.

### DIFF
--- a/plugins/tile-marker-metronome
+++ b/plugins/tile-marker-metronome
@@ -1,2 +1,2 @@
 repository=https://github.com/RenaudBernon/tile-marker-metronome.git
-commit=b58e39961c55276a0d9a04023b40eed016da3fa0
+commit=46dc642def4816a69717474755a96aadbf94f3c3
--- a/plugins/tile-marker-metronome
+++ b/plugins/tile-marker-metronome
@@ -1,2 +1,2 @@
 repository=https://github.com/RenaudBernon/tile-marker-metronome.git
-commit=46dc642def4816a69717474755a96aadbf94f3c3
+commit=e9a25f92e777dc8d777f40f7c66a35470a146836
--- a/plugins/tile-marker-metronome
+++ b/plugins/tile-marker-metronome
@@ -1,2 +1,2 @@
-repository=https://github.com/abd-khrubi/better-godwars-overlay.git
-commit=a7da67b9bddf5b14aeb54d789b1dfe02965d473f
+repository=https://github.com/RenaudBernon/tile-marker-metronome.git
+commit=b58e39961c55276a0d9a04023b40eed016da3fa0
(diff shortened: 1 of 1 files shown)


plugins/tile-marker-metronome
```diff
@@ -0,0 +1,2 @@
+repository=https://github.com/abd-khrubi/better-godwars-overlay.git
+commit=a7da67b9bddf5b14aeb54d789b1dfe02965d473f
```
iProdigy, line 2:
```suggestion
repository=https://github.com/RenaudBernon/tile-marker-metronome.git
commit=b58e39961c55276a0d9a04023b40eed016da3fa0
```

iProdigy:
```
tile-marker-metronome: Missing LICENSE file
All plugins must be licensed under a license that allows us to freely distribute the plugin jar standalone.
We recommend the BSD 2 Clause license.
```

iProdigy:
not a complete review but: on plugin shutDown, please remove navigationButton from the ClientToolbar

not as important: your game state logic will be called more frequently than you think due to the intermediate LOADING state (e.g., LOGGED_IN -> LOADING -> LOGGED_IN)

RenaudBernon:
> not as important: your game state logic will be called more frequently than you think due to the intermediate LOADING state (e.g., LOGGED_IN -> LOADING -> LOGGED_IN)

Doesnt the intermediate `loading` state also load new regions?
Granted, I shouldn't be doing `loadGroups();` but `loadPoints();` seems necessary?

RenaudBernon:
Ill remove the load groups()
If you think the IO for the load points is taxing I can also add a check for the region so I only load the points here when necessary.

iProdigy:
another option is to inject ScheduledExecutorService and perform that IO off of the client thread, but I doubt there is enough of a performance hit to require you to make that change

raiyni:
Is there a serious use for this outside of timing some kind of boss/skill/minigame mechanic and knowing where to stand?

RenaudBernon:
@iProdigy, I removed the navigation on shutdown, and removed loading groups on gamestate changed
@raiyni, The main usecase I see indeed is timing during PvM encounters, maybe sepulchure.
You can also turn the Falador party room into a disco floor if thats your thing, but I recon thats not what most people would do ;)

raiyni:
I'm struggling to see the validity of this plugin outside of using it for boss/minigame mechanics. I think this would  need to be heavily gated to be allowed but other contributors can chime in I guess.

LlemonDuck:
If I'm understanding correctly, this is just visual metronome but on ground tiles, right?

Does the player have to manually start the timers each time or are they started when the map scene is loaded?

LlemonDuck:
yeah so since this would effectively synchronize the tile colours with the map loading, that's not gonna be allowed as it could be used to derive many boss mechanic timings

LlemonDuck:
if you want to recreate it somehow that it can't provide that benefit, feel free to resubmit

PR #7017, opened 2024-11-20
Update Inventory Setups Fuzzy Changes

dillydill123:
Small bugfix. Fuzzy items are added at the bottom. I recently made it so the placeholders would be added to the bottom as well. However, when updating a setup, the placeholders would be deleted then re-added, causing them to teleport to the end of the layout. This fixes that issue by not removing them in the first place if the non placeholder ID is in the setup.


geheur:
lgtm

PR #7022, opened 2024-11-21
Update: Obelisk protection

Car-Role:
- Fixed "Protection Active" text appearing outside POH
- Added isInPOH() check to verify POH obelisk presence
- updated how the overlay is being displayed. now shows it only while the obelisk is being moused over
- fixed how stackable items were being handled for wealth calculation and potentially losing them in the wilderness


raiyni:
You don't need to ping individual mergers as soon as you open the plugin

iProdigy:
Couldn't you use region IDs to check POH status rather than tile iteration? https://github.com/runelite/runelite/blob/890acd684cabfce17bbbe0a648662906050855e6/runelite-client/src/main/java/net/runelite/client/plugins/roofremoval/RoofRemovalConfigOverride.java#L35

Car-Role:
> Couldn't you use region IDs to check POH status rather than tile iteration? https://github.com/runelite/runelite/blob/890acd684cabfce17bbbe0a648662906050855e6/runelite-client/src/main/java/net/runelite/client/plugins/roofremoval/RoofRemovalConfigOverride.java#L35

That's the first thing I tried but for some reason, I wasn't able to make it work properly. This approach does work though

iProdigy:
> That's the first thing I tried but for some reason, I wasn't able to make it work properly

did you remember to use `isInInstancedRegion`/`fromLocalInstance`?

Car-Role:
> > That's the first thing I tried but for some reason, I wasn't able to make it work properly
> 
> did you remember to use `isInInstancedRegion`/`fromLocalInstance`?

Changed it to this approach and got it working. Thanks

Car-Role:
The version that's live right now is fairly buggy. I would really appreciate someone taking a look at this pr that fixes all these critical bugs

raiyni:
You should be using GameObjectSpawned/Despawned events instead of iterating every tile in the scene on every menu entry added

iProdigy:
They're referring to the `findObelisk` call in `onMenuEntryAdded` (that still exists)

iProdigy (approved):
my last performance nit: `isInPOH` gets called multiple times in a given tick - you could update this once a tick (or consider whether it's even necessary given `pohObelisk` is now a field)

tylerwgrass (approved):
Feel free to address the perf comment in a separate PR.

PR #7030, opened 2024-11-23
add BotHunter

olipayne:


raiyni:
We already have a plugin like this, bot detector, and it causes problems of people falsely accusing legitimate players of being bots and using that information to harass them.

olipayne:
> We already have a plugin like this, bot detector, and it causes problems of people falsely accusing legitimate players of being bots and using that information to harass them.

Well aware of this, the existing Bot Detector plugin does extremely rudimentary analysis and forced modelling which will:
 - misclassify a lot of humans as bots, just because they do the same thing
  - misclassify a lot of bots as humans, simply because a profile for them has never been built.

My proposed platform is entirely open source and I intend to make the raw data available to data scientists who want to investigate it further, there are too many bots out there to try to profile them all, we need to crowd-source data to identify anomalies, not specific patterns already pre-defined.

extreme4all:
@olipayne would love to collab, your intentions are basically the same we had when we started but somewhere due to lack of skill & time we ended up where we are now, while we achieve some of what we intend to we'd wish we could achieve more... happy to hear your vision and see how we can improve things
you can contact me easily via discord https://discord.gg/bot-detector-plugin-817916789668708384

LlemonDuck:
Can we close this PR then? Are you two collaborating on this?

PR #7046, opened 2024-11-25
Add rsvr

ekqrnw:
A diorama-style VR plugin.


raiyni:
Programmatic clicking is not allowed

Cyborger1:
I think the only way something like this could get added is if the VR part is purely graphical. 

The effort of going through being able to click world elements in 3D space and projecting that back into the corresponding 2D interface point is commendable. However, the programmatic clicking required to make this work cannot be allowed, as raiyni said. 

If you instead control the actual mouse onto the 2D plane of the interface window and project a line from the UI plane to the 3D environment onto the object being clicked, this may be fine?

Basically the controls would be no different than using Virtual Desktop to control your mouse on Windows, and playing without the controllers (mouse+kb) would also be an option. I do understand this might be unwieldy to control, but this is likely the only avenue this has any chance of being accepted onto the plugin hub.

LlemonDuck:

PR #7052, opened 2024-11-26
Update amulet of chemistry alerter

josmoo:
added new alchemist's amulet, huasca, and aldarium. also added new banks


TheStonedTurtle (approved):
LGTM but just a heads up that you have a reference to `21163` as a magic constant when you can use `ItemID.AMULET_OF_CHEMISTRY` instead.

tylerwgrass (approved):

PR #7056, opened 2024-11-27
Create clanchatlogger

cchvezj:
Attempting to log when users enter and exit a clan chat, as well as how long they are in said chat. (This is a work in progress and can be ignored if needs be)

### DIFF
--- a/plugins/clanchatlogger
+++ b/plugins/clanchatlogger
@@ -1,0 +1,2 @@
+repository=https://github.com/cchvezj/clanchatlogger.git
+commit=dd8a30a29657785d63caed8f4f82bd773f234a61


iProdigy:
This can already be achieved with https://runelite.net/plugin-hub/show/clan-chat-webhook

cchvezj:
> This can already be achieved with https://runelite.net/plugin-hub/show/clan-chat-webhook

I wanted it to only track when a person enters and exits, that won't pull that data (at least from what I am seeing while using it) It appears to only populate the data if a person actually leaves the clan or is invited into it, not logging in/out.

iProdigy:
Have you considered PRing the login/logout faeture to that plugin?

cchvezj:
> Have you considered PRing the login/logout feature to that plugin?

That would probably honestly be easier than doing it from scratch, I didn't necessarily want a public version just something to help with the clan I am in on the administrative side. Clan is full and we need to see which members are active and which aren't to clear up space.

LlemonDuck:
Based on your last response, I assume you'll be trying to PR it and we can close this PR?

PR #7058, opened 2024-11-27
add poison-moo

remindful7560:
Plugin request for Poison Moo.

Causes your character to moo (using overheads) on the poison timer, like cows do (according to Mod Ash).


raiyni:
java code needs to target 11

remindful7560:
Whoops. Switched to Java 11.

PR #7059, opened 2024-11-27
Add bank-locks plugin

vartan:
which allows users to prevent accidentally depositing items into the bank.


raiyni:
Can't you just menu swap an option in the interface instead?

vartan:
>Can't you just menu swap an option in the interface instead?

Sort of, but you also have to remember to also menu swap deposit-all items/equipment. And remember to swap them both back when you're done.

PR #7089, opened 2024-11-30
Update Dodgy Deals

MouldySamuel:
Updated to V1.2


iProdigy (approved):
btw you can use `@Range` to enforce a minimum on `pluginHideTime`

tylerwgrass (approved):

PR #7091, opened 2024-12-01
Add laststand plugin

ghost:
Adds an infobox timer for the Last Stand relic.

### DIFF
--- a/plugins/laststand
+++ b/plugins/laststand
@@ -1,0 +1,2 @@
+repository=https://github.com/roundshoe/laststand.git
+commit=658c565adb01df127b551bab2a4f4246347890bb


DominickCobb-rs:
Hi! I made the Undying Retribution timer plugin last year. I didn't have the message from the relic yet this year, but given I just have to change out the message I'd like to just re-enable the one I created last year as it covers the edge cases and is basically feature complete.

DominickCobb-rs:
I'd like to collaborate if possible. I've updated [my repo](https://github.com/DominickCobb-rs/undying-retribution-timer) to use the new messages which should work in place. Could you test this and confirm?

PR #7093, opened 2024-12-01
Create forager-pouch-alerts

austinleblanc:
This is a simple plugin that watches chat to see if you receive the notification that you failed to insert an herb into your forager's pouch, and then notifies you.


Felanbird:
Do we need a plugin that only reads a single chat message? Would `Chat Notifier` not suffice for this feature?

austinleblanc:
> Do we need a plugin that only reads a single chat message? Would `Chat Notifications*` not suffice for this feature?

Sure 🤷

PR #7095, opened 2024-12-01
Update valuable drop prices

Anderzenn:
Minor quality of life changes

- Added basic colour configuration.
- Added option to apply different colours to different sections of the message.
- Added number formatting
- Fixed bug where if "BOTH" was selected as display type but there was no ha value, it would still print ha - but with no value.
- Added/refined options for debugging/development of plugin.


Anderzenn:
Sorry for the general weirdness of this pull request, there was a conflict for whatever reason, should be good now.

PR #7096, opened 2024-12-02
Add load-time plugin

rsfost:
This plugin attempts to calculate the client's load time when loading into a new area. Load time can be useful awareness in situations where you have to react to something on the first tick of loading into an area.


rsfost:
It has some niche uses in barbarian assault speed running

PR #7097, opened 2024-12-02
Create last-stand-timer

BtwDas:
This simple plugin adds a 3min timer when the player dies. 

This is made to give players an indication when their last stand timer is about to come up.


raiyni:
See https://github.com/runelite/plugin-hub/pull/7091#issuecomment-2510201293

PR #7101, opened 2024-12-02
New plugin/rune lite duplicator osx

NickSpaghetti:
This plugin allows Mac Os Users to open up a new runelite client without having to go into the terminal and use open command or shift click on Runelite.app.


raiyni:
We don't allow plugins that use `exec()` calls

NickSpaghetti:
> We don't allow plugins that use `exec()` calls

I assume then ProcessBuilder is off the table as well?

PR #7113, opened 2024-12-06
Clue scroll notifier

SoaresPT:
Unsure if it's allowed since the readme prefers us not to download resources on the user's machines but tried to implement this as per this issue: https://github.com/SoaresPT/clue-scroll-notifier/issues/1

Changes:

- Plugin now downloads 5 "default" .wav files. They won't be replaced if the user keeps the same name allowing them to have their own unique sounds
- Added the ability to switch which sound get played.
- Added ability to toggle notifications for Scroll Boxes for leagues.


LlemonDuck:
Provided it is only 5 sound files, it would be nicer if you can include them in the plugin resources and just unpack them into the dir on startup instead of fetching remotely.

SoaresPT:
No longer grabs the sound files from remote and instead has them packed on the resources dir as per readme instructions

PR #7120, opened 2024-12-07
Update `hunter-rumours` to `1.3.2`

geel9:
- Correctly tracks hunter creatures caught while a league XP multiplier is active


tylerwgrass:
Did you verify normal worlds still work as expected? (I'm not sure how the leagues varbits work in this case).

tylerwgrass (approved):

iProdigy (approved):

PR #7125, opened 2024-12-08
Object Highlighter

ShaneeexD:
Enables players to highlight any in-game object with customizable colors and styles through a simple shift + right-click interface.

Includes individual highlighting or "All"

### DIFF
--- a/plugins/object-highlighter
+++ b/plugins/object-highlighter
@@ -1,0 +1,2 @@
+repository=https://github.com/ShaneeexD/ObjectHighlighter.git
+commit=7069dcab4210d8776c2b92db40440936e8f7abce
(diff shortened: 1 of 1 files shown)


YvesW:
Why are you trying to remake object indicators? What does this plugin do significantly different that the core plugin doesn't already do (better)?

ShaneeexD:
> Why are you trying to remake object indicators? What does this plugin do significantly different that the core plugin doesn't already do (better)?

Mine allows to tag "All" of that object ID as well as individually, which can be useful for certain cases

raiyni:
We no longer allow highlight by id plugins

ShaneeexD:
> We no longer allow highlight by id plugins

Ah okay then, sorry for asking but I'm curious as to why

cdfisher:
> > We no longer allow highlight by id plugins
> 
> Ah okay then, sorry for asking but I'm curious as to why

There's too much potential for abuse, and it puts a big burden on reviewers.

PR #7126, opened 2024-12-09
BugFix to Prevent Leagues Tracking

OSRSKoeppy:
Added a few lines of code to prevent Leagues drops and achievements from being tracked


YvesW:
You are not added as author. You need to be added to the author field of the plugin file, and the original author has to confirm in this PR that you are allowed to submit plugin updates.

joelhalen:
I've added @OSRSKoeppy as an author, and this acts as me confirming he is authorized to submit updates on the plugin's behalf

Just need him to pull the updated plugin properties file to this request.

TIA!

Alexsuperfly:
add your name in the authors field ([example](https://github.com/runelite/plugin-hub/blob/master/plugins/117hd#L3)) then since the original author has given the go ahead for additional ones in the future.

PR #7127, opened 2024-12-09
Discord collection logger

PJGJ210:
Set to not message by default on seasonal worlds


iProdigy (approved):
btw you can use `Set#contains` to convert the world checks from $O(n)$ to $O(1)$

PR #7133, opened 2024-12-10
Update tzhaar-hp-tracker

MoreBuchus:
Update to 1.0.3
Adds math for equilibrium - Thanks to Umbra


iProdigy (approved):
note: relic status and total level could be obtained dynamically but your approach is fine for a hub plugin

tylerwgrass (approved):

PR #7134, opened 2024-12-11
Update world-heatmap

GrandTheftWalrus:
- Enabled automatic uploading of anonymous user heatmap data to osrsworldheatmap.com
- Changed image output filenames for better sorting
- Changed PLACES_SPOKEN_AT to include messages by players other than the local player

### DIFF
--- a/plugins/world-heatmap
+++ b/plugins/world-heatmap
@@ -1,4 +1,4 @@
 repository=https://github.com/GrandTheftWalrus/RuneLite-World-Heatmap.git
 commit=dff24021d0e30b89ef839fc2f2c31a0fd72d3405
-warning=By default (but with the option to opt-out), this plugin uploads anonymized user data to the global heatmap at osrsworldheatmap.com. User data is only displayed in aggregate with all global data.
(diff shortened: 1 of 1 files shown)


cdfisher:
Your warning is a good start but should include something along the lines of `This plugin submits your IP address to a 3rd party website not controlled or verified by the RuneLite Developers.` as well.

GrandTheftWalrus:
> Your warning is a good start but should include something along the lines of `This plugin submits your IP address to a 3rd party website not controlled or verified by the RuneLite Developers.` as well.

Tanks for the feedback. I added it.

LlemonDuck:
We cannot show warnings to users after they have already installed the plugin. Since you have active users, you'll need to make this off by default and instead add a warning to the now opt-in config option.

GrandTheftWalrus:
@LlemonDuck 
I've changed the data upload to be turned off by default with the warning on the checkbox:
image

LlemonDuck:
don't do your network i/o in onGameTick, and switch it to use the injectable OkHttp client.

GrandTheftWalrus:
> don't do your network i/o in onGameTick, and switch it to use the injectable OkHttp client.

Oopsies. I fixed it now

PR #7137, opened 2024-12-11
Add Hide League Icons plugin

ThePharros:
Hub plugin to force hide the league icons in Raging Echoes League so that the currently unsaved vanilla setting no longer needs to be re-set upon each login.

### DIFF
--- a/plugins/hide-league-icons
+++ b/plugins/hide-league-icons
@@ -1,0 +1,3 @@
+repository=https://github.com/ThePharros/hide-league-icons.git
+commit=374944fc7dca9e0c161631178435e31c35c94937
+authors=ThePharros


raiyni:
This looks fine, does it still hide the icons if you get a pm from split pm mode?

ThePharros:
> This looks fine, does it still hide the icons if you get a pm from split pm mode?

It does. The blue outline is from the Widget Inspector showing the borders of the target widget 651.3.
image

ThePharros:
Turns out the vanilla bug that prevents the in-game setting of hiding these is a result from using the Settings' search function. If the setting is manually pathed to and toggled, then the visibility boolean seems to persist and this plugin is thus no longer needed. May be best to just close this PR and inform people of the in-game solution instead.

PR #7141, opened 2024-12-11
Create RuneLingual

YS-jack:
### Translation Plugin: RuneLingual

This is a translation plugin, which hopefully puts an end to the ever-increasing number of translation plugins.
It translates everything, which surprisingly no other plugins have managed to do.

- **Universal Translation**: Other translation plugins only translate one or a few elements, such as dialogues, chats, or menu entries, and most of them only support one or two languages. RuneLingual translates everything, and can easily expand supported language via DeepL.
- **Non-Latin Characters**: Can type and chat in non-Latin characters, which has never been done before.
- **Multiple Translation Methods**: Translates via manual transcripts or DeepL.

Currently, RuneLingual only supports Japanese, but it should be easy to implement other languages, especially those using Latin characters.

For more detailed information, please refer to the [README](https://github.com/YS-jack/RuneLingual-Plugin/blob/master/README.md) and the [README for developers](https://github.com/YS-jack/RuneLingual-Plugin/blob/master/README_deveploers.md).

There are some improvements (like adding other languages) that could be done, but the code is large enough already, I thought I should do it after initial launch.


YvesW:
Do we really need a 6th translation plugin?
> It translates everything

> Currently, RuneLingual only supports Japanese

Right now this just looks like https://xkcd.com/927/

YS-jack:
What if I came back with this plugin translating more languages? Adding languages which are supported by deepl and use latin characters would be simple to add. Languages using non latin characters will take more time but is doable. I just wanted the review to be easier

raiyni:
polywoof already supports deepl integration

YS-jack:
Polywoof only translates dialogues, mine can translate other elements too

YS-jack:
I could ask if polywoof's developer if he's interested in upgrading his plugin, but ive talked to him before and he didnt seem interested in doing it by himself

Felanbird:
As this is a fork, issues are not enabled by default, so users will currently be unable to open issue reports
380997175-658b6239-0107-425c-8ff2-44bec8c19e77

YS-jack:
> Ok, that's fine. Thanks for checking though! I'll review this one when I get a chance.
Thanks to all RuneLite developers for their time!



> As this is a fork, issues are not enabled by default, so users will currently be unable to open issue reports
OK I think I enabled it just now

LlemonDuck:
Just FYI this is probably going to be many many rounds of review as I chip away at it, but here's a few things to start:

* This isn't needed https://github.com/YS-jack/RuneLingual-Plugin/blob/7cc428e8d37e0ad78370e39366b9e8bd29b027da/build.gradle#L18-L24
* ReadmeImages shouldn't be in your plugin resources. It'll get packaged into the final JAR
* You should move all the H2 database connection stuff onto its own executor

YS-jack:
> Just FYI this is probably going to be many many rounds of review as I chip away at it

understood.

YS-jack:
just realized that my last commit broke something, trying to fix

YS-jack:
>just realized that my last commit broke something, trying to fix

the issue has has been fixed

LlemonDuck:
Which dependency of yours requires the apache maven2 repo? If it was the commented-out google-cloud-translate, can you remove it?

LlemonDuck:
* Don't log spammy things like this at a level higher than debug (or tbh, at this frequency use trace) https://github.com/YS-jack/RuneLingual-Plugin/blob/4ca8f611a56bd080d028578178c4d45d54f097bc/src/main/java/com/RuneLingual/RuneLingualPlugin.java#L206
* onConfigChanged is fired for all config changes across the entire client, make sure the first thing you do is check whether the event is actually your config group or not https://github.com/YS-jack/RuneLingual-Plugin/blob/4ca8f611a56bd080d028578178c4d45d54f097bc/src/main/java/com/RuneLingual/RuneLingualPlugin.java#L268
* Looks like your shutDown might be missing `h2Manager.closeConn()`?
* Two things on MenuCapture#handleMenuEvent:
  * Don't use setDaemon(false) since that will prevent the JVM from exiting if your thread hangs for whatever reason
  * You're very likely to hold the MenuEntry reference longer than it is valid if you're waiting on a translation. You'll need some other mechanism for processing translations in the background, then propagating them back to matching entries after it completes. Make sure you do that last part on the client thread

YS-jack:
this took longer than i would've liked, but it should have fixed the issues you have mentioned

LlemonDuck:
Instead of performing HTTP requests manually with HttpURLConnection and such, you should use an injected OkHttpClient. Make sure the requests are done asynchronously via enqueue, and also make sure you set a proper user agent.

YS-jack:
I have now replaced it with okhttp

LlemonDuck:
you're mis-logging exceptions in a few different ways across the codebase. Don't use `printStackTrace` and don't use `log.something(String.valueOf(e))`, just log the exception as a vararg `log.error(e)` or `log.error("some message", e)`

LlemonDuck:
don't sleep the okhttp workers, use a scheduled executor service and provide a delay to a callback through that instead

LlemonDuck:
Thanks for your patience, here's the last thing:

Add the following to your plugin-hub file:
```
warning = This plugin submits your chat messages, ... to a third party service (DeepL) for translation, which is not maintained or verified by RuneLite.
```

---

Non-blocking:

I don't think you can rely on this to render well cross-platform https://github.com/YS-jack/RuneLingual-Plugin/blob/c819ad77496601e35781f241d4d94e38e862ee09/src/main/java/com/RuneLingual/RuneLingualConfig.java#L66

You should look into using PreparedStatement to pass parameters in your queries, instead of string concatenation. You're not really at risk of injection per se but queries may fail if they contain control characters you aren't expecting. https://github.com/YS-jack/RuneLingual-Plugin/blob/c819ad77496601e35781f241d4d94e38e862ee09/src/main/java/com/RuneLingual/SidePanelComponents/SearchSection.java#L182

You can use LinkBrowser instead of Desktop#browse for better cross-platform support https://github.com/YS-jack/RuneLingual-Plugin/blob/c819ad77496601e35781f241d4d94e38e862ee09/src/main/java/com/RuneLingual/SidePanelComponents/SidePanel.java#L97

LlemonDuck:
Ideally, just update the manifest now with that warning and then I'll merge. Then you can approach the non-blocking items in another PR where my review workload is much smaller

YS-jack:
> Add the following to your plugin-hub file: ...

I have added this now


> Ideally, just update the manifest now with that warning and then I'll merge. Then you can approach the non-blocking items in another PR where my review workload is much smaller

I will do just that, ill make a pr when im done.
I will make another pr for translating UI later

thank you  for reviewing!

YS-jack:
i editied the warning message to be more descriptive

PR #7142, opened 2024-12-11
Create rogue-husher

LeeT96:
Simple plugin that removes "Someone's stealing from us, get them!" when stealing from the chest at the Rogue's castle.


raiyni:
Can this be accomplished with chat filter

LlemonDuck:
You should remove your config class (and the one in test source) altogether

YvesW:
You are adding 2 plugins here. Please rebase and only add 1.

Also, just make this 1 plugin instead of 2, and like I suggested before: wouldn't it be nice to e.g. PR this to other plugins like chop chop stop?

LlemonDuck:

PR #7144, opened 2024-12-12
Create sound-alerts-expanded

lewislarsen:
This plugin expands upon the sound alerts offered by plugins such as C-Engineer's Announcer plugin.

It offers sound alerts when you're missing a farming/max cape at herb patches currently but I'd like to add more in the future when I think of some more that'd be useful. :)

https://github.com/lewislarsen/sound-alerts-expanded

### DIFF
--- a/plugins/sound-alerts-expanded
+++ b/plugins/sound-alerts-expanded
@@ -1,0 +1,2 @@
+repository=https://github.com/lewislarsen/sound-alerts-expanded
+commit=6845394b216bd14dcd127057869fcec7645048c3
--- a/plugins/sound-alerts-expanded
+++ b/plugins/sound-alerts-expanded
@@ -1,2 +1,2 @@
 repository=https://github.com/lewislarsen/sound-alerts-expanded.git
-commit=6845394b216bd14dcd127057869fcec7645048c3
+commit=8082263d14c042623e2a522f5acb2a8a2b2f4895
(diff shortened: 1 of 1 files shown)


YvesW:
Wouldn't it be nice to PR this to the C engineer plugin, as it already does the infernal cape parchment?

lewislarsen:
> Wouldn't it be nice to PR this to the C engineer plugin, as it already does the infernal cape parchment?

Great question!

I had contacted the author of the C-Engineer plugin prior to development but both the creator and C-Engineer are strict as to what gets implemented into C-Engineer Announcements. This plugin wouldn't have those restrictions and as an added benefit this plugin would also allow him to refer people to for potential future features.

You can see an example of rejected features by checking the C-Engineer's plugin issues. (https://github.com/m0bilebtw/c-engineer-completed/issues?q=is%3Aissue+is%3Aclosed)

m0bilebtw:
Hi all, author of C Engineer: Completed here - we are looking to keep the C Engi plugin relatively small in terms of what it announces rather than making it super general purpose which would somewhat explode the scope of the plugin and result in majority of announcements not being C-centric. The QoL such as infernal parchment announcement that may already seem not very C-centric were explicitly requested by C. I can confirm I briefly discussed this with the PR author over discord - happy to discuss further if anyone feels it would be beneficial

PR #7145, opened 2024-12-12
Update Rooftop Agility Improved to v0.6

TicTac7x:


iProdigy (approved):
fyi standard java convention is to not use Optional as fields, but this doesn't matter for hub plugins: https://stackoverflow.com/a/26328555

tylerwgrass (approved):
You don't need to create an entry in config file to read/write config values (regarding your version ConfigItem)

PR #7155, opened 2024-12-14
Update tictac7x-rooftops to v0.6.1

TicTac7x:


pajlada (approved):

iProdigy:
Since `pluginVersion` was changed but `pluginMessage` wasn't, aren't users going to see a message they've seen before?

TicTac7x:
> Since `pluginVersion` was changed but `pluginMessage` wasn't, aren't users going to see a message they've seen before?
> 
> If intentional, lmk & i'll merge

The message would be same yes. Wanted to keep the same one to make sure players who haven't updsted yet can see that  both the colossal wyrm and ape atoll are added, otherwise the message would be only about ape atoll.

iProdigy (approved):

PR #7163, opened 2024-12-18
Create combat-level-indicators

Tyler5934:
Adds a plugin similar to Player Indicators, which highlights players in the wilderness that can attack the local player.


raiyni:
This is disallowed per scouting plugins

PR #7167, opened 2024-12-20
Update soulreaper-axe-qol

seacelery:
Small fix to make the overlay use runescapeSmallFont rather than Runelite's overlay setting.


tylerwgrass (approved):

iProdigy (approved):
consider caching this work that's done every frame https://github.com/seacelery/soulreaper-axe-qol/blob/42d882f3f9870b08b0fda65b6e231997878d8704/src/main/java/com/soulreaperaxeqol/SoulreaperAxeQoLOverlay.java#L175-L180

PR #7173, opened 2024-12-22
DropTracker 3.1

OSRSKoeppy:
Re-wrote the Time tracking functionality to be able to support bosses that send a time message prior to the kill count message.

Also added a plugin version to be sent across with the embeds to help recognize what verison users are on.

Also added in support for Team sizes


LlemonDuck:
> Use of java.lang.System.out is not allowed: system i/o

OSRSKoeppy:
Updated it to remove the print statements

PR #7178, opened 2024-12-26
DropTracker 3.2

OSRSKoeppy:
Just added in the ability to hide PMs. It may be a while before my next push so wanted to get this change out there


OSRSKoeppy:
Just going to redo this

PR #7179, opened 2024-12-26
DropTracker Version 3.2

OSRSKoeppy:
Just added in the ability to send screenshots with hiding PMs


iProdigy:
The new commit hash is invalid (does not exist for that repo)

OSRSKoeppy:
@iProdigy Apologizes on that got it fixed

PR #7180, opened 2024-12-26
Drop party chest version 1.0

pgroenbaek:
Very simple plugin that adds total chest value to the title of the drop party chest interfaces in the party room and clan hall. 

This is quite similar to what the bank plugin does to the bank interface title.

### DIFF
--- a/plugins/drop-party-chest
+++ b/plugins/drop-party-chest
@@ -1,0 +1,3 @@
+repository=https://github.com/pgroenbaek/drop-party-chest.git
+commit=0f05e0998dd504241ce11e69ba33e784cf129325
+authors=pgroenbaek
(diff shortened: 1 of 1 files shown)


YvesW:
Seems like this is something we might want to add to core?

pgroenbaek:
> Seems like this is something we might want to add to core?

Definitely could be added there, but that's completely up to you guys to decide.

I can make a PR with this in the main runelite repo if you want it there instead. :)

YvesW:
I'd make a core PR, or at least ask in #dev if the maintainers think it belongs in core.

I'd personally also remove the log.info calls (or makke them log.debug) but that's my 0.02

pgroenbaek:
> I'd make a core PR, or at least ask in #dev if the maintainers think it belongs in core.
> 
> I'd personally also remove the log.info calls (or makke them log.debug) but that's my 0.02

Okay, I'll ask there.

And yeah, the log.info calls are from the plugin generator template. I'll be removing those.

pgroenbaek:
I have created a [PR with this as a core plugin in the runelite repo](https://github.com/runelite/runelite/pull/18652).

pgroenbaek:
Adam was of the opinion that this would probably fit better as a plugin hub plugin. 

I've updated this PR to the latest commit with changes suggested by Yves if you think this kind of functionality is relevant to have as a plugin hub plugin. This is completely up to you of course. :)

PR #7182, opened 2024-12-27
Create music-track-display

Tommynaut:
A simple plugin that displays the current music track as a game message in the console. 
Fair warning that this was created by someone relatively novice.

### DIFF
--- a/plugins/music-track-display
+++ b/plugins/music-track-display
@@ -1,0 +1,2 @@
+repository=https://github.com/Tommynaut/RL_music.git
+commit=c85002b351203e28701907882b7393ab6979b5b3
--- a/plugins/music-track-display
+++ b/plugins/music-track-display
@@ -1,2 +1,2 @@
 repository=https://github.com/Tommynaut/RL_music.git
-commit=c85002b351203e28701907882b7393ab6979b5b3
+commit=2e9140ece0745123d0baf7a116f50b08431d1f52
--- a/plugins/music-track-display
+++ b/plugins/music-track-display
@@ -1,2 +1,2 @@
 repository=https://github.com/Tommynaut/RL_music.git
-commit=2e9140ece0745123d0baf7a116f50b08431d1f52
+commit=0477e72eb44be0dd43155901dec2ada8fdce7293


cdfisher:
This isn't at all a full review but doing this every game tick is VERY quickly going to spam the user's chatbox with nothing but this message

Tommynaut:
Hi Chris,

Thanks for your message.
Apologies if I'm replying the wrong way, I couldn't see your message on
GitHub and can only see it in my emails. It's probably obvious that I'm new
to this.

With regard to what you said, I have tested this and it doesn't seem to
spam the chatbox - it only sends the message when the music track changes.
I have it running every tick because there didn't seem to be a better way
to check for the track change as an event (there's no associated variable
client-side).

Did you test it and see it spam the chatbox?

Kind regards,
Thomas

raiyni:
I would assume there is a varbit changed or client script that's called.

LlemonDuck:
The checks you're performing are simple enough and I wouldn't worry about trying to optimize it prematurely. But you _should_ remove your config since it is unused.

Tommynaut:
> The checks you're performing are simple enough and I wouldn't worry about trying to optimize it prematurely. But you _should_ remove your config since it is unused.

Thank you for your suggestion. 
Is it okay if I leave the config file in there and comment out everything related to it, including its contents? I was thinking of using the config later when I have more time to add some other features and my intention was to keep it as a template for when I do make changes.

LlemonDuck:
Sure, but in that case you should also comment out its `@Provides` block in the plugin class.

I say this only because having it present means there will be a config gear icon beside the plugin name, which can be confusing for users if it opens to a blank page. If you want to add the config later it won't cause any issues to re-add the provider.

Tommynaut:
> Sure, but in that case you should also comment out its `@Provides` block in the plugin class.
> 
> I say this only because having it present means there will be a config gear icon beside the plugin name, which can be confusing for users if it opens to a blank page. If you want to add the config later it won't cause any issues to re-add the provider.

Thank you LlemonDuck, that makes sense. I have updated the file.

PR #7183, opened 2024-12-28
Create wiki-lookup-plugin

Jacobfinn123:


tylerwgrass:
This appears to just be a clone of the example plugin

Felanbird:
Also, this is already a feature of both the core game, and our original default `Wiki` plugin.
image

Jacobfinn123:
> Also, this is already a feature of both the core game, and our original default `Wiki` plugin. image

Oh cool, wasn't aware of this feature.

PR #7184, opened 2024-12-28
Update flipping-copilot to v1.6.0

cbrewitt:
Major refactor and improve transaction inference
- Refactor all depedencies to use constructor based injection where possible
- Restructure code to more cleanly follow MVC pattern:
  - Controller/Handler classes should have minimal state
  - State should be stored and managed in '<model name>Manager' classes
  - Manager classes should not reference Controllers or Views (JPanels)
  - Controllers and Views (JPanels) may reference each other
- Disentangle state from Controller/View classes
- Update the handling of offer events for inferring transactions based on GrandExchangePlugin code
- Construct the AccountStatus on the fly directly from client.getGrandExchangeOffers() to avoid sync issues
- Track the uncollected quantities explciitly in an independent Manager
- Use accountHash instead of displayName where possible
- Include a machine ID + additional flags on transactions to help track/clean up bad transactions
- Simplify the osrs login management
- Add suggestion settings tab in suggestions panel
- Move sell only toggle to suggestion settings
- Add blacklist UI feature

### DIFF
--- a/plugins/flipping-copilot
+++ b/plugins/flipping-copilot
@@ -1,3 +1,3 @@
 repository=https://github.com/cbrewitt/flipping-copilot.git
-commit=f53a52eaa962cd976a931b77fd6cef52abe49954
+commit=b0bf18e3227b43e58fc539bf244d3fbe755dc946
 warning=This plugin submits your grand exchange offers, grand exchange transactions, and IP address to a 3rd party server not controlled or verified by the RuneLite Developers.
--- a/plugins/flipping-copilot
+++ b/plugins/flipping-copilot
@@ -1,3 +1,3 @@
 repository=https://github.com/cbrewitt/flipping-copilot.git
-commit=74b567cf386eeb28709c814afa5026ad76225f3f
+commit=f53a52eaa962cd976a931b77fd6cef52abe49954
 warning=This plugin submits your grand exchange offers, grand exchange transactions, and IP address to a 3rd party server not controlled or verified by the RuneLite Developers.
--- a/plugins/flipping-copilot
+++ b/plugins/flipping-copilot
@@ -1,3 +1,3 @@
 repository=https://github.com/cbrewitt/flipping-copilot.git
-commit=5b68ad752278acdb6e2aef4d3f0aa65119cf7c6a
+commit=6ad8e3cf036788182b5e0101d48936dd189e6561
 warning=This plugin submits your grand exchange offers, grand exchange transactions, and IP address to a 3rd party server not controlled or verified by the RuneLite Developers.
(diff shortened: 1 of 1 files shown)


LlemonDuck:
You've done a good job of wrapping things in ClientThread#invokeLater where needed, but you need to be more cognizant of getting back _off_ the client thread as soon as possible. You're doing a whole bunch of I/O and Swing things on the client thread in a lot of places. Use SwingUtilities#invokeLater and a parallel executor instead.

Tip btw: you can use `@RequiredArgsConstructor(onConstructor_ = @Inject)` if you're just injecting a bunch of `private final` fields and don't need to do any other actions in the ctor. Then if you extend it you don't have to write it in 3 places.

cbrewitt:
Thank you for the feedback, we've made the following changes to address it:
- Ensure all ui refresh methods execute in swing thread
- Switch disk writes in most Manager classes to async
- Use @RequiredArgsConstructor where possible

We've also added a few other changes with the latest commits:
- add f2p only toggle
- fix "collect items" bug

LlemonDuck:
what do you need this for? https://github.com/cbrewitt/flipping-copilot/compare/5b68ad752278acdb6e2aef4d3f0aa65119cf7c6a...cbrewitt:74b567cf386eeb28709c814afa5026ad76225f3f#diff-143066f71dc6a3c22ce21eaf87df1118bdbed9c53de9e4d22568a65829f77952R223-R263

cbrewitt:
> what do you need this for? [cbrewitt/flipping-copilot@5b68ad7...cbrewitt:74b567cf386eeb28709c814afa5026ad76225f3f#diff-143066f71dc6a3c22ce21eaf87df1118bdbed9c53de9e4d22568a65829f77952R223-R263](https://github.com/cbrewitt/flipping-copilot/compare/5b68ad752278acdb6e2aef4d3f0aa65119cf7c6a...cbrewitt:74b567cf386eeb28709c814afa5026ad76225f3f#diff-143066f71dc6a3c22ce21eaf87df1118bdbed9c53de9e4d22568a65829f77952R223-R263)

This is actually borrowed from in the GrandExchangePlugin code https://github.com/runelite/runelite/blob/0c0c41d5b7d6b1314d2257ffb9c10cf82b2ec7b9/runelite-client/src/main/java/net/runelite/client/plugins/grandexchange/GrandExchangePlugin.java#L938-L981. We thought that it could be useful in certain cases for ensuring transaction consistency and debugging user reported issues related to transactions. 

Right now our backend isn't actually doing anything with it. If there are any concerns relating to user privacy/security we can remove it.

LlemonDuck:
If you aren't using it now I would prefer if you could remove it, yes.

cbrewitt:
MachineID has been removed now, as requested

cbrewitt:
We've made a some extra changes with an additional commit:
- Add flipping.gg to the list of price tracker websites
- bugfix: GrandExchangeOfferEventHandler prev offer event NPE issues
- Support configuring chat text color
- Cycle the highlight overlays through different colors 
- Don't get new suggestion on item inventory changed if GE is not open
- Add additional protection against collect items bug
- Add a new field `requestedSuggestionTypes` and set it when not at the GE

PR #7185, opened 2024-12-29

adhansen:
## Summary
New version of the Wilderness Player Alarm with improvements for plugin inter-operability

## Detail
There are two main issues with the plugin that I'm trying to fix here:
* https://github.com/adhansen/plugin-repo/issues/46 \
I was only able to reproduce this using a user-provided profile which pulled in a truly obscene amount of external plugins. I don't know how people play the game like this, and I don't have insight into which plugin could have caused this. But it turned out to be a simple [one-line change](https://github.com/adhansen/plugin-repo/commit/adaa209820d9b3f552e0cf090f01d8196d81a490) to fix it.
* https://github.com/adhansen/plugin-repo/issues/42 \
This one also weirdly only repro'd with a large profile. It involves GUI elements from other plugins on the same layer getting pushed around by the alarm. My solution to mitigate this is to expose the layer that the alarm is on to the user and allow control of that. Sort of like the [ping grapher plugin](https://github.com/yuh25/Ping-Grapher/blob/master/src/main/java/com/pinggraph/PingGraphOverlay.java#L49) allows for, but preferring an enum over a bool.

### DIFF
--- a/plugins/wilderness-player-alarm
+++ b/plugins/wilderness-player-alarm
@@ -1,2 +1,2 @@
 repository=https://github.com/adhansen/plugin-repo.git
-commit=18a3ebd819f261553f00cdb421d4a7864c3a2f5a
+commit=48df31d70c34ab46c2608636b1924118c27e264b
--- a/plugins/wilderness-player-alarm
+++ b/plugins/wilderness-player-alarm
@@ -1,2 +1,2 @@
 repository=https://github.com/adhansen/plugin-repo.git
-commit=438d44140068b64520cbdc22400b4e8a3e5b9d38
+commit=18a3ebd819f261553f00cdb421d4a7864c3a2f5a


iProdigy:
Are you intentionally resetting `flashControl` to NORMAL for plugin users who previously configured this value to something else? if not, you can revert the `flashSpeed` config key name to `flashControl`

adhansen:
> Are you intentionally resetting `flashControl` to NORMAL for plugin users who previously configured this value to something else? if not, you can revert the `flashSpeed` config key name to `flashControl`

I'm aware that it will reset this config item. I guess I could revert the `keyName`, but I definitely would like to change the `name` as it appears in the user-facing settings. Alright hold off on merge I'll change that back...

iProdigy (approved):

tylerwgrass (approved):

PR #7186, opened 2024-12-29
Added prayer flick helper info

BunnyBytes:
A plugin which helps the user with prayer flick timing by visualising their inputs against the game ticks on a graph.

### DIFF
--- a/plugins/prayer-flick-helper
+++ b/plugins/prayer-flick-helper
@@ -1,0 +1,2 @@
+repository=https://github.com/BunnyBytes/Prayer-Flick-Helper.git
+commit=238d8e13266e744f43d3e6abba33af7c95fe72e6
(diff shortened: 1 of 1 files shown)


iProdigy:
How does this plugin compare with https://github.com/JarateKing/histogram? We want to avoid having multiple plugins with duplicate functionality - if you've identified an improvement over the existing plugin, could it be PRed to their repo?

BunnyBytes:
> How does this plugin compare with https://github.com/JarateKing/histogram? We want to avoid having multiple plugins with duplicate functionality - if you've identified an improvement over the existing plugin, could it be PRed to their repo?

Oh no! I wasn't actually aware of that plugin. I suppose I can have a look and see if there's anything I can contribute!

PR #7189, opened 2024-12-30
Create dialog-font-changer plugin

andrewmenich:
This plugin replaces the standard dialog box quill font with a list of other font options. The goal of this plugin is to make the dialog box more legible and customizable, particularly for those with vision impairments. It works by hiding the original widget and creates a new overlay in its place. It currently only handles NPC and optionally Player dialog.

### DIFF
--- a/plugins/dialog-font-changer
+++ b/plugins/dialog-font-changer
@@ -1,0 +1,2 @@
+repository=https://github.com/andrewmenich/dialog-font-changer.git
+commit=ed4f5ef3fdc15c33bec9c4bbacd420d7b75cd2ba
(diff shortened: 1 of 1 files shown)


andrewmenich:
> Please don't open new prs. You can push changes to your current branch and update the commit hash here.

Sorry about that! Was just rereading the contribution docs and realized my mistake.

raiyni:
Is there a reason you didn't just change the FontID? https://static.runelite.net/api/runelite-api/net/runelite/api/FontID.html

andrewmenich:
I explored that initially and it's definitely the path of least resistance. `PLAIN_12`, `VERDANA_13`, and their bold counterparts work well, I just didn't find them to be as legible as I was hoping for (namely for quest dialog where there is lots of reading). I started playing around with rendering my own overlay and found that using more common fonts like Roboto/Lato were exactly what I was looking for. If this approach is a blocker I'd definitely settle for the former.

LlemonDuck:
You're not providing proper attribution for the font files as are required by their licenses. Regardless, it seems a lot simpler (and I would _much_ prefer it to avoid licensing requirements) to not package any built-in fonts and just let users specify a font that exists on the system.

LlemonDuck:

PR #7193, opened 2024-12-31
New plugin 'gauntlet-chest-value'

ghost:
This plugin displays the total value of the loot you receive from the gauntlet chest. Tested and verified in-game.

### DIFF
--- a/plugins/gauntlet-chest-value
+++ b/plugins/gauntlet-chest-value
@@ -1,3 +1,3 @@
 repository=https://github.com/RealityWinner/gauntlet-chest-value.git
-commit=c9411147b1b0353fbe1e15b0acbfa45801541dea
+commit=61fe7e9931b086694a9794bc0b86af1eb2a3883c
 authors=RealityWinner


YvesW:
Isn't this something that might be worthwhile to PR to core? I'd at least discuss it in #development

YvesW:
Yeah, that one should also just be added to core IMO instead of adding a hub plugin for every chest. Alternatively if Adam wouldn't want that, we could make 1 chest value hub plugin.

tylerwgrass:
Seems there isn't (currently) opposition to adding this to the core loot tracker, so I'll move this to waiting for now.
[This](https://github.com/runelite/runelite/blob/f654e1ccbaa6cff23e50c32fde876aea9e15a00f/runelite-client/src/main/java/net/runelite/client/plugins/loottracker/LootTrackerPlugin.java#L972) is a likely reference

If something changes this PR can be given another look.

tylerwgrass:
You are welcome to open a PR to core yourself

PR #7200, opened 2025-01-02
Create animation-indicator

Need-Femboy:
Highlights an NPC for the tick its animation changes, useful for learning some fights


LlemonDuck:
This violates the following rule quite aggressively:

PR #7206, opened 2025-01-04
Update Last Man Standing v1.7.2

pwatts6060:
- Minor changes to autodetect new variant of terminator bots


LlemonDuck:
Not new to this PR but can you make sure this gets removed on shutDown https://github.com/pwatts6060/runelite-plugins/blob/lms/src/main/java/com/lmsnotifier/LMSPlugin.java#L181

pwatts6060:
> Not new to this PR but can you make sure this gets removed on shutDown https://github.com/pwatts6060/runelite-plugins/blob/lms/src/main/java/com/lmsnotifier/LMSPlugin.java#L181

It is a bit confusing but it is removed already in a subroutine called in shutdown. ItemUpgrade.reset()

LlemonDuck:
Thanks! That's easier to follow

PR #7207, opened 2025-01-05
OSRS-Splits-TheKodai

KeyboardIsMagic:
OSRS Splits – The Kodai
A custom RuneLite plugin for managing parties, verifying members, and automatically capturing Nex loot screenshots—improving safety and trust in OSRS Splits and RuneWatch communities. Includes straightforward creation/joining of parties, split confirmations, and quick screenshot uploads to Discord. Name changes won’t hide a bad actor—verified members are tracked permanently. Enjoy Nex with peace of mind!


tylerwgrass:
Is there a reason you cannot use already included versions of the dependencies you have changed?

raiyni:
This should be added to the nex plugin you already have

KeyboardIsMagic:
> Is there a reason you cannot use already included versions of the dependencies you have changed, and how is this different from your already existing [Nex Split Tracker](https://github.com/KeyboardIsMagic/NexSplitTracker) plugin?



> This should be added to the nex plugin you already have

The original Nex Split Tracker is used as a in client spread sheet to track drops from random teams and display profit from splits.
The proposed plugin unites the community of OSRS Splits The Kodai and RuneWatch to help prevent and instill trust into the community members. Community members voluntarily sign up through discord so their accounts may be tracked, these accounts are tracked regardless of x name changes so scamming is highly discouraged as members who do not split valuable drops with their party members will be tracked via RuneWatch. This plugin does not track the splits in a spreadsheet format, rather posts the confirmations of party members agreeing to the TOS of the community of RuneWatch and OSRS Splits, and any and all unique drops from the Nex encounter itself. All parties have clear and automated evidence to help maintain transparency when playing with strangers and help with any dispute among RuneWatch cases. 

This plugin aims to help the community while discouraging scamming by unavoidable long term consequences via RuneWatch.

YvesW:
I still don't see a reason this can't be added to the nex split plugin that you already own with the same name. Just add a config section for it and it works.

Also, inject our okhttp and use that instead of what you are doing now for http calls.

Still got the dependency changes label because you are changing stuff.

I've not looked if what you are doing is ok or just weird. Usually proper authentication of an osrs account doesn't really work unless you do a skilling challenge with hiscore checks or something like that.

KeyboardIsMagic:
Merge with existing Nex Split Plugin:
The Nex Split tracker works much like the Raids Loot Tracker, however both are independent plugins rather than added to the built in Loot Tracker. Likewise, the Loot Tracker is an independent plugin, and not added to the plugin Party Panel or vice versa. These plugins serve different purposes, although they have a common theme of killing Nex, their purposes and functionality are very different. The OSRS Splits – The Kodai plugin takes inspiration from plugins like Dink, Party Panel, and Discord in aims to link a community of Nex killer together. While both relate to Nex, they serve different audiences and use-cases: the existing plugin is for personal loot tracking; this one is for a specialized community that uses Discord-based verification. They could technically be combined, but it might bloat the current Nex Split plugin with a large, community-focused workflow that not all users need. If this is something the Runelite requires, I would rather take down the Nex Splits Tracker(spreadsheet) and push this instead, then at a later date, integrate the spreadsheet into this plugin, which is something I really do not want to do as they are separate and would feel more comfortable if it was removed from the plugin hub entirely.

Dependency:
I’ve updated dependencies to use versions already included in the plugin hub. For HTTP calls, I will update the code to inject the provided okhttp from RuneLite (rather than bundling my own).

Account authentication:
Accounts are registered (not required for functionality) on our discord which is linked to a database that other discord bots access to preform skilling checks via the hiscores. These accounts are tracked indefinitely regardless of name changes since the accounts are linked by their Discord Ids. True, scammers could create multiple discord accounts but I feel this is taking a step in the right direction and something RuneWatch staff members specifically requested to help mitigate scams as users who are registered through Discord are given a rank based on their trustworthiness/amount they have split with other community members. Evidence is taken in the form of screenshots when all party members have confirmed to split via the plugin and sent to Discord, this happens again when any unique drop happens at Nex. This plugin will support a community of 8,000+ people, something that the legacy friends chat or clan system cannot support.

YvesW:
Please stop using an LLM to write your messages if you are. Keep it clear and concise.

Currently multiple reviewers think it's best to integrate the plugins, so I strongly recommend just doing that. You don't have to make arbitrary differences between them. For example the ToA plugin by Llemon also does it all in 1 plugin. No reason to split it out in 20 plugins.

There's still a dependency changes tag on this PR. I think the bot will remove it when there are actual no dependency changes (e.g. the exclude you added).

KeyboardIsMagic:
As mentioned before I did not have any LLMs write the response. I was under the impression since there was a discrepancy, more information regarding the matter would be useful to identify key differences between the two plugins. It is mere coincidence that the community OSRS Splits and Nex Split Tracker have similar names, and boss Nex.

I am currently updating the code for okHttp.

raiyni:
This can be added to the existing plugin as a config option, which would make it not required for every user of the plugin.

KeyboardIsMagic:
> This can be added to the existing plugin as a config option, which would make it not required for every user of the plugin.

Alright, if that is what you want I will add them together. A new PR will be generated with the reflected incoming changes.

KeyboardIsMagic:
> Alright, if that is what you want I will add them together. A new PR will be generated with the reflected incoming changes.

PR #7213, opened 2025-01-07
Fix Bosses without Time

OSRSKoeppy:
Was still trying to send Boss kill messages for bosses without PBs. This is only used for bosses with PB so cleaned this up and removed it accordingly.

Also Fixed Combat Achievements not properly sending a screenshot when achieved. Also added in additional logging when webhooks fail

Incremented version to 3.2.2


OSRSKoeppy:
The previous version was unknowingly broken (Before this pull request) so if someone can look at it quickly it would be greatly appreciated. Right now its causing our users to get rate limited in discord

PR #7214, opened 2025-01-08
Submit tile marker metronome

RenaudBernon:
Resubmit of https://github.com/runelite/plugin-hub/pull/6998

This plugin combines Tile markers with the visual metronome.
Marked tiles change color every X ticks.

In the last MR the comment was made that since I was only keeping track of ticks for loaded tiles, this would effectively sync the metronome with some boss encounters, making mechanics that happen after X ticks trivial.
To combat this I am now keeping a global tick counter. This hopefully alleviates that concern.

### DIFF
--- a/plugins/tile-marker-metronome
+++ b/plugins/tile-marker-metronome
@@ -1,2 +1,2 @@
 repository=https://github.com/RenaudBernon/tile-marker-metronome.git
-commit=aa37949e9a22577b5f8872048240046674ffc178
+commit=5f01834ffce27271d91c5ceaf80edde8da4f6d7b
--- a/plugins/tile-marker-metronome
+++ b/plugins/tile-marker-metronome
@@ -1,2 +1,2 @@
 repository=https://github.com/RenaudBernon/tile-marker-metronome.git
-commit=5f01834ffce27271d91c5ceaf80edde8da4f6d7b
+commit=7d5e294d8bc6679c42df6a79bb781b64964ab9b2
--- a/plugins/tile-marker-metronome
+++ b/plugins/tile-marker-metronome
@@ -1,2 +1,2 @@
 repository=https://github.com/RenaudBernon/tile-marker-metronome.git
-commit=d2e7848fbe0ce5d932e10bc7e8e021a0fe18dfbe
+commit=aa37949e9a22577b5f8872048240046674ffc178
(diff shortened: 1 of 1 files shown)


LlemonDuck:
These aren't blockers but you should want to change them anyway:

This is going to cause parity issues where some groups don't roll over when they should if gcd(100, tickCount) is 1. If you were concerned about the variable becoming too large, it just won't. You'd need like 40 years of gametime to overflow the variable:
```
if (currentTick > 100) {
    currentTick = 1;
}
```

Your `currentTick++` should be in `onGameTick` not `tickGroup`, otherwise you're ticking each group each tick times the number of groups, not by the number of ticks that have happened.

Marking currentTick as transient does nothing, 

https://github.com/RenaudBernon/tile-marker-metronome/blob/aa37949e9a22577b5f8872048240046674ffc178/src/main/java/com/tilemarkermetronome/TileMarkerMetronomePlugin.java#L121-L128

---

To avoid your menu entry from looking identical to the core options, you should change your "Mark" and "Unmark" to something like "Tick-Mark" or literally anything else that isn't "Mark".

https://github.com/RenaudBernon/tile-marker-metronome/blob/aa37949e9a22577b5f8872048240046674ffc178/src/main/java/com/tilemarkermetronome/TileMarkerMetronomePlugin.java#L164-L166

---

This doesn't need to be atomic, there's no multithreading going on:

https://github.com/RenaudBernon/tile-marker-metronome/blob/aa37949e9a22577b5f8872048240046674ffc178/src/main/java/com/tilemarkermetronome/TileMarkerMetronomeGroup.java#L64

RenaudBernon:
Thanks for the feedback, Ive made changes according to the comments.

Except for 
> This doesn't need to be atomic, there's no multithreading going on:
>
> https://github.com/RenaudBernon/tile-marker-metronome/blob/aa37949e9a22577b5f8872048240046674ffc178/src/main/java/com/tilemarkermetronome/TileMarkerMetronomeGroup.java#L64+

I'm using an `AtomicInteger` because variables used in streams need to be effectively final. I wouldn't be able to increment a normal `int`.

LlemonDuck:
Don't do Swing work on the client thread, use SwingUtilities#invokeLater for e.g. https://github.com/RenaudBernon/tile-marker-metronome/blob/5f01834ffce27271d91c5ceaf80edde8da4f6d7b/src/main/java/com/tilemarkermetronome/TileMarkerMetronomePlugin.java#L140

---

Downgrade this and other frequent log.info calls to log.debug instead https://github.com/RenaudBernon/tile-marker-metronome/blob/5f01834ffce27271d91c5ceaf80edde8da4f6d7b/src/main/java/com/tilemarkermetronome/TileMarkerMetronomeGroup.java#L73

---

```
for (int i = 0; i < pointIndex; i++) {
  currentTileColor++;
  currentTileColor %= colors.size();
}
```

PR #7215, opened 2025-01-09
Create click-move-options

DuncanClotfelter:
Adds the ability to disable movement options and menu items to prevent misclicks

Allows the following changes to movement:

 - Toggling left click to move
 - Toggling the function of Walk Here in the right click menu
 - Toggling middle click to move
 - Toggling Shift + Click to move
 - Toggling left click to move on tiles marked by the Ground Markers plugin

Allows the user to declutter menu items via:

 - Deprioritizing ground item/NPC/object options
 - Removing "Walk Here", Cancel, and Examine options

### DIFF
--- a/plugins/click-move-options
+++ b/plugins/click-move-options
@@ -1,2 +1,2 @@
 repository=https://github.com/DuncanClotfelter/click-move-options.git
-commit=2dc9c74b3922b2daacff87aa60d562053509672e
+commit=f8cf6f0c6886fb818d23ba32f73e0141f4f0d08c
--- a/plugins/click-move-options
+++ b/plugins/click-move-options
@@ -1,0 +1,2 @@
+repository=https://github.com/DuncanClotfelter/click-move-options.git
+commit=2dc9c74b3922b2daacff87aa60d562053509672e


DuncanClotfelter:
> Where is the intended use of this plugin?

This plugin allows users to denominate movement clicks from other "selection" clicks and can be used anywhere as a QoL change.
For example, you would no longer need to right click->Walk Here on titles that are obscured by dropped items or attackable NPCs, as you can set movement to Shift+Left Click instead.

Disabling left click movement in general, but enabling it on marked tiles also provides immediate confirmation (in the form of the yellow X) that you have clicked on the correct square. This is a small QoL change for combat, especially for those like me that often misclick and have high latency - where you are unsure if the initial click was on the correct tile, and any additional clicks will be on a moving screen with a strange delay that is likely to disrupt your next action(s).

Disabling menu options should reduce screen clutter and provide less misclick potential as well.

Some of this functionality has been partially implemented across other plugins, (see: Red Clicks Only, Ground Markers) but without as much user customization.

Let me know if I should update the README to reflect this, I have been more verbose here and in the initial pull request.

SaltisRS:
> > Where is the intended use of this plugin?
> 
> This plugin allows users to denominate movement clicks from other "selection" clicks and can be used anywhere as a QoL change. For example, you would no longer need to right click->Walk Here on titles that are obscured by dropped items or attackable NPCs, as you can set movement to Shift+Left Click instead.
> 
> Disabling left click movement in general, but enabling it on marked tiles also provides immediate confirmation (in the form of the yellow X) that you have clicked on the correct square. This is a small QoL change for combat, especially for those like me that often misclick and have high latency - where you are unsure if the initial click was on the correct tile, and any additional clicks will be on a moving screen with a strange delay that is likely to disrupt your next action(s).
> 
> Disabling menu options should reduce screen clutter and provide less misclick potential as well.


Isn't most of this functionality already available via Menu Entry Swapper and their various addon plugins?

DuncanClotfelter:
> > > Where is the intended use of this plugin?
> > 
> > 
> > This plugin allows users to denominate movement clicks from other "selection" clicks and can be used anywhere as a QoL change. For example, you would no longer need to right click->Walk Here on titles that are obscured by dropped items or attackable NPCs, as you can set movement to Shift+Left Click instead.
> > Disabling left click movement in general, but enabling it on marked tiles also provides immediate confirmation (in the form of the yellow X) that you have clicked on the correct square. This is a small QoL change for combat, especially for those like me that often misclick and have high latency - where you are unsure if the initial click was on the correct tile, and any additional clicks will be on a moving screen with a strange delay that is likely to disrupt your next action(s).
> > Disabling menu options should reduce screen clutter and provide less misclick potential as well.
> 
> Isn't most of this functionality already available via Menu Entry Swapper and their various addon plugins?

You are right, Menu Entry Swapper does already support shift-click click-through.

The original idea behind the plugin was to allow users to choose a niche method of global movement (such as Shift-Click or Middle Click) while providing additional methods to marked tiles. This would help avoid misclicks on adjacent tiles, or give immediate feedback when one occurs. I think this could be useful in both PvM situations and other activities where you want to stay on a path.

I would be fine with removing the additional features to avoid redundancy, but they do have a couple of unique points:

The current click-through options have the benefit of allowing click-through by default - so that one could always left-click move without pressing Shift or tagging each individual entity. The global removal of "Walk Here" and "Cancel" seem to be unique as well.

LlemonDuck:
@DuncanClotfelter are you planning to make any changes related to the above or is it ready to review as-is?

LlemonDuck:

PR #7219, opened 2025-01-10
 impling-saver takeover

TheLope:
I would like to takeover the plugin.

I created an issue here that was not responded to in over 7 days. https://github.com/AggieNateHarris/ImplingSaver/issues/14


LlemonDuck:
@AggieNateHarris Last call before we transfer the plugin!

PR #7223, opened 2025-01-10
Create Stopwatch and Timer plugin

pr0f3ss:
This plugin adds a simple stopwatch and countdown timer to the plugin panel, ideal for speedrunning, skilling, and PvM encounters. The countdown timer finds use in specific PvM mechanics, such as ghost skipping at Cerberus or set spawning in the Inferno. It comes with multiple presets and optional sound notifications.

### DIFF
--- a/plugins/stopwatch-timer
+++ b/plugins/stopwatch-timer
@@ -1,0 +1,2 @@
+repository=https://github.com/pr0f3ss/StopWatch.git
+commit=e99aebef26e58e6ccc635066a74f3c9c3946f414
--- a/plugins/stopwatch-timer
+++ b/plugins/stopwatch-timer
@@ -1,2 +1,2 @@
 repository=https://github.com/pr0f3ss/StopWatch.git
-commit=e99aebef26e58e6ccc635066a74f3c9c3946f414
+commit=e63db3d21dba130f3bcb253630f6088d1cb04047
--- a/plugins/stopwatch-timer
+++ b/plugins/stopwatch-timer
@@ -1,2 +1,2 @@
 repository=https://github.com/pr0f3ss/StopWatch.git
-commit=cc38d5c8ee2f5139c138fb6c6c620a39779ad701
+commit=8799b9c45ce137901a1371339d13248178777d40
--- a/plugins/stopwatch-timer
+++ b/plugins/stopwatch-timer
@@ -1,2 +1,2 @@
 repository=https://github.com/pr0f3ss/StopWatch.git
-commit=e63db3d21dba130f3bcb253630f6088d1cb04047
+commit=cc38d5c8ee2f5139c138fb6c6c620a39779ad701


DominickCobb-rs:
This violates Jagex's rules about plugins assisting with PvM.

pr0f3ss:
> This violates Jagex's rules about plugins assisting with PvM.

This plugin certainly does not violate any Jagex rules. It works as a standalone feature in the plugin panel and does not pull data from the core client. The set-timer plugin #3188 offers similar functionality, albeit limited to the inferno. 

Below you can see how the panel plugin is implemented.
image
image

raiyni:
The default timers plugin already has these features?

pr0f3ss:
> The default timers plugin already has these features?

This plugin is designed with speedrunners and PvMers in mind, offering larger UI elements for clearer visibility during high-paced gameplay. Moreover, the larger UI and integration into the plugin panel is very helpful for streamers and content creators.  It also includes predefined presets for specific PvM encounters without requiring manual adjustments. While the default timers plugin may serve general needs, this plugin is tailored for those who require quick and efficient time control. 

A similar argument for the need of separate plugins can be made for set-timers (#3188).

DominickCobb-rs:
> This violates Jagex's rules about plugins assisting with PvM.

I didn't see it was *manually activated* timers. Disregard my comment

LlemonDuck:
Don't construct your panel on the client thread, construct it on the Swing thread (just move the new StopWatchPanel outside of the invokeLater): https://github.com/pr0f3ss/StopWatch/blob/cc38d5c8ee2f5139c138fb6c6c620a39779ad701/src/main/java/com/stopwatch/StopWatchPlugin.java#L41

Same goes for ClientToolbar#removeNavigation, that should be on Swing thread as well https://github.com/pr0f3ss/StopWatch/blob/cc38d5c8ee2f5139c138fb6c6c620a39779ad701/src/main/java/com/stopwatch/StopWatchPlugin.java#L60

You need to make sure this clip is closed to avoid leaking resources (check something like c-engineer plugin for an example? I'm not sure exactly where to look): https://github.com/pr0f3ss/StopWatch/blob/cc38d5c8ee2f5139c138fb6c6c620a39779ad701/src/main/java/com/stopwatch/StopWatchPanel.java#L356-L358

pr0f3ss:
> Don't construct your panel on the client thread, construct it on the Swing thread (just move the new StopWatchPanel outside of the invokeLater): https://github.com/pr0f3ss/StopWatch/blob/cc38d5c8ee2f5139c138fb6c6c620a39779ad701/src/main/java/com/stopwatch/StopWatchPlugin.java#L41
> 
> Same goes for ClientToolbar#removeNavigation, that should be on Swing thread as well https://github.com/pr0f3ss/StopWatch/blob/cc38d5c8ee2f5139c138fb6c6c620a39779ad701/src/main/java/com/stopwatch/StopWatchPlugin.java#L60
> 
> You need to make sure this clip is closed to avoid leaking resources (check something like c-engineer plugin for an example? I'm not sure exactly where to look): https://github.com/pr0f3ss/StopWatch/blob/cc38d5c8ee2f5139c138fb6c6c620a39779ad701/src/main/java/com/stopwatch/StopWatchPanel.java#L356-L358

Thanks for the review. 

I've adjusted all of your raised concerns. 

For the clip player, I looked into the c-engineer plugin to see how they do it. Essentially, I created a sound player class that holds a single clip instance, which is created/destroyed appropriately.

LlemonDuck:
Can you describe where you got the alert.wav sound file and whether it is appropriately licensable?

pr0f3ss:
> Can you describe where you got the alert.wav sound file and whether it is appropriately licensable?

Sure. The audio file can be found [here](https://pixabay.com/sound-effects/level-up-191997/). The license states that the content is free to use without having to attribute the author.

PR #7229, opened 2025-01-11
Add raids death indicator

vikke1234:
Similarly to nylo death indicators and tzhaar hp, this plugin is a death indicator for ToA. The idea is to make it able to reliably confirm whether an enemy dies or not. It highlights akkha as a range (e.g. shadow) hit procs shadows.

It accomplishes this by computing the internal fraction for hitpoints.

Currently this only supports Ampken and Akkha. The enemy data for other rooms is a placeholder for v1.x, this is an initial PoC.

### DIFF
--- a/plugins/raids-death-indicator
+++ b/plugins/raids-death-indicator
@@ -1,2 +1,2 @@
 repository=https://github.com/vikke1234/raids-death-indicator.git
-commit=bc9e3cb81e4f34f7693a36f0cd9f6ec4b0bbbfc2
+commit=1f1282c9d6c9450882e4a1c55cc9ef72f06e63c1
--- a/plugins/raids-death-indicator
+++ b/plugins/raids-death-indicator
@@ -1,5 +1,2 @@
 repository=https://github.com/vikke1234/raids-death-indicator.git
 commit=55732dd9b9883c50fa76cb8c271ee36f0f9b4aa7
-authors=Fisu
-tags=raids
-description=Tries to accurately show when mobs die in raids (currently only ToA)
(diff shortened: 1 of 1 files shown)


vikke1234:
re-running workflows

iProdigy:
not a full review but: on plugin shutdown you should do `overlayManager.remove(overlay);` and `wsClient.unregisterMessage(EntityDamaged.class);`

vikke1234:
alright, I'll fix it tomorrow 👍

vikke1234:
re-running workflows

LlemonDuck:
Pushing a commit will already rerun workflows, you don't need to close/open the PR unless it's very stale

LlemonDuck:
Downgrade to log.debug https://github.com/vikke1234/raids-death-indicator/blob/42e6b82f1ccdeaf8fce7f5bc4b89b3d0e51ec9a9/src/main/java/com/example/utils/PredictionTree.java#L129

You may also be interested in the scaling work we did in the wiki dps calculator since the raids HP values are rounded in an interesting way: https://github.com/weirdgloop/osrs-dps-calc/blob/main/src/lib/scaling/TombsOfAmascut.ts#L36-L40

vikke1234:
Hi,

Sorry for the delay, I'm in the middle of a move.

Is the rounding done for all monsters (baba monkeys etc.)? I added that for akkhas shadow, as I found the same thing in the ToA plugin.

vikke1234:
> Pushing a commit will already rerun workflows, you don't need to close/open the PR unless it's very stale

I did this because I had forgot to push to the plugin repo, so the commit didn't exist and I needed to re-run the workflows.

LlemonDuck:
Path of Apmeken monkeys aren't hp-rounded (they aren't scaled at all anymore). I'm not aware of any other relevant exceptions off the top of my head.

vikke1234:
Oh alright, that's good to know, I'll see what I can do later, I need to figure out how xp scaling works on swarms and jugs as well before it'll matter. Might have to study leagues videos on kephri/zebak.

Does it look fine other than that?

PR #7233, opened 2025-01-12
Create ba-heal-codes

evaan:


tylerwgrass (commented):
- Can you move this out of `render`? One option would be to handle in `onGameTick()` instead https://github.com/evaan/ba-heal-codes/blob/c4f430e2482de6811dda376cb0c32d50d577be76/src/main/java/dev/evaan/HealCodeOverlay.java#L34-L47
- You might want to also put the code parts in something that is already iterable instead of parsing the strings every render

Some other non-blockers, but you may want to include a README to provide some example screenshots/usage/etc, as well as tags in `runelite-plugin.properties` to make it more discoverable

PR #7235, opened 2025-01-14
feat: remaining amethyst plugin

marshallez:
lets get 99 mining

- Shows the number of amethyst ore remaining in a rock while mining it.


iProdigy:
Avoid looping through all tiles please: https://github.com/marshallez/remaining-amethyst/blob/2741e33f4e45db580157f7de549fc57e2b5b2b8a/src/main/java/com/remainingamethyst/RemainingAmethystPlugin.java#L139

e.g., try `LocalPoint local = LocalPoint.fromWorld(client, facingPoint); Tile tile = tiles[facingPoint.getPlane()][local.getSceneX()][local.getSceneY()];`

PR #7239, opened 2025-01-16
tasks-tracker: improve profile & panel performance

tylerthardy:
## Main Changes
Reduced hang on profile switch with:
- Reduced calls to redraw task panels
https://github.com/osrs-reldo/tasks-tracker-plugin/pull/95
- Hid plugin panel during plugin shutdown (JPanel GC was contributing to hang, but not when hidden)
https://github.com/osrs-reldo/tasks-tracker-plugin/pull/96)

## Testing Details
- Noted significant improvements switching between profiles
- Evidence & full testing details on our internal PR: [LINK](https://github.com/osrs-reldo/tasks-tracker-plugin/pull/95)

# 
🧁🧁 THANK YOU Runelite team for working with us during this League, despite the issues. 🧁🧁

### DIFF
--- a/plugins/tasks-tracker
+++ b/plugins/tasks-tracker
@@ -1,3 +1,3 @@
 repository=https://github.com/osrs-reldo/tasks-tracker-plugin.git
-commit=1f8d2b36e44ef80b88345a3b38117424a8983324
+commit=12666e5c46bf320f75534038a15085711e54fb42
 authors=tylerthardy


iProdigy (approved):
thanks for continuing to make performance improvements - I'd suggest creating a `TaskListPanel#refreshMultiple(Collection<TaskFromStruct>)` method to reduce the number of `taskPanels` iterations (and `Stream<T>` creations)

then you could remove this for loop that creates a bunch of swing tasks: https://github.com/osrs-reldo/tasks-tracker-plugin/blob/12666e5c46bf320f75534038a15085711e54fb42/src/main/java/net/reldo/taskstracker/TasksTrackerPlugin.java#L522-L525

tylerthardy:
> thanks for continuing to make performance improvements - I'd suggest creating a `TaskListPanel#refreshMultiple(Collection<TaskFromStruct>)` method to reduce the number of `taskPanels` iterations (and `Stream<T>` creations)
> 
> then you could remove this for loop that creates a bunch of swing tasks: https://github.com/osrs-reldo/tasks-tracker-plugin/blob/12666e5c46bf320f75534038a15085711e54fb42/src/main/java/net/reldo/taskstracker/TasksTrackerPlugin.java#L522-L525

Thanks. A bit of naivety on my side, I assumed putting them on the thread in a loop would run them non-concurrently, but I suppose there's only 1 thread, eh?

Will do!

iProdigy:
requires rebase

PR #7242, opened 2025-01-16
Upgrade IRC to v0.4 (6d90a43)

ryanwohara:
- Improved side panel 
- Join/part/quit/kick messages
- ;;ms for MemoServ
- View the topic
- Previous startup messages are hidden and replaced with the MOTD
- Rejoin channel on kick
- Renamed delimiter to prefix

### DIFF
--- a/plugins/irc
+++ b/plugins/irc
@@ -1,2 +1,2 @@
 repository=https://github.com/ryanwohara/irc-plugin.git
-commit=636f6b940d73176974497b1fb961dad4b55b1e26
+commit=6d90a4386ab0ec8bd14681a7f87981c8ab065b3c


ryanwohara:
Sorry for the big diffs on two files - had to change the line endings from CRLF to LF

LlemonDuck:
Can you please escape the incoming messages with `StringEscapeUtils.escapeHtml4()`?
https://github.com/ryanwohara/irc-plugin/compare/636f6b940d73176974497b1fb961dad4b55b1e26...ryanwohara:5b8fc54c039ff4d581c352f61c1024e867afe113?w=1#diff-b9c77d977dbe541b12889ed0b50b9406cb2597a878881c8970015319bf21c796R82

ryanwohara:
@LlemonDuck Thank you. Just added it here: https://github.com/ryanwohara/irc-plugin/compare/636f6b940d73176974497b1fb961dad4b55b1e26...ryanwohara:8a1c1820dd7cacefd1faf815ae84fed5e42f555b#diff-b9c77d977dbe541b12889ed0b50b9406cb2597a878881c8970015319bf21c796R83

ryanwohara:
Actually that seems to break my app
image

Without that tweak it renders as expected. I'm not familiar enough with JLabel to understand why :grimacing: 

Can I revert this change?

ryanwohara:
Converting this to a draft until I can revert this change. I don't think it's usable as it is now.

ryanwohara:
Moved back out of draft status. 

Just to reiterate: I cannot use that `StringEscapeUtils.escapeHtml4()` method as < will display as &lt; in a literal sense.

LlemonDuck:
That is the goal, yes. Unsanitized IRC messages shouldn't be placed into html tags. If you want to do any other formatting first, you should do it after escaping.

ryanwohara:
@LlemonDuck No, this is not the goal. I want things to be readable.

This is what happens without your suggestion:
image

This is what happens with your suggestion:
image


It's not evaluating the HTML entities. 
Also, this was fine as per the previous reviewers on this plugin.

LlemonDuck:
That behaviour is surprising to me, can you point to where it's being escaped then? This would even conflict with your desired URL replacing since that relies on the JLabel interpreting that anchor tag correctly.

This snippet _does_ perform rich formatting based on `message`'s contents, which is what I would expect to happen based on your code:

```java
String message = "Dragon: <div>foo</foo>";

JLabel label = new JLabel("<html>" + message + "</html>");
add(label);
```

Additionally, using `message = "Dragon: &lt;div&gt;foo&lt;/foo&gt;";` renders correctly with the characters `<div>foo</foo>` shown on the label.

I didn't see any preprocessing of control characters in Message.java either.

---

In terms of other notes:
* Rich-replacing links with anchor elements (if I'm interpreting your intent correctly) is a bad idea as it hides the destination URL. In a browser one can hover the link to see a preview or copy the link without loading it to view it in full. Neither are possible by default in Swing and you provide no alternative means either.
* Since you're doing I/O, you need to add a warning to the plugin hub manifest ([example](https://github.com/runelite/plugin-hub/blob/3b294207cf0682f19683398a69215b00e482859a/plugins/flipper#L3) describing the data you're egressing. Something like `This plugin submits your IP address and applicable IRC messages to a third party server not controlled by the RuneLite developers.`
* Use RuneLite's `LinkBrowser` utility class instead of Desktop to open links.

ryanwohara:
Making me jump through hoops not previously required is rude.

LlemonDuck:
Some of these changes are still going to be required even if you don't want to go forward with the substantive update or else it puts your plugin at risk of being disabled. Of particular importance would be the hub manifest warning.

PR #7246, opened 2025-01-18
Create 1.5 Woodcutting

BenLite0:
My plugin adds an overlay timer to teak and mahogany stumps on fossil island that counts down its respawn time.


LlemonDuck:
This is more of an issue on our end, and we'll look into fixing it, but can you rename your repository to not include a `.`? `BenLite/1-5-Woodcutting` would be great.

BenLite0:
> This is more of an issue on our end, and we'll look into fixing it, but can you rename your repository to not include a `.`? `BenLite/1-5-Woodcutting` would be great.

Apologies i thought it had me do this before but i guess i only just changed the files name and not my repository, hopefully its correct now

tylerwgrass:
Can you please replace your `log.info()` calls with `log.debug()`?

BenLite0:
> Can you please replace your `log.info()` calls with `log.debug()`?

I believe its done, when i did the commit i got some wierd message about merging remote, and in the commits its showing something about the license.  but the changes have been made.  If it doesnt show i can try 1 commit hash older

PR #7247, opened 2025-01-18
Update shortest-path to v1.16.8

Skretzo:
- Fix paths that should go through caves  
image
- Rewrite to create less WorldPoints
- Limit clue compass transports to league worlds

### DIFF
--- a/plugins/shortest-path
+++ b/plugins/shortest-path
@@ -1,2 +1,2 @@
 repository=https://github.com/Skretzo/shortest-path.git
-commit=5714c5a0f9aa102fcb0f280c2f78a58cb2c1d56d
+commit=a4828d35672f7228d10fb664af0dfa50586932ab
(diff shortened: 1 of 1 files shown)


pajlada (approved):
Just some notes, seems good

`onPluginMessage` `path` now accepts an Integer instead of a WorldPoint, is it possible to be backwards-compatible & accept WorldPoint here for a while? Mostly thinking in the context of Quest Helper which uses this and sends WorldPoints

PathTileOverlay::drawTile no longer compares planes, is that intended?

Your `LOCATION_PERMUTATION` now uses plane `1` as its placeholder instead of `-1`, is that intended?

Skretzo:
> `onPluginMessage` `path` now accepts an Integer instead of a WorldPoint, is it possible to be backwards-compatible & accept WorldPoint here for a while? Mostly thinking in the context of Quest Helper which uses this and sends WorldPoints

Good catch, I will change it to also accept WorldPoint.

> PathTileOverlay::drawTile no longer compares planes, is that intended?

Yes, the plane comparison is also performed in `WorldPointUtil.toLocalPoint`.

> Your `LOCATION_PERMUTATION` now uses plane `1` as its placeholder instead of `-1`, is that intended?

Yes, otherwise the integer overflows in `WorldPointUtil.packWorldPoint`.

pajlada (approved):

iProdigy (approved):

PR #7248, opened 2025-01-18
Add remove-chat-options plugin

96jonesa:
Removes all chatbox menu options.

Useful when the chatbox is transparent, allowing the user to interact with the rest
of the game as if the chatbox were not there at all.

### DIFF
--- a/plugins/remove-chat-options
+++ b/plugins/remove-chat-options
@@ -1,2 +1,2 @@
 repository=https://github.com/96jonesa/remove-chat-options.git
-commit=84afa66dd865eb3f53cd5ff341268b587922efdc
+commit=879a2a323f2cd83f800a507d0715fdbfa6518ea1
--- a/plugins/remove-chat-options
+++ b/plugins/remove-chat-options
@@ -1,2 +1,2 @@
 repository=https://github.com/96jonesa/remove-chat-options.git
-commit=881f8ef8a478503cd02f6554a6bcbf9d194ad85d
+commit=d1070ef0568fd3d412a2808eb88182d2e366b02b
--- a/plugins/remove-chat-options
+++ b/plugins/remove-chat-options
@@ -1,2 +1,2 @@
 repository=https://github.com/96jonesa/remove-chat-options.git
-commit=5542319e4a6bae39f1b6d00210b8e1213f2edf9a
+commit=4f1e8c39233361c53b9f8f9774296d53b32d2c2f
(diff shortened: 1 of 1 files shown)


YvesW:
Could you give some visual examples of the plugin working? Having a bit of a hard time understanding the use case and the support traffic we'll get because of the plugin. Or maybe I do, but then a menuswap might be more appropriate than hiding the options outright?

96jonesa:
The motivation for the plugin came from speed-running Barbarian Assault, in which it is preferable to have as small of a viewport as possible while also maximizing the amount of that viewport in which you can interact with various entities

An example of this is when stocking food as the Healer - the player will right-click the dispenser and select the correct food several times without losing a tick

Without this plugin, a transparent chatbox makes this possible but do to messages in the chatbox / private messages this can be interfered with as various "Add friend", "Add ignore", etc. options appear in the middle of the options list

Even if all of these options were moved to the bottom of the list, depending on where the cursor is when the right-lick menu is opened this could change the position of the desired menu option relative to the cursor (if the cursor is too close to the bottom of the viewport)

will add images in a moment to help explain

96jonesa:
image

image

96jonesa:
in the first image I open the right-click menu close to the bottom of the viewport, and it opens hovering over option 4

in the second image, I open the right-click menu higher up and it opens hovering over option 2

with more options in the list, even if all the undesired options are at the bottom, the length of the list increases and thus results in the position of options relative to the cursor changing when opening the right-click menu near the bottom of the viewport

this plugin removes the undesired options entirely, avoiding this

96jonesa:
when trying to select the same option once per tick from a right-click menu e.g. 6 ticks in a row, not having this plugin and instead using menu swaps makes it unreliable since (1) the contents of the right-click menu can vary based on the contents of the chatbox; and (2) a longer list of options means the player has to right-click further from the bottom of the viewport to ensure a consistent cursor position relative to the menu options

tylerwgrass:
Why not just hide the chat?

96jonesa:
it is still useful to see the chat - other players may be communicating information to you that you need to see

96jonesa:
there are also game-related messages that can be useful e.g. it is common for Barbarian Assault players to alch their horn on the first tick of a wave to make room for more food, then instead of navigating back to their inventory from their spellbook they can just look at the game messages that indicate if they are stocking food correctly each tick

PR #7253, opened 2025-01-19
Update profit-tracker to latest commit

mrhappyasthma:
Significant improvements to the profit calculation. And better support for looting bag, bank, and other similar container items.  Also removed deprecated APIs.


iProdigy (approved):
code could be streamlined in some places, but lgtm for the hub

e.g. `getItemCollectionDifference` could be simplified [like so](https://github.com/pajlads/DinkPlugin/blob/66b80d8171b6caf002171f306eb06f41c2510487/src/main/java/dinkplugin/notifiers/GroupStorageNotifier.java#L259)

tylerwgrass (approved):

PR #7257, opened 2025-01-20
add loot-filters plugin

riktenx:
Adds the "Loot Filters" plugin, a re-imagined implementation of the base Ground Items plugin experience which supports user-scriptable item overlay rules known as loot filters.

The upstream [readme](https://github.com/riktenx/loot-filters/blob/main/README.md) describes the plugin in greater detail and includes a link to a guide for the filter scripting language.

### DIFF
--- a/plugins/loot-filters
+++ b/plugins/loot-filters
@@ -1,0 +1,2 @@
+repository=https://github.com/riktenx/loot-filters.git
+commit=46599d566a5ec7b4620c5c21bd8cbc4843db97d6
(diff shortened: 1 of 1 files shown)


LlemonDuck:
Nothing is a blocker, so to make my life easier reviewing any future smaller diffs, you might want to look into:
* moving the parsing into its own thread. I'm sure the parsing isn't too expensive, but doing the i/o could lead to hiccups
* think about storing the scripts on disk. The config manager has a maximum value size that is fairly restrictive, and we try not to store large entries if we can avoid it. Obviously this makes syncing more difficult, so weigh the tradeoffs.

PR #7261, opened 2025-01-22
Update impling-saver to 2.0.0

TheLope:
Complete refactor of the plugin to provide functionality of saving implings only when the associated clue is found in the bank
This is because the prior functionality was only checking the players inventory, which was since added to the core game.
Large portions of the functionality mirror what was done for the Casket Saver refactor.


tylerwgrass:
Given that this was transferred to you from another author, you should at least properly attribute them credit in the copyright

TheLope:
> Given that this was transferred to you from another author, you should at least properly attribute them credit in the copyright

Okay yeah totally happy to do so just wasn't sure the proper formatting, expected feedback on that one

PR #7268, opened 2025-01-22
Create YouMissedThatOne

hyppis1:
This plugin enables you to play custom sound effects for both hit and miss actions during attacks. You can configure whether these sound effects should apply to normal attacks, special attacks, or both. You can also swap these custom sound effects to what ever audio file you want. Similar to sound swapper plugin.

Original plan was just to play custom sound effects, and then use the sound swapper plugin itself to swap these to what ever audio files. But the sound swapper plugin cant hear the sound effects this plugin makes, so i had to code small, lesser version of sound swapper to this plugin itself.

The way users configure which weapons should have what sound effects played, is ID based. I have read the rules and i know ID based plugins are not really welcome thing. But this is only to play/swap custom sound effects. Additionally there is just sound effect randomizer users can turn on.

Also a bit of an backstory why i made this plugin. Content creator Molgoatkirby uses sound swapper plugin to play custom sound on his zcb special attacks. And he has wished for a while to have different sound play depending on whether the special attack hits or misses. So i started making this plugin, but it felt it would be a "waste" for the plugin to only work on zcb. So i made it to work solely on user provided ID's so they can add any weapon and to play any sound effect with said weapons.


tylerwgrass:
In addition to ID based plugins typically being disallowed due to moderation challenges, the overlay providing soundIds also poses issues as it could be abused for indicating boss mechanics.

It also seems this is doing a lot of what other plugins already do (sound swapper, visual sounds, and annoyance mute to name a few).

hyppis1:
I will happily remove the sound ID display. Reasoning for it was to mute the default attack sounds if players want to do so.

Also about the many similarities between plugins. 
- Removing sound ID display fixes the problem with visual sounds plugin.
- I will remove the muting sounds options happily. Reason for it was to be more handy for users to mute unwanted sounds on this plugin, but  annoyance mute plugin already does that, and does that better. So yeah, I will remove that option also.
- But the sound swapper similarity sadly needs to stay. I hoped that sound swapper plugin could swap the sound effects this plugin produces, but sadly it cant pick up the sound effects this plugin makes. So I had to implement my own, smaller version of sound swapper plugin. Sound swapper plugin is still the superior choice for everything else.

LlemonDuck:
It looks like you carried this over from sound swapper, but make sure you're closing your Clip instances to release the underlying resources. Typically in other plugins that means closing the last clip before the next time you play one, as well as on shutDown

PR #7269, opened 2025-01-23
add chatbox-scroll-to-zoom plugin

96jonesa:
When enabled, if scroll-to-zoom is enabled (in the game's settings), then scrolling the mouse wheel while hovering over the chatbox will zoom instead of scrolling through the chat history.

If the user is holding the CONTROL key, then scrolling the mouse wheel while hovering over the chatbox will scroll through the chat history instead of zooming.

### DIFF
--- a/plugins/chatbox-scroll-to-zoom
+++ b/plugins/chatbox-scroll-to-zoom
@@ -1,2 +1,2 @@
 repository=https://github.com/96jonesa/chatbox-scroll-to-zoom.git
-commit=ef91495caf123921edee79adc48acf5a0e730c7f
+commit=5ddfe8bb42f816a6de7539e7877ca1701b43494f
--- a/plugins/chatbox-scroll-to-zoom
+++ b/plugins/chatbox-scroll-to-zoom
@@ -1,2 +1,2 @@
 repository=https://github.com/96jonesa/chatbox-scroll-to-zoom.git
-commit=ce119deb4111133924155651c1c7c26c3ae58668
+commit=22db1cf19a847647f0a049ff9f6eb8246fedd76c
--- a/plugins/chatbox-scroll-to-zoom
+++ b/plugins/chatbox-scroll-to-zoom
@@ -1,0 +1,2 @@
+repository=https://github.com/96jonesa/chatbox-scroll-to-zoom.git
+commit=ef91495caf123921edee79adc48acf5a0e730c7f
--- a/plugins/chatbox-scroll-to-zoom
+++ b/plugins/chatbox-scroll-to-zoom
@@ -1,2 +1,2 @@
 repository=https://github.com/96jonesa/chatbox-scroll-to-zoom.git
-commit=5ddfe8bb42f816a6de7539e7877ca1701b43494f
+commit=ce119deb4111133924155651c1c7c26c3ae58668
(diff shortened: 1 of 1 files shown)


YvesW:
Does losing focus properly fire the key released event? Otherwise you might have to check for that

charlieradford:
Following, def would love this!

96jonesa:
> Does losing focus properly fire the key released event? Otherwise you might have to check for that

Can you swing losing focus? Like pressing CONTROL, then clicking some other window and realizing the key?

I initially tried setting the listener to a JavascriptCallback lambda function that would call one script if CONTROL was currently keyed down and another if it was not, but for some reason even though I could confirm it was reaching those logical branches it wouldn't actually run either of the scripts

96jonesa:
unclear what the behavior should be if the scroll wheel is used while hovering over the client _while not focused on the client_

interestingly, even while not focused on the client the scrollWheelListener still fires if you are hovering over the client

96jonesa:
if anyone knows why this doesn't work lemme know:

```
chatboxWidget.setOnScrollWheelListener((JavaScriptCallback) event -> {
    if (client.isKeyPressed(KeyCode.KC_CONTROL)) {
        client.runScript(SCROLL_SCRIPT_ARGS);
    } else {
        client.runScript(ZOOM_SCRIPT_ARGS);
    }
};
```

debugging shows that it does get into the correct if-else branch and the correct scripts are running, it just seems like the behavior of the scripts themselves is getting messed up by virtue of being wrapped in a `JavaScriptCallback`, and not due to the if-else branching

e.g. if I make this change:

```
@Subscribe
public void onScriptPreFired(ScriptPreFired scriptPreFired) {
    if (scriptPreFired.getScriptId() == ZOOM_SCRIPT_ARGS[0]) {
        System.out.println("hi");
    }
}

private void zoomMode() {
    Widget chatboxWidget = client.getWidget(ComponentID.CHATBOX_MESSAGE_LINES);
    if (chatboxWidget != null) {
        // chatboxWidget.setOnScrollWheelListener(ZOOM_SCRIPT_ARGS);
        chatboxWidget.setOnScrollWheelListener((JavaScriptCallback) event -> {
            client.runScript(ZOOM_SCRIPT_ARGS);
        });
    }
}
```

it still runs that script but the actual behavior of the script is very weird

96jonesa:
got the JavaScriptCallback with conditional logic to determine the Script to run working, so the implementation is much simpler now

YvesW:
I don't have time to really check this out and/or dig into this, but keep in mind we got <https://static.runelite.net/api/runelite-api/net/runelite/api/events/FocusChanged.html> You might also be able to check how the OverlayManager handles alt dragging of widgets (IIRC it's that Manager that handles that).

Additionally, make sure to run stuff on client thread that should be ran on client thread. People sometimes have problems with that.

96jonesa:
ye previous commit subscribed to FocusChanged events but testing showed that while the OnScrollWheelListener still picks up scroll wheel events when out of focus, no KeyEvents / client.isKeyPressed registers when out of focus

hence settling for just having it always act as if the CONTROL key is not pressed when out of focus

all scripts are running in a JavaScriptCallback as part of the OnScrollWheelListener, so they're always running on the client thread

PR #7275, opened 2025-01-23
Version 3.5 DropTracker

OSRSKoeppy:
Fix for Side Panel API request (just needed to update URL) Encrypting Webhooks used to prevent malicious users from accessing the webhook

### DIFF
--- a/plugins/droptracker
+++ b/plugins/droptracker
@@ -1,3 +1,3 @@
 repository=https://github.com/joelhalen/droptracker-plugin.git
-commit=189a7880072c8765a2124fb47f607bb712e9623f
+commit=dbd2fd7e54a5f5e1b2e4224ea1f09e19c0900569
 authors=joelhalen,OSRSKoeppy


raiyni:
1. Please don't close and open new PRs. You can push updated hashes here.
2. Also your encryption isn't going to protect you from real malicious actors

joelhalen:
> 1. Please don't close and open new PRs. You can push updated hashes here.
> 
>     2. Also your encryption isn't going to protect you from real malicious actors

Of course, somebody who had RuneLite and really wanted to get to the end urls would have no problem doing so. The code is right there. That's a non-issue for us; it's something that's unavoidable. We are cycling through webhooks, creating new ones on a fairly regular basis and deleting old ones. This is the only way our plugin is able to communicate with our backend, and so hashing the webhooks when we store them on github means that web scraping libraries like the ones recently giving us problems won't be able to determine what they're looking at, meaning they're at least more secure than posting a raw discord url--and secure enough for our purposes.

OSRSKoeppy:
Apologizes on opening and closing my PR. Noticed it had some weirdness on my fork that made it seem like there was a lot of changes I made since my last pull request which wasn't accurate

PR #7277, opened 2025-01-24
Update maple-xp-bar

SirHitech:
- add config to change the bar's placement
- add config to change the size of the bar
- add config to change the xp progress text color
- add config sections for a slight organization improvement


SirHitech:
Gathering further feedback from users - will re-open with more changes

PR #7286, opened 2025-01-25
Update maple-xp-bar

SirHitech:
added configs to customize the size and position of the xp bar


iProdigy:
avoid reconstructing the font every frame (e.g., save to a field that is updated as the config is changed): https://github.com/SirHitech/MapleXPBar/blob/f0f6d0fde43838fadcf25038fca31c68f04207ea/src/main/java/com/maplexpbar/MapleXPBarPlugin.java#L271

also consider using [`Font#deriveFont(float)`](https://docs.oracle.com/javase/8/docs/api/java/awt/Font.html#deriveFont-float-) instead of directly calling the constructor

iProdigy:
instead of `new Font("Runescape Small", ...)` you can use `FontManager.getRunescapeSmallFont()` (and `deriveFont` it)

iProdigy (approved):

PR #7292, opened 2025-01-25
update remove-chat-options

96jonesa:
don't remove chat options if CONTROL key is pressed

provides the user with a convenient hotkey (CONTROL) to temporarily toggle the plugin's behavior off


iProdigy (approved):
would be nice to document this behavior somewhere users can read

96jonesa:
> would be nice to document this behavior somewhere users can read

added this to the plugin description

iProdigy (approved):

tylerwgrass (approved):

PR #7297, opened 2025-01-26
Create `wilderness-map-locations` plugin

DapperMickie:


tylerwgrass:
Some perf considerations:
- Getting the closest point and direction in [ClosestLocationOverlay](https://github.com/DapperMickie/wilderness-map-locations/blob/0f53c5008214f892c77953a06f6f4e648a9355e6/src/main/java/com/github/dappermickie/wildernessmaplocations/ClosestLocationOverlay.java#L48-L54) could be run per-tick instead of per-render
- [getMapPoints](https://github.com/DapperMickie/wilderness-map-locations/blob/master/src/main/java/com/github/dappermickie/wildernessmaplocations/WildernessMapLocationsPlugin.java#L57) is regenerating every map point multiple times per render, but could instead be initialized once and referenced

LlemonDuck:
what are you hoping to use this for?

DapperMickie:
It's used for callouts during Multi PvP fights, shows your current location so you can easily call out where you are. It also helps out for newer players to understand all these callouts and to easily find the locations on the map.

PR #7302, opened 2025-01-29
Create vorkath-run-warning

jkvlntn:
Plugin to warn players if they are fighting Vorkath while running. If sprint is enabled while in the fight, the whole screen will be tinted with configurable color.


raiyni:
Can you add an indicator in your overlay that this is coming from this plugin please

jkvlntn:
I added a panel, not sure if that'ts what you meant.

Felanbird:
> I added a panel, not sure if that'ts what you meant.

He means an indication that the warning is coming from THIS plugin, so as long as the indication says "vorkath run warning" somewhere, then it's good.

jkvlntn:
Thanks, I added that.

PR #7310, opened 2025-01-30
Update GP-Per-Hour to 1.10

MosheBenZacharia:
- Support new bank deposit interface
- Fix large loot grid expanding the window
- Add right click menu to delete trips below the clicked trip
- Adjust compost bucket tracking logic


MosheBenZacharia:
Thanks so much for the quick review lemon!!

PR #7311, opened 2025-01-30
added new plugin runelite-duplicator

NickSpaghetti:


LlemonDuck:
we're not interested in hosting this for two reasons:
* we try avoid external process interaction in hub plugins as much as possible
* not enough practical use

PR #7312, opened 2025-01-30
add mlm-despawn-timer

xetrics:
A plugin to estimate/display when a vein will despawn in the Motherlode Mine.


LlemonDuck:
since you aren't currently using your config class you should remove (at least) the `@Provides` block in order to not show a settings icon beside your plugin in the config panel

xetrics:
I wasn't sure if the config implementation was needed for the plugin to function properly. I've removed it for now.

iProdigy:
for completeness, could you remove from `availableVeins` upon the `WallObjectDespawned` event

xetrics:
I cannot, due to how MLM works. Veins are removed and readded every tick.

iProdigy:
no worries, the `availableVeins.clear()` on `LOADING` gamestate should be sufficient (if not you can consider `WeakHashMap` & `newSetFromMap`)

iProdigy (approved):

PR #7321, opened 2025-01-31
Update Log Locked to work with new Log interface.

Mrnice98:
Update Log Locked to work with new Log interface.


iProdigy:
```
loglocked: plugin uses terminally deprecated APIs:
```

Mrnice98:
Thank you for the heads up. I didn't see it failed the build check.

PR #7326, opened 2025-01-31
Ectoplasmator reminder

staytheknight:
Changed the structure of the plugin to fetch a list of NPC names from another branch, and compare the name of the NPC targets (Instead of IDs). 

This is to set up a structure for future updates, where only the text file on the other branch needs to be updated with new Spectral NPCs, and no changes need to be made to the function or structure of the code on main.


LlemonDuck:
You're not closing this reader: https://github.com/staytheknight/ectoplasmator-reminder/compare/7b8c95980a813861de9ed61853ebe6d4dbcd58ba...staytheknight:830e603cfc5589c707a37a6db21d5cd1a3073e12#diff-10693da25e57a0233faf318338fc9fb18252e411e8edc6a852d65756c8fdc750R53-R54

Also you should not block the EDT while it downloads these files on plugin start up. You can inject the OkHttpClient and use that which has support for async operations built-in.

staytheknight:
> You're not closing this reader: [staytheknight/ectoplasmator-reminder@7b8c959...staytheknight:830e603cfc5589c707a37a6db21d5cd1a3073e12#diff-10693da25e57a0233faf318338fc9fb18252e411e8edc6a852d65756c8fdc750R53-R54](https://github.com/staytheknight/ectoplasmator-reminder/compare/7b8c95980a813861de9ed61853ebe6d4dbcd58ba...staytheknight:830e603cfc5589c707a37a6db21d5cd1a3073e12#diff-10693da25e57a0233faf318338fc9fb18252e411e8edc6a852d65756c8fdc750R53-R54)
> 
> Also you should not block the EDT while it downloads these files on plugin start up. You can inject the OkHttpClient and use that which has support for async operations built-in.

Great, thank you for the feedback, I will look into fixing these issues, and trying to async this process.
I appreciate the heads up as to what I should look into as I'm not very familiar with I/O.

iProdigy:
Please use the version of okhttp already provided by the runelite client rather than depending on `4.12.0`

staytheknight:
> Please use the version of okhttp already provided by the runelite client rather than depending on `4.12.0` (simply inject `OkHttpClient` rather than try to construct it)

Thanks for the heads up, first time using OkHTTP for me, had a lot to figure out :)
Changing it to @Inject now.

staytheknight:
Looks like it didn't like me calling 'new OkHttpClient()'. 
Trying to fix that now, I'm just confused how to properly use @ Inject.

tylerwgrass:
https://github.com/runelite/runelite/blob/13073fd235b386f17640bdef53468d23194446c8/runelite-client/src/main/java/net/runelite/client/plugins/crowdsourcing/CrowdsourcingManager.java#L50-L51

just inject and then you can use it, you do not need to call `new`

staytheknight:
> https://github.com/runelite/runelite/blob/13073fd235b386f17640bdef53468d23194446c8/runelite-client/src/main/java/net/runelite/client/plugins/crowdsourcing/CrowdsourcingManager.java#L50-L51
> 
> just inject and then you can use it, you do not need to call `new`

Thank you very much for the examples, examples help me tremendously.
I had my new code set up exactly as yours was with the Inject.
My Inject was not working as it kept giving me a null pointer. 

Looking more closely at the differences between our code, I found out the class storing my OkHTTPClient inject was 'abstract' and my IDE kept telling me to add static in front of a lot of my function calls. 

Removing Abstract and Static modifiers and changing how the SpectralCreature URL reader is called in the Plugin seems to have fixed my issues on my side of things after testing, we'll now see if the automatic checks on GitHub like it or not. 

Java is not a language I use often if at all, so this has been quite the adventure.

LlemonDuck:
One last change: Use slf4j instead of JUL for logging. You can use `@Slf4j` at the top of your class to automatically generate the `log` variable in your class:
```java
@Slf4j
public class SpectralCreatures
```
or you can do it manually with `Logger log = LoggerFactory.getLogger(SpectralCreatures.class);` where both of the logging classes are from the `org.slf4j` package.

Then use `log.info` or `log.warn` here (warn is fine since this won't spam) https://github.com/staytheknight/ectoplasmator-reminder/compare/7b8c95980a813861de9ed61853ebe6d4dbcd58ba...staytheknight:f9a0c97edf5fc5a5892f75684f56be59f4b852d3#diff-10693da25e57a0233faf318338fc9fb18252e411e8edc6a852d65756c8fdc750R88-R89

PR #7328, opened 2025-02-01
Create dorgeshkaan-agility-course

Anderzenn:
Plugin that adds some quality of life features to the Dorgesh-Kaan Agility course.

Apologies if the code is a little messy, it's my first time making a plugin using widgets. I also had some issues and I do plan on doing a refactor of the code in the future. But for now I wanted to get an initial release out.

Release v0.1
- Adds a movable widget that tells you which items Turgall requested
- Highlights the requested items from Turgall from the boilers.
- Lets you know if you have a spanner or not
- Widgets are removed from the screen when you're not in the area

### DIFF
--- a/plugins/dorgeshkaan-agility-course
+++ b/plugins/dorgeshkaan-agility-course
@@ -1,2 +1,2 @@
 repository=https://github.com/Anderzenn/dorgesh-kaan-agility-course.git
-commit=15256e53ea37a2bf025f872b8cedf2fcc26889d0
+commit=f42af0e6a1dd2130726a219dac22670248e0a0a9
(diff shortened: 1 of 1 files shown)


LlemonDuck:
Use the RuneLite ImageIO#load utility + don't use printStackTrace to blindly "handle" errors

Anderzenn:
> Use the RuneLite ImageIO#load utility + don't use printStackTrace to blindly "handle" errors https://github.com/Anderzenn/dorgesh-kaan-agility-course/blob/f42af0e6a1dd2130726a219dac22670248e0a0a9/src/main/java/com/anderzenn/dorgeshkaanagilitycourse/DorgeshKaanOverlay.java#L49-L56

These changes should now be in place :)

Anderzenn:
I just noticed in my spanner warning overlay class for whatever reason decided to make a call to the plugin class instead of using the boolean I had in the actual class. Ignore this, that's being fixed in a future update. 

I've fixed it locally, but I don't see a reason to commit it when it's still working. At least not until I have some more stuff to commit.

PR #7332, opened 2025-02-01
Create improved-fixed-hide-chat plugin file

MetalMan196:
This is a clone of the Fixed Mode Hide Chat plugin. The base plugin gives users the ability to collapse the chatbox while in "Fixed - Classic" display mode. This plugin adds a feature that automatically un-hides the chatbox when a chatbox-dialog is opened requiring user input.


Felanbird:
We don't want copies of plugins on the hub, we have a [plugin takeover policy](https://github.com/runelite/runelite/wiki/Plugin-takeover-policy).

PR #7338, opened 2025-02-02
Update another-bronzeman-mode to include the new sidepanel.

CodePanter:
This update gets rid of all the hacky collection log code; and shifts all the item listing and sorting to a new side-panel.
To make sure the users are aware of where to find this functionality, I have also included a new 'version update' system, that writes a chat message on log in to anyone who hasn't used the panel since the current release.


iProdigy:
not blocking: `buildDir` is deprecated by gradle in favor of `layout.buildDirectory`

CodePanter:
> not blocking: `buildDir` is deprecated by gradle in favor of `layout.buildDirectory`

Thank you for pointing that out! I'd rather avoid including new deprecated code in my project; so I just replaced it right away!

PR #7340, opened 2025-02-02
Update radius-markers

Skretzo:
- Fix `client.getCachedNPCs()` deprecation


iProdigy (approved):
consider saving `getSearchText().toLowerCase()` to a variable in `filterMarkersAndDo` rather than creating the new string every loop iteration

iProdigy (approved):

PR #7350, opened 2025-02-04
Raid data tracker 1.6.0 update. The ToA update.

Raitab:
Large PR for the Raid Data Tracker plugin. 

- Tombs of Amascut support
- Maintenance
- Code reformatting


iProdigy:
is the branch change intentional? i.e., are you purposely not merging raid-data-tracker-main onto raid-data-tracker?

Raitab:
Branch change is intentional for now. The next update will revert back to the default, but there are extenuating circumstances with this project and I want to make sure this change gets pushed before I effectively replace that branch with this one.

raiyni:
I can be removed from all of the license headers fyi

mcdub123:
any update on when the plugin will be available by chance?

Felanbird:
> any update on when the plugin will be available by chance?

It's +6000 lines of changes, so when someone has the time to review that.

LlemonDuck:
* log.debug https://github.com/Raitab/raid-data-tracker/compare/0386eba23ce63b06c0de1e6776b1b80552521d86...Raitab:7a125e83fb6df44c3a571400d2bc230947975175#diff-321833cb2db743f83003c80d235683d6cdf0c364aac38158392dfa0e75352a3cR325
* Use a logger https://github.com/Raitab/raid-data-tracker/compare/0386eba23ce63b06c0de1e6776b1b80552521d86...Raitab:7a125e83fb6df44c3a571400d2bc230947975175#diff-deae8f2b14dd3bbf986fa4d7f8007737fe7b49672ddc6a183c44d6ca152e8c77R110
* use a logger https://github.com/Raitab/raid-data-tracker/compare/0386eba23ce63b06c0de1e6776b1b80552521d86...Raitab:7a125e83fb6df44c3a571400d2bc230947975175#diff-deae8f2b14dd3bbf986fa4d7f8007737fe7b49672ddc6a183c44d6ca152e8c77R147
you get the gist

LlemonDuck:
Changing the config group will reset your settings, but in this case I'd say you should still do it since "example" is potentially overloaded. You should try to migrate your users' settings to the new config group if they exist on startup

Raitab:
The config group hasn't changed since 2020, you're seeing a remnant of cas's previous commit.
https://github.com/Raitab/raid-data-tracker/commit/ea6ed5bca7b17319fc0ed252be349ada0bbf5cd4#diff-13e9c16ff85012ded06b67cce1867abd988ab65c764391e92c51777537bff159

LlemonDuck:
mm, note to self ..

LlemonDuck:
This is both a) not descriptive enough of what it's doing, and b) shouldn't be automatic. You can direct players to change the settings for themselves instead and listen for the config changed event.

https://github.com/Raitab/raid-data-tracker/compare/0386eba23ce63b06c0de1e6776b1b80552521d86..Raitab:3398de9093cf0f666a6eb7f786858a375c968dcc#diff-10497be07dfe504e818c4725eaa74f5c4b94079c32071f3a0583d6c52f22fd48R270-R284

Also I'm bit confused why you're using both onPluginMessage _and_ vendoring the entire points tracker to have it locally. All the plugin message provides is points, so why both?

null-zero:
> This is both a) not descriptive enough of what it's doing, and b) shouldn't be automatic. You can direct players to change the settings for themselves instead and listen for the config changed event.
> 
> https://github.com/Raitab/raid-data-tracker/compare/0386eba23ce63b06c0de1e6776b1b80552521d86..Raitab:3398de9093cf0f666a6eb7f786858a375c968dcc#diff-10497be07dfe504e818c4725eaa74f5c4b94079c32071f3a0583d6c52f22fd48R270-R284

I'm happy to make it more descriptive, I realize the code itself could be more clear about what it's doing but I would like to ask for clarification on whether you also meant the messages that get presented to the users.

As for the automatic setting of configs, I assumed it would be okay to be added due to the existence of the "Plugin Presets" hub plugin and that they are behind an interface that requires a user to explicitly interact with a button to enable the ToA plugin and the onPluginMessage setting. But I also realize that the Runelite team is probably cautious about letting plugins change other plugin configs, so I can update those as well.

> Also I'm bit confused why you're using both onPluginMessage _and_ vendoring the entire points tracker to have it locally. All the plugin message provides is points, so why both?

To explain this, the vendoring happened before we got into contact with you about adding the onPluginMessage. I realized that we couldn't easily piggy-back the party points easily. If we vendored the entire points tracker module , it would require everyone in the raid to be using Raid Data Tracker to track properly, making it less useful due to our significantly decreased user base compared to the ToA plugin's user base. After you were kind enough to work on adding the onPluginMessage that allows us to utilize the part points, the thought to keep the vendored code was based on how we would like the plugin to still moderately function for ToA without requiring users to install another plugin while also not having to reinvent the wheel to track personal points in ToA. The vendored code tracks only personal points, which we figured is a fair bare minimum.

PR #7357, opened 2025-02-05
Update chat-success-rates

Skretzo:
- Only show the chat box copy data menu option when there is data to copy and the chat box is visible


iProdigy (approved):
if u want to micro-optimize, `DUPLICATE_CACHE` could be an EnumMap and you could use entrySet instead of keySet + Map#get

tylerwgrass (approved):

PR #7374, opened 2025-02-08
Create dizana-alert

Servilabs:
This plugin is meant for people who use the quiver for raids and wish to avoid going into it and find it is empty because of a recent death or other event.


Servilabs:
fixing initial commit soon

LlemonDuck:
Call OverlayManager#remove in shutDown as well to make sure it is removed if the plugin is disabled even while wearing an empty quiver

LlemonDuck:
You may also want to checkQuiver during startUp for the inverse scenario. Just make sure to do it on the ClientThread

Servilabs:
Thank you very much, I have updated to reflect both changes.

PR #7383, opened 2025-02-09
Upgrade naturalspeech

phyce:
Major upgrade to Natural Speech plugin

+ Removed warning as all voice models are now hosted on github.


abextm:
You can't use different versions of dependencies we already ship, it will break on the release.

PR #7384, opened 2025-02-09
Create removeupperfloorstob

ppyLEK:
This plugin removes tiles on upper floors (1-3) at tob bank improving visibility at the bank and preventing clicks on scenery.


ppyLEK:
Had no idea that its ok to completely remove config class as I assumed that it would cause issues but seems its fine.

PR #7386, opened 2025-02-09
Update loot-filters: add sound playback support

riktenx:


riktenx:
Witholding this for the moment, as arbitrary sound playback from file would put the plugin into restricted review mode which I don't necessarily want to do right now. Reviewing options.

PR #7395, opened 2025-02-11
Update steps-per-clue

Chubs1:
Updating commit to fix a bug where if it was the first message it would spam the user with errors.


iProdigy (approved):
you probably want a null check for `message` on L46

pajlada (approved):

PR #7400, opened 2025-02-11
Update toaghost-plugin

Langesund:
Updated for ironmen/hcim and possible other problems when a player randomly dc's


pajlada:
It doesn't look like you've actually updated any code, just removed comments. Please clarify

PR #7414, opened 2025-02-13
[clan-halls-plugin] ✨ Feature: auto send settings report if clan has not been synced

raphpion:
* Includes a rework of `ClanHallsClient` with typed `APIResponses` without callbacks.

* Requests are now sent using basic authorization header instead of putting the credentials in the body.

* Request payloads have been simplified to reflect this change.


raphpion:
I have closed this PR as sending the `{ clientId, clientSecret }` pair as authorization header isn't very secure! I will rework this approach and open another PR later.

PR #7426, opened 2025-02-15
Add process-priority plugin

rsfost:
This plugin lets Windows users set the priority class of the RuneLite process without having to open Task Manager.


LlemonDuck:
Thank you for the submission but we're not interested in hosting hub plugins using JNA. There are alternative options available for users to set the permanently set the process priority without doing it from within RuneLite which we'd prefer are used.

PR #7429, opened 2025-02-15
Create ChatHighlightPlayer

scriptnapps:


raiyni:
This looks like it's primary use is targeting people in pvp?

scriptnapps:
> This looks like it's primary use is targeting people in pvp?

NO in G.E tag people so you can trade them if you read chat and can't find them or missed chat above their heads?

scriptnapps:
here is a video on how this works https://youtu.be/8tBZ5xTB9oo (it dissapiers after x amount of seconds)

scriptnapps:
UPDATED LICENCE

scriptnapps:
ready to be used :)

PR #7444, opened 2025-02-17
Create npc-stack-reorderer

Skretzo:
Reorders NPC stacks with invisible NPCs from showing the oldest NPCs to instead show the newest NPCs.

There is a limit of 5 NPCs bigger than 1x1 that can be rendered in the same stack. The 6th, 7th, ... will turn invisible, even in the right-click menu. This plugin uses entity hiding to hide the 1st, 2nd, ... NPCs to make the 6th, 7th, ... NPCs become visible. This is useful for niche accounts without access to aoe attacks that e.g. wants to tag magers during zuk.


Skretzo:
Jamflex changed how npc stacks work today, presumably as preparation for sailing...

PR #7447, opened 2025-02-17
Update tictac7x-charges to v0.5.19

TicTac7x:


pajlada (approved):
This resets the fish barrel storage since the key has changed, I assume you're aware and are fine with that. Normally I think we'd prefer a migration but since this is such an easily fixed/low risk config it seems fine

TicTac7x:
Bit confusing, but no, the key is the same, all storages are automatically saved to key + _storage ending dynamically, the old key was just a leftover from when fish barrel didnt have storage info yet, so should be good.

iProdigy (dismissed):

pajlada:
@TicTac7x With your latest commit, you've undone some of the previous work (so this doesn't actually update to v0.5.19, no Flamtaer bag support) - is this intended?

TicTac7x:
Not intended, rechecking...

TicTac7x:
Forgot to merge my local branch to the main charges branch... Thanks for noticing it, hopefully all good now.

pajlada (approved):

PR #7450, opened 2025-02-17
Update emblem-trader-skull-timer

Teekiz:
Updated to version 2.0:
Added timer checks for entering the abyss, wearing items that provide skulls and PVP scenarios. Renamed plugin to 'Skulled Timer' to reflect new features.


LlemonDuck:
FYI renaming your config group will reset all your users' config options to the default so you probably want to keep it as its old value

PR #7458, opened 2025-02-18
Collection Log Luck v1.2.0

peanubnutter:
Detect stale collectionlog.net data to fix calculations when viewing log 
Update luck without having to relog after each kill / drop Hueycoatl and Royal Titans
Wintertodt rework, Moxi, special items like Tempoross, GoTR 
Misc bug fixes


LlemonDuck:
I'm going to merge this since it's not new, but you're still using System.out or System.err somewhere if you can please switch them to slf4j or remove them in another change.

peanubnutter:
Thanks. I don't see any usages besides two commented-out lines in CollectionLogDeserializer.javs which I temporarily uncomment once per update to detect new items. It never makes it into a release.

Not sure why the tool is complaining about commented out lines...

PR #7459, opened 2025-02-18
RuneGPT Submission

Kamran14:
RuneGPT connects with LLM providers to help RuneLite users achieve their goal. With stats preloaded, each response is tailored to the user providing a unique experience.

Currently it only supports Gemini which is available to all users free of charge.
Time constraint


Kamran14:
Will update once build failures are addressed.

Kamran14:
Failures have been addressed and fixed in new commit.

iProdigy:
* this chat message is unnecessary; please remove - https://github.com/Kamran14/RuneGPT/blob/0a33f830df8df8a039485cca67dfc81b1c3cd2b4/src/main/java/com/kamrant/runegpt/RuneGPT.java#L66
* please downgrade this logging to the debug (or trace) level - https://github.com/Kamran14/RuneGPT/blob/0a33f830df8df8a039485cca67dfc81b1c3cd2b4/src/main/java/com/kamrant/runegpt/service/PlayerStats.java#L30
* prefer using the injectable ExecutorService to spawning new threads - https://github.com/Kamran14/RuneGPT/blob/0a33f830df8df8a039485cca67dfc81b1c3cd2b4/src/main/java/com/kamrant/runegpt/panels/GPTPanel.java#L75
* you probably want to create this string using gson so the prompt is properly escaped https://github.com/Kamran14/RuneGPT/blob/0a33f830df8df8a039485cca67dfc81b1c3cd2b4/src/main/java/com/kamrant/runegpt/builders/GPTRequest.java#L16C10-L16C10
* consider compressing this logo file - https://github.com/Kamran14/RuneGPT/blob/0a33f830df8df8a039485cca67dfc81b1c3cd2b4/src/main/resources/com/kamrant/runegpt/logo.png
* what's the point of the enable config option? - https://github.com/Kamran14/RuneGPT/blob/0a33f830df8df8a039485cca67dfc81b1c3cd2b4/src/main/java/com/kamrant/runegpt/RuneGPT.java#L42
* the runegpt file in this repo needs a `warning=` line that explains what data is sent over http; examples can be found at https://github.com/search?q=repo%3Arunelite%2Fplugin-hub%20warning%3D&type=code

iProdigy:
also https://github.com/polyphilz/scapegpt already exists; does this need to be a separate plugin (rather than contributing improvements to the existing plugin)

Kamran14:
- Fixed mentioned issues
- config option was introduced while testing api key, forgot to remove
- compressed img to 18kb, let me know if this is sufficient
- added warning

regarding existing plugin;
- ScapeGPT exists but is not a functional plugin as it hasn't been working for months possibly a year now based on user reviews and my own test
- RuneGPT is open source and time to introduce new features and test end-to-end is fairly easy compared to ScapeGPT due to it having a requirement of a third party server
- ScapeGPT's third party has not been active for a while so likely the author has abandoned the project

YvesW:
You should create an issue and/or a PR to know for sure as described in the take-over policy. Also, we have a take-over policy for this specific reason.

Kamran14:
iProdigy had submitted a PR on that repo 2 weeks ago and there has been no activity from the author; https://github.com/polyphilz/scapegpt/pull/2

- I'd rather not take over the plugin as that would require me to spin up an instance and maintain that for the backend, and likely will have to pay down the road as well
- The approaches these two plugins take is quite different and I'd have to change most of the logic that ScapeGPT currently holds
- ScapeGPT only works with openai currently and requires payment to use, which is against the model I'm trying to use for RuneGPT where players bring their own key

All in all, taking over ScapeGPT isn’t ideal since it would require me to maintain a backend (with potential future costs), its approach differs significantly from RuneGPT's, and it relies solely on OpenAI with a paid model, whereas RuneGPT lets players bring their own API key via Gemini which is free.

raiyni:
We aren't asking you to never change the code of the plugin you takeover.

LlemonDuck:
Don't block the shared ScheduledExecutorService with a call to OkHttpClient#execute. Either create your own executor which you're free to block, or use okhttp's async via enqueue (this would fit your current use case already, just switch to a callback style)

LlemonDuck:

PR #7460, opened 2025-02-18
[Update] - Varlamore House Thieving - v1.5.1 - Sololegends

sololegends:
# Release v1.5.1

Resolves Runelite breaking API change


pajlada:
> Use of java.lang.System.err is not allowed: system i/o

@sololegends  Can you get rid of the `System.err.println` call in `VarlamoreHouseThievingPlugin.java`?

pajlada (approved):

PR #7474, opened 2025-02-19
Update IRC to v1.0.2 (9ef5d8d)

ryanwohara:
Complete refactor of the architecture.
More configuration options.
Multiple channel support.
Server selection.
Links are clickable.
Images can be previewed without your browser.
Font selector.
Best effort emoji support.
Ctrl+k, Ctrl+i, Ctrl+b, Ctrl+u support in the side panel.

Includes authors and new warning in the plugin metadata.

### DIFF
--- a/plugins/irc
+++ b/plugins/irc
@@ -1,4 +1,4 @@
 repository=https://github.com/ryanwohara/irc-plugin.git
-commit=0543db5a8dd48ec9fc53909ff8560652d055b34d
+commit=9ef5d8d7dce23c3e1999c914f0a7a8f20f6c0f75
 warning=This plugin submits your IP address and applicable IRC messages to a third-party server not controlled by the RuneLite developers.
 authors=ryanwohara
--- a/plugins/irc
+++ b/plugins/irc
@@ -1,2 +1,4 @@
 repository=https://github.com/ryanwohara/irc-plugin.git
-commit=636f6b940d73176974497b1fb961dad4b55b1e26
+commit=7be0991a861b1d60d2fbcaf996e12bc88730e9e0
+warning=This plugin submits your IP address and applicable IRC messages to a third-party server not controlled by the RuneLite developers.
+authors=ryanwohara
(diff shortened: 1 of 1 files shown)


ryanwohara:
(This was discussed in Discord but I'm noting here for the sake of everyone)

I cannot find any instance of System.out calls in my codebase :thinking: this might be a false positive in the automated checks

ryanwohara:
Trying to remove the json dependency as it was duplicated and unnecessary for this work

ryanwohara:
Downgrading kitteh irc seemed to have solved my issues - about to push up my changes.

package/gradle/verification-metadata.xml
```diff
@@ -4,6 +4,11 @@
       <verify-metadata>true</verify-metadata>
       <verify-signatures>false</verify-signatures>
       <trusted-artifacts>
+         <trust group="net.runelite"/>
```
ryanwohara, line 7: Idk what this is; going to rectify momentarily

ryanwohara:
image

This PR isn't as bad as it looks on the surface, I promise

ryanwohara:
```
Showing  with 3,003 additions and 997 deletions.
```

Should be less daunting to review now. 

Thank you in advance.

pajlada (changes requested):
You've got a lot of regexes that are used a lot of times without pre-compiling (e.g. stripStyles & the part that prevents sending `;P` as an IRC message, message boldness/underline/italic/color matching, link matching). It might be an idea too compile them once & re-use them

an NPE is fired if a user tries to send an IRC message by typing `;hi` with no username set (i.e. before ircAdapter is set). Mark `ircAdapter` as nullable to get hints for other places that might be without checks.

The link parsing makes it possible to do some funky stuff with the `<a href` element that's created. That regex should probably be locked down a bit more. (example message: `https://f'>a</a><b>hi`)

I think it'd be a good idea to leave link previews for a separate PR - they're pretty janky right now:
 - multiple link previews can get stuck on screen
 - link preview says it accepts avif/webp but no animation works
 - link preview user agent feels weird

non-blocking:
- You can parse emojis straight into the final map or list instead of parsing them into a temporary map/list then copying them into the final map
- DMs can have dupe entries in the panel
- `closePane` function looks a bit rough, around ~366 you'd probably want to null-check to avoid NPE

PR #7479, opened 2025-02-19
Hold your ground

skeldoor:
Update hold your ground to remove cached npcs deprecated function


iProdigy (approved):
could be simplified via `event.getMenuEntry().getNpc()`

PR #7491, opened 2025-02-20
Update nekos

Ondra-D:
- Improved overlay rendering for better performance.
- Added the option to change the image type displayed (Nekos, Kitsune and cats (for the normies))


Ondra-D:
If I try to render just the buffered image the game freezes for a second, does not happen with the volatile one.

LlemonDuck:
Instead of using a VolatileImage, try something like setting a "loaded" flag after the image has finished loading, and don't attempt to draw the image until that is set. You're probably just hanging the render thread waiting on the image to load.

iProdigy:
also avoid locking in `Overlay#render`

PR #7492, opened 2025-02-21
Update xp goalz

hybridbrizz:
- add Total XP goal


iProdigy:
fyi overall experience can overflow an integer https://github.com/hybridbrizz/Xp-Goals/blob/56b04caae6fc93cbfd38292d2e5404f1b9b123d5/src/main/java/com/ericversteeg/XpGoalsPlugin.java#L94

iProdigy (approved):
would recommend creating another pr with just the int->long change for skillsXp map

PR #7494, opened 2025-02-21
add playernames plugin

MatthewA:
PlayerNames Plugin is a player name indicator, shows player name info and color codes the text based on combat brackets.
The colors are customisable in the plugin settings.


tylerwgrass:
This falls under the rejected features listed [here](https://github.com/runelite/runelite/wiki/Rejected-or-Rolled-Back-Features):
> Level-based PvP player indicators (eg. highlighting players who can be attacked, or are within X levels range)

PR #7504, opened 2025-02-22
Added Wine Tracker Plugin

lb4292:
This plugin tracks wine fermentation time, calculates wines per hour, and estimates the number of wines needed to reach a target level.


tylerwgrass:
You should have generated your plugin from the plugin template as listed [here](https://github.com/runelite/plugin-hub?tab=readme-ov-file#creating-new-plugins)

Also the fermentation time and math used for bad wines are just wrong.

Did you write this with ChatGPT?

lb4292:
I am new to this and thought of a good idea that could help people, I used ai to help

PR #7509, opened 2025-02-22
Create tob-predicted-hit

JaccodR:
A plugin that shows party members predicted hits above their heads in ToB.

### DIFF
--- a/plugins/tob-predicted-hit
+++ b/plugins/tob-predicted-hit
@@ -1,2 +1,2 @@
 repository=https://github.com/JaccodR/party-hits.git
-commit=71afdd38a6bc35a96a945dd150be4f5d417a4cc8
+commit=9e6d03f8a7535797df4c652eb06c21b9cf3c44ce
--- a/plugins/tob-predicted-hit
+++ b/plugins/tob-predicted-hit
@@ -1,0 +1,2 @@
+repository=https://github.com/JaccodR/party-hits.git
+commit=71afdd38a6bc35a96a945dd150be4f5d417a4cc8


pajlada:
This plugin feels pretty aligned with what [Raids death indicator](https://github.com/vikke1234/raids-death-indicator) tries to achieve, did you look into contributing/collaborating with that plugin? Or do the plugins differ in ways I haven't thought about

JaccodR:
> This plugin feels pretty aligned with what [Raids death indicator](https://github.com/vikke1234/raids-death-indicator) tries to achieve, did you look into contributing/collaborating with that plugin? Or do the plugins differ in ways I haven't thought about

I hadn’t heard of this plugin before, but after reviewing it, it appears to be focused on predicting deaths for NPCs in ToA. My plugin aims to provide early visibility of teammates' hits, which is particularly useful for maging in the Maiden room, as it removes the need to manually call out bow hits. Additionally, my plugin includes an estimated HP feature for Maiden that does not require everyone to be in the RuneLite party or any calibration, unlike the Raids Death Indicator.

pajlada:
> > This plugin feels pretty aligned with what [Raids death indicator](https://github.com/vikke1234/raids-death-indicator?rgh-link-date=2025-02-23T12%3A25%3A01Z) tries to achieve, did you look into contributing/collaborating with that plugin? Or do the plugins differ in ways I haven't thought about
> 
> I hadn’t heard of this plugin before, but after reviewing it, it appears to be focused on predicting deaths for NPCs in ToA. My plugin aims to provide early visibility of teammates' hits, which is particularly useful for maging in the Maiden room, as it removes the need to manually call out bow hits. Additionally, my plugin includes an estimated HP feature for Maiden that does not require everyone to be in the RuneLite party or any calibration, unlike the Raids Death Indicator.

Thank you for the extra context - @vikke1234 can you confirm this from your perspective too?

vikke1234:
Hi @pajlada,

I think "raids death indicator" is a little miss-leading, sure it does predict deaths like with "nylo death indicators" and the like. But it also does a similar thing with Akkha and when it would proc the shadows, which was it's initial purpose.

I spoke with @JaccodR  on discord, in my opinion it would fit into my plugin. Though as he mentioned on discord, it would probably fit into the "ToB qol" one as well.

Though there's the argument to be made that if you don't need e.g. chambers/toa, you get a bunch of other stuff bundled that you don't want. For this I wouldn't mind adding toggles for raids.

pajlada:
@JaccodR Would you like to keep your plugin as its own plugin, or would you prefer to contribute it to another plugin? Both are fine options

JaccodR:
> @JaccodR Would you like to keep your plugin as its own plugin, or would you prefer to contribute it to another plugin? Both are fine options

I would prefer to keep it as its own plugin if possible as this would be easier for me (don't have to read up on the entire functionality of the raids death indicator).

PR #7511, opened 2025-02-23
Update BAS Queue Helper

BA-Services:
Added 1 3rd party dependancy


tylerwgrass:
Is there a reason you can't move your S3 calls behind your APIGW or use pre-signed URLs instead of adding the aws sdk as a dependency?

SkylerPIlot:
Pre-signed URLs expire after 1 week iirc.


I think we might be able to alter it to not use SDK functionality, let me play with it a little more and talk to the other dev. 👍

SkylerPIlot:
@tylerwgrass was able to put it all behind the APIGW, and should now be ready for a normal review

pajlada:
@SkylerPIlot The plugin currently does not compile, resolve the issues you see in the build action https://github.com/runelite/plugin-hub/pull/7511/checks

pajlada:
@SkylerPIlot The plugin now compiles, but uses functions we disallow - see the Checks tab for more details, or click the check under the comments stating "Action required" https://github.com/runelite/plugin-hub/pull/7511/checks?check_run_id=37929509057

SkylerPIlot:
Okay, looks like I got all the checks and whatnot correct, please let me know if there's anything else needed otherwise.

pajlada (changes requested):
Please clarify in your config that specifying a value in the `KEY` entry that entering an API key means it'll send network requests to an external API

`client.getLocalPlayer()` can return null, you should check for that in `onGameTick` and stop/reset your logic

`client.getWidget(...)` can also return null, you should check for this too before using the result

Less important: If the google form code isn't used (it looks like it's deprecated according to the top of the form), you can remove it

PR #7519, opened 2025-02-24
Add barrows-path plugin

HectorMF:
This plugin highlights the doorways that lead to the barrows chest.

### DIFF
--- a/plugins/barrows-path
+++ b/plugins/barrows-path
@@ -1,0 +1,2 @@
+repository=https://github.com/HectorMF/barrows-path.git
+commit=bdf10aa868eca973253035425b08ef68938f9b91


HectorMF:
That's fine, there is a lot of overlap with the functionality.  Can someone check that this doesn't break the rules? 
While this change wouldn't speed up barrows runs or makes them any easier, it does solve the maze.  My worry is it may break the rule, specifically "Reveals the maze layout in the Sotetseg boss fight in Theatre of Blood".

LlemonDuck:
Barrows is not part of the Theatre of Blood so I'm pretty sure you're good on that front. Hopefully we'll see this back here with a PR from the Barrows Door Highlighter plugin, but if that doesn't work out feel free to resubmit this.

PR #7524, opened 2025-02-25
Update loot-filters to v1.3.0

riktenx:


iProdigy (approved):
I'd change this line to `ids.contains(item.getId())` to avoid stream overhead: https://github.com/riktenx/loot-filters/blob/ea9e0b2e70637f9a2b1fc7971fd5b95bd9b2d673/src/main/java/com/lootfilters/rule/ItemIdRule.java#L29

tylerwgrass (approved):

PR #7527, opened 2025-02-25
Update Monster Stats

Koitere:
Changed now deprecated call of getCachedNPCS() to the proper updated call getTopLevelWorldView().npcs().byIndex()


Koitere:
Apologies forgot to make this from separate branch. Will close this and open with proper branch.

PR #7537, opened 2025-02-26
AutoTranslator Plugin

Alluneve:
AutoTranslator is an implementation of [DeepL java library](https://github.com/DeepLcom/deepl-java) with [Optimaize](https://github.com/optimaize/language-detector) for language detection in runelite.

it subscribes to chat messages and DeepL will translate them and add a chat message with the translated text if:
- they are at least 2 words
- they match one of your configured languages _default fr, de, es and it_

the plugin requires bringing your own DeepL API key for the translation services

### DIFF
--- a/plugins/autotranslator
+++ b/plugins/autotranslator
@@ -1,0 +1,2 @@
+repository=https://github.com/Alluneve/AutoTranslator
+commit=5b6d9e51fc4055116ff39ded89f67193d30aed5e


raiyni:
As mentioned in discord, we already have multiple translation plugins available. Thus we aren't interested in distributing more of them. I believe the https://runelite.net/plugin-hub/show/polywoof plugin already supports deepl. If there is a translation plugin that you would rather collaborate on that is missing some specific feature, we would prefer you go that route.

PR #7540, opened 2025-02-27
Update loot-filters to v1.3.3

riktenx:


iProdigy (approved):
you may also want to wrap your `clearIndices` call in `shutDown` within a clientThread runnable

tylerwgrass (approved):

PR #7542, opened 2025-02-27
Update tob-predicted-hit

JaccodR:
Updates:
* Fix fake xp drops for multi-hit weapons like scythe being grouped instead of processing seperately
* Exclude multi tick weapons from maidens damage queue
* Add a config option to only update maidens live hp every tick, instead of on every xp drop
* Make maiden only option still sent hits to other party members, but not display any
* Ensure maidenHandler deactivates when the party wipes


tylerwgrass (dismissed):

iProdigy:
your `VarbitChanged` handler should check whether the varb you care about is the value that changed: `event.getVarbitId() == Varbits.THEATRE_OF_BLOOD`. then you can use `event.getValue()` instead of `Client#getVarbitValue`

iProdigy (approved):

tylerwgrass (approved):

PR #7543, opened 2025-02-27
Update raids death indicator with chambers support

vikke1234:
This PR adds chambers support (olm mage hand and vespula portal) along with some misc fixes.


iProdigy (approved):

LlemonDuck:
btw if you just use log.debug instead of println then you won't need to comment out your debug statements before submission. You can add a logger to any class by annotating it with `@Slf4j`.

vikke1234:
@LlemonDuck I don't really like the log.debug as it's cluttered by a ton of other logs from other plugins. I've been meaning to try to add what you suggested a while back like the toa plugin does but I haven't gotten around to it yet.

PR #7545, opened 2025-02-27
RuneMod. A new graphics plugin for the phub.

RuneMod:
This plugin modifies RS graphics by rendering the 3D world in an external program. The external program's graphics are overlayed on the runelite canvas.

The plugin is not ready for a full public release. For this reason, the plugin requires a private beta key in order to work. You can use this key for testing purposes "DMXgvAS2mCPXQMKnY3qmUCtthZO0R8Si".


tylerwgrass:
You are adding new dependencies that will require verification. 
See [here](https://github.com/runelite/plugin-hub/tree/master/package/verification-template) for info and [here](https://github.com/runelite/plugin-hub/pulls?q=is%3Apr+label%3A%22dependency+change%22+) to see what others have done.
Keep in mind this will significantly add time to review as the new dependencies need to be individually verified that they are okay, so you may want to remove these dependency additions unless you absolutely need them.

LlemonDuck:
As discussed in Discord, downloading additional software at runtime is not allowed and we cannot host the plugin on the hub in this state. In order to resubmit you would need to be able to package everything into easily-reviewable, build-time, reproducible-build-capable dependencies. Note that this basically means you would need to be able to build Unreal Engine from source, which is not possible and we wouldn't review it even if it was.

You are free to do whatever you wish with self-hosting, but I will warn you that as of the most recent client statement:
> the only third-party clients that are approved for use are [RuneLite and HDOS]. The use of any other third-party client may result in permanent action being taken against your account.

PR #7554, opened 2025-02-28
Create Unmuted Jingles

hootisman:
Plugin that plays jingles while the game music is muted. If the game is muted, it will unmute for a short duration to let the jingle play.


vikke1234 (commented):
lgtm, I can't approve though

PR #7561, opened 2025-03-02
Update Dink to v1.11.3

iProdigy:
Dink v1.11.0 allows other plugins to trigger webhook notifications via `PluginMessage`

https://github.com/pajlads/DinkPlugin/releases/tag/v1.11.0
https://github.com/pajlads/DinkPlugin/releases/tag/v1.11.1
https://github.com/pajlads/DinkPlugin/releases/tag/v1.11.2
https://github.com/pajlads/DinkPlugin/releases/tag/v1.11.3

### DIFF
--- a/plugins/dink
+++ b/plugins/dink
@@ -1,3 +1,3 @@
 repository=https://github.com/pajlads/DinkPlugin.git
-commit=2ca2ed1627bd35573ab62e4ee30851d6eae681ca
+commit=2a348498862a7808e8abc479d00718757b36fa9a
 authors=Mm2PL,pajlada,Felanbird,iProdigy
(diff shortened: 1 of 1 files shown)


LlemonDuck:
I have some concerns here, primarily around the by-default enablement and inclusion of screenshots. I'd prefer that it was opt-in since the URL is externally provided. Otherwise this introduces a non-obvious control flow which makes other plugins harder to review.

Watchdog can be responsible for informing users that they will need to manually enable this behaviour in Dink's config.

iProdigy:
Updated to require user opt-in & force external plugins that specify a url to supply `HttpUrl` rather than string in order to flag the code scanner

PR #7562, opened 2025-03-02
Create special-attack-infobox

iZupe:
Adds an infobox that displays the current special attack energy

### DIFF
--- a/plugins/special-attack-infobox
+++ b/plugins/special-attack-infobox
@@ -1,0 +1,2 @@
+repository=https://github.com/iZupe/Special-Attack-Infobox.git
+commit=a4dbf4ae8ebbb13d066a74a60e12073253bdf0b9
--- a/plugins/special-attack-infobox
+++ b/plugins/special-attack-infobox
@@ -1,2 +1,2 @@
 repository=https://github.com/iZupe/Special-Attack-Infobox.git
-commit=a4dbf4ae8ebbb13d066a74a60e12073253bdf0b9
+commit=20873a491a0d0572421b1e1b1a2d1c506cce60ce
--- a/plugins/special-attack-infobox
+++ b/plugins/special-attack-infobox
@@ -1,2 +1,2 @@
 repository=https://github.com/iZupe/Special-Attack-Infobox.git
-commit=20873a491a0d0572421b1e1b1a2d1c506cce60ce
+commit=a1a33dbe90a44e9035096099b1a969b9459b5907


cdfisher:
Not a full review but this should remove the overlay on shutdown.

Additionally, it would probably be preferable to have an "always on" option added for the infobox created by the existing "Spec Overlay" plugin, so consider contributing to the existing plugin rather than submitting an entirely new one.

iZupe:
> Not a full review but this should remove the overlay on shutdown.
> 
> Additionally, it would probably be preferable to have an "always on" option added for the infobox created by the existing "Spec Overlay" plugin, so consider contributing to the existing plugin rather than submitting an entirely new one.

I've updated the plugin to remove the overlay on shutdown. I created this plugin because the spec overlay plugin is for when soulreaper axe. This plugin can be used in all situations when the resizable inventory obstructs the special attack minimap orb and the current energy can't be seen.

cdfisher:
> > Not a full review but this should remove the overlay on shutdown.
> > Additionally, it would probably be preferable to have an "always on" option added for the infobox created by the existing "Spec Overlay" plugin, so consider contributing to the existing plugin rather than submitting an entirely new one.
> 
> I've updated the plugin to remove the overlay on shutdown. I created this plugin because the spec overlay plugin is for when soulreaper axe. This plugin can be used in all situations when the resizable inventory obstructs the special attack minimap orb and the current energy can't be seen.

I understand what the Spec Overlay plugin currently does. I was suggesting that you could contribute to that plugin to add an option to always show the infobox, regardless of what the player's equipped items are.

iZupe:
> > > Not a full review but this should remove the overlay on shutdown.
> > > Additionally, it would probably be preferable to have an "always on" option added for the infobox created by the existing "Spec Overlay" plugin, so consider contributing to the existing plugin rather than submitting an entirely new one.
> > 
> > 
> > I've updated the plugin to remove the overlay on shutdown. I created this plugin because the spec overlay plugin is for when soulreaper axe. This plugin can be used in all situations when the resizable inventory obstructs the special attack minimap orb and the current energy can't be seen.
> 
> I understand what the Spec Overlay plugin currently does. I was suggesting that you could contribute to that plugin to add an option to always show the infobox, regardless of what the player's equipped items are.

I see. I do not know how to do that since my coding abilities are almost nonexistent.

LlemonDuck:
https://runelite.net/plugin-hub/show/soulreaper-axe-qol ?

LlemonDuck:
Otherwise, you would go to https://github.com/OmbudRov/Spec-Overlay, clone it (or edit via github web ui), add a new config option to show regardless of weapon, and you'd add it to this if https://github.com/OmbudRov/Spec-Overlay/blob/master/src/main/java/com/specoverlay/SpecOverlayPlugin.java#L67

Then you just make a pull request to the Spec-Overlay repository in the github ui, and they would hopefully accept it and submit it as a change to us.

I can't imagine we need a _third_ sra spec energy plugin.

iZupe:
> Otherwise, you would go to https://github.com/OmbudRov/Spec-Overlay, clone it (or edit via github web ui), add a new config option to show regardless of weapon, and you'd add it to this if https://github.com/OmbudRov/Spec-Overlay/blob/master/src/main/java/com/specoverlay/SpecOverlayPlugin.java#L67
> 
> Then you just make a pull request to the Spec-Overlay repository in the github ui, and they would hopefully accept it and submit it as a change to us.
> 
> I can't imagine we need a _third_ sra spec energy plugin.

I appreciate the help, but this isn't for the SRA spec. I play in resizable in the default game window size and the spec minimap orb is covered by the inventory. You cannot see the number for the current spec amount. I wanted an infobox that shows the current spec energy so I created one because it didn't exist on the plugin hub. I'm not a coder. I don't know what I'm doing and it's a miracle I got it to work after spending hours putting this together. I was thrilled to get this working and would greatly appreciate it if this was approved since I do not know how to change another plugin.

PR #7571, opened 2025-03-03
Raids death indicator party bugfix

vikke1234:
Adds a hp overlay for debug purposes and fixes party.


tylerwgrass (approved):

iProdigy (approved):
would recommend `.mapToInt(Player::getCombatLevel).max().getAsInt()` at https://github.com/vikke1234/raids-death-indicator/blob/ebd0f054f7a35409d761f062136deac576832b38/src/main/java/com/example/raids/Cox.java#L72

PR #7578, opened 2025-03-04
Create prop-hunt-2

willrobaggins:
The original Prop hunt plugin no longer works, so I have pulled Idyl's code and updated it to retrieve player info from a new server location I am hosting.


Felanbird:
https://github.com/runelite/runelite/wiki/Plugin-takeover-policy

PR #7589, opened 2025-03-06
Update prop-hunt

willrobaggins:
Updating plugin file to contain the new hash and location of the repository.

Idyl has transferred ownership to me, the only change been committed to the plugin is to route the traffic to a different server as the previous was taken offline.

### DIFF
--- a/plugins/prop-hunt
+++ b/plugins/prop-hunt
@@ -1,4 +1,4 @@
-repository=https://github.com/IdylRS/prop-hunt.git
+repository=https://github.com/willrobaggins/prop-hunt.git
 commit=40d757eec5c7214ba3faa7a50b4b566766edc280
 warning=This plugin sends player data, including usernames, plugin configuration, and IP address to a third-party server for communication between clients.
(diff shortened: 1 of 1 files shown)


plugins/prop-hunt
```diff
@@ -1,3 +1,4 @@
-repository=https://github.com/IdylRS/prop-hunt.git
-commit=767fdfccd180ec0255766dea3e0de5ab9f8943b1
+repository=https://github.com/willrobaggins/prop-hunt.git
+commit=40d757eec5c7214ba3faa7a50b4b566766edc280
 warning=This plugin sends player data, including usernames, plugin configuration, and IP address to a third-party server for communication between clients.
+authors=IdylRS, willrobaggins
```
Felanbird, line 4: An authors field isn't necessary for a repo transfer, unless you'd like Idyl to be able to update the plugin on your behalf

willrobaggins: Understood, I will remove it.

PR #7596, opened 2025-03-07
Add Missed Clues plugin

lalochazia:
Adds a small fun plugin that, upon receiving a "sneaking suspicion" message, simulates a clue of that tier. This then displays the items generated in an "incinerator clue scroll" overlay, alongside the items values.


iProdigy:
```
missed-clues: plugin uses terminally deprecated APIs:
Do not create fresh Gson instances, always @Inject the client's Gson.
You can customize it by calling .newBuilder() on it
```

LlemonDuck:
You should remove the dependency on jsoup if it's no longer being used

PR #7597, opened 2025-03-08
Create tictac7x-capslock v0.1

TicTac7x:
NEW PLUGIN THAT ALLOWS MESSAGES TO BE SENT IN CAPS LOCK. Why? Because I need to yell at people.

### DIFF
--- a/plugins/tictac7x-capslock
+++ b/plugins/tictac7x-capslock
@@ -1,2 +1,2 @@
 repository=https://github.com/TicTac7x/runelite-plugins.git
-commit=0ee83578a782bed2303cf6c5a07a710cf64e02fd
+commit=e01a1a8bd8e8b67b26cf9780b97e557ffc44afc9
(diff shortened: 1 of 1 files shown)


qt31415926535-femboy:
will I need this plugin if I want to hear other people shouting in all caps as well or does it trick the server with some Alt code like it used to be possible with?

LlemonDuck:
Based on the implementation, you'll need the plugin. I think it also will turn all one-word messages into all caps lol

TicTac7x:
> Based on the implementation, you'll need the plugin. I think it also will turn all one-word messages into all caps lol

One word is not caps locked, since it checks that the worda array is more than 1

TicTac7x:
> will I need this plugin if I want to hear other people shouting in all caps as well or does it trick the server with some Alt code like it used to be possible with?

Yes

LlemonDuck:
remove the empty config

PR #7601, opened 2025-03-08
Update temple-osrs

44Mikael:
Added collection log sync feature based on WikiSync implementation.


iProdigy:
Please update the User-Agent on your http calls so the wiki knows where the requests are coming from

44Mikael:
Added hopefully descriptive enough User-Agent based on Wiki policy

PR #7609, opened 2025-03-09
Update missed-clues v1.1

lalochazia:
v1.1 changes:
- Changed the configuration for each tier from Yes/No to whether you want the Overlay, Chat Message, Both or Neither
- Changed the loot message to be a single line, no more filling up the chat box
- Added a "kill count" for each missed clue tier
- Fixed Issue with some items appearing above the clue (e.g. Prayer Icons, Player Indicators plugin)
- Updated Readme with explanation of full configuration and added features


pajlada (changes requested):
This change to your config will mean that your users will have to reconfigure your plugin. It's best if you migrate the users' old config to respect those that have disabled certain tiers already

See how RuneLite's IdleNotifier plugin does this here: https://github.com/runelite/runelite/blob/1.11.2.1/runelite-client/src/main/java/net/runelite/client/plugins/idlenotifier/IdleNotifierPlugin.java#L611

PR #7615, opened 2025-03-10
Exclude NPCs, Reset on NPC change, Add Overlay Config Option

geeckon:
This PR adds the following:

Toggle for displaying current hit in damage overlay Added config option to exclude specified NPCs from total damage overlay Added config option to reset total damage on opponent change

### DIFF
--- a/plugins/instant-damage-calculator
+++ b/plugins/instant-damage-calculator
@@ -1,2 +1,2 @@
 repository=https://github.com/geeckon/instant-damage-calculator.git
-commit=230a2073a123952ef8421b59a6fede42159667bf
+commit=aedddf96e85bbf8675daa888b16b3e2914c72a63


TheStonedTurtle:
Per <https://github.com/runelite/runelite/wiki/Rejected-or-Rolled-Back-Features>:

The NPC ID exclusion, while not the entirety of the functionality, could be problematic. Is there a reason why this plugin specifically needs to use IDs for this instead of basing it on the NPC ID?

RonnyTommy548:
> The NPC ID exclusion, while not the entirety of the functionality, could be problematic. Is there a reason why this plugin specifically needs to use IDs for this instead of basing it on the NPC Name?

The reason for using IDs instead of names is for the NPCs that are multi-phased with the same name (such as Zulrah, Phantom Muspah, or Verzik for example). 

The exclusion by ID allows for tracking damage done just to specific phases of a boss, such as Mage phase zulrah.
The resetting by ID allows for tracking damage done for each phase separately, such as P1-P3 verzik, or when Phantom Muspah changes forms

TheStonedTurtle:
While those are potential use cases for needing this to be done by ID, due to the limited number of examples and the potential for abuse we'd prefer if you used checkbox styled config options instead for these scenarios. Allowing users to specify any NPC ID opens the plugin up for potential abuse that we can not foresee, especially on future PvM related content.

geeckon:
@RonnyTommy548 please rework it according to runelite policies then. You can either do a checkbox for each boss that has a use case for this or combine them into one.

PR #7627, opened 2025-03-12
Delete plugins/ironclad-clan-goals

dextermb:
Remove our clan plugin while we work on sorting out hosting and plugin maintenance


raiyni:
The disabled flag should be set with a reason instead of deleting the file.

dextermb:
> The disabled flag should be set with a reason instead of deleting the file.

Thanks for raising this, I assume I should just need to add `disabled=true` to the file?

Felanbird:
[yes](https://github.com/search?q=repo%3Arunelite%2Fplugin-hub+disabled%3D&type=code)

PR #7628, opened 2025-03-12
Create current-spellbook

IceIceFuego:
Adds current-spellbook, a plugin which adds an overlay to tell you the spellbook you are currently on

image


IceIceFuego:
Thanks for the review and merge :)

PR #7634, opened 2025-03-14
Update custom-death-sound

ruttie3:
Various improvements and fixes an issue with new death animations not triggering sound. More info here: https://github.com/ruttie3/runelite-custom-death-sound/pull/2.


LlemonDuck:
changing the config group will reset all your users' settings; you should probably not do this

ruttie3:
The plugin's settings have changed, files are now loaded from a directory inside the .runelite directory so resetting those settings seems fine to me. Their sound file won't work after this change anyway.

LlemonDuck:
In that case it would be good to add some sort of notice to people who already had the plugin installed and probably won't notice that it has reset. You could check for the old config group/key values being set and send a chat message on login until they play a sound successfully or something similar.

LlemonDuck:

PR #7636, opened 2025-03-14
Introduce Valiance Clan plugin

Sicryption:
The Valiance clan has been needing some serious automation help for clan events. This is the beginning of that process. This plugin currently supports:
```
- Submitting item drops to Valiance Server (configurable)
- Collection log data to Valiance Server
```

Since the collection log plugins existing (WikiSync / Temple), don't support quantity, this was a good time for us to get started. Dink was close to what we wanted, but ultimately we want more control. Submitting personal bests (ensuring its a team of all irons) is one of our near-term goals which other plugins do not support. Big part of this plugin for the clan is to have it configured from the get-go, no need to mess with any settings. Install and clan members can instantly partake in clan events.

First plugin submission for the plugin hub, so please judge & correct!


coopermor:
Take a look at how plugins should be submitted to the hub. https://github.com/runelite/plugin-hub/blob/master/README.md#submitting-a-plugin

PR #7640, opened 2025-03-14
Update local-consciousness to 2.1

YeahImRose:
- Fixes bug where displayed item name does not change in some situations
- Fixes bug where nav button does not have icon depending on image data load speed


YeahImRose:
Opened from wrong branch

PR #7652, opened 2025-03-16
Update unmovable to v1.1.1

fireflowerr:
* Add strategy config to control whether walking should be filtered or deprioritized by swapping with cancel
* Bump junit to 5 and add tests
* Fix a bug where filtering would not work for more than 2 items

### DIFF
--- a/plugins/unmovable
+++ b/plugins/unmovable
@@ -1,2 +1,2 @@
 repository=https://github.com/fireflowerr/unmovable.git
-commit=4968db9e7c22b648714c8d2e41275dc14a5bfa7c
+commit=618fb421e37bc700df5f36fc678d7b748a9773b8


pajlada:
You can inject your config instead, that way you don't have to set config in the provides function.

Similarly, you can `@Inject` the client directly instead of using the inject the set function

See how those two things are done in the template plugin https://github.com/runelite/plugin-hub/blob/master/templateplugin/src/main/java/_(package_path)_/_(plugin_prefix)_Plugin.java#L21-L25

Not blockers, but maybe good for next time you make changes

pajlada:
All of your tests are failing, is that something you want to look at before we merge this in?

fireflowerr:
> You can inject your config instead, that way you don't have to set config in the provides function.
> 
> Similarly, you can `@Inject` the client directly instead of using the inject the set function
> 
> See how those two things are done in the template plugin https://github.com/runelite/plugin-hub/blob/master/templateplugin/src/main/java/_(package_path)_/_(plugin_prefix)_Plugin.java#L21-L25
> 
> Not blockers, but maybe good for next time you make changes

I used setters over field injection so that I have the ability to mock the dependency in unit tests.

I see no test failures locally running `./gradlew :test`. I will try to reproduce the failure. Could you provide any information on the test failure so I can reproduce it?

pajlada:
nvm all good, it was me doing things - sorry for the confusion @fireflowerr 

and all good regarding injection, just wanted to double check

PR #7656, opened 2025-03-16
Update author list for fixed-mode-hide-chat plugin

supyosup:
Hi @deathbeam, the last thing I need from you is that you approve the addition of myself as an author on this plugin. Just a comment with an OK is good.

Thank you.


deathbeam:
OK

PR #7663, opened 2025-03-18
Create chance-man

ChunkyAtlas:
Chance Man is a plugin that locks tradeable items until they are unlocked with a random roll. The plugin tracks rolled and unlocked items per player and saves progress using JSON files. It also includes a roll animation and a simple user interface panel for viewing rolled and unlocked items

### DIFF
--- a/plugins/chance-man
+++ b/plugins/chance-man
@@ -1,2 +1,2 @@
 repository=https://github.com/ChunkyAtlas/chance-man.git
-commit=fab560b4bf9925e3ac73221c8fb4a29674004091
+commit=8743d002e7ee6783bbc0b9c483e1c5712abcdf6a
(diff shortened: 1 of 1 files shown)


LlemonDuck:
Use `RuneLite.RUNELITE_DIR` https://github.com/ChunkyAtlas/chance-man/blob/2f4743dd2b039a322971593c68c2142e313efcad/src/main/java/com/chanceman/UnlockedItemsManager.java#L41-L43

LlemonDuck:
`loadUnlockedItems` should also use your executor service, and RolledItemsManager should also be on a thread. They can share the same executor service if you move it somewhere common

non-blocking:
ChatMessageManager#queueMessage is already thread-safe and you don't need to invokeLater for it

PR #7664, opened 2025-03-18
Add Quick Search plugin

hirzalla:
# Quick Search Plugin

Adds chat commands for quick searching of OSRS-related platforms:
- OSRS Wiki (::wiki, ::osrs)
- YouTube (::yt, ::youtube) 
- Twitch (::twitch, ::ttv)
- Kick (::kick)

Each platform can be toggled on/off.

Examples:
- `::wiki abyssal whip`
- `::yt zulrah guide`
- `::ttv woox`
- `::kick odablock`

### DIFF
--- a/plugins/quick-search
+++ b/plugins/quick-search
@@ -1,0 +1,2 @@
+repository=https://github.com/hirzalla/quick-search.git
+commit=7b22611b9008a693ec4d5aed4b3caab2ed0337da
--- a/plugins/quick-search
+++ b/plugins/quick-search
@@ -1,2 +1,2 @@
 repository=https://github.com/hirzalla/quick-search.git
-commit=7fe7b771f7fb7a8ab5e92dd4efd27ac5a9783217
+commit=ddd20911a28c39efb19944811a74732fcf507813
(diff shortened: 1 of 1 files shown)


iProdigy:
Please use the `CommandExecuted` event instead of `ChatMessage` (i.e., `::command` instead of `!command`)

hirzalla:
@iProdigy will do. Out of curiosity, what's the reason behind that? Does it perform better or are `!` commands reserved for specific things?

iProdigy:
my last nit: could you update these messages to explicitly indicate they are coming from the `Quick Search` plugin
https://github.com/hirzalla/quick-search/blob/ddd20911a28c39efb19944811a74732fcf507813/src/main/java/com/quicksearch/QuickSearchPlugin.java#L65
https://github.com/hirzalla/quick-search/blob/ddd20911a28c39efb19944811a74732fcf507813/src/main/java/com/quicksearch/QuickSearchPlugin.java#L70

hirzalla:
@iProdigy yep, good call. Also added an additional config for suppressing messages to avoid unwanted noise

iProdigy (dismissed):

LlemonDuck:
::wiki is already a vanilla command handled by script 3299, you shouldn't override it

hirzalla:
@LlemonDuck updated, thanks for spotting

PR #7667, opened 2025-03-18
Create stats-lookup

MKbrun:
Plugin which allows the user to look up the stats on a monsters gathered from osrswiki

### DIFF
--- a/plugins/stats-lookup
+++ b/plugins/stats-lookup
@@ -1,0 +1,2 @@
+repository=https://github.com/MKbrun/stats-lookup.git
+commit=47d6eb6aaea7e7555fd80c84571d0884095214e0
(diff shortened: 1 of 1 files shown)


iProdigy:
Please use the version of okhttp that is already bundled by the runelite client

Nightfirecat:
This really should not query the wiki every time a lookup is performed; instead, it would be much better to scrape the wiki for this data periodically and have the plugin fetch that scraped data.
(separately, I don't really understand the point of this plugin when the wiki lookup button exists)

coopermor:
The wiki dps calc already scrapes the wiki to create a monsters.json which is kept up to date. If you did want the plugin to exist you can probably just parse the json data that is already being scraped. https://raw.githubusercontent.com/weirdgloop/osrs-dps-calc/refs/heads/main/cdn/json/monsters.json

MKbrun:
> The wiki dps calc already scrapes the wiki to create a monsters.json which is kept up to date. If you did want the plugin to exist you can probably just parse the json data that is already being scraped. https://raw.githubusercontent.com/weirdgloop/osrs-dps-calc/refs/heads/main/cdn/json/monsters.json

Thanks for this. Didn't know, but will try to use that instead

LlemonDuck:

PR #7671, opened 2025-03-19
Updated empty-hands-indicator

ciaagent:
Added the ability to exclude players from being highlighted.


iProdigy:
please cache `playersExcludedSet` rather than recreating it on each render (also I think you meant to do `new HashSet` instead of `new ArrayList`. lastly consider applying `toLowerCase`)

ciaagent:
Thank you for the feedback. Updated

iProdigy:
* you've changed the config group - all users will be reset to the default config. is this intentional?
* micro-optimization: your `ConfigChanged` handler can check whether `excludedPlayers` changed (rather than a different config setting)
* micro-optimization: your `String#trim` calls are unnecessary

ciaagent:
I see. No, resetting to default wasn't intentional. 
I'm more than happy to make the micro-optimizations suggested.
Thank you!

ciaagent:
Done. Thank you!

iProdigy (approved):

pajlada (approved):

PR #7676, opened 2025-03-20
Updated empty-hands-indicator plugin

ciaagent:
Added the ability for the user to get notified when their hands are empty.
Fixed local player being outlined if empty handed and a part of a clan.


tylerwgrass:
Please remove this overlay on shutdown: https://github.com/ciaagent/Empty-Hands-Indicator/blob/master/src/main/java/emptyhandsindicator/EmptyHandsIndicatorPlugin.java#L68

ciaagent:
Good catch! Done.

tylerwgrass (approved):

PR #7680, opened 2025-03-20
Added bingoscape-companion

kaffeed:
Participate in bingo events with your clan or friends. 
View custom bingo boards, complete tasks, and submit screenshots as evidence to bingoscape (https://next.bingoscape.org)
Track your team progress and submit screenshots directly from runelite.

### DIFF
--- a/plugins/bingoscape-companion
+++ b/plugins/bingoscape-companion
@@ -1,2 +1,2 @@
 repository=https://github.com/bingoscape/bingoscape-companion.git
-commit=1d4105c93836d39ac3f3cc3a036d496cb091f2f8
+commit=8852f3d6c527c0c74773b058e4bcabca13903287
--- a/plugins/bingoscape-companion
+++ b/plugins/bingoscape-companion
@@ -1,2 +1,2 @@
 repository=https://github.com/bingoscape/bingoscape-companion.git
-commit=3311b170a8e41d1e5a6d171ba7c15db3d0f94279
+commit=771547f5ff7aecd0f2f37ec11f740d894b9e64b4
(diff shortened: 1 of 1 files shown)


raiyni:
There are already 4 bingo related plugins, do they not fit your needs? Is this something that can be contributed to an existing plugin?

kaffeed:
So, the closest i've seen is the "Better Bingo" Plugin although that kinda fits a different use case.. I've developed a platform to host/manage bingos, e.g. 
image
which can be used as a standalone, but the plugin I've developed would be a huge QoL for the participants/users of bingoscape.
Also, I plan to keep bingoscape free/[open source](https://github.com/bingoscape/bingoscape) forever, should this no longer be the case, I totally understand not hosting the plugin on here anymore.
Better Bingo for example only tracks dropped items, while bingoscape is a bit more free in that regard as you can just take a screenshot from the client and directly submit it, regardless of a drop

plugins/bingoscape-companion
```diff
@@ -0,0 +1,3 @@
+repository=https://github.com/bingoscape/bingoscape-companion.git
+commit=7fd25ead4b469b6aed24413438c8baf5cfa56b7e
+authors=kaffeed
```
LlemonDuck, line 3: You'll need a warning here since your domain is filled by default:
```suggestion
authors=kaffeed
warning=This plugin submits <...> and your IP to a server not controlled or verified by the RuneLite developers.
```

LlemonDuck (changes requested):

kaffeed:
Fixed the warning :)

PR #7683, opened 2025-03-20
Death Dotter Plugin

dmgear:
This is a simple plugin that compares the radii of two players, and if the distance between two players falls below the threshold that is the sum of their radii, we hide the local client player similar to entity hider.

The purpose of this plugin is to perform the same function as entity hider + player outline in pvp, but without having to play with an invisible character.

### DIFF
--- a/plugins/deathdotter
+++ b/plugins/deathdotter
@@ -1,2 +1,2 @@
 repository=https://github.com/dmgear/deathdotterplugin.git
-commit=404148cfe54656a921f42c0d36ad1e0ddc87bb51
+commit=bc4cf26b98bdf4c6e3c98d9b91c94d55fa97590d
--- a/plugins/deathdotter
+++ b/plugins/deathdotter
@@ -1,0 +1,2 @@
+repository=https://github.com/dmgear/deathdotterplugin.git
+commit=4e9d611ae87b972973b2c82cfea5d4439cf7ec00
--- a/plugins/deathdotter
+++ b/plugins/deathdotter
@@ -1,2 +1,2 @@
 repository=https://github.com/dmgear/deathdotterplugin.git
-commit=bc4cf26b98bdf4c6e3c98d9b91c94d55fa97590d
+commit=ef7b4ba604f28944fb980dbb11e0a119258be4f7
--- a/plugins/deathdotter
+++ b/plugins/deathdotter
@@ -1,2 +1,2 @@
 repository=https://github.com/dmgear/deathdotterplugin.git
-commit=4e9d611ae87b972973b2c82cfea5d4439cf7ec00
+commit=f86baa5dbbe2bf67f4593b94e738bd1f3a5ab437
--- a/plugins/deathdotter
+++ b/plugins/deathdotter
@@ -1,2 +1,2 @@
 repository=https://github.com/dmgear/deathdotterplugin.git
-commit=f86baa5dbbe2bf67f4593b94e738bd1f3a5ab437
+commit=404148cfe54656a921f42c0d36ad1e0ddc87bb51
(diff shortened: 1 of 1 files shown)


tylerwgrass:
Please do not close/reopen your PR. If there are build issues, you can make a commit to fix and push it to this PR branch.

dmgear:
> Please do not close/reopen your PR. If there are build issues, you can make a commit to fix and push it to this PR branch.

Understood, my apologies. Thank you.

cdfisher:
Not a thorough review, and both of these are fairly minor but:
* You can just remove `enabledByDefault = true` entirely since that's the default behavior
* It doesn't look like your config key actually gets used anywhere so you can probably just remove the config altogether.

dmgear:
> Not a thorough review, and both of these are fairly minor but:
> 
> * You can just remove `enabledByDefault = true` entirely since that's the default behavior
> * It doesn't look like your config key actually gets used anywhere so you can probably just remove the config altogether.

Thank you for your feedback. I will remove `enabledByDefault = true`. However, the config key is being used in the updateConfig() method in my plugin class.

cdfisher:
> However, the config key is being used in the updateConfig() method in my plugin class.

The only thing you're doing with `updateConfig()` is updating the config value which you don't actually use anywhere. Your config, `updateConfig()`, `onConfigChanged()`, and all the references to them could all be removed while having no effect on your plugin because none of them are used in any meaningful way.

dmgear:
> > Not a thorough review, and both of these are fairly minor but:
> > 
> > * You can just remove `enabledByDefault = true` entirely since that's the default behavior
> > * It doesn't look like your config key actually gets used anywhere so you can probably just remove the config altogether.
> 
> Thank you for your feedback. I will remove `enabledByDefault = true`. However, the config key is being used in the updateConfig() method in my plugin class.

Ah! I see what you mean now! I am new to runelite development so please bear with me. To my understanding, to register the plugin with runelite to get it to show up in the client I must use the config. I notice now that if I remove the config and its associated methods my plugin no longer shows up in the runelite client. What should I do instead?

dmgear:
> > However, the config key is being used in the updateConfig() method in my plugin class.
> 
> The only thing you're doing with `updateConfig()` is updating the config value which you don't actually use anywhere. Your config, `updateConfig()`, `onConfigChanged()`, and all the references to them could all be removed while having no effect on your plugin because none of them are used in any meaningful way.

I understand. I was misunderstanding the purpose of the config class. I have removed my config and also refactored my plugin.

tylerwgrass:
You have added a bunch of files that shouldn't be here.
You should only be adding your `plugins/deathdotter` file.

dmgear:
> You have added a bunch of files that shouldn't be here. You should only be adding your `plugins/deathdotter` file.

Unsure what happened, I didn't make changes in any of those files. I unstaged those changes by resetting my commit. It asked me to do a pull before I pushed again, and I pushed just now. But I think it committed a bunch of extra files again.

dmgear:
On this commit: "https://github.com/runelite/plugin-hub/pull/7683/commits/b29e66e9ab32e3d6c798ca31351269e363411795"

You can see I just pushed my plugin. But then here: "https://github.com/runelite/plugin-hub/pull/7683/commits/086cacaa367b5997a0bb69fc844e1019a51d651c"

There's a bunch of extra files from the merge. Please advise.

tylerwgrass:
Delete the files and push a new commit instead of resetting.

dmgear:
Done.

Felanbird:
(it's not done)

dmgear:
Reset to previous commit. I think this is good now?

PR #7684, opened 2025-03-20
Event calendar plugin

k7lw:
Attempting to add an Event Calendar plugin to integrate with discord events

### DIFF
--- a/plugins/Event-Calendar
+++ b/plugins/Event-Calendar
@@ -1,0 +1,2 @@
+repository=https://github.com/k7lw/Event-Calendar.git
+commit=5ae069c77b130e922fb439d3249dd04e2f5138bc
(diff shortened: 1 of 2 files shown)


LlemonDuck:
use an injected okhttpclient for http requests

LlemonDuck:
Check the scanner feedback: don't use System.err, use a logger instead

LlemonDuck:
since you're using a third party service, add a warning to your plugin hub manifest e.g. https://github.com/runelite/plugin-hub/blob/7113b4ffc58fc7394c16eb50e42e019e62da0a12/plugins/ntfyer#L3

non blocking: you should look into gson for serializing/deserializing json values instead of manually parsing the string

LlemonDuck:

PR #7689, opened 2025-03-22
Update flipper v1.0.1

UmaLPZ:
Updated and remodeled Flipper plugin. Track your buys, sells, flips, and in-progress offers.


LlemonDuck:
FYI you don't need this, we don't log at debug level when you're not in --debug mode (and it's also set to true despite your comment :P). You're free to log.debug to your heart's content, just `@Slf4j` each class you want it in

PR #7712, opened 2025-03-25
Add npc-max-hit

hirzalla:
Displays an overlay/infobox with maximum hit values for NPCs during combat. Pulls data from the Wiki and parses it.

### DIFF
--- a/plugins/npc-max-hit
+++ b/plugins/npc-max-hit
@@ -1,2 +1,2 @@
 repository=https://github.com/hirzalla/npc-max-hit.git
-commit=3d5def575e51b1267dc23e1f4e3a703ffac9de1a
+commit=a65b8b167d1f77244c7e06c1f66fd3ddce9e4453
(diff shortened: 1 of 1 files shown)


hirzalla:
should be good for a review ^ 

a _potential_ improvement in the future might be to use weirdgloop's osrs-dps-calc [scraped monsters data](https://raw.githubusercontent.com/weirdgloop/osrs-dps-calc/refs/heads/main/cdn/json/monsters.json) instead of asking wiki on demand, but as of now that data doesn't properly include the only data I need for this plugin pretty much (I do have a [PR](https://github.com/weirdgloop/osrs-dps-calc/pull/629) to sort that out) and has more than what the plugin needs atm - it'd also be a another dependency to keep in mind, but could be worth it at some point if it turns out to be that way

currently I do have caching going on by npc ID for requests fired so _shouldn't_ be too bad I don't think, don't expect too many happening quickly. getting this out to the wild should help get some user feedback and possibly other improvements/changes

PR #7725, opened 2025-03-27
Create better-xp-tracker

chiefbean:
XP tracker plugin that brings back the kills to next level/goal that were removed from official runelite tracker plugin


tylerwgrass:
As mentioned, this should be a contribution to the core xp tracker plugin instead of an entirely new hub plugin

chiefbean:
🫡

PR #7726, opened 2025-03-28
Update Creator's Kit to v1.5.20

ScreteMonge:
Bugfix: Fix concurrent access of DataFinder lists during async population & Make DataFinder callbacks thread-agnostic


pajlada (commented):
[ScreteMonge/creators-kit@`806e9a3e8d45f5675904b2fa95b126f02beb38f5...ScreteMonge:1e7c07d83bf7cb9b336b53eb02b2e7aa8eee4389`#diff-8819ad27c3](https://github.com/ScreteMonge/creators-kit/compare/806e9a3e8d45f5675904b2fa95b126f02beb38f5...ScreteMonge:1e7c07d83bf7cb9b336b53eb02b2e7aa8eee4389#diff-8819ad27c388c11bae5549b86567d5f1af263a0a3e1ffb7981adaba40a123734R173) The `onResponse` callbacks that were changed in `DataFinder.java` have had their successful/body checks changed. Previously you would only load the data if the response was successful AND the response had a non-null body. Now you try to load data from the body if the response was successful OR the response had a non-null body. This seems unintended and could lead to a NullPointerException if a successful response has no body (which would be unlikely)

Not blocking, but the concurrency access of `loadCallbacks` and `loadState` is unsound. If it improves the way things work that's fine, but know you might run into sneaky bugs in the future. (unless I misunderstand java threading/data structures, in which case feel free to ignore me)

PR #7750, opened 2025-04-01
Add Lucky Clan Bingo plugin

EwwItsMike:
A plugin intended for usage in Clan events.

When players receive a drop, it is checked against a list of commonly used items for Bingo-type clan events. 
Plugin sends a screenshot and drop information to a Discord webhook.

The Discord webhook link is the ONLY adjustable setting for this plugin, meaning the items list cannot be altered (at least not easily, some tech-savvy people could find out they can alter it within the .jar file).
This is intentional, and the main point of the plugin, to prevent cheating in certain types of clan events (where people not submitting a drop because it doesn't benefit them is prohibited, for example).

The plugin is originally intended for the Lucky clan, but I hope other clans/communities will find this plugin useful for their events too.

### DIFF
--- a/plugins/lucky-clan-bingo
+++ b/plugins/lucky-clan-bingo
@@ -1,2 +1,2 @@
 repository=https://github.com/EwwItsMike/LuckyClanBingo.git
-commit=561c41e72f5eb9f2c4b7ce86551bf6f891cd916f
+commit=0c5b30e3f8865b6da889b188133d74e4e63d5148
--- a/plugins/lucky-clan-bingo
+++ b/plugins/lucky-clan-bingo
@@ -1,2 +1,2 @@
 repository=https://github.com/EwwItsMike/LuckyClanBingo.git
-commit=d4c0cc2f49d9e0df08d70371dba0ca99e30e4179
+commit=13e104894aa00a83dbaff84a806cb627a91772d6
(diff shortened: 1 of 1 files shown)


tylerwgrass:
There are a few existing bingo plugins already, is there a particular reason to add this over the existing ones?

This plugin also seems it would be more restrictive for general use as everyone would need to use your bingo list, which may or may not be what different groups want and adding/removing items would affect everyone and require updating your plugin. 

It is also not obvious to users what those items are as 1) it lives in your repo and 2) you provide no indication what/where the list is

EwwItsMike:
As far as I can find, there are currently 2 specific "Bingo" plugins in the plugin hub:
One limited to a board of 25 items, and one that requires event organisers to set up a backend server.

This plugin functions roughly the same as a Discord Loot Logger or Dink plugin, but it has the ability to select which items to send removed as an anti-cheating measure.
The fact users cannot easily find the list of items is a fair point, I will update the README to provide useful information for both event organisers and end-users.

Also @ the restrictiveness part, I've tried to include as many items as I could think of, but it's very possible I may have missed some. I will also link to a way of requesting item additions in the README.

pajlada:
> I've added the README explaining how to use and set up the plugin, and given free-to-use Google Drive resources for other communities to use for their events as well. Also explained how the event I'm planning to hold in my clan is going to go, which I hope provides enough context as to why the currently existing Loot logging/Bingo plugins are not suited for it.
> 
> I also keep seeing one of the Plugin Hub Checks failing, but I cannot see the details of it. It only states a maintainer has to review it and they might request changed, are there any changes I need to make?

It just means it needs extra review + approval - no extra action to be taken on your end.

LlemonDuck:
Don't use printStackTrace, just append it to your log statement `log.error("message", e)` https://github.com/EwwItsMike/LuckyClanBingo/blob/d4c0cc2f49d9e0df08d70371dba0ca99e30e4179/src/main/java/com/LuckyClan/Bingo/LuckyClanBingoPlugin.java#L104

PR #7756, opened 2025-04-02
Create party-emotes

Headline:
Hey all - this is a simple plugin that draws a limited set of emotes (free & use is allowed openmoji.org) above players heads via the party plugin. Looking for ways to be a bit more expressive in raids :)

If there's any changes you'd like to see let me know, as far as I'm aware this doesn't violate any rules and I followed instructions pretty closely. I'd appreciate another set of eyes on the plugin, though, before merge.

Thanks!

### DIFF
--- a/plugins/party-emotes
+++ b/plugins/party-emotes
@@ -1,2 +1,2 @@
 repository=https://github.com/Headline/PartyEmotes.git
-commit=623c0c80e2f9f6bb319fe5b395bd20b3350a8338
+commit=0e7dd12a0a7e10a20de67618f54a920e05eca007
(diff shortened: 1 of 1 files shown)


tylerwgrass:
Not a full review, but you should add throttling so users can't spam events.
You also probably want to change the `PluginDescriptor` name to `Party Emotes` to make the intent more clear as right now it shows up as `Emotes`

Headline:
> Not a full review, but you should add throttling so users can't spam events. You also probably want to change the `PluginDescriptor` name to `Party Emotes` to make the intent more clear as right now it shows up as `Emotes`

Hey- Yeah, missed that. They get sent as fast as the user clicked like the party ping, and on the receive side the map entry gets replaced so there's not too much of concern for EmoteEvents getting backed up. 

I'd still like to keep it relatively responsive for users who want to keep the emote overhead, once per tick or two seem alright to you? @tylerwgrass

tylerwgrass:
Your messages are still being sent over the websocket whether or not you are storing them in the map when received.

I do not have detailed info on the limits/restrictions/etc for the service myself, so someone else may provide a more accurate review, but in my opinion I don't think these events need to be sent more than every 500ms.

PR #7760, opened 2025-04-04
Add Cannon Damage plugin metadata

DunkingOreos:
This plugin is specifically made to get correct cannonball statistics for use when calculating untrimmed slayer.

The plugin will start when it detects the placement of a cannon, including the shattered relics ornament kit cannon, and will stop once the cannon is picked up. The overlay will remain on screen for 60 seconds after the cannon is picked up to allow time for a screenshot or the user to write down the numbers.

ONLY damage from the cannon is supposed tracked, though if you are ranging as your main attack style when using the cannon it will count that xp as well. This is because you should NOT be ranging when going for untrimmed slayer WHILE using a cannon, magic will suffice.

### DIFF
--- a/plugins/cannon-damage
+++ b/plugins/cannon-damage
@@ -1,3 +1,3 @@
 repository=https://github.com/DunkingOreos/Cannon-Damage.git
-commit=e17b91e6c73235a2d37e36c7bc5f836caa87b8f0
+commit=1668e512ae4c859a9ee2e5182c70862a1de79b3b
 authors=DunkingOreos,OzanKurt7,cultbaus
--- a/plugins/cannon-damage
+++ b/plugins/cannon-damage
@@ -1,0 +1,3 @@
+repository=https://github.com/DunkingOreos/Cannon-Damage.git
+commit=41546aacc30b04bffc75c3e45a58b04af8bfe36a
+authors=DunkingOreos,OzanKurt7,cultbaus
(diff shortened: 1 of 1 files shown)


tylerwgrass:
You should remove stale entries from `trackedProjectiles` at some point as you are only ever adding to it, and only clear it on shutdown meaning this may cause issues in longer sessions.

DunkingOreos:
The `trackedProjectiles ` should now remove after getting to 5. Also, the activation of other people's cannons is an interesting assessment that I didn't take into consideration. If the first check after cannon placed is if the player gets ranged experience would that make it unique enough to only trigger on your own cannon?

LlemonDuck:
You'd end up with a potential edge case if two cannons are fired at the same time. You could track where the cannon was placed and then check where the projectile source is, but it's probably impossible to have a perfectly accurate plugin for all cases here.

DunkingOreos:
The intention of the plugin (mainly for untrimmed slayer data calculations) is for use in cannon-able slayer areas which _should_ help ensure there is only 1 cannon placed. Although I don't foresee many instances where 2 cannons are down in a close relation during slayer, I do acknowledge your potential edge case issue, and would agree that it's most likely improbable to have a perfectly accurate plugin for all cases.

I'm okay with you doing a review as is and certainly could revisit/update in future if feedback suggests different than my foresight.

PR #7765, opened 2025-04-05
Create homeassistant

db1996:
This is a home assistant integration to have farming patch information available in [home assistant](https://www.home-assistant.io/).

This allows you to send notifications to your phone if you're not playing the game to notify you of patches being ready etc.

### DIFF
--- a/plugins/homeassistant
+++ b/plugins/homeassistant
@@ -1,2 +1,2 @@
 repository=https://github.com/db1996/homeassistant.git
-commit=e6f0328c71fbf8185d6937d4909a507a00e54620
+commit=72bef7cc562315dc18b75c6dd5397e37e27a514b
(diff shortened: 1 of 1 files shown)


LlemonDuck:
Use an injected OkHttpClient instead of manually handling the requests

LlemonDuck:
You can lean on the runelite farming plugin instead of cloning it all yourself by just reading config keys. Were you trying to do anything that wouldn't be covered by that?

db1996:
@LlemonDuck I really tried to convert this to use the config. But it seems to me like the time tracking plugin does not save this information in the config. And uses it's internal classes to calculate the next completion time. 

What they save in the config is the starting time of patches. Which means I would still need to copy all of the logic for calculating completion times. Which are in the same classes. 

When looking at other plugins that also use the time tracking from the same plugin, they also copy the logic directly

PR #7771, opened 2025-04-06
DropTracker - Adventure Log support

joelhalen:
Added support to the plugin for tracking Adventure Log entries when a player opens the log in their own POH, allowing us to grab all existing stored PBs for use in our clans' "hall of fames" instantly, without requiring a re-submission of boss kills.

Alongside this, we're also able to parse players' acquired pets and send this data with the gathered PBs.

### DIFF
--- a/plugins/droptracker
+++ b/plugins/droptracker
@@ -1,3 +1,3 @@
 repository=https://github.com/joelhalen/droptracker-plugin.git
-commit=98d36df63e8c48dd981c6837b03c15d815e49c02
+commit=5afb3776f77b242dd704725e377afad9569fbd88
 authors=joelhalen,OSRSKoeppy
--- a/plugins/droptracker
+++ b/plugins/droptracker
@@ -1,3 +1,3 @@
 repository=https://github.com/joelhalen/droptracker-plugin.git
-commit=a991f7c1a96e02cc2929864c3e0ae4ff50e8547b
+commit=9fbddd158934fa5e8d91c6e0db7cc1e2db1d3673
 authors=joelhalen,OSRSKoeppy
--- a/plugins/droptracker
+++ b/plugins/droptracker
@@ -1,3 +1,3 @@
 repository=https://github.com/joelhalen/droptracker-plugin.git
-commit=5afb3776f77b242dd704725e377afad9569fbd88
+commit=01f4f5c534222cdca83bbfeac4b7c2f3c7593183
 authors=joelhalen,OSRSKoeppy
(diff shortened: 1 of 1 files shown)


pajlada (changes requested):
You shouldn't set core plugin config fields like you're doing.

You're relying on the Chat Commands plugin elsewhere, which already does almost all of the work from your WidgetEvent file, could you not just read those config fields a tick after the chat commands plugin has saved its data?

LlemonDuck:
don't do this https://github.com/joelhalen/droptracker-plugin/compare/3e3d3b6ec6e3b8b5b9aaf330be0c8db9f49ba36e...joelhalen:a991f7c1a96e02cc2929864c3e0ae4ff50e8547b#diff-579f6c75d7593f434a291e31e3d36e393f1545a9600f7b25835685ed5d9ab410R176

it's fine if you want to gate it behind debug mode or something (or I'd specifically recommend just using [a debug plugin](https://github.com/LlemonDuck/tombs-of-amascut/blob/main/src/test/java/launcher/debugplugins/ToaDebugPlugin.java#L35-L36) and [loading it in your test launcher](https://github.com/LlemonDuck/tombs-of-amascut/blob/main/src/test/java/launcher/TombsOfAmascutPluginTest.java#L12))

joelhalen:
> don't do this [joelhalen/droptracker-plugin@3e3d3b6...joelhalen:a991f7c1a96e02cc2929864c3e0ae4ff50e8547b#diff-579f6c75d7593f434a291e31e3d36e393f1545a9600f7b25835685ed5d9ab410R176](https://github.com/joelhalen/droptracker-plugin/compare/3e3d3b6ec6e3b8b5b9aaf330be0c8db9f49ba36e...joelhalen:a991f7c1a96e02cc2929864c3e0ae4ff50e8547b#diff-579f6c75d7593f434a291e31e3d36e393f1545a9600f7b25835685ed5d9ab410R176)
> 
> it's fine if you want to gate it behind debug mode or something (or I'd specifically recommend just using [a debug plugin](https://github.com/LlemonDuck/tombs-of-amascut/blob/main/src/test/java/launcher/debugplugins/ToaDebugPlugin.java#L35-L36) and [loading it in your test launcher](https://github.com/LlemonDuck/tombs-of-amascut/blob/main/src/test/java/launcher/TombsOfAmascutPluginTest.java#L12))

Very interesting.... not sure how I missed & left that one in there. 
Sorry about that; thanks!

PR #7775, opened 2025-04-06
Create plugin Tob utilities

NCG-RS:
# Tob Utilities

The primary goal of this plugin is to minimize the need for mid-raid configuration changes as well as add some new functionalities

### Maiden
- Chat message on entry regarding the chance of a scuff based on maidens index
- Highlighting of scuffed spawns

### Bloat
- The ability to hide self or other players while in bloat room

### Nylocas
- Aggressive highlighting

### Verzik
- The ability to hide verzik with a hotkey 
(specifically because it is something where you need to toggle entity hider when yellows spawn under verz in hmt)
- An overlay to indicate which player is holding dawnbringer staff using the player orb

### Metronome
- A visual metronome which changes tick count based on room
Note: This does not sync to the instance timer so does not indicate the timing of game mechanics any more than the visual metronome plugin. Just removes the need for mid raid configuration changes

I've tried to ensure everything is compliant, if there's anything specifically that is a concern I'm happy to make any required changes to make this happen.


LlemonDuck:
> Chat message on entry regarding the chance of a scuff based on maidens index
how does this work? I'm just not aware

NCG-RS:
> > Chat message on entry regarding the chance of a scuff based on maidens index
> 
> how does this work? I'm just not aware

Each NPC that spawns has an index between ~25000 and ~65000 which increments with each NPC spawn on the world. Nylos being "scuffed" is a bug that occurs when the Nylocas spawns have a lower index than Maiden, caused by the index exceeding 65000 mid fight and wrapping around to the beginning. This makes the nylos spawn 1 tile north of where they are supposed to.

Currently meta for handling this, is dropping a pet to check the index of it. This feature isn't necessarily as good as that (because you don't know maidens index until entering the room) but it does allow you to fast reset if desired when you see that it is going to occur

LlemonDuck:
this will leak npc instances, since ActorDeath is triggered when the NPC's health bar hits zero, but the following GameTick will re-add it as the NPC still exists for its death animation. Use onNpcSpawned and onNpcChanged instead of onGameTick to track insertions, and always additionally remove in onNpcDespawned and plugin shutdown.

https://github.com/NCG-RS/TobUtilities/blob/d920768faa0786f64c4e99330f4319ca90abedba/src/main/java/com/tobutilities/nylocas/NylocasHandler.java#L38

PR #7787, opened 2025-04-08
Added a few new functions

daanbom:
Added option for coll log sounds, max hit sounds and death sounds


daanbom:
I think it gives error cause of namechange of plugin, anything i need to do?

raiyni:
Do not change the name of the file in this pr. You can change the name in your @PluginDescriptor or runelite-plugin.properties

daanbom:
> Do not change the name of the file in this pr. You can change the name in your @PluginDescriptor or runelite-plugin.properties

I changed back the name of the file, works like this?

LlemonDuck:
fyi all 3 sounds added are the same, in case that wasn't intentional

PR #7806, opened 2025-04-10
World Map Runes 0.1.1 creation

weiss-e:
This is a very simple plugin that shows the runes required for all spellbook teleport destinations in a human-friendly readable format within the existing tooltip generated by the World Map RuneLite plugin. It's compatible with the default World Map plugin as well as all other plugins. It should be helpful to those who jump around the map a lot and are not always immediately aware of what runes are needed for a specific teleport - now you can just hover your mouse over any teleport icon on the map and see exactly what you need!

Example screenshots:
ss_1
ss_2
ss_3


weiss-e (commented):
deprecated

tylerwgrass:
This should match your plugin class path: 
https://github.com/weiss-e/world-map-runes/blob/db1e18e9b6adc55592f037c7b4a4ef336c3fbb7d/runelite-plugin.properties#L6

PR #7816, opened 2025-04-11
Create perryventory plugin

Lunitar8:
Creating perryventory plugin for Permanent Inventory Mode. Whenever an item is dropped, consumed, equipped etc., the plugin leaves behind a placeholder shadow of that item, which limits your total inventory space to 28 over the course of the account.


tylerwgrass:
>  net.runelite.pluginhub.packager.PluginBuildException: perryventory: plugin uses terminally deprecated APIs:
Do not create fresh Gson instances, always @Inject the client's Gson.
You can customize it by calling .newBuilder() on it

PR #7824, opened 2025-04-13
Collection tracker

voltmaister:
This plugin works similarly to the original Collection Log plugin, where typing !log boss in the chat fetches and displays collection log progress with item icons directly in the message.

It uses the TempleOSRS API to fetch and sync collection data and stores results locally to reduce load on TempleOSRS servers and speed up future responses.

Data is cached per player and automatically updated only when the remote data has changed.

### DIFF
--- a/plugins/collection-tracker
+++ b/plugins/collection-tracker
@@ -1,2 +1,2 @@
 repository=https://github.com/voltmaister/collection-tracker.git
-commit=5653e948d1821b1b1bf15640b52d0be89faf6694
+commit=7c3841e5172368d6c75dae165f73fdd7df362530
--- a/plugins/collection-tracker
+++ b/plugins/collection-tracker
@@ -1,2 +1,2 @@
 repository=https://github.com/voltmaister/collection-tracker.git
-commit=17a2b5093c3df7f0b957c45df32732be3a727746
+commit=347b96859a3dbef63eb21acaf2c589a595312e71
--- a/plugins/collection-tracker
+++ b/plugins/collection-tracker
@@ -1,2 +1,2 @@
 repository=https://github.com/voltmaister/collection-tracker.git
-commit=7c3841e5172368d6c75dae165f73fdd7df362530
+commit=17a2b5093c3df7f0b957c45df32732be3a727746
--- a/plugins/collection-tracker
+++ b/plugins/collection-tracker
@@ -1,3 +1,3 @@
 repository=https://github.com/voltmaister/collection-tracker.git
-commit=bca438675ca699d534c1d1a9e0c3f86035a3f8ba
+commit=0f055d58a355922b450721dbf460ede5537985ba
 warning=This plugin submits your username and IP address to a server not controlled or verified by the RuneLite developers.
--- a/plugins/collection-tracker
+++ b/plugins/collection-tracker
@@ -1,2 +1,2 @@
 repository=https://github.com/voltmaister/collection-tracker.git
-commit=347b96859a3dbef63eb21acaf2c589a595312e71
+commit=685c14a83357ae945b9b72e653500d7af1617505
--- a/plugins/collection-tracker
+++ b/plugins/collection-tracker
@@ -1,0 +1,2 @@
+repository=https://github.com/voltmaister/collection-tracker.git
+commit=5653e948d1821b1b1bf15640b52d0be89faf6694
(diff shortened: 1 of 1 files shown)


raiyni:
Initial things:

* Inject our okhttp client and use that
* switch your  `log.info` calls to `log.debug` or remove them
* add a warning about 3rd party information like `This plugin submits your username and IP address to a server not controlled or verified by the RuneLite developers.`
* You need to save your Executor service you create instead of doing `Executors.newSingleThreadExecutor().execute`
* You need to `Class.forName` your h2 instance like https://github.com/bigdiesel2m/pageless-things/compare/0992fcc77c59462a8b387a2f3868825731c81640...bigdiesel2m:352ec7cedec7b984d337c6c2e9382ceb81110e83

voltmaister:
Fixed everything mentioned in the review, not sure about the warning though as I can't test it locally.

tylerwgrass:
The warning should be in the plugin file in this pr. Example:

https://github.com/runelite/plugin-hub/blob/master/plugins/temple-osrs

voltmaister:
Thanks for the quick reply just fixed it.

raiyni:
Have you talked to the developers of the temple osrs plugin about adding this feature to their plugin?

voltmaister:
Yes, before the developing of the plugin I asked for permission and when I finished it I sent it to them to test it and see if it's ok by them. I got the green light to add it to plugin hub by them

LlemonDuck:
The RuneProfile plugin just merged a change adding a !log command, you might want to either change yours to not conflict or otherwise work something out for people with both installed

voltmaister:
Just updated from !log to !col hope now everything is ok.

LlemonDuck:
don't add an explicit dependency on gson, use the transitive dependency from runelite-client

voltmaister:
Hope this fixed any remaining issues.

LlemonDuck:
use slf4j instead of JUL for logging

voltmaister:
just pushed the changes

pajlada (changes requested):
## Requested changes
your plugin breaks in some parts if you disable then re-enable the plugin. you should create the executor in the startUp function instead

you don't need to invoke chatMessageManager.queue on the swing thread
https://github.com/voltmaister/collection-tracker/blob/0f055d58a355922b450721dbf460ede5537985ba/src/main/java/com/voltmaister/CollectionTrackerPlugin.java#L383

you should save the database inside the `RuneLite.RUNELITE_DIR`
https://github.com/voltmaister/collection-tracker/blob/0f055d58a355922b450721dbf460ede5537985ba/src/main/java/com/voltmaister/db/CollectionDatabase.java#L15
preferably in a subdirectory with your plugin name (so final db path would be `RuneLite.RUNELITE_DIR + "/collection-tracker/runelite-collections.mv.db"`)

## Questions
You have some unused functions in CollectionTrackerPlugin (isDataOutdated, printCollectionForCategory). Is this intended?

you're calling init on your database whenever a user clicks the "Sync Collection Log" button, is it not enough for it to be called on plugin startup?

## nits that you don't have to action
you don't have to specify a user or password for the h2 database

black text (which is used for the `!col toa <other_player>` use case) is very difficult to read with transparent chat background

You already have gson in the same class - no need to get it from the TempleApiClient
https://github.com/voltmaister/collection-tracker/blob/0f055d58a355922b450721dbf460ede5537985ba/src/main/java/com/voltmaister/CollectionTrackerPlugin.java#L399

CollectionItem could use lombok's `@Data`

You could use lombok's `@Getter`/`@Setter` on gson, and `@Setter` on httpClient. this would allow you to remove your custom getters/setters in TempleApiClient.java

voltmaister:
@pajlada Thanks for the detailed review on my code as for the questions the functions are for future things for now I commented them out... also you were right about the database init it should be just in the startup

pajlada:
## nits you don't have to do anything about
you could use some more examples for how to use the `!col` command in your readme. e.g. if i want to show my fishing trawler logs I have to do `!col fishing_trawler`

CollectionParser.java: You don't have to call setRuneLiteFormatMessage & client.refreshChat in the swing thread. If you don't want to run it in your executor thread, you could run it in the client thread. I'm not aware of any threading restrictions of those two functions
Line 52 and 81 of TempleApiClient.java you're not checking if body is null, so you _could_ throw an NPE here if the API misbehaves
In TempleApiClient.java, you could use an actual HttpUrl to build your Temple OSRS API URLs, rather than using string concats
You could make a panel class that contains all your swing logic, leaving your main plugin file a bit cleaner. this is what most other plugins do

PR #7827, opened 2025-04-13
Update = Custom Sounds fixed sound issues

daanbom:
certain issues arrived when sounds were placed close together, did a revamp on the sound system and should work perfectly now


tylerwgrass:
Your branch has conflicts that need resolving

LlemonDuck:
You should check out [AudioPlayer](https://github.com/runelite/runelite/blob/8f044e5a75d96e637ef5cfaff6fb79d726a84572/runelite-client/src/main/java/net/runelite/client/audio/AudioPlayer.java), which provides a comprehensive API for playing sounds so you don't have to manage them yourself.

daanbom:
> You should check out [AudioPlayer](https://github.com/runelite/runelite/blob/8f044e5a75d96e637ef5cfaff6fb79d726a84572/runelite-client/src/main/java/net/runelite/client/audio/AudioPlayer.java), which provides a comprehensive API for playing sounds so you don't have to manage them yourself.

I thought I did it the same as the other plugins which use sound. Do I need to change it to audioplayer or is this fine?

LlemonDuck:
It's fine as-is, but AudioPlayer is a new built-in API that aims to handle this exact use case for plugins. If you're willing, it would be better to take advantage of its implementation to a) reduce complexity in your plugin, b) not require ongoing maintenance, c) integrate better with niche setups that were accounted for here.

daanbom:
> It's fine as-is, but AudioPlayer is a new built-in API that aims to handle this exact use case for plugins. If you're willing, it would be better to take advantage of its implementation to a) reduce complexity in your plugin, b) not require ongoing maintenance, c) integrate better with niche setups that were accounted for here.

Sounds good. I will implement it then. But I would suggest still merging first since right now it has some significant bugs that are fixed on this patch. But I will see if I can rewrite it fast

PR #7828, opened 2025-04-14
Create Balloon Hider 1.0

amooose:
Hides the massive Hot Air Balloons in Entrana, Taverley, Castle Wars, Gnome Stronghold, Crafting Guild, and Varrock.  
Includes an option to keep or hide the basket.  
Supports removal in the extended map region too.

### DIFF
--- a/plugins/balloon-hider
+++ b/plugins/balloon-hider
@@ -1,2 +1,2 @@
 repository=https://github.com/amooose/Balloon-Hider-Runelite.git
-commit=47d104d28fabd4c7dd954a65ba3051328b13900d
+commit=8254e95b5cd44863e6dafa798cb73aaa026c8f16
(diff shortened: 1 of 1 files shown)


LlemonDuck:
Are you sure the map load subscriber is necessary when you have the onGameObjectSpawned handler? We mock GameObjectSpawned events during plugin startup as well

amooose:
Calling the removal also within onPreMapLoad was the only way I was able to remove the object in the extended region.
For example removing the onPreMapLoad and teleporting to Varrock, then running east to the Sawmill, it wont be removed.
It doesn't seem to trigger any sort onGameObjectSpawned event until you get close and the scene loads more.

https://github.com/user-attachments/assets/77a49248-ca02-4807-adac-51d610145c81

LlemonDuck:
Ah, I guess that makes sense. Thanks for the explanation.

PR #7856, opened 2025-04-19
Add Gymlocked Tracker plugin

pdewit:
This PR adds a plugin called Gymlocked Tracker. This is a plugin made for an XP locked account by Zidria with the following rules:

You can only gain XP in game if you unlock them IRL by getting fit. For example:

(Body)Weight training (per rep) = 2,000xp  
Running (per 1km) = 15,000xp  
Walking (per step) = 5xp  
Eating protein (per gram) = 1,000xp  
Calories (per calorie overage) = -25xp  
Sleeping & waking on time = 50,000xp  
Meditate (per session) = 25,000xp  

Once you've built up XP IRL you can add them via the plugin and start playing. When the counter hits 0 you'll be Gymlocked.

Game your way towards a healthy life!

### DIFF
--- a/plugins/gymlocked-tracker
+++ b/plugins/gymlocked-tracker
@@ -1,2 +1,2 @@
 repository=https://github.com/pdewit/gymlocked-tracker.git
-commit=f4a4cbd79d12bda1b77d8bdd825046c2eec26e25
+commit=26964e5f6203ef789574fe03ce76ec45e032b955
--- a/plugins/gymlocked-tracker
+++ b/plugins/gymlocked-tracker
@@ -1,2 +1,2 @@
 repository=https://github.com/pdewit/gymlocked-tracker.git
-commit=26964e5f6203ef789574fe03ce76ec45e032b955
+commit=da0b67cf8011a525e50e32108803be2828bce603
--- a/plugins/gymlocked-tracker
+++ b/plugins/gymlocked-tracker
@@ -1,0 +1,2 @@
+repository=https://github.com/pdewit/gymlocked-tracker.git
+commit=f4a4cbd79d12bda1b77d8bdd825046c2eec26e25


raiyni:
Not required but you are aware we have an on xp drop event right? That way you aren't fetching total XP every game tick.

pdewit:
> Not required but you are aware we have an on xp drop event right? That way you aren't fetching total XP every game tick.

Yeah good point, I used to have it like that but I ran into an issue where onStatChanged would increase with giants 'chunks' of stats on log in and wasn't very reliable to retrieve total xp when logging in. I reworked it to use game ticks when logging in and onStatChanged afterwards, thank you.

raiyni:
That was a known issue in the master client too, you'd need to wait some number of ticks to not see that. https://github.com/runelite/runelite/blob/8f044e5a75d96e637ef5cfaff6fb79d726a84572/runelite-client/src/main/java/net/runelite/client/plugins/xptracker/XpTrackerPlugin.java#L205

https://github.com/runelite/runelite/blob/8f044e5a75d96e637ef5cfaff6fb79d726a84572/runelite-client/src/main/java/net/runelite/client/plugins/xptracker/XpTrackerPlugin.java#L418

looks like it was waiting 2 ticks but this works too.

LlemonDuck:
If this is for a specific user, do we need to host it on the plugin hub? You can send it to them with a shadowJar build. `./gradlew shadowJar`

pdewit:
> If this is for a specific user, do we need to host it on the plugin hub? You can send it to them with a shadowJar build. `./gradlew shadowJar`

Hey, fair point, there were a couple of people asking for it on the [ongoing YouTube series](https://youtu.be/O0wqqHNwvw4?si=zzjSGoUGMuVFTqfJ) so the creator asked me if I can make a plugin for it. I figure you guys get a lot of 'snowflake' account plugins but I thought this one might be inspirational enough for people to want to try out as well 😄

PR #7857, opened 2025-04-19
Create skill-data-exporter.properties

cmhainesii:
Skill Data Exporter is a plugin that allows you export your characters skill level's to a CSV file. Optionally, you can also include XP values in the exported data.

Export to CSV file by typing: ::export in the chat. You will see a message confirming the export and the path to the file. The path and filename can be changed in the plugin configuration.


LlemonDuck:
You should use RuneLite.RUNELITE_DIR as the base for saving files. Since the default is evaluated once and then stored in the config, this would break for users where they share a synced config across two machines with different OSes or usernames since it would use the value from the previous machine.

cmhainesii:
Updated output directory to RuneLite.RUNELITE_DIR.

cmhainesii:
So I need to remove the ability for a user to set the csv file from being saved where they want it because of the case where a user explicitly changes the directory from the default, and at the same time they are syncing configs between two machines? 

Wouldn't it be more reasonable to set it to a subfolder in the runelite dir as default and I can add code to check before writing the file for a non existent directory and it will automatically set the configuration settings back to default at that point? I can put a message in the chat letting them know that the directory was reset to default in that case.

My issue with your approach is were only talking about cases where it was explicitly changed by the user. The default case always works and some additional logic can prevent errors from being thrown.

LlemonDuck:
Notifying the user is another valid alternative. You could also just use something like a ~ to reference user.home.

cmhainesii:
Ok so I can add the additional logic to make sure there's no scenario where
errors are thrown about invalid directories and add a notification to the
user telling them that the default directory was used instead. I mean I can
stick to runelite directory in a plugin specific subfolder. I originally
had it go to the user's home directory. Either works for me. I just want to
prevent any issues by ensuring the default is dir is always valid as a
fallback.

cmhainesii:
Implemented logic to check that output directory is valid prior to attempting to write CSV file. If invalid, it will revert the config setting to the default value, RUNELITE_DIR/skill_data_exporter.

PR #7858, opened 2025-04-19
Create Player Attack Tracker

Heavy-Flop:
Plugin that counts various helpful player combat metrics for tracking, including:

Attempted Attacks [Any attempted attack on an enemy]
Successful Attacks [Any attacks that successfully dealt damage to an enemy]
Damage Dealt [Any damage dealt to enemies]

Additionally, thresholds can be set at which these values will reset and measurements will begin anew (i.e., reset when attempted attacks reach 10 and/or damage dealt reaches 30)

### DIFF
--- a/plugins/Player Attack Tracker
+++ b/plugins/Player Attack Tracker
@@ -1,0 +1,2 @@
+repository=https://github.com/Heavy-Flop/Attack-Counter.git
+commit=8d8163ff9855f2f15c5216451f5d7fd2d060e1c7


Felanbird:
As we explained in the discord, this is clearly against Jagex's plugin rules:


>Guidelines:
>Any features which aid any boss fights by doing any of the following are prohibited (this includes all Raids sub-bosses, Slayer bosses, Demi-bosses, and wave-based minigames, including the Fight Caves and Inferno):

> Attack counters

cdfisher:
Why are you submitting this when we already told you this was not allowed?

Felanbird:
Right below [the attack counter entry](https://secure.runescape.com/m=news/third-party-client-guidelines?oldschool=1) is a special note for the tile marker rule, so if they meant attack counters for boss attacks, they would probably say that.

image

another "rule, except for in this case" example
image

Heavy-Flop:
Totally fair and I'm fine with it being rejected it was just a learning experience in making plugins, I mean again it wasn't my intention to pass or use anything against TOS. I just wanted to give you guys a proper opportunity to look at the repo rather than just the name since I figured you guys had more experience interpreting that document than me and this way I could maybe see/understand the reasoning for future reference. Thanks.

PR #7868, opened 2025-04-21
Disable equipment-lock

iProdigy:
Requested by @Matimatician


Matimatician:
I approve of the change.

PR #7880, opened 2025-04-23
Create sandcrabhelper

Az3rate:
Ready for PR


coopermor:
- Why do you continue submitting AI generated plugins?
- How is this an improvement over the core plugin NPC Aggression Timer?
- You didn't submit your plugin with a valid git repository link
- Your plugin repository still contains example plugin code
- This is not accurate as NPC aggression starts decreasing when you enter a region, not when you start combat
- The plugin does not even reset the timer properly
image

PR #7895, opened 2025-04-27
Add Ground Marker Profiles plugin

DangItOSRS:
- Expected use case is for thieving room in cox, switching between sets of tiles based on team size and whether or not telegrab is used
- Based off of tile packs/nyloer
- Sidebar for switching between sets of tiles with the click of a button
- Up to 10 sets of tiles, able to name buttons via the config
- Only one set of tiles is selected at a time

### DIFF
--- a/plugins/ground-marker-profiles
+++ b/plugins/ground-marker-profiles
@@ -1,0 +1,2 @@
+repository=https://github.com/DangItOSRS/ground-marker-roles.git
+commit=ec274f97cc684e90a1461d3b691faf51b52ece85


LlemonDuck:
FYI you may run into issues with single config entries being too large to store if you're using the JSON-formatted tile markers. There is a per-record size limit to each config value. You should consider looking into ways to shrink that value by using another format, or just save them externally.

If it's of any context, the core plugin also uses JSON but shards the keys by regionId, although this wouldn't work for your use case.

PR #7901, opened 2025-04-28
Added !log command to pull from TempleOSRS api

touchtouch105:
Pulls from the TempleOSRS Api when typing !log to pull collection log data and display it in chat

### DIFF
--- a/plugins/templecollectionlog
+++ b/plugins/templecollectionlog
@@ -1,0 +1,2 @@
+repository=https://github.com/touchtouch105/TempleLogCommand.git
+commit=4225fad34bd30894e727b55cbf49e3f62e46eb24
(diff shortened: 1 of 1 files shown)


cdfisher:
I think we would prefer for this to be added to the existing TempleOSRS plugin rather than adding it as its own separate plugin

touchtouch105:
Definitely agree, i feel like this would just act like a decent interim to get something working until the temple team creates a properly fleshed out !log. Also have a merge request to use base runelite api's instead of temple that im trying to get merged into chat commands, if you want i can remake a diff pull request using the base runelite apis so that im not stepping on any temple toes

LlemonDuck:
!log is already used by RuneProfile, you should consider using a different one so that users with both plugins do not have conflicts. 

Also, if you start pulling an install base, you risk pulling it out from under them if you later defer to the TempleOSRS plugin, so again it'd be really nice if you can have it added directly to that plugin. Have you attempted to PR it to that plugin repo?

touchtouch105:
i have not, let me see if i can get it incorporated with templeosrs plugin and see where it goes

PR #7908, opened 2025-04-29
Add time to max plugin information

Hamelot:
A RuneLite plugin that helps you track your progress towards maxing all skills in Old School RuneScape.

### XP Tracking and Forecasting

- **Interval-Based XP Tracking**: Choose to track your XP gains on a daily, weekly, or monthly basis
- **Target Date Setting**: Set a specific date by which you want to max all your skills
- **Progress Calculation**: Automatically calculates how much XP you need per day/week/month to achieve your max goal
- **Automatic Reset**: XP tracking automatically resets at the start of each new interval (day/week/month)

### Progress Visualization

- **Skill Sorting**: Skills are organized into "Active" and "Completed" categories based on whether you've met your XP goal for the current interval
- **XP Gained Display**: Shows how much XP you've gained in each skill during the current interval
- **Required XP Display**: Shows how much XP you need to gain for each skill to stay on track
- **Progress Bar**: Visual representation of your progress towards each skill's interval goal
- **Remaining Time**: Shows the number of days/weeks/months remaining until your target date

### Notifications

- **Goal Completion**: Receive chat notifications when you've gained enough XP in a skill for the current interval
- **Max Level Alert**: Updates the UI when a skill reaches level 99 to focus on remaining skills

### Commands

- **::resetxp**: Reset your XP baseline at any time with the `::resetxp` command
- **Confirmation Dialogs**: Prevents accidental resets with confirmation prompts

### Configuration

- **Target Date**: Set your goal completion date in YYYY-MM-DD format
- **Tracking Interval**: Choose between daily, weekly, or monthly tracking
- **Chat Notifications**: Toggle notifications when you reach your XP goals

### DIFF
--- a/plugins/time-to-max
+++ b/plugins/time-to-max
@@ -1,2 +1,2 @@
 repository=https://github.com/Hamelot/time-to-max.git
-commit=b4fecfc7e2ac5adb1b2bc11e7fc772525f94ee17
+commit=67c19511a7255ca16d88464513e5cfe2e0151633
--- a/plugins/time-to-max
+++ b/plugins/time-to-max
@@ -1,0 +1,2 @@
+repository=https://github.com/Hamelot/time-to-max.git
+commit=360d647b6ffb71035fac31525de7fa23fb625d6b
--- a/plugins/time-to-max
+++ b/plugins/time-to-max
@@ -1,2 +1,2 @@
 repository=https://github.com/Hamelot/time-to-max.git
-commit=792413d66e65257af1df7fd165b7012025b11142
+commit=d3e32f4a055837b56f8ed924de702f6b76685b89
--- a/plugins/time-to-max
+++ b/plugins/time-to-max
@@ -1,2 +1,2 @@
 repository=https://github.com/Hamelot/time-to-max.git
-commit=d89451517895788133556ce3746a8fbac3982d40
+commit=6604e6a8e391795919b9ff3a4a146d2069761a00
--- a/plugins/time-to-max
+++ b/plugins/time-to-max
@@ -1,2 +1,2 @@
 repository=https://github.com/Hamelot/time-to-max.git
-commit=b7bc3be204e838f9a21422d03fbcc9959b41a9c1
+commit=49df3836fc9137857e5f3b3ac50a170cb8cb8e84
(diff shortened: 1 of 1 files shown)


Hamelot:
Updated plugin to use injected client instead of gson as per initial build failure

LlemonDuck:
Instead of keying your config by username, you should consider using `{get,set}RsProfileConfiguration` which will automatically key based on username + world type, which also makes sure that it correctly partitions on beta worlds/leagues/etc.

Hamelot:
This is my first plugin, so I wouldn't have known about that. Do you know of any plugins that use that so I can review?

raiyni:
Core https://github.com/search?q=repo%3Arunelite%2Frunelite%20RsProfileConfiguration&type=code

LlemonDuck:
don't do all this file i/o on the client thread, inject the shared scheduled executor service or make your own and clean it up on plugin shutdown.

LlemonDuck:
Future.get blocks the caller thread until it's done, so this still isn't async https://github.com/Hamelot/time-to-max/blob/b7bc3be204e838f9a21422d03fbcc9959b41a9c1/src/main/java/com/timetomax/SkillsTracker.java#L83-L84

LlemonDuck:
you're calling things on the right thread now, which is great, although two things:
* in startUp, `if (!baselineSet)` will always be false since the async task won't have completed by the time that check is called, which then means you're overwriting the data every time. You can move all of that to another method and then call it from the tail of the async task to make sure it runs while also giving the loader a chance to finish first.
* Non-blocking: it's better to reuse the executors than to spawn one and immediately shut it down afterward. Just make sure to call shutdown in your plugin's shutDown.

LlemonDuck:
fyi we have Client#getAccountHash which can give you a unique account identifier that persists across name changes, instead of checking by username

it's also distinct for other profiles on the same username like leagues, dmm, quest speedrunning, etc

Hamelot:
> fyi we have Client#getAccountHash which can give you a unique account identifier that persists across name changes, instead of checking by username
> 
> 
> 
> it's also distinct for other profiles on the same username like leagues, dmm, quest speedrunning, etc

---

> Instead of keying your config by username, you should consider using `{get,set}RsProfileConfiguration` which will automatically key based on username + world type, which also makes sure that it correctly partitions on beta worlds/leagues/etc.

---

Which should I do? I've already made the previous change and tested on different modes. Would a player name change really be a huge concern?

Even if the player is using month based intervals, more than 2 or 3 name changes would seem like a lot to me.

Felanbird:
Players change their names all the time

LlemonDuck:
you should use ConfigManager's rsProfile stuff _and_ use account hash. You're just still doing some manual username checks which you can replace with account hash

Hamelot:
Noted. I'll go through and clear any manual checks and use the account hash.

PR #7911, opened 2025-04-30
Created Nightmare Shadows plugin

touchtouch105:
Created plugin that highlights nightmare shadow tiles and tags them with text for how many ticks are left on them


coopermor:
Have you read Jagex's statement for third party clients? These are prohibited


Indicates the time where a boss mechanic may start or end

Adds additional visual or audio indicators of a boss mechanic except in cases where this is a manually triggered external helper.

touchtouch105:
am i able to do it so that it just makes the swirly a solid black tile with no red tile highlight or tick counters?

coopermor:
I would assume that would be considered "Adds additional visual indicator"

PR #7916, opened 2025-04-30
Update clean-chat to 1.0.3

ldavid432:
- Add icon
- Some readme updates
- Fix chat commands breaking
  - Fixes ldavid432/chat-cleanup#2


ldavid432:
Just realized I probably need to shutdown the executor after it's done but won't be able to fix that till tonight, will reopen when it's fixed

PR #7920, opened 2025-04-30
Update tictac7x-charges to v0.6.3

TicTac7x:


raiyni:
After some discussion, we would rather not have conflicts defined in hub plugins that aren't  incompatible with the main client plugin. Eg. Mostly reserved for plugins that rely on modifying GPU sources.

TicTac7x:
> After some discussion, we would rather not have conflicts defined in hub plugins that aren't incompatible with the main client plugin. Eg. Mostly reserved for plugins that rely on modifying GPU sources.

This sounds like something that should be like this for most plugins, but since mine and the official charges heavily overlap, then this sounds like a conflict to me. But meh, i am willing to remove the conflict, but all future issue reports how these 2 plugins interfere will be ignored tho.

PR #7922, opened 2025-04-30
Update chance-man to v2.1.5

ChunkyAtlas:
- bypass spell restrictions for Blighted Surge Sack
- add option to disable rolling f2p trade only items
- add FreeToPlayBlockedItems enum for items that cannot be self-sufficiently obtained on f2p
- add option to filter out f2p items that cannot be self-sufficiently obtained on f2p


pajlada:
## nits you don't have to action now
In your `onConfigChanged` function, even if you hit the correct key (e.g. `freeToPlay`) and call `refreshTradeableitems`, you'll continue to check if the key matches any of the other

PR #7926, opened 2025-05-01
Create hide-chat

Jordanza21:
A plugin to hide the Chatbox widget from the user interface via a toggle option in the configuration of the plugin.


raiyni:
You can already hide chat by clicking on the stones?

Jordanza21:
That is true, however it still leaves the stone visible. For a more role-play style, I would like the ability to hide the chat only while still maintaining inventory functionality.
I believe it is already possible to hide only the minimap, which also aligns with my intent.

PR #7927, opened 2025-05-01
add discord multi webhook plugin

malkailany:
Another discord notification plugin but with multiple webhook support for specific events


cdfisher:
Your repository link should end in `.git`

Additionally, does this plugin do anything that isn't already doable with Dink?

malkailany:
Well this is awkward! I've looked for a plugin like this for some time and never found anything - thanks for suggesting Dink

Might cancel this pull request if Dink does work as expected

pajlada:
If Dink doesn't work as expected, we're super open feature requests/bug reports, and more contributors https://github.com/pajlads/DinkPlugin
We're trying to make it work for as many use cases as possible

PR #7932, opened 2025-05-01
Created simple sound swapper

touchtouch105:
Implemented basic plugin that allows for simple substitution of sounds based off sound id. 
Ex: Swap sound id 2508 with sound id 1326


Felanbird:
From the README:

## Reviewing
We will review your plugin to ensure it isn't malicious, doesn't [break Jagex's rules](https://secure.runescape.com/m=news/third-party-client-guidelines?oldschool=1), 
or isn't one of our previously [Rejected/Rolledback features](https://github.com/runelite/runelite/wiki/Rejected-or-Rolled-Back-Features).  
__If it is difficult for us to ensure the plugin isn't against the rules we will not merge it__. 


From our Rejected/Rolledback features:

PR #7933, opened 2025-05-01
Created mace sound swapper

touchtouch105:
Replaces sound of mace attacks (more specifically inquistors mace) with girthier sounds


cdfisher:
I'm fairly certain this can already be done with the existing `Sound Swapper` plugin on the Plugin Hub.

touchtouch105:
While technically achievable, it wouldnt allow for easy enough swapping between sound effects for the mace and would require a different .wav for each sound effect. Aim for the plugin is to solve a specific problem with a simple solution.

image

PR #7936, opened 2025-05-02
Update clean-chat to 1.0.7

ldavid432:
Wouldn't let me reopen cause I force pushed, oops

- Add icon
- Some readme updates
- Fix chat commands breaking
  - Fixes ldavid432/chat-cleanup#2
- Fix timestamps
- Fix right-click username being wrong
  - Fixes ldavid432/chat-cleanup#4

### DIFF
--- a/plugins/clean-chat
+++ b/plugins/clean-chat
@@ -1,2 +1,2 @@
 repository=https://github.com/ldavid432/chat-cleanup.git
-commit=ad1c185b8ac89a34b7765b498d86cab65489e152
+commit=c56ddedadc5c522f4a959a7653d4cd60d226c367


LlemonDuck:
do you need _5_ threads in your executor service?

ldavid432:
> do you need _5_ threads in your executor service?

I just wanted to protect against a bunch of chat commands being sent at once (mainly possibly an incorrect one, which would block the executor), although I think spamming commands typically happens in public which I don't deal with so probably not as common as I'm thinking.

Would it be better to just have 1 or 2?

ldavid432:
Switched to using a Timer (which should just be 1 thread) and scheduling the task instead of blocking threads so should be much better now

LlemonDuck:
fwiw, Timer is strictly worse than `Executors.newSingleThreadScheduledExecutor()` but it's not blocking

PR #7937, opened 2025-05-02
Implemented Metrics Tracker to track monster kph

touchtouch105:
Removing comment due to feature overhaul

### DIFF
--- a/plugins/data-logger
+++ b/plugins/data-logger
@@ -1,2 +1,0 @@
-repository=https://github.com/touchtouch105/DataLogger.git
-commit=3f92cd2075d5f2fdab3fd3ed51e077d907c809cf
--- a/plugins/data-logger
+++ b/plugins/data-logger
@@ -1,0 +1,2 @@
+repository=https://github.com/touchtouch105/DataLogger.git
+commit=3f92cd2075d5f2fdab3fd3ed51e077d907c809cf
--- a/plugins/metrics-tracker
+++ b/plugins/metrics-tracker
@@ -1,2 +1,2 @@
 repository=https://github.com/touchtouch105/metricstracker.git
-commit=44106dac1f3de680910b8f236b3dce4c203a78ef
+commit=eda4aa9fb738bc80542af68bdb77948a498a9ffe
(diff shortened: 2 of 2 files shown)


raiyni:
There exists a plugin that logs CSV data `csvexport`

touchtouch105:
> There exists a plugin that logs CSV data `csvexport`
That plugin logs every data point that can reasonably be queried every tick, not really equivalent

tylerwgrass:
If you wanted a smaller subset you could PR/suggest that the existing plugin has config options to toggle different info from being saved.
To me this seems more useful than a separate plugin with the random subset of fields you have picked.

touchtouch105:
> If you wanted a smaller subset you could PR/suggest that the existing plugin has config options to toggle different info from being saved. To me this seems more useful than a separate plugin with the random subset of fields you have picked.

The existing plugin is both broken and not currently being supported as per the issue tickets that have been open for 5 months

tylerwgrass:
I would not assume that it is broken or unsupported just from the issues, granted I haven't used the plugin.
It is still preferred to contribute to existing plugins instead of creating new ones that practically do the same thing as multiple identical plugins.
If it is unmaintained, that is the purpose of the [Plugin Takeover Policy](https://github.com/runelite/runelite/wiki/Plugin-takeover-policy).

touchtouch105:
Renamed data logger to metrics tracker, plugin now tracks non boss NPC kills per hour instead of the original concept. Its a complete pivot from the original design but didnt want to open a diff PR since this one was still open

ChunkyAtlas:
I've got a similar plugin already in the hub, although it does not have the KPH. I'd be more than willing to accept a PR.
https://github.com/ChunkyAtlas/monster-monitor

touchtouch105:
> I've got a similar plugin already in the hub, although it does not have the KPH. I'd be more than willing to accept a PR. https://github.com/ChunkyAtlas/monster-monitor

Ideally Id like to expand upon this plugin to track quite a few metrics, not just npc related, so id pref it to be standalone so i can continue to expand upon it :)

touchtouch105:
Refactored a bit to clean up unnecessarily lengthy code and to standardize formatting. Ran prolonged testing on this rev and am happy where its at, no further updates coming for this rev

LlemonDuck:
Please don't purge your comment history even if things have changed. Just explain the change and allow me the opportunity to still get the former context without having to manually view each comment's edit history.

PR #7939, opened 2025-05-03
Add Kill Count Viewer plugin

RedSparr0w:
# Kill Count Viewer

A plugin to display players kill counts at various bosses throughout the game

image

image
image


LlemonDuck:
it seems like this would hammer the hiscores and potentially get you rate limited, can you add a minimum cooldown between lookups?

RedSparr0w:
Thanks for the review @LlemonDuck 
It only checks new highscores every 30 minutes
https://github.com/RedSparr0w/runelite-plugins/blob/dc12b715c9eddf4edd28954b6d7ae2e06a06f009/src/main/java/com/redprisonsentence/RedPrisonSentenceOverlay.java#L124-L125

tylerwgrass:
It seems like you would still be making (potentially large) request bursts if many players are loaded at once, and looks like you might not have restricted fetching to the area, which would make this even more likely.

LlemonDuck:
you can also remove bin/

RedSparr0w:
I've updated the plugin to only send 1 lookup request every 5 seconds (feel like this should be fairly reasonable) let me know if it need adjusting further.

The plugin currently by default only looks up players while in the gauntlet lobby, but the player can enable an option to enable everywhere (should be okay now with the rate limiting).

Removed the `bin/` folder.

RedSparr0w:
Any other changes needed for approval?

LlemonDuck:
injection is resolved when your plugin is loaded, not when it's started, so your scheduling in the overlay class will cause the task to fire a) before your plugin is ever started, and b) even after your plugin is stopped.

Make sure you initialize the scheduler separately during startUp, and also clean it up in shutDown

RedSparr0w:
Thanks for the review ❤️ 

Updated plugin.
Now only runs while enabled, makes use of the gametick function.
I've expanded the plugin to now be enabled in majority of boss areas and mini games rather than just the gauntlet.
Please let me know if any other changes required.

tylerwgrass:
I would really advise not adding a bunch of new functionality to something that is in the process of being reviewed

RedSparr0w:
Yeah sorry, I was going to do it as a separate plugin, but doesn't make sense

LlemonDuck:
we cannot accept GPL-licensed plugins as the copyleft constraints are not suitable for us. we recommend the BSD 2-Clause license if you're willing to relicense it.

RedSparr0w:
No worries, have updated LICENSE to BSD 2

LlemonDuck:
use log.warn or log.error instead of printStackTrace https://github.com/RedSparr0w/runelite-plugins/blob/fbcd5721b37d80fadfc45ff633a4b93f929cdc25/src/main/java/com/killcountviewer/KillCountViewerOverlay.java#L335

RedSparr0w:
Thank you, removed `printStackTrace` and replaced with `log.warn`

LlemonDuck:
annotating the class with `@Slf4j` is a nice shorthand for setting up a logger btw

RedSparr0w:
Did not know this, will keep in mind for any updates, thank you for the merge!

PR #7943, opened 2025-05-03
Create Glassman

Mattb2291:
A plugin which replicates the 'Nightmare Mode' or '1hp' mode.

Simple and functional, this plugin will simulate life with only 1 hitpoint - damage taken from any source will cause the gamemode to finish and return the account to normal.

Glassman mode


Mattb2291:
Apologies,  I changed the commit hash (not sure if that was the right way to do it or not...)

LlemonDuck:
changing the hash is correct, but also you'll need to use a lowercase g in your manifest file name (the one in this repo), so `glassman` instead of `Glassman`

pajlada:
Still need to lowercase the plugin manifest file, which is the file you're pushing changes to in this repo (`runelite/plugin-hub`)
image

Mattb2291:
> Still need to lowercase the plugin manifest file, which is the file you're pushing changes to in this repo (`runelite/plugin-hub`) image

Thank you :) I changed all the others to lowercase but didn't see that one (sorry!)

pajlada:
Your plugin fails to build, see the error at https://github.com/runelite/plugin-hub/actions/runs/14981524544/job/42091074308?pr=7943

you'll need to update the `plugins` field in your `runelite-plugin.properties` file in your plugin repo to correctly point at the plugin class itself (so `com.glassman.GlassmanPlugin` if I looked correctly), then update the hash in this PR again

pajlada:
Plugin still fails to build, reason is the same as I mentioned above

Mattb2291:
Sorry, when I tried that last time it wasn't running on my PC. I think it's resolved...

Mattb2291:
I'm not sure if you can see this here, but if I want to update my plugin, do I do it in this thread or create a new pull request?

pajlada:
@Mattb2291 You'd create a new pull request updating the commit hash in the file you created in this pull request

PR #7945, opened 2025-05-03
Added PvP Zone Warning plugin

XirevGH:
This plugin enhances the visibility of the skull icon that tells the player that they are in a PvP enabled area by triggering a notification when the state of the icon changes. This is intended to remind players that are not paying attention to the skull icon that they can now be attacked. For example if the player logged out in a PvP world, forgot and didn't notice the skull icon, then went on to do something without the intent of PKing, this would give a clear reminder that they are about to do something dangerous. (Totally not what happened to me and the reason I made this plugin)

### DIFF
--- a/plugins/pvp-zone-warning
+++ b/plugins/pvp-zone-warning
@@ -1,2 +1,2 @@
 repository=https://github.com/XirevGH/pvp-zone-warning.git
-commit=ac12d40c7f61de7ce22e4aaea3ed025a32496450
+commit=0588223179454369f504aec16f5f989324193d34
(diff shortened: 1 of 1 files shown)


LlemonDuck:
you can replace this boolean with a Notification type to allow the users to customize it further https://github.com/XirevGH/pvp-zone-warning/blob/0588223179454369f504aec16f5f989324193d34/src/test/java/com/pvpzonewarning/PvPZoneWarningConfig.java#L29

LlemonDuck:

PR #7947, opened 2025-05-04
Update 1.1 Collection-Tracker

voltmaister:
Fixed names with multiple spaces in the name (this time for sure)
changed normalize function to replace non-breaking spaces with regular spaces


voltmaister:
Apologies for the frequent PRs — I discovered that the bug had been fixed in the test environment but not in production, so I needed to push a quick fix to address it.

PR #7950, opened 2025-05-04
Create special-attack-orb-hider

adnapPanda:
Added a new plugin that hides the special attack orb for specified weapons. This will allow players to not accidentally spec with undesired weapons like the abyssal whip.

Alternatively there's also an option to hide the orb for all weapons if desired


LlemonDuck:
requireNonNull doesn't guard you against throwing NPEs, you should still check whether the widget is null and just return early if it is

PR #7959, opened 2025-05-06
Update shortest-path to v1.17.2

Skretzo:
- Correctly identify built quetzal landing sites


LlemonDuck:
we don't allow reflection in hub plugins

perhaps we can expose a way to reverse map gameval names to varb ids, but unless that happens please use the id values instead

iProdigy (approved):

tylerwgrass (approved):

PR #7962, opened 2025-05-06
Create Storage Reminder

Yondem:
On logout spawns a text box in the middle of the screen with customizable text. Intended to serve as a reminder to put items into group storage before logging out.

Code is AI slop.


cdfisher:
Please don't close PRs and open new ones, just push to the existing branch. 

We also really  would prefer not to have AI-generated plugins submitted because even in the rare cases that they do work, they often do unintended things or - like in this case - use deprecated features from the API which shouldn't be used anymore.

Yondem:
> Please don't close PRs and open new ones, just push to the existing branch.

I think I had to rename the plugin to lower case, and upon doing so it auto-closed the previous PR. Sorry!

> We also really would prefer not to have AI-generated plugins submitted because even in the rare cases that they do work, they often do unintended things or - like in this case - use deprecated features from the API which shouldn't be used anymore.

I understand. I'm getting the help of someone who understands this better than me to rewrite this and fix this issue.

RedSparr0w:
@Yondem I have created a similar plugin here: #8267

LlemonDuck:

PR #7974, opened 2025-05-08
Update npc-contact-filter

OldShrimp:


iProdigy (approved):
avoid dashes in the group name https://github.com/OldShrimp/npc-contact-filter/blob/c56d90e6b41ffa3f8512c397251a87aee446c1d8/build.gradle#L29

tylerwgrass (approved):

PR #7976, opened 2025-05-08
Create Loot Chance plugin

Sloshedd:
Loot Chance is a plugin that helps you track base drop rates and calculate your odds of receiving unique items based on your current kill or loot count while factoring in table chance. Additionally, features a probability calculator with inputs for base chance, table chance, and kill count. It supports data from the OSRS Wiki and presents it in a clean, OSRS-themed panel. Data is pulled from a local Json and png icons for offline use and less-frequent updates. Locally re-running a scraper for new content (loot sources/data and icons) and committing on major updates is all that is required. Using local custom scrapers, it gathers not just monsters but anything that might give you loot including scenery (minigames and clues in an update).

### DIFF
--- a/plugins/lootchance
+++ b/plugins/lootchance
@@ -1,2 +1,2 @@
 repository=https://github.com/Sloshedd/Loot-Chance.git
-commit=63b3a32141fe601ac28584d50ef0c06f58fe0e54
+commit=95b65dc5328d5fcc57ae8858edd131fdda4f883c
--- a/plugins/lootchance
+++ b/plugins/lootchance
@@ -1,0 +1,2 @@
+repository=https://github.com/Sloshedd/Loot-Chance.git
+commit=63b3a32141fe601ac28584d50ef0c06f58fe0e54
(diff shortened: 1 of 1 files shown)


riktenx:
This is mathematically wrong or at the least, incredibly misleading. Individual drop probability doesn't increase in the way your panel implies as kill count increases. Your loot tables are also almost completely broken, most of the individual "drop tables" have 100% chances.

Did you write this code yourself?

image

Sloshedd:
Thanks for your review. I agree, the individual drop table chance scraping is inconsistent at the moment and fallsback to 1/1 for any wiki entry that doesn't directly display a table chance in percentage or fractional format. The probability calculation also relies on that value (instead of base chances per item) in turn reporting a false calculation. Apologies, as I am a beginner coming from C++ but I did write this

Sloshedd:
Using your example above, and the formula shown here, this was my result which aligns with the number you're seeing. This was calculated using Mathway

cdfisher:
That doesn't mean that's the chance of getting that drop on the next kill. They're independent trials, so it doesn't change from trial to trial.

Sloshedd:
Using a binomial calculator from stattrek,  I got the same result. Taking individual trials into account
image

cdfisher:
You may be calculating the value itself correctly but you're misinterpreting what that value actually means

Sloshedd:
I understand it is poor user experience choice of words.

Sloshedd:
How's this? 
image
image

PR #7980, opened 2025-05-09
ScaleUp 1.0.0

CursedMika:
Allows you to record any activity and then scale its requirements (either material or time) up or down as needed.


CursedMika:
Forgot to clean dependencies...

PR #7981, opened 2025-05-09
ScaleUp 1.0.1

CursedMika:
Allows you to record any activity and then scale its requirements (either material or time) up or down as needed.

### DIFF
--- a/plugins/scaleup
+++ b/plugins/scaleup
@@ -1,0 +1,2 @@
+repository=https://github.com/CursedMika/ScaleUp
+commit=908f8aed266e87f16085cfc5de8346b1bdf46ed8


LlemonDuck:
Swing actions need to be on the swing thread, not the client thread. Use SwingUtilities.invokeLater to switch back to the client thread here (and other places) https://github.com/CursedMika/ScaleUp/blob/e19b9f1cbfc82ba27ea5134f4471b6b14d85bfd4/src/main/java/eu/cursedmika/ui/ScaleUpPanel.java#L80-L86

CursedMika:
Tried to rewrite it so UI logic and RuneLite API access are separated as much as possible. I hope I understood how to switch between client thread and UI thread properly now.

PR #7983, opened 2025-05-10
Add agilityfc plugin

afctun:


plugins/agilityfc
```diff
@@ -0,0 +1,2 @@
+repository=https://github.com/agilityfc/plugin.git
+commit=597fe8dad78e519ebbb59c6051392e4924766eb0
```
tylerwgrass, line 2:
```suggestion
commit=597fe8dad78e519ebbb59c6051392e4924766eb0
authors=afctun
```

afctun:
I've got fixes for both of the tasks in the failed check. Are these ok?

- https://github.com/agilityfc/plugin/commit/763eeeb8836cb85186aedc229b3c5efeb44e7ec9
- https://github.com/agilityfc/plugin/commit/e12d26985dd42cb1b10288d115ed184da37072cc

tylerwgrass:
you can push an updated commit to this pr and it will rerun the checks

LlemonDuck:
What's the point of all this manual SSL verification?

afctun:
> What's the point of all this manual SSL verification?

Due to financial and privacy reasons I don't own a domain name that I can use for the plugin's "donation service", but without one it's not possible to get an SSL certificate from any public CA, even a free one like Let's Encrypt. I do need a secure connection however (due to exchanging sensitive data like authentication tokens), so I went down the route of using a self-signed certificate. This turned out to be quite simple and doesn't require all that much extra code.

All of the SSL heavy lifting is still left up to the OkHttpClient. I only configure it with a trust store that includes the given self-signed certificate in addition to all of the default ones present on the system.

raiyni:
You should be able to use one of the various DDNS services to get an accessible name.

afctun:
> You should be able to use one of the various DDNS services to get an accessible name.

I will look into that in the future, though is it ok if it stays as it is for now?

afctun:
On a second thought, I looked into it and it was simpler than I expected. :-)

iProdigy:
> I believe the "waiting for author" label is an error? I've pushed the necessary change that gets rid of the custom SSL stuff.

Sorry, didn't see your commit

afctun:
Hi, been a few days. Let me know if you have any other comments! :-)

LlemonDuck:
we cannot accept GPL-licensed plugins as the copyleft constraints are not suitable for us. we recommend the BSD 2-Clause license if you're willing to relicense it.

afctun:
> we cannot accept GPL-licensed plugins as the copyleft constraints are not suitable for us. we recommend the BSD 2-Clause license if you're willing to relicense it.

Oh I see, I thought both were fine. Relicensed as requested.

PR #7993, opened 2025-05-12
Tormented Demon counter plugin

mpr120:
This plugin will count the number of attacks a Tormented Demon has done during each attack style phase. It does not "pre-determine" attack style or attacks at all. This is a QOL improvement so players do not have to physically count the attacks that a tormented demon has already used every time the Tormented Demon switchs attack styles.


cdfisher:
Attack counters are not permitted per Jagex's Third Party Client Guidelines.

PR #7995, opened 2025-05-12
Update ca-export to 1.2.0

cdfisher:
- Migrates to using gamevals
- Cleans up CAFileWriter a little
- Adds an identifier to the exported filename to denote what sort of game profile it came from (Main game, DMM, Leagues, etc)


raiyni:
If you c are enough CAFileWriter can be injected so you aren't managing the ctor but it doesn't really matter

PR #8018, opened 2025-05-15
Update Flipping copilot to v1.7.0

cbrewitt:
- Allow user to select timeframe
- Add pop-up with price graphs and statistics
- Add pop-up to manage accounts
- Update readme and plugin description

### DIFF
--- a/plugins/flipping-copilot
+++ b/plugins/flipping-copilot
@@ -1,3 +1,3 @@
 repository=https://github.com/cbrewitt/flipping-copilot.git
-commit=4758ff663bc849e1204c674de762f6d8597215fd
+commit=8afa4f5e536a85fa203eb9280c924f9e79737578
 warning=This plugin submits your grand exchange offers, grand exchange transactions, and IP address to a 3rd party server not controlled or verified by the RuneLite Developers.
--- a/plugins/flipping-copilot
+++ b/plugins/flipping-copilot
@@ -1,3 +1,3 @@
 repository=https://github.com/cbrewitt/flipping-copilot.git
-commit=711c61e0b662d9244ea7f72f919783b0ef3b4bee
+commit=4758ff663bc849e1204c674de762f6d8597215fd
 warning=This plugin submits your grand exchange offers, grand exchange transactions, and IP address to a 3rd party server not controlled or verified by the RuneLite Developers.


raiyni:
You shouldn't need to use reflection to deserialize/serialize objects

raiyni:
The point is, you know what classes you are unmarshalling to. So you should be able to create them. Those libraries rely on reflection under the hood for generic implementations.

cbrewitt:
> The point is, you know what classes you are unmarshalling to. So you should be able to create them. Those libraries rely on reflection under the hood for generic implementations.

Yep we understand that. Our implementation ⁠ MsgpackDeserializer ⁠/ ⁠ MsgpackSerializer ⁠ is intended to be a generic implementation to make it easily re-usable across different target types. Even though we know the target types, there could  potentially be many target types in the future and it is cleaner and more robust to have a generic implementation that is tested once than dedicated code for each type. However, if this is a blocker for you we are happy to migrate to dedicated code per type.

Adam-:
> > You shouldn't need to use reflection to deserialize/serialize objects
> 
> We are leveraging MessagePack (https://msgpack.org/index.html) format for more efficient transmission of larger bits of data from our server. Ideally, we would use an off the shelf library like https://github.com/msgpack/msgpack-java to do the serialization/deserialization. However, no such dependency already exists in the project (unlike for JSON where the ⁠ gson ⁠ library is already included) and we are under the understanding that there are restrictions on adding new dependencies. So we have added our own minimal lightweight implementation and that is the reason for the reflection. The efficiency improvements were large enough for us to justify this additional effort rather than just sticking with JSON.
> 
> We agree the use of reflection should be minimised but unfortunately it is fairly standard in any general purpose serialisation/deserialisation code. The ⁠ gson ⁠ library relies on it under the hood. And for example, our custom annotation for easily mapping the message pack field names simply can't be implemented without reflection
> 
> ```
> @Retention(RetentionPolicy.RUNTIME)
> @Target(ElementType.FIELD)
> public @interface MsgpackName {
>     String value();
> }
> ```

If you are really unable to use either gson or protobuf, you should be adding the msgpack dependency and using that.

cbrewitt:
We've migrated to the msgpack dependency, as suggested.

Adam-:
One of your dependencies requires java 21, but we only support 11

cbrewitt:
> One of your dependencies requires java 21, but we only support 11

Would you mind sharing more details? 

When building and running the plugin with Java 11 and the `org.msgpack:jackson-dataformat-msgpack:0.9.9` dependancy included, everything seemed to function correctly.

Based on their [release notes](https://github.com/msgpack/msgpack-java/releases/tag/v0.9.6), it looks like msgpack-java added support for Java 21 and dropped support for Java 7 in version 0.9.6, but it doesn't look like they have dropped support for Java 11.

cbrewitt:
I've switched to an older version of msgpack-java, before Java 21 support was added.

cbrewitt:
> Feedback:
> 
> * Use of `java.lang.System.out` is not allowed: system i/o
> * Use of `java.lang.System.err` is not allowed: system i/o
> 
> A new scan will be automatically performed on the next push to the PR branch.
> 
> Internal use only: [Reviewer details](https://discord.com/channels/301497432909414422/1262533853570928790/1375834064979890198) [Maintainer details](https://discord.com/channels/301497432909414422/1075111769867034745/1375834064426242179)

It looks like the msgpack-java depedency was rejected by the code scanner because it contained system i/o.

We've reverted to our custom msgpack implementation, with reflection removed as recommended by [raiyni](https://github.com/raiyni).

raiyni:
Dependencies *can* have things flagged, it just takes time for someone to review the dependency. 

This is reflection and isn't allowed [cbrewitt/flipping-copilot@`4758ff663bc849e1204c674de762f6d8597215fd...cbrewitt:3bb0212932a1c955469eb27df5a9bcedb1b46318`#diff-6f67a66f77](https://github.com/cbrewitt/flipping-copilot/compare/4758ff663bc849e1204c674de762f6d8597215fd...cbrewitt:3bb0212932a1c955469eb27df5a9bcedb1b46318#diff-6f67a66f77efc3762d7dee1c753f1c2cc46211cdc345d83c980e25c0c641777fR53)

cbrewitt:
Thank you for the clarification. I've now removed the reflection. When making the initial pull request, we were unaware of the policy on reflection. 

We appreciate all the work the Runelite maintainers are doing to review pull requests, and want to make the job as easy as possible. I've left the custom msgpack implementation for now, would it make things easier for you if we used the-off-the shelf library?

PR #8022, opened 2025-05-15
Update Chat Panel to v2.2

Yenof:
Updated changes:

Add Split PMs (PM Tabs)

Tab name color, font size, font family, and underline color all configurable.

Add Filtering/censoring

Add per-character random colors

Add name and timestamp event color overrides

<br>
<br>
<br>
<br>

Previous message:

~~Add Chat Out!~~

~~Add input font size and color configs~~
~~Add channel selector font config~~

~~Add name and Timestamp event color override~~

~~Add Split PMs (PM Tabs)~~
~~Add reMessage~~

<br>
<br>

~~Anti-autotyping considerations made:~~

~~No pasting or import~~
~~No invalid character entry~~
~~No typing when logged out~~
~~80 character limit like RL~~
~~keyListeners for RL to track inputs into the popout(?)~~
~~Error messages can't be sent by user spamming buttons~~
~~Keypress repeat speed same as RL (same as OS)~~
~~No auto-channel switching (input required)~~

~~Notes:~~
~~While having no understanding of bots or auto-typers, I have made all the protections I could think to add to prevent~~ ~~these things and tested relentlessly.  ~~
~~Couldn't test on mac, if someone wants to verify there is no unintended text entry permitted on there, that would be~~ ~~awesome.~~
~~If there is anything that can be added to reduce potential malicious usage of this plugin, please let me know.~~
~~I feel like at it's core this update follows all rules and guidelines for plugins, I just hope I've made it possible for that to be forwarded to and monitored by the client.~~

### DIFF
--- a/plugins/chatpanel
+++ b/plugins/chatpanel
@@ -1,3 +1,3 @@
 repository=https://github.com/Yenof/Chat-Panel.git
-commit=8e18f46f53107ebc0de4d4b74386ff27e210f3d2
+commit=6a30c500248f1611d14abcbab9139bf6d806eae3
 authors=Yenof
--- a/plugins/chatpanel
+++ b/plugins/chatpanel
@@ -1,3 +1,3 @@
 repository=https://github.com/Yenof/Chat-Panel.git
-commit=6a30c500248f1611d14abcbab9139bf6d806eae3
+commit=3411108d95414d2487fd4561aaefda6b85d26060
 authors=Yenof


raiyni:
I don't see this being something we could reasonably allow, considering we don't allow any other types of auto-typing plugins.

Yenof:
Hi, yeah I wouldn't want any auto-typers allowed in the plugin hub either. 

I've tried pretty hard to make this not auto-typing in any way, and nothing should be able to be automatically typed.

As far as I can tell keypresses should be sent 1by1 to the client and should be able to be monitored by and seen by the client in realtime.
My hopes are this makes it protected from auto typing like the in-game text field.

I don't understand how the anti-botting stuff works, but is there a better listener I can use or something to better allow the client to monitor and verify that this is not auto typing? 

I think this feature would be super amazing and I'd like to do everything I can to make it safe and functional.

Yenof:
Good day maintainers! I'll try to add some detail to my previous reply.


This is not an auto-typing plugin.


The user manually enters text into the text box, and manually presses enter to send to the server using ScriptID_CHAT_SEND.

This is the same as how the RuneLite sendChatBoxInput in ChatInputManager functions. 
Get string from box, runScript CHAT_SEND.

No auto-typing at all, everything is manual. Typing, sending, channel switching, all manual.
This just adds a new InputBox into RuneLite, the rules of which should be equal to or more restrictive than the original.

I have added what I think allows RuneLite to directly 'watch' keypresses, but from what I understand RuneLite can already monitor actions and events in it's plugins, so it should be able to verify all inputs are manual. 
If the current protections on the InputBox are insufficient, I'd like to fix that. 


The "reMessage" part is for internally moving messages between PM tabs, nothing to do with the Chat Out/Sending if that was a concern. I can see why that name seems alarming though haha.


Attached is a video displaying keyboard events, RuneLite, Chat Panel, and the console output to show how they function together. 
The RL ChatInputBox is unaffacted by Chat Panel events, nothing is being automatically entered, mirrored, or transferred into it.
The RL Console receives keypresses as they are manually typed into the Plugin window, and then ScriptID_CHAT_SEND takes it from the Plugin to the server when the user clicks send or presses enter. 


This plugin is for accessibility, and contains no mechanisms to automate typing or inputs. 


If anyone can figure out any improvements to or an alternative way to monitor this new input box so we can verify there is no auto-typing, I would be extremely grateful. :D 

https://github.com/user-attachments/assets/59c0cd77-2f87-4d62-9d6a-abda89044895

LlemonDuck:
This is still not going to fly: I understand where you're coming from, but it _is_ moving keypresses from the game window to another, unwatchable panel, bypassing any potential anticheat within the client and also spoofing actions on the players' behalf within the window.

If someone made a program where you manually type in what you want to say, and it then simulated the keypresses into the RuneLite window, that would also be a problem, and I hope it's more obvious why, but it's in the same vein of issues.

We can't feasibly host this on the plugin hub given the risk it would open users up to.

I'll leave the PR open since it looks like you have other changes you want merged as well, but please remove this option.

Yenof:
Hi, thanks so much for the reply. I think I understand, but just so I have no doubts:

By "unwatchable panel" that means there is nothing that can be done to allow it to be watched? 

But I totally understand if that might involve dipping into secret sauce, and I wouldn't want to compromise anything about RuneLite's anti-cheating systems. 

Thanks again! :D 
I'll start removing this feature.

scottreel11:
> As a third-party client Jagex can set different rules than what they do for their own client.

I understand, I was providing context since I didn't see anyone mentioning that Jagex had already brought similar features to their client.

PR #8024, opened 2025-05-15
Adding warmth alert plugin to plugin hub

graceulinski:
Warmth Alert plugin for Wintertodt allows you to set notifications and/or a colored overlay when your warmth falls below a specified level at the skilling boss.


coopermor:
This seems like a good candidate to PR to core's Idle Notifier in my opinion

PR #8026, opened 2025-05-16
Adding overhead-notifications to pluginhub

Shuddie:
Notification plugin that lets you do overhead text for stuff such as thralls, saturated heart, overloads and divine potions.

This is similar to the overhead option in Watchdog, but as Watchdog is restricted in most endgame pvm locations it's hard to benefit from it. this plugin uses hardcoded checks to avoid misuse for game message checks.

This restricts the overhead option to only be used with things that doesn't go against the game integrity and isn't limited to watchdog's limited areas.

### DIFF
--- a/plugins/overhead-notifications
+++ b/plugins/overhead-notifications
@@ -1,0 +1,2 @@
+repository=https://github.com/Shuddie/overhead-notifications.git
+commit=7cb6673318d2bd71e9a521b5fc9396d0e79cf49c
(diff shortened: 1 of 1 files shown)


pajlada:
Your `runelite-plugin.properties` file is misconfigured. Your `plugins` key needs to point to the class that extends `Plugin`

Shuddie:
> Your `runelite-plugin.properties` file is misconfigured. Your `plugins` key needs to point to the class that extends `Plugin`

thanks, noticed it after the build failed. got it fixed now

PR #8028, opened 2025-05-16
Update Ge-Filters Support Link

Nick2bad4u:
Support link for the old repo was never changed, this commit only changed the support url to the correct URL:

Commit in GE-Filter Repo: https://github.com/Nick2bad4u/GE-Filters/commit/54c7886eceef69724608a988f95cc8175d50cfe7

```support=https://github.com/salverrs/GE-Filters```
Changed to:
```support=https://github.com/Nick2bad4u/GE-Filters```

No other changes made.

### DIFF
--- a/plugins/ge-filters
+++ b/plugins/ge-filters
@@ -1,3 +1,3 @@
 repository=https://github.com/nick2bad4u/GE-Filters.git
-commit=5851376b18e91a65a942eb221b739c643902ab9b
+commit=54c7886eceef69724608a988f95cc8175d50cfe7
 authors=nick2bad4u, Salverrs


pajlada:
The `support` key in the `runelite-plugin.properties` file is actually not used since 2023.
Are you OK with me closing this PR, so the change you made to the field can roll into the next change you make to the plugin?

Nick2bad4u:
> The `support` key in the `runelite-plugin.properties` file is actually not used since 2023.
> Are you OK with me closing this PR, so the change you made to the field can roll into the next change you make to the plugin?

Sure 

I only opened it because someone on the thread on my last update told me support issues would be filed in the wrong place without it. 

So i guess that's not true anymore?

pajlada:
Correct - you can confirm this by checking the plugin page on the website (https://runelite.net/plugin-hub/show/ge-filters) and clicking the "Report an issue" link. You'll see it correctly links to your repository already.

I'm closing this PR in that case, letting the support fix commit be merged in next time you have another update to the plugin.

Sorry for the confusion!

PR #8037, opened 2025-05-17
Create horn-sound

Faaruz:
Play an audio file when the SoulFlame horn is activated. You can replace the file inside the Runelite folder with a different .wav file.

### DIFF
--- a/plugins/horn-sound
+++ b/plugins/horn-sound
@@ -1,3 +1,3 @@
 repository=https://github.com/Faaruz/horn-sound.git
-commit=807546c5b79c8be51e9575e800c75c85e5e368bb
+commit=b76c67ee47984af53ab5e6fc780b5e3f01515a91
 authors=Darkwarden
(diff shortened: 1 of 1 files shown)


Faaruz:
Audio from https://www.youtube.com/watch?v=UaUa_0qPPgc
If i need to change the audio due to licencing, I can. Couldn't find anything in description as its a very old / popular video.

Nightfirecat:
Something being old/popular/meme-worthy doesn't mean it's licensed such that we can distribute it. I'd suggest you use a different file. If you want to offer users the ability to have this exact sound, you could add functionality to read from a file in `~/.runelite/` if it is present

Faaruz:
> Something being old/popular/meme-worthy doesn't mean it's licensed such that we can distribute it. I'd suggest you use a different file. If you want to offer users the ability to have this exact sound, you could add functionality to read from a file in `~/.runelite/` if it is present

I believe ive adjusted it to allow for people to add in their own audio inside the folder created. Further ive replaced the file to something im allowed to use and adjusted the licence.

PR #8039, opened 2025-05-18
Update calculator-pro

kanderson66:
Add Custom Functions and RunTime/CustomTag list viewing via "!calc !list CustomTags"


kanderson66:
Also includes hash 3de6745c80e3b1fc59cbf90922bf56203f1c3487

iProdigy:
fyi none of the images in your readme will render because they are not direct image links

kanderson66:
> fyi none of the images in your readme will render because they are not direct image links

Thanks for that, I'll update the links once I get on my laptop

PR #8042, opened 2025-05-18
Update arceuus-timers 2.0

Mantic-says-hi:


tylerwgrass (approved):

iProdigy (approved):
nit: double-brace initialization should be considered as an anti-pattern (use `Map.of` instead)

https://github.com/Mantic-says-hi/arceuus-timers/blob/407fe6ac0c84159824d055170e426a3a321f235b/src/main/java/com/arceuustimers/ArceuusTimersPlugin.java#L86

https://stackoverflow.com/a/27521360

PR #8046, opened 2025-05-18
Add Yama Utilities plugin

sreilly64:
This plugin contains various utility or QoL features surrounding the Yama encounter:
- Your damage contribution will be printed to chat when Yama is defeated.
- When in a Party with someone via the Party Plugin/Party Hub Panel, their display name will be automatically added the "Join" chat interface at the Voice of Yama as well as the "Add Friend" chat interface.
- When in a Party with someone via the Party Plugin/Party Hub Panel, and they also have the Yama Utilities plugin enabled, if they enter Yama's Domain the Voice of Yama will display overhead text indicating that they are inside.
- When you step on and activate a glyph during the Yama encounter, your character will be highlighted for the duration of the glyphs protection. Each glyph (or combination of both glyphs) has its own color highlight which can be configured in the plugin setting.


tylerwgrass:
> When you step on and activate a glyph during the Yama encounter, your character will be highlighted for the duration of the glyphs protection. Each glyph (or combination of both glyphs) has its own color highlight which can be configured in the plugin setting.

This feature is not allowed as it breaks the following guideline. Please remove it.

sreilly64:
Acknowledged. Will update later this evening.

sreilly64:
Found some time to update it now. I have removed the highlighting feature as requested. Please let me know if there is anything else @tylerwgrass. Thank you.

sreilly64:
@rdutta done!

sreilly64:
Pushed a small bug fix. Damage tracking was counting the healing hitsplats (when Void Flare explodes) as part of total damage dealt.

LlemonDuck:
you should attribute vikke1234 properly in your license if you are using derivative code from their plugin

just copy this line into your license below your own name line: https://github.com/vikke1234/charge-calc/blob/master/LICENSE#L3

sreilly64:
Good idea, thanks @LlemonDuck. Just pushed the change.

LlemonDuck:
Plugin looks fine, although before I merge I want to just double check that your duo name autofill won't trigger any auto-typing anticheat. Please don't push another update in the meantime so that it doesn't require a fresh review.

PR #8054, opened 2025-05-20
Update timers-ca

PaJauKat:
Added timers for Yama


Nightfirecat:
Yama doesn't have CAs yet does he?

Sharpienero:
> Yama doesn't have CAs yet does he?

Not yet

PaJauKat:
image

iProdigy:
shouldn't these maps be cleared upon `Yama#reset`? https://github.com/PaJauKat/Timers-CA/blob/0d1a2b4c337a39c2b2e88a67b80a2b6ef0a640de/src/main/java/com/TimersCA/Bosses/Yama.java#L36-L40

PaJauKat:
> shouldn't these maps be cleared upon `Yama#reset`? https://github.com/PaJauKat/Timers-CA/blob/0d1a2b4c337a39c2b2e88a67b80a2b6ef0a640de/src/main/java/com/TimersCA/Bosses/Yama.java#L36-L40

They reset at the start https://github.com/PaJauKat/Timers-CA/blob/0d1a2b4c337a39c2b2e88a67b80a2b6ef0a640de/src/main/java/com/TimersCA/Bosses/Yama.java#L78C41-L79C17 I prefer this way because they are related with minionTimes and phaseTimes.

iProdigy:
> They reset at the start

Yeah that's good - if someone started the yama fight but left before completion, would the maps never get cleared? (until they start another yama fight)

iProdigy:
> Yes, the times persist until you start another KC, which lets you review them while preparing for the next attempt.

Makes sense, though the maps may not get cleared if the user switches to fighting another  boss, no?

iProdigy (approved):

PaJauKat:
> > Yes, the times persist until you start another KC, which lets you review them while preparing for the next attempt.
> 
> Makes sense, though the maps may not get cleared if the user switches to fighting another boss, no?

No, map data isn’t cleared when changing bosses. Each boss maintains its own timer logic based on its specific mechanics. So if you switch to a different boss and later return, the last KC time for the original boss will still be shown once you're back in its region ID.

iProdigy:
Thanks for clarifying it's intentional (not an unintended memory leak)

PR #8056, opened 2025-05-20
Submitting ironwill plugin

seniir:
Records boss/raids new collection items with kill count and date

### DIFF
--- a/plugins/ironwill
+++ b/plugins/ironwill
@@ -1,0 +1,2 @@
+repository=https://github.com/seniir/Ironwill-plugin.git
+commit=5a6ffb3a4b27d07e0ab163c6c9d19552371829ad


tylerwgrass:
- The strings here should be sanitized/escaped: https://github.com/seniir/Ironwill-plugin/blob/5a6ffb3a4b27d07e0ab163c6c9d19552371829ad/src/main/java/com/ironwill/IronwillPanel.java#L48
- The config file is meant for user-set configuration which doesn't really make sense (players could just change their kc/drop to be whatever). You should probably remove your [Config](https://github.com/seniir/Ironwill-plugin/blob/5a6ffb3a4b27d07e0ab163c6c9d19552371829ad/src/main/java/com/ironwill/IronwillConfig.java) file and use [ConfigManager](https://github.com/runelite/runelite/blob/f31b51aecef47bcfdfdf1401aa102abe7966b149/runelite-client/src/main/java/net/runelite/client/config/ConfigManager.java#L96) instead.

PR #8060, opened 2025-05-21
Update `custom-item-hovers` to `1.3.0`

geel9:
- Adds new variables and functions available to expressions
	- Variables
		- Items
			- `name`
		- Player
			- `player_hp`
			- `player_prayer`
			- `{skill}_level`
			- `{skill}_level_real`
			- `{skill}_xp`
			- `player_x`
			- `player_y`
			- `player_region`
			- `player_region_x`
			- `player_region_y`
	- Functions
		- Inventories
			- `[inv/equipment/bank]_qty()`
			- `[inv/equipment/bank]_contains()`
		- Varbits
			- `varbit()`
			- `varplayer()`
			- `varcint()`
			- `varcstr()`
		- Math
			- `abs()`
			- `pow()`
			- `max()`
			- `min()`
		- Strings
			- `len()`
			- `contains()`
		- Colors
			- `[red/yellow/blue/green]()`
			- `[red/yellow/blue/green]Start()`
			- `rgb()`
			- `rgbStart()`
			- `colorEnd()`
- Removes "helpful" tutorial message which spammed users
- Improves expression engine
	- Refactored to be easier to reason about
	    - Apologies, reviewers -- it felt necessary
	- Adds static evaluation pass during load to reduce expression complexity during runtime to only the dynamic components
	- Adds support for functions with no arguments, a fixed number of arguments, or infinite arguments

### DIFF
--- a/plugins/custom-item-hovers
+++ b/plugins/custom-item-hovers
@@ -1,2 +1,2 @@
 repository=https://github.com/geel9/runelite-custom-item-hovers.git
-commit=2f77348646ee11b5885584256932efae08d8465e
+commit=afa69698ade3271fda17a96737b1256a54b17f55
--- a/plugins/custom-item-hovers
+++ b/plugins/custom-item-hovers
@@ -1,2 +1,2 @@
 repository=https://github.com/geel9/runelite-custom-item-hovers.git
-commit=afa69698ade3271fda17a96737b1256a54b17f55
+commit=4ce92b2aca10eeba3df0f27586d89d5b995164aa
--- a/plugins/custom-item-hovers
+++ b/plugins/custom-item-hovers
@@ -1,2 +1,2 @@
 repository=https://github.com/geel9/runelite-custom-item-hovers.git
-commit=82fe4f0f8b1f44f26cbdc3a1964724ec3cb9ecf6
+commit=2f77348646ee11b5885584256932efae08d8465e
--- a/plugins/custom-item-hovers
+++ b/plugins/custom-item-hovers
@@ -1,2 +1,2 @@
 repository=https://github.com/geel9/runelite-custom-item-hovers.git
-commit=4ce92b2aca10eeba3df0f27586d89d5b995164aa
+commit=88becd5cdd4919f79074f3a5704e1ffbe8e8859e


pajlada:
## Questions
https://github.com/geel9/runelite-custom-item-hovers/compare/82fe4f0f8b1f44f26cbdc3a1964724ec3cb9ecf6...geel9:88becd5cdd4919f79074f3a5704e1ffbe8e8859e#diff-8d344f20a984928a674df1cfbeb850621acd36fcfd98f97e01e31ffeb24e73dbR151
If the Compiler.Compile commands succeeds, but the EvaluateStatically call fails, compiledExpression can point at the output of Compiler.Compile - is this fine?

------------------

Do you have any examples or ideas of what would make use of the varbit/varplayer/varc functions?

------------------

https://github.com/geel9/runelite-custom-item-hovers/compare/82fe4f0f8b1f44f26cbdc3a1964724ec3cb9ecf6...geel9:88becd5cdd4919f79074f3a5704e1ffbe8e8859e#diff-65bff2ef85ace558d974f8428a34730c407286f7856746553ae061e73520dd4aR137
client.getVarbitValue can throw an exception if the varbit doesn't exist, have you taken this into consideration?

## Non-blockers
https://github.com/geel9/runelite-custom-item-hovers/compare/82fe4f0f8b1f44f26cbdc3a1964724ec3cb9ecf6...geel9:88becd5cdd4919f79074f3a5704e1ffbe8e8859e#diff-c7fb0ec88e87aff313c3af1fc65c9327162d2f5c4908993cb555c084f1084e09R326
you updated the logic but didn't update the exception text

------------------

https://github.com/geel9/runelite-custom-item-hovers/compare/82fe4f0f8b1f44f26cbdc3a1964724ec3cb9ecf6...geel9:88becd5cdd4919f79074f3a5704e1ffbe8e8859e#diff-65bff2ef85ace558d974f8428a34730c407286f7856746553ae061e73520dd4aR56
you could move this to below the fixedFunction check

------------------

This is not new with this PR, but you're doing file reading/parsing _inside_ the render function of your overlay, which can throw exceptions. This is not fatal anymore but you should consider doing this in an executor thread where possible. I understand evaluating might require client threads since you're using client-thread-only functions there.

PR #8061, opened 2025-05-21
Update schboop

katlande:
Apologies for updating again so immediately. I have to replace the newest .wav files with 16-bit ones as the plugin cannot handle the 32-bit version I had uploaded last commit.

I also removed a niche joke from the plugin that I didn't think anyone knew about, but the community noticed within an hour and is now requesting I restore it lol.


tylerwgrass (approved):

LlemonDuck:
also, instead of using regex just use `string.contains("meow") || string.contains("Meow")` there

PR #8062, opened 2025-05-21
Add Sote Ball Highlighter

1jz:
This allows players to add an outline highlight to Sotetseg projectiles so they can see them under the boss which can be hidden due to render order. The plugin highlights default sote balls along with the ToA and Inferno themes from the TOB QoL plugin. The plugin only highlights these projectiles within the Sotetseg instance.


tylerwgrass:
Plugins that provide additional visual indicators for boss mechanics are prohibited:

PR #8066, opened 2025-05-21
Add Ironman Planner plugin to Plugin Hub, All credit to Boaty

Ceepee-OSRS:
I've used Boaty's ironman V2 guide as the basis for this tracker plugin, it tracks progress and saves the progress exactly the way it does within the OSRS Wiki i'm hopefuly if accepted then it will then be able to be used within runelite instead of the OSRS wiki.


cdfisher:
For information on how to properly submit a plugin to the hub, see the plugin-hub README [here](https://github.com/runelite/plugin-hub?tab=readme-ov-file)

raiyni:
You will also need to rewrite this to not produce thousands of config keys.

tylerwgrass:
Closing as this will need a fair bit of work before being accepted.
Please review the comments above and you can open again when addressed.

PR #8070, opened 2025-05-22
disable tictac7x-charges

Alexsuperfly:
multiple reports of confused users having all their infoboxes hidden, even ones not created/managed by this plugin.


TicTac7x:
https://github.com/runelite/plugin-hub/pull/8082

PR #8072, opened 2025-05-23
Add Shortest Clue plugin

KeiranY:
Plugin that integrates with the Shortest Path plugin to find and display the shortest path to your current clue scroll location.


LlemonDuck:
Since your config is empty, you should remove it to prevent a gear icon appearing beside your plugin entry.

PR #8075, opened 2025-05-23
Update glassman to v1d

Mattb2291:


iProdigy (approved):
nit: double-brace initialization should be treated as an anti-pattern (use Map.of instead)

https://github.com/Mattb2291/Glassman/blob/4511693ec6b6c1baa9f7add505c676e693e26d94/src/main/java/com/glassman/GlassmanPlugin.java#L47-L51

https://stackoverflow.com/a/27521360

iProdigy (approved):

riktenx (approved):

PR #8077, opened 2025-05-23
Update pvp-performance-tracker to v1.7.0

Matsyir:
Yet another new statistic! Along with various improvements and fixes.

The new statistic "hits on robes", it tracks how many times you were hit by melee/range while in robes. You don't want to get hit in robes. You can configure if this looks at only robe bottom, top, both, or either.
Screenshot 2025-05-23 185915

Also added a very basic new Frame/Window to display a summary of all the attacks used, in a more concise way than the fight log.
Screenshot 2025-05-23 185259

Mostly changes from these PRs, many thanks to the contributors! More details in regards to the various improvements/fixes can be found in the PRs.

https://github.com/Matsyir/pvp-performance-tracker/pull/54
https://github.com/Matsyir/pvp-performance-tracker/pull/55
https://github.com/Matsyir/pvp-performance-tracker/pull/56

### DIFF
--- a/plugins/pvp-performance-tracker
+++ b/plugins/pvp-performance-tracker
@@ -1,2 +1,2 @@
 repository=https://github.com/Matsyir/pvp-performance-tracker.git
-commit=efcaa7a79044b171edbec1bfcb696ce65a9c364c
+commit=b8d457a9b33e6f5df9734a0a0a4551b6586c70a0


Matsyir:
Alright I included the 1.6.3 threading crash fix in this, plus some other minor things, and also fixed the multi-commit nonsense 👍

Matsyir:
Alright, should be ready again.

No more silly calc loops in Render().

Also, some recent API changes seem to have made it so we don't even need to use that @ Schedule thing for the hitsplat/healthbar delays anymore (https://github.com/Matsyir/pvp-performance-tracker/issues/57), so that's also simply been removed now.

Still some small optimizations possible, but nothing major. It's already currently live with some stuff in Render() and seemingly works fine.

PR #8079, opened 2025-05-24
Add bingosrs plugin

mistereman22:
This plugin is for use alongside the BingOSRS website: https://bingosrs.com/.

It will grab data on required drops from the server based on user configuration (bingo ID), and then listen for drops and submit a screenshot along with drop data when the player receives one of the required drops.

### DIFF
--- a/plugins/bingosrs
+++ b/plugins/bingosrs
@@ -1,0 +1,4 @@
+repository=https://github.com/mistereman22/bingosrs-runelite-plugin.git
+commit=728f3184c6bc8924c13760bf0794bb53d4d6a83c
+authors=mistereman22
+warning=This plugin sends player data, including usernames, screenshots, and IP address to a third-party server not controlled or verified by the RuneLite Developers, for tile verification.


cdfisher:
Have you considered collaborating with the authors of any of the existing bingo and/or Discord webhook plugins? There are already a lot of those on the hub, and (just based on your description as I haven't looked at the plugin code) at least a large subset of your functionality is already stuff that's doable with something like Dink.

mistereman22:
> Have you considered collaborating with the authors of any of the existing bingo and/or Discord webhook plugins? There are already a lot of those on the hub, and (just based on your description as I haven't looked at the plugin code) at least a large subset of your functionality is already stuff that's doable with something like Dink.

I looked into utilizing Dink but it doesn't easily fit the use case for a couple reasons, I haven't fully looked into all discord webhook plugins but the ones I have are similar:
- Lack of support for custom data models.
- Lack of support for arbitrary item lists.

I think both of these can be rigged up to work in base Dink (using dynamic config url and modifying the API endpoints to accept the payload that Dink sends) but requires the user to set fields that are not necessarily intuitive to non-developers. Also (maybe more importantly), this plugin currently only supports submitting drops, but the intention is that it will support other functionality (e.g. displaying bingo information) in the future that falls outside the scope of what Dink does.

As for existing bingo plugins, https://github.com/bingoscape/bingoscape-companion is the most similar, but I don't think there's enough overlap to warrant cramming both into a single plugin.

Right now, this plugin does duplicate some logic from Dink for identifying drops from different events (e.g. `ServerNpcLoot`, `NpcLootReceived`, `LootReceived`). From my perspective, an ideal solution would have that logic output available to consume in my plugin. I hadn't looked at any plugin code prior to a week ago though, so I'm likely missing a lot of context too.

LlemonDuck:
If it helps your own code organization, you can register classes to the EventBus directly by injecting the EventBus to your plugin and using EventBus#register. Just make sure to unregister in shutDown. Only the plugin class is handled automatically.

You should also consider OkHttpClient#enqueue to get async for free, but it's not a blocker since you are still using it async.

PR #8080, opened 2025-05-24
add yama-blindfold

rdutta:
- Hides scenery (GameObject and GroundObject removal from tiles w/ hardcoded whitelist)
- Hides terrain (custom DrawCallbacks w/ hardcoded whitelist of scene tiles to draw)
- Hide fade transitions (modifies fadeout cs2 script args)
- Only hides objects/terrain that are loaded prior to boss encounter start

demo


Adam-:
Hi. Twice you have submitted plugins & then deleted them (the gauntlet and underwater agility, under the alias "roundshoe") after they gain some popularity. This is extremely disruptive for us and causes me to have to promote unwilling people to be maintainers of your code. As such, I am no longer willing to accept plugins from you.

PR #8082, opened 2025-05-24
Update tictac7x-charges to v0.6.5

TicTac7x:


raiyni:
You should consider writing your base classes so that it's not possible for one of your infoboxes to be an issue in the future. Considering how many there are, it's kind of hard to spot these issues at a glance and it's taken us a while to find someone with a client log.

PR #8083, opened 2025-05-24
adding plugin Ti-Kai la mitraille with sounds

Steevvy:
Plugin de notifications audio personnalisées. Joue des sons lors de trades, pets, level-up, mort, prière à 0, etc.
Tous les sons `.wav` sont inclus dans `src/main/resources/` et embarqués dans le `.jar`. 
Ce plugin ne fait appel à aucune ressource externe ni interaction disque hors du dossier embarqué.

### DIFF
--- a/plugins/attack-ranges
+++ b/plugins/attack-ranges
@@ -1,2 +1,2 @@
 repository=https://github.com/tylerwgrass/attack-ranges.git
-commit=75243d497d5121813f0b81aa19682527ccaab0f1
+commit=364602f61c01c3b61e1752c8d9007063807b01c9


riktenx:
This is not a valid plugin hub submission. Plugins live in their own repositories.

Please review https://github.com/runelite/plugin-hub?tab=readme-ov-file#creating-new-plugins.

riktenx:
The actual functionality in this plugin also seems to be covered by several other existing ones on the hub such as [watchdog](https://github.com/adamk33n3r/runelite-watchdog). We'd rather not have duplicates. If those plugins are missing features you're looking for then we'd prefer you contribute to them instead.

PR #8084, opened 2025-05-24
Ajout du plugin Ti-Kai la mitraille

Steevvy:
Ce plugin joue des sons personnalisés lors de certains événements comme les trades, drops de familiers, mort, prière vide, etc. Les fichiers .wav sont embarqués dans le .jar.


Steevvy:
Totally understand your decision — thank you for the feedback.

My intention wasn't to duplicate existing functionality, but rather to have my own set of personalized sound cues to better distinguish specific drops and events during gameplay.

I’ll explore whether contributing to existing plugins like Watchdog makes sense for those needs, or rethink how I can bring something more unique.

Appreciate the review and all the work you do on RuneLite

PR #8085, opened 2025-05-25
Update sote-wall-remover

Need-Femboy:
Adds Gauntlet support

### DIFF
--- a/plugins/sote-wall-remover
+++ b/plugins/sote-wall-remover
@@ -1,2 +1,2 @@
 repository=https://github.com/Need-Femboy/SoteWallRemover.git
-commit=6a2974797c8a42e36434e15bfbaeff7a7885e960
+commit=6daaca43c90717e5ddf49fb1223caba49d4938bb


Felanbird:
We've gotten a lot of people in the discord confused why their nex walls were suddenly gone, as they installed this plugin as a plugin for sote walls, so I would prefer any new additions be locked behind disabled-by-default configs.

Need-Femboy:
I added a quick config, I'm not really sure what is best practice for them so I just went with what works

riktenx (commented):
re: [onConfigChanged](https://github.com/Need-Femboy/SoteWallRemover/compare/6a2974797c8a42e36434e15bfbaeff7a7885e960...Need-Femboy:2eed7d09d6919f4a44e75491f1796f342d6ca5f7#diff-5070c13e7b6a8f12ce396e9526eb6254ea0985d610dc101bb1273c81a187e0d5R76):

There should probably be player coordinate checks here, right now you're always going to  force a reload whenever someone toggles one of your new configs off (even if they're not in a relevant zone and thus no actual reload is required).

riktenx (approved):

tylerwgrass:
This will now show nex walls for everyone, who might expect it to be on. You can do this if you really want to but it would be preferred if either:
1. You display a one-time message or something to let people know this is happening
2. Set the default config for nex to true

PR #8086, opened 2025-05-25
Update Embargo to 1.4.6

Sharpienero:
- Adds a command to display clan rank inside of the text chat.
    - There is a cache server side that caches the original request for n seconds to allow the command to be processed by all clients running 1.4.6

image

Configurable in the plugin settings: 
image

### DIFF
--- a/plugins/embargo-clan-plugin
+++ b/plugins/embargo-clan-plugin
@@ -1,3 +1,3 @@
 repository=https://github.com/Sharpienero/Embargo-Plugin.git
-commit=b9434a87cf6d3afb664525aa790018452b3ef770
+commit=bfb2084cb10ebbb853f0b4c3f8bf75b05d65ea3f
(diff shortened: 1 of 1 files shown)


patrickgillan:
Hey, just wanna apologize for pushing another commit 2 weeks after the original but one of my developers found a bug with the previous command code that we wanted to address sooner rather than later.

I also noticed that after adding the new commit, the SoundManager code above popped. Rather than address that, I'd rather remove the code completely if that makes the PR acceptable.

Sharpienero:
Hey, just wanna apologize for pushing another commit 2 weeks after the original but one of my developers found a bug with the previous command code that we wanted to address sooner rather than later.

I also noticed that after adding the new commit, the SoundManager code above popped. Rather than address that, I'd rather remove the code completely if that makes the PR acceptable. What do you recommend regarding that?

LlemonDuck:
Don't worry about pushing things, we'll let you know if we want you to freeze the PR

PR #8106, opened 2025-05-27
Update markofdarknesshelper plugin to 1.2

Nick2bad4u:
Updated plugin version to 1.2 to include new features:

Notification Timeout config
Notification color settings
Fixed robustness of timer


riktenx (approved):
Overlays are not typically stateful, although if your plugin is small and they're just scalar types then it's not the end of the world.

Nick2bad4u:
> Overlays are not typically stateful, although if your plugin is small and they're just scalar types then it's not the end of the world.

It's just based off chat messages. I did have a timer built directly based off magic level and such but it was easier to just track the chat messages.

Nick2bad4u:
> Overlays are not typically stateful, although if your plugin is small and they're just scalar types then it's not the end of the world.

Its just based off chat messages. I did have a timer built directly but it was easier to just track the chat messages.

tylerwgrass (approved):

PR #8107, opened 2025-05-27
Slayer Master Plugin

QualitySushi:
A RuneLite plugin that reminds you to visit the correct Slayer Master at task milestones to maximize your slayer point rewards.

### DIFF
--- a/plugins/slayer-master
+++ b/plugins/slayer-master
@@ -1,0 +1,2 @@
+repository=https://github.com/QualitySushi/slayer-master.git
+commit=3c69950aa4fc8025011bb875957abd1bd404b812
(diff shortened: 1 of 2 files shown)


QualitySushi:
Oh my bad, as you can tell, still fairly new to this. I was just trying to throw together some suggestions from friends quickly, but I'll have to take some time later and properly read through everything. Thank you for the heads up!

Also yeah it's already a fair bit different as the way I'm tracking slayer tasks is a bit different and it's more customizable, as you can swap which slayer master you get for each milestone, as well as personalize a small overlay with customizable text, etc etc. I'm also still getting more feedback and planning out more additions so I believe it will be significantly different, if it's not enough already. Will be happy to share some pictures if you'd like once I actually get stuck in

Appreciate the patience!

PR #8109, opened 2025-05-27
Update log-master

ImTedious:
Update task list
Fix compatibility with other plugins


ImTedious:
Added another change to the log-master plugin to fix scrolling through tasks in resizable

tylerwgrass:
Can you move the plugin transfer to its own PR please?

PR #8110, opened 2025-05-27
Clan Spam Filter V1.2

andreasx23:
1. Refactored code for improved maintainability of the plugin
2. Added an icon to enhance the plugin’s visual identity
3. Revised names and descriptions for various configuration settings
4. Introduced a new setting: "Filter Combat Diary Tasks"
5. Added additional tags to make the plugin easier to find in searches
6. Updated the plugin description for clearer explanation of its purpose

Edit:
7. Fixed a bug where the total level-up broadcast didn't extract the correct text for long.parse

### DIFF
--- a/plugins/guild-spam-filter
+++ b/plugins/guild-spam-filter
@@ -1,3 +1,3 @@
 repository=https://github.com/andreasx23/GuildSpamFilter.git
-commit=48e35d51b58ebc6a0cf07472c99983498ead85da
+commit=a2383bcff57080c0247ac37a02f26563e4fc039d
 authors=andreasx23


riktenx (approved):
I am not usually one to nit code but in this case it pertains to the reviewability of your patches:

You can greatly simplify `shouldFilterCombatDiary` and `shouldFilterAchievementDiary` using enum ordinal comparison and `String.format`. I recommend you DRY because the length of the patch you give us is directly proportional to how long it takes us to get to/review it. That is not an invitation to cram things onto a single line, just inviting you to think about it since it will benefit everyone involved.

PR #8121, opened 2025-05-29
Create CARoadmap

itohti:
This plugin will fetch relevant player stats, kill counts, and combat achievement completions. This plugin also uses Wise Old Man API to fetch EHB, and boss personal bests. All of this data is stored into a cloud database and then used to run a machine learning algorithm to recommend a list of combat achievements to reach a point threshold the player inputs in the configuration.


raiyni:
This is outside our scope but your web scraper should be setting a user agent to let WOM/Wiki know your plugin is doing that scraping.

PR #8125, opened 2025-05-30
Create ThemeLite

Yenof:
A plugin that recolors elements of the client UI. Kinda silly.

Are you tired of RuneLite's sleek, functional, color-coordinated interface? Ruin it all with ThemeLite!

themeLitepicfix


raiyni:
Personally I don't see this being added, but it wouldn't be feasible anyway if you are clobbering the theme and unable to reproduce it.

PR #8126, opened 2025-05-31
Create in-range

thebeanbag:
Checks the range of equipped weapon, when you hover over an attackable npc it'll highlight them in either green, yellow, or red. Green if they are in attack range, orange is you are in range but don't have line of sight, and red if out of range. Have it configurable so you can just enable it in relevant content like inferno or colo.

### DIFF
--- a/plugins/in-range
+++ b/plugins/in-range
@@ -1,2 +1,2 @@
 repository=https://github.com/thebeanbag/In-Range.git
-commit=45e8744d64cba8b2c44a2e4bc4f2650abf6bf347
+commit=875abfb2f7d29775a729d718a31e2ab786f007d0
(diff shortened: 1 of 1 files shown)


tylerwgrass:
The [attack ranges](https://github.com/tylerwgrass/attack-ranges) plugin does most of what you are looking for.
I am open to contributions if there is something you would like implemented

tylerwgrass:
I am going to close this request but feel free to take a look at mine and open an issue or pr if you want to add something and we can discuss there.

> I do like the simplicity of my implementation in that you can quickly hover over an enemy to see if youre in attack range and its pretty unobtrusive outside of that.

This would be a welcome addition for example

PR #8128, opened 2025-05-31
Update time-to-max to v1.0.5

Hamelot:
- correct tooltip in side-panel
- remove unused methods
- logic correction on goal xp
- removal of unnecessary XpInfoBox logic
- addition of endXp to XpSave


Hamelot:
Going to close this as I found some issues with this build I want to fix first

PR #8134, opened 2025-06-01
Create yama-phase-reminder

AbusiveTuna:
Displays an info box that tells you what phase you are currently on.

### DIFF
--- a/plugins/yama-phase-reminder
+++ b/plugins/yama-phase-reminder
@@ -1,0 +1,2 @@
+repository=https://github.com/AbusiveTuna/yamaPhaseReminder
+commit=9bfaac59fc1a83a6f79d3aa7583a12a91c549e2b


AbusiveTuna:
I disagree, the transitions between phases aren't minor and can't be possibly unknown. You are teleported to another area, complete a small side fight, then teleport back. 

I would agree with you if there was no transition between phases, and someone not paying attention could suddenly be facing different mechanics without realizing the phase transition happened. However that just isn't possible for this fight.

Felanbird:
> I disagree, the transitions between phases aren't minor and can't be possibly unknown. You are teleported to another area, complete a small side fight, then teleport back.

It seems like you've just argued for why this plugin should not be added? Your point is that it's very obvious that you have phase transitioned, so there's no point in being told what phase it is.

AbusiveTuna:
Then why are others with similar utility allowed? It's obvious that you need to re-summon a thrall, or death charge, or to know what olm phase we're on? It's nice to have, and I see no reason why we can't have the same here.

PR #8142, opened 2025-06-02
Soulflame airhorn

Brankhout:
This plugin replaces the normal special attack sound of the soulflame horn with 4 different airhorn sounds


raiyni:
This can be accomplished with the sound swapper plugin?

PR #8143, opened 2025-06-02
Soulflame airhorn

Brankhout:
Hi! sorry for the spam earlier, i f'ed up.

This plugin Replaces the default spec sound of the soulflame horn with 1 of 4 random airhorn sounds
I made this instead of using the sound swapper plugin to have the random effect.
it has the same purpose as the ImbuedFart plugin


LlemonDuck:
Use AudioSystem instead of managing the clips yourself.

Also the config item description still says fart.

If you're copying code from imbued fart, you should credit the author in your license file as well by adding `Copyright (c) 2022, Himonn` beside your own line.

PR #8145, opened 2025-06-02
Add Tick Beats Metronome Plugin

HarperDevLab:
This plugin adds a metronome with advanced audio and visual customization options not found in existing metronome plugins.

Features:

- A synchronized audio and visual metronome
- More customizable sound options for a richer, better-sounding metronome experience
- Optional tick smoothing: plays ticks when they *should* occur based on average game tick rhythm, ignoring jitter or lag. Ideal for situations where averaged timing is preferred over absolute server accuracy (that includes lag).


tylerwgrass:
It would be preferred that you contribute to an existing plugin instead of creating a new one when most of the feature set overlaps.
Also when would you not want the metronome to match the server?

HarperDevLab:
rereading, I worded this badly:" Every other metronome seems to just have specialized use cases and less functionality" by less functionality, I meant they're more specialized whereas this functionality is meant to be more general use anywhere tick timing may be helpful.

I should also mention that having more than 2 sound effects is a nice feature in that you can align different actions with different sound effects which also helps when getting into a flow for tick based activities that are more than just 2 steps

LlemonDuck:
downgrade your log.infos to log.debugs

HarperDevLab:
I'm not entirely sure what you mean, can I have a bit more info on what isn't allowed in the log file and why?

Felanbird:
Change the `log.info` lines to `log.debug` so they don't spam the users logs during general use. The logging will still work when you run the client with `--debug` during development.

HarperDevLab:
oh gotcha, thank you, will do!

LlemonDuck:
writing content to a file every single game tick is going to cause choppiness. What's the point of the log file for end users anyway? Can this just be a log.debug call, or can it be removed? Can it be gated behind development mode?

If you want to keep it, put it on an async thread

HarperDevLab:
Thanks for the feedback on making log file writes async, that makes way more sense.

I figured if there was gonna be a problem with the plugin it would be the log file, that's why when you mentioned log.infos to log.debugs, after not looking at the plugin for a couple weeks my mental reflex was that you were talking about the log file having too much info, over the obvious console logging (I usually use C# where it's usually just Console.WriteLine()), I felt like a bit of a dingus afterword.

The log is primarily for a companion music player application I'm working on that will play music (any music that's a steady 100 beats per minute) in time with game ticks . 

The original goal of the plugin was to allow tick playing in time with music, but then I learned that audio playing is now limited to runelite's audio library which is too limited to allow for what I was aiming for, (it only plays an audio clip from start to finish with little to no control over the clip during playback) so a companion app that has more advanced audio functionality, and can keep in sync with game ticks seemed like my next best option, and I decided the likely most compliant way to share game tick info with that app would probably be just to make a simple text log file that would contain any info the app might need to stay in sync. That way the application doesn't directly interact with Runelite in any way, it just reads the text file it creates.

As far as the music player application, I've been testing with Sea Shanty 2 and a modified inferno track that runs at 100bpm, and I feel like it makes tick playing a lot more fun and engaging, (gives it a bit of a rhythm game feel). The plan is for the app to be portable (no install, just download and run) open source and hopefully cross platform.

Syncing manually to game ticks is a bit of a pain to setup and get accurate, to the point where I think most people wouldn't bother, having the log file to sync up automatically would be awesome, so players can just start up the application and automatically have the music play in sync with Runescape game ticks, and I'm hoping now that writing the file is async it won't be a problem.  The music playing application should be in a good proof of concept/ minimally viable state early this coming week.

LlemonDuck:
Do you want to look into the above or shall I review the plugin as-is?

LlemonDuck:
It can live in your main repo too if you want, just not in the src/ directory. Or it could be in a separate branch. Use an injected OkHttpClient, not java.net. If you zip everything then you don't need to worry about rate limiting (not that github really seems to care either). Just download and unzip into .runelite/tick-beats or something.

HarperDevLab:
I ended up doing 1 download at a time for a bit more granular control, the plugin checks what files are missing and downloads them individually. I also did end up throttling the downloads just in case, aiming for better safe than sorry. The user will get an audio message saying that files are downloading if they select a song that hasn't been downloaded yet, all in all the low quality audio should download within 15-20minutes. (If the user checks that they want to download the higher quality versions, the low quality version will play until the high quality download is ready so I figure there's not too much of a rush on those)

Ended up getting into converting some songs into 100BPM versions that will work with the plugin, some were quick and easy, others turned into time investments, hopefully 25 built in songs is good enough for the first version of the plugin.

Let me know if there are any questions, hopefully I didn't mess anything up and everything is looking good to go.

HarperDevLab:
Was hoping to get it done a bit earlier today, but just finished, I ended up getting some ideas on where text overlay messages could be useful and added them there as well. I feel like the thing that could likely trip up users the most is making sure wav files are 16-bit. In making the plugin, I stumbled on some 24 bit sounds which for an end user would just cause things to fail silently.

Also added some download information to the tick beats info box to give users a better idea of what's going on with downloads while they're happening, once everything is downloaded they're meant to disappear as to not clutter up the ui.

All in all I do think the overlay messages are a better user experience, giving better tailored feedback when something isn't working right.

Rushed this one a bit at the end there, and being past my usual bedtime, I'm hoping I didn't do anything silly

PR #8154, opened 2025-06-03
Update skills-tab-progress-bars to v1.4.1

m0bilebtw:
Remove migration of deprecated configs
Replace use of deprecated RL ID classes
Minor refactors
Run formatter


m0bilebtw:
- Understood- will avoid in future
- The original 'darken' settings were two separate checkboxes (darken at level 99 and at 200m xp), it is now an enum - the migration was simply choosing the correct enum value based on the no longer visible checkboxes. If anyone hasn't run the migration by now, they'll instead have the default darken behaviour selected. No other settings were affected.

riktenx (approved):

iProdigy (approved):

PR #8155, opened 2025-06-04
Update emoji-palette

hjdarnel:


Nightfirecat:
Reflection is not permitted on the plugin hub.

hjdarnel:
Bummer, not sure how else to maintain the plugin. I guess I could keep my own set of triggers and images, but that doesn't seem great. Open to suggestions (I noticed it's broken in the live client now.)

PR #8158, opened 2025-06-04
update time-to-max to v1.1.1

Hamelot:
Utilize state to determine start date for interval reset.

previously the reset would only trigger if you were online crossing an interval threshold.
I unfortunately was only testing in goblin mode and would always have my testing client open across midnight...

### DIFF
--- a/plugins/time-to-max
+++ b/plugins/time-to-max
@@ -1,2 +1,2 @@
 repository=https://github.com/Hamelot/time-to-max.git
-commit=88d809dfb70305af04f10f4490bae0500ad4a30e
+commit=d62edfbe7f767e8e025727b0dfe69121a7c27db2


iProdigy (approved):
consider downgrading this to `log.debug`: https://github.com/Hamelot/time-to-max/blob/d62edfbe7f767e8e025727b0dfe69121a7c27db2/src/main/java/com/timetomax/TimeToMaxPlugin.java#L693

and you could store `LocalDate.now()` to a local variable instead of repeatedly calling it in a single line: https://github.com/Hamelot/time-to-max/blob/d62edfbe7f767e8e025727b0dfe69121a7c27db2/src/main/java/com/timetomax/TimeToMaxPlugin.java#L420

PR #8162, opened 2025-06-05
Update flipping-utilities

AntonyGarand:
Updated tax handling to better account for items bought prior to tax update.


AntonyGarand:
Will re-send and add myself as author to prevent the non-author tag

Felanbird:
that can be done in this PR

AntonyGarand:
Ahh right 🤦my PR game is rusty

AntonyGarand:
I see the build is failing with the terminally deprecated APIs: I guess I'll have to fix these as well

LlemonDuck:
@zumaad Do you agree to adding @AntonyGarand as an author of the plugin? This gives them the permission to submit changes to the plugin hub now as well as in the future.

zumaad:
@LlemonDuck yep, I agree to adding @AntonyGarand as an author of the plugin!

PR #8163, opened 2025-06-05
Update temple-osrs

44Mikael:
Added !col command and collection log auto sync for counts. Both made by @joemckie

### DIFF
--- a/plugins/temple-osrs
+++ b/plugins/temple-osrs
@@ -1,4 +1,4 @@
 repository=https://github.com/SMaloney2017/Temple-OSRS-Plugin.git
-commit=f1651d3b432f76fa16b2d6c74a773d80c2e8bb3f
+commit=c2729269f464d35e6fc357b712226cb6851119f9
 authors=SMaloney2017,44Mikael
 warning=This plugin submits your IP address to a server not controlled or verified by the RuneLite developers.
--- a/plugins/temple-osrs
+++ b/plugins/temple-osrs
@@ -1,4 +1,4 @@
 repository=https://github.com/SMaloney2017/Temple-OSRS-Plugin.git
-commit=cd8f0e826b007498cdeaa6f55671dddf45d7baad
+commit=e51bf78ef8878fa51912530a975faf182977d8e7
 authors=SMaloney2017,44Mikael
 warning=This plugin submits your IP address to a server not controlled or verified by the RuneLite developers.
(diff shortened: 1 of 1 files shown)


joemckie:
A few comments from me on this one:

First off, apologies for another big PR 😅 this will likely be the last of them as our feature set is mostly done now with complete auto-sync capabilities.

A lot of the code that has been added was adapted, with permission, from @voltmaister's [`collection-tracker`](https://github.com/voltmaister/collection-tracker), as we felt it made more sense to include this functionality in a single plugin, rather than requiring players to install (and the devs to maintain) multiple plugins. Hopefully, given a lot of the code already exists in some form in the Plugin Hub, it should ease the burden of reviewing this.

@voltmaister said they'd disable `collection-tracker` when this is released, so it won't be duplicated 😄

A brief rundown of what I've changed here and how it differs from the existing plugin:

- Collection log categories and their items are now derived from the in-game structs, which should reduce the maintenance needed when the game updates (though we may need to tweak it in the future, as it's working off an enum that provides the struct ID). It also means we no longer use the input from the chat command as the title.
- A few changes have been made to the database structure, and it now doubles up as a cache for the current player's log. This is used to determine item diffs from the TempleOSRS API and the in-game log.
- A bug was fixed (or should have been fixed) that caused items to be duplicated in the cache if the API was called again (it was just performing `INSERT` before; now it's an upsert).

The auto-sync works when a player has the `autoSyncClog` config option enabled and opens their log. It uses the local database (which is synchronised with the API data) to determine which items have changed since their last sync, and only calls the endpoint if items have been found. I added a backoff strategy to include retry attempts, incase of unexpected failures (expected failures, like the endpoint returning `403`, will not retry) 😄

Aside from that, I refactored some existing things to avoid messy callback patterns.

I plan to add tests in the near future, but didn't get round to it this time round, as I'm not too familiar with mocking HTTP requests or databases in Java. I presume a PR that just includes tests will be a very quick approval, so will do that separately :)

Thank you! ❤️

LlemonDuck:
@44Mikael the PR sat here long enough to break on some client script engine changes in the most recent update and it'll need to be updated. If you can submit another PR with those fixes, I'll try to have it merged as soon as possible.

joemckie:
@LlemonDuck we have a fix for that on the way 😄 thanks!

PR #8166, opened 2025-06-05
Collection tracker is getting Removed

voltmaister:
The collectrion tracker plugin is getting removed as it gets integrated to templeosrs directly #8163


plugins/collection-tracker
```diff
@@ -1,3 +1,4 @@
 repository=https://github.com/voltmaister/collection-tracker.git
 commit=a65dd733b1a58a35ec6823ed1658cc8207423e4c
 warning=This plugin submits your username and IP address to a server not controlled or verified by the RuneLite developers.
+disabled=due to merge with templeOSRS plugin
```
LlemonDuck, line 4:
```suggestion
disabled=Use TempleOSRS instead.
```
This message is shown to users on the plugin hub panel, so an explicit alternative is useful.

LlemonDuck (changes requested):

PR #8168, opened 2025-06-05
update Preventsgm

Zippilipy:
Update prevent misclicks (preventsgm before) so that if you have sulphuorous essence you cannot teleport away. The way I've implemented this is really horrible, since I don't know of any good way to check if an event is teleporting you somewhere, so I just made a huge list of teleport items and check if you clicked it... If u think of a better solution please let me know


tylerwgrass:
Can you include info in your chat message that says which plugin is sending the message? (maybe just prefix with your plugin name or something)

Zippilipy:
I have added this in https://github.com/Zippilipy/prevent-sgm/tree/929871cb03558825f2db062ff28589c96fd87ea6. I'm not really sure how to change a merge request though.

PR #8184, opened 2025-06-08
Update auto-retaliate-warning

ste-h:
Added more config options - custom text for each warning, and custom text colours for each warning


riktenx:
Your group in build.gradle no longer matches what your plugin is actually using.

riktenx (approved):

PR #8185, opened 2025-06-08
Create rogue-silencer

Jbasilious:
Toggle to hide rogues models or overhead text at the rogues castle to make thieving rogues chests less annoying.


coopermor:
This is already an option in [Public Chat And NPC Overhead Chat Remover](https://runelite.net/plugin-hub/show/public-chat-and-npc-overhead-remover).

PR #8190, opened 2025-06-09
Update chugging-barrel to v1.3.0

Syhlex:
v1.3.0 changes:
- Adds reminder to bank barrel after use (https://github.com/Syhlex/chugging-barrel/issues/4)

Screenshot 2025-06-08 222943


LlemonDuck:
fwiw if you can remove the dependency on guice-multibindings, your plugin will probably scan green and won't need maintainer review. Then you should get quicker reviews.

You can replace it with e.g. ```
@Provides
public List<PluginLifecycleComponent> provideComponents(
    PLC1 plc1,
    PLC2 plc2,
    ...
) {
    return List.of(plc1, plc2, ...);
}
```

PR #8198, opened 2025-06-09
Update linux-custom-cursor

nwerblun:
Integrating memory leak fix from rcnoob
Changed deprecated Inventory ID to new version
Added mirroring across x/y options for images that are not in the correct direction by default.
Fixed mirroring implementation to be cached and not have memory leaks.


nwerblun:
Added a requested feature to have scaling as an option, as UI scaling affects the cursor size. Also updated readme and picture used in the readme.

PR #8199, opened 2025-06-10
Complete overhaul of clean-chat

ldavid432:
- Totally refactor channel name removal
  - It now edits the chat widgets instead of re-sending chat messages as different types
- Chat message blocking has mostly stayed the same
- All user-facing functionality should remain the same

Sorry for the huge PR but this needed to be done, the old method was not looking very sustainable. I kept having to fix small bugs with not scalable methods.

### DIFF
--- a/plugins/clean-chat
+++ b/plugins/clean-chat
@@ -1,2 +1,2 @@
 repository=https://github.com/ldavid432/chat-cleanup.git
-commit=7b59f8649533fce047abb02ca8eac129aa001be7
+commit=42cd64db187b837d4114f1fbb0b30e4a9ff5cbd7


ldavid432:
Closing cause I noticed some bugs. Also going to try and rework it so the diff isn't as bad

riktenx:
Aren't [these sets](https://github.com/ldavid432/chat-cleanup/compare/7b59f8649533fce047abb02ca8eac129aa001be7...ldavid432:0e3c565672d9db4d3756c930e01ae18e14ac0bd4#diff-407df4d5ec02c0109b9576e1a09525a5edb9beb6fe0eebc53f784ab9bb00e3b5R31) leaking entries? You're never clearing them as far as I can see, just adding to them when channels change.

ldavid432:
> Aren't [these sets](https://github.com/ldavid432/chat-cleanup/compare/7b59f8649533fce047abb02ca8eac129aa001be7...ldavid432:0e3c565672d9db4d3756c930e01ae18e14ac0bd4#diff-407df4d5ec02c0109b9576e1a09525a5edb9beb6fe0eebc53f784ab9bb00e3b5R31) leaking entries? You're never clearing them as far as I can see, just adding to them when channels change.

I guess I should clear them on shutdown yeah, but it's intentional that they're not cleared anywhere else so that even if you have old messages from a chat you left in your history they'll still get processed

riktenx:
Just clear them _on_ channel change / leave?

riktenx:
Then you're going to have to maintain a history of channel names, and stale out old ones when all of the messages from that channel scroll out of the client's own history.

My point is, you can't just have a set that grows unbounded. Unless another reviewer/maintainer comes in here and says that's okay.

riktenx:
Even a sane cap would be fine (like, 128 channels, and then if they literally swap channels that many times they'll just have to deal with it). My issue is the unboundedness. There's a big difference between cap of N and no cap.

ldavid432:
Updated with a cap of 128

riktenx (approved):

PR #8202, opened 2025-06-10
adding dry-rate-tracker plugin

georgebunn97:
A RuneLite plugin that tracks your dry streaks for OSRS raids (Theatre of Blood, Tombs of Amascut, and Chambers of Xeric).

Originally came about as my friend kept saying he was '150' raids dry, but we had no way to check this and tell him he's wrong..

### DIFF
--- a/plugins/dry-rate-tracker
+++ b/plugins/dry-rate-tracker
@@ -1,0 +1,2 @@
+repository=https://github.com/georgebunn97/dry-rate.git
+commit=fff8b4697c7600135cf99c865b1f067496cf08d2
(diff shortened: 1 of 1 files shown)


LlemonDuck:
downgrade logging statements from log.info to log.debug

consider using lombok to simplify

don't mix `@Inject` on a constructor and fields in the same class

georgebunn97:
> downgrade logging statements from log.info to log.debug
> 
> consider using lombok to simplify
> 
> don't mix `@Inject` on a constructor and fields in the same class

Thanks @LlemonDuck! I _think_ I've fixed those issues

PR #8215, opened 2025-06-12
Watchdog v3.10.1

adamk33n3r:
- add cardinal option to location alert
- add fix for dink's regression
- add title option to popup
- update popup to use capture groups
- update location alerts to handle instances

### DIFF
--- a/plugins/watchdog
+++ b/plugins/watchdog
@@ -1,2 +1,2 @@
 repository=https://github.com/adamk33n3r/runelite-watchdog.git
-commit=9fff05d80751b2264721275b42ead9288b9bdde1
+commit=025bda837ef92e0f5bd7ba9836a81ae929572eca


riktenx:
We've started hosting our own build of jaco-mp3 on rrn, I've just cut you a PR to switch over to that: https://github.com/adamk33n3r/runelite-watchdog/pull/184 -- please double-check that everything is working as you expect and then include that patch in your release here.

re: feedback about AudioPlayer: you can ignore that for your plugin for the time being. It's going to flag those APIs as being used anyway because of jaco. For your direct use of Clip, I don't think AudioPlayer supports everything you need to replicate your current behavior 1:1 so you can leave that be for now as well.

adamk33n3r:
Oh thanks, I'll take a look. Where is that updated version of jaco from? Or is runelite developing on it itself? Just wondering cause I thought I had the newest version. But ya looking at the new AudioPlayer I don't think it exposes enough for it to be used by me. I know I at least do my own queuing and stopping etc.

PR #8220, opened 2025-06-12
Update steps-per-clue

Chubs1:
Fixed the issue with using methods that not longer exist, also fixed it to only run when you check steps on clue instead of always running on any chat message like it was before.


Chubs1:
uhhh i was gonna use it and imported it but they i didn't oops

Chubs1:
Okay yeah sorry i have 0 idea what im doing

tylerwgrass (approved):

iProdigy (approved):

PR #8233, opened 2025-06-15
Add Unbonded Log plugin

Xi2ediviist:
Adds a plugin that hides members-only Collection Log sections for pure F2P players.


cdfisher:
Instructions for how to properly submit plugins to the Plugin Hub can be found [here](https://github.com/runelite/plugin-hub?tab=readme-ov-file#plugin-hub). LLMs are not going to give you a properly structured hub plugin.

PR #8238, opened 2025-06-16
Create alch-copilot

rcnoob:
Alch Copilot: a more profit-driven fork of ABC Alchemy

The plugin will recommend the player items to purchase and HA based on multiple variables set in the configuration

### DIFF
--- a/plugins/alch-copilot
+++ b/plugins/alch-copilot
@@ -1,2 +1,2 @@
 repository=https://github.com/rcnoob/alch-copilot.git
-commit=eb12912db2f312ad2cd24efa0672f3f538562f46
+commit=7cfdcc9dcb63f1736e5c812c7c8e69616b998574


raiyni:
Can you please contribute to the already existing alch profit plugin?

raiyni:
Again, we prefer you contribute to existing plugins due to hub bloat and user fragmentation. If the owner is awol, there are transfer options for the plugin. But completely rewriting for no reason isn't looked at as a valid reason to duplicate plugins if they serve the same purpose. 

As such, what is different? It looks like your plugin downloads prices and shows best alchs for profits, how is that different?

rcnoob:
> Again, we prefer you contribute to existing plugins due to hub bloat and user fragmentation. If the owner is awol, there are transfer options for the plugin. But completely rewriting for no reason isn't looked at as a valid reason to duplicate plugins if they serve the same purpose.
> 
> As such, what is different? It looks like your plugin downloads prices and shows best alchs for profits, how is that different?

the focus is for actual profit based on volume instead of a gigantic unsorted list; I may as well look at the wiki prices and plan from there at that point. my plugin is aimed on reducing that need to remove yourself from the client and instead integrating information and configuration beyond that of ABC alc. feel free to test it and the differences would be clear

raiyni:
The list is sorted by profit per item on ABC, an option to sort by total profit would be a valid contribution to it. It also appears you are using a large portion of their code anyway.

rcnoob:
> The list is sorted by profit per item on ABC, an option to sort by total profit would be a valid contribution to it. It also appears you are using a large portion of their code anyway.

this tracks, because it is a fork. the reason I am not submitting a pr for ABC is because the way I have it working now is ideal for me, and I'm guessing many others. if it makes you feel any better I can also submit a separate pr for ABC to add more variables to search by. there is no profit to be found when an item cannot be purchased for hours at a time; my plugin aims to solve that issue.

rcnoob:
Okay, I guess I will not contribute and instead use private plugins; feels cheaty and wrong but oh well. Thank you for the inconvenience

PR #8247, opened 2025-06-18
Create visual-timer

Divine-Dream:
displays an in-game timer for people to keep track of their timers without looking away from the game


Adam-:
Please update this to use the `AudioPlayer` class to play clips

PR #8250, opened 2025-06-18
Goggles reminder

Noatel:
Fixed an issue because i was checking if the inventory had a newly create potion (3-dose). When drinking a 4 dose potion, the notifcation activated (Because you get a 3 dose potion agian). 

The notification now requires:
- decrease in an unfinished potion
- increase in a finished potion

Only triggers reminder when an actual potion is created (and not drank).


iProdigy (approved):
nit: you could add `break;` after L77 and L86 of `GogglesReminderPlugin`

PR #8074, opened 2025-05-23
Update glassman to v1d

Mattb2291:

PR #8251, opened 2025-06-18
Visual timer

Divine-Dream:
I was asked to create a new branch to submit my plugin. Hopefully this is correct.

### DIFF
--- a/plugins/overall-tob-timer
+++ b/plugins/overall-tob-timer
@@ -1,0 +1,2 @@
+repository=https://github.com/Divine-Dream/overall-timer-tob.git
+commit=256e6cd0a3b458325db618b0d392219631770f8e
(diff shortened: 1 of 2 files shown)


Felanbird:
See my comment here https://github.com/runelite/plugin-hub/pull/8074#issuecomment-2905711276

Divine-Dream:
> See my comment here [#8074 (comment)](https://github.com/runelite/plugin-hub/pull/8074#issuecomment-2905711276)
> 
> but use your own /plugin-hub repo, not the one I mentioned explicitly for that user.

So is my most recent PR correct?

I made a new branch in my fork of plugin-hub called "visual-timer" and submitted a PR request comparing plugin-hub/master with MY plugin-hub/visual-timer

Felanbird:
It's not correct technically, but if you were to remove the file `overall-tob-timer` from the `plugins` folder, on this branch, it can be correct-enough.

Divine-Dream:
> It's not correct technically, but if you were to remove the file `overall-tob-timer` from the `plugins` folder, on this branch, it can be correct-enough.
> 
> edit: BUT, if you were to take this approach, you would absolutely need to do what I suggested in my other comment, otherwise updating either of your plugins will cause a whole bunch of conflicts just like this.

https://github.com/runelite/plugin-hub/pull/8252

Is this done right?

PR #8252, opened 2025-06-18
Create visual-timer

Divine-Dream:
creates an on-screen timer. I was asked to resubmit this. Hopefully I have it right now.


LlemonDuck:
you're not removing the screen overlay on shutDown

Divine-Dream:
Alright, I went ahead and downgraded log.infos to log.debugs and I also removed the overlay on shutDown().

PR #8256, opened 2025-06-20
Add Clue Path Finder to the Plugin Hub

Infinitay:
This plugin helps automatically path to an active clue scrolls destination(s) by utilizing the [Shortest Path plugin](https://github.com/Skretzo/shortest-path).

_Shameless request for an `onClueScrollUpdated` event either to be subscribed to or hooked onto via it's service if possible_


Infinitay:
Closing because someone apparently was also working on this and their plugin "Shortest Clue" just got merged #8072. Use that instead.

PR #8261, opened 2025-06-20
Soulflame Horn Buff - Fix default sound

sim-macdonald:
The default sound (only the default and not custom sounds) is not playing on the user client. It was working fine on the dev client, but didn't work on the user client.

The change will hopefully fix this issue.

Also made a small update to readme and description.


LlemonDuck:
I don't see why your first revision shouldn't work tbh. Have you tested it with a shadowJar build?

PR #8266, opened 2025-06-21
Add Harder PvM

crscillitoe:
This plugin adds "features" which, when enabled, will make the act of PvMing in runescape significantly more difficult. User discretion is advised.

Presently the functionality is rather minimal.

### DIFF
--- a/plugins/harder-pvm
+++ b/plugins/harder-pvm
@@ -1,0 +1,2 @@
+repository=https://github.com/crscillitoe/HarderPvM.git
+commit=0be1b843ee22c2cc715ab16401181441b5368659


cdfisher:
Ignoring that this feels virtually guaranteed to cause some support traffic when someone inevitably manages to lose items they care about, could this not in its current state just be done with Customizable Menu Swaps?

crscillitoe:
I think that's a fair and valid critique. This plugin will be acting as a "challenge" plugin for PvM streamers so the menu swap approach would sort of give away the bit.

I'm going to refactor to swapping left click to examine, as that should be far less destructive should anyone be experimenting with arbitrary plugins in their free time.

I understand if the use case is a bit too niche for plugin hub adoption, let me know what you think!

tylerwgrass:
Someone else will likely weigh in, but imo a plugin that intentionally disrupts/inhibits gameplay would be a risky addition to the plugin hub.
At a minimum, I would recommend replacing the current feature to force-drop any inventory item to instead consume the click (actively "doing nothing" on click), as the alternative could lead to users losing very expensive items as well as potentially making scamming/lures easier.

I would also recommend that features are well described, have multiple friction points before enabling, and ways of alerting users that this plugin is active (lest they forget and do some risky activity not expecting disruption).
Some of these you have already, but this would be things such as:

- Features disabled by default
- Detailed description of the feature in the config toggle
- Something like an infobox to warn that the plugin/a certain feature is enabled

LlemonDuck:
Yeah, as-is this plugin is only disruptive and likely to confuse a lot of users. Feel free to resubmit if you implement other ideas

PR #8272, opened 2025-06-22
Create Yama-Voiceover

lFlames:
My plugin detects phrases from Yama ,and then plays the voice over corresponding to that phrase

### DIFF
--- a/plugins/Yama-Voiceover
+++ b/plugins/Yama-Voiceover
@@ -1,0 +1,2 @@
+repository=https://github.com/lFlames/Yama-Voiceover.git
+commit=92ff3cf0685f34cd65f99f1d81e25dc52767407c


cdfisher:
This would probably be better as a contribution to the existing "Boss voice" plugin.

pajlada:
I feel like you've misinterpreted the Boss Voicer plugin maintainer's messages regarding "not wanting outside contributors". No PRs have been declined. No feature requests have been declined. If you've talked to them and they have stated something else to you that's fine, but that's not apparent to me.

> i would kindly ask you to rethink this and let this plugin exist as its own plugin ,since merging this plugin with the voicer one would require more unnecessary work on my end ,which would be difficult for me to accomplish ,since before working on this plugin i had no background on java/runelite coding and i needed alot of external help to make this plugin happen.

Just as you value your time, we value our time. We strongly prefer not to maintain plugins that perform, or can perform (with a PR), the same task. If you can show that your plugin does something unique that would not make sense in the boss voicer plugin, we will reconsider. Your reasons above do not convince me.

lFlames:
i am closing this request since gael the creator of boss voicer is not responding ,and if i want to take over his plugin i need to wait a week at least

PSA ,i meant no disrespect by my comment ,and i did not mean to imply that i do not value someone else's time ,sorry if i offended anyone.

PR #8274, opened 2025-06-22
add prayer-markers

coopermor:
Prayer markers allows you to add dynamic screen markers to the prayer interface. This plugin does **NOT** indicate what prayers to use.

Colored prayed markers can be added for any prayer with a configurable stroke thickness and color.


LlemonDuck:
remove unused config

PR #8284, opened 2025-06-25
Update essence-running

AaronPoon:


cdfisher:
> Use ComponentID or InterfaceID instead of WidgetInfo or WidgetID

Except don't use ComponentID anymore either, just use the nested classes of InterfaceID instead. We should probably update that now that we have gamevals.

PR #8302, opened 2025-06-28
Update large-logout

dekvall:


kobunlaunch:
image

I believe the update has caused issues within the world hop menu.

PR #8303, opened 2025-06-28
feat: :sparkles: Add POH Portal Labels plugin

jcbmcn:
- add new poh portal label plugin

### DIFF
--- a/plugins/poh-portal-labels
+++ b/plugins/poh-portal-labels
@@ -1,0 +1,2 @@
+repository=https://github.com/jcbmcn/poh-portal-labels.git
+commit=2c9c8cd5b56599f6fe249d81be4e88448d3bd199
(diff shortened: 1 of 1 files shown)


jcbmcn:
> You may find this tool useful: https://abextm.github.io/cache2/#/viewer/obj/Portal

super helpful, thank you @tylerwgrass

Felanbird:
> Ready for review

I think the reason tyler gave you that link was to subtly hint about adding the other materials portals, so this plugin is not doing some of the teak portals, and none of the mahog/marble

jcbmcn:
@Felanbird, totally agree. Original thought was to just get something out the door and see if there was interest in adding more. But after your comment, I made the effort to get them all added.

Definitely the right call, appreciate you and tyler giving that push. 

Ready for next review.

jcbmcn:
@LlemonDuck, thank you for the catch. This is now updated, I also did some additional cleanup.

PR #8305, opened 2025-06-28
Update Creator's Kit to v2.0

ScreteMonge:
The long-prepared for Creator's Kit 2.0 update. This update activates a number of previously added but hidden features including the TimeSheet and all keyframing capabilities. It allows for the creation of keyframes for different aspects of RLObjects, like animations, movement, and 2D elements like overheads and text.

ItemData, BodyPart, DataFinder, & ModelGetter:
- now includes wear positions per item and names them appropriately when using the CacheSearcher

Character:
- Now uses Keyframing instead of previous Programs. Program, ProgramComp, and the ProgrammerPanel have been removed
- Now directly handles rgb value instead of ProgramComp. This change is reflected as well in CharacterSave

Pathfinder:
- Methods now return int[][] instead of Coordinate[] for finding paths, and accommodates instances properly, with separate methods for poh/gauntlet
- Transplant method now accounts for all MovementKeyFrames

Programmer
- Now properly acknowledges whether the Character is in the scene or not

AttributePanel:
- Temporarily disabled MovementKeyFrame looping
- Added a number of tooltips for better referencing different features
- Cleaned up some wording and formatting regarding SpawnKeyFrames and ModelKeyFrames
- Tweaked SpotAnimCard and HitsplatCard creation such that you cannot duplicate to the currently open KeyFrameType, instead showing a label instead of a button

CacheSearcherTab
- Now shows animations for an NPC when the NPC option is clicked

SubMenuCreator:
- Now properly handles initialization of Movement or Orientation keyframes through TimeSheetPanel methods of initialization

ManagerTree
- Removed unused methods relating to previous Programs
- Updated JoptionPane.showConfirmDialog to a null parent
- ManagerTreeTransferHandler now properly handles whether the drag/drop location is hovering over an inappropriate target

OrientationKeyFrames
- Now consider 32 to be the default turn rate. All references to turn rate of -1 have been updated to a static value of 32 from OrientationKeyFrame

Hitsplats:
- Now applies different variants, including Normal, Max, and Other for different hitsplats/tints
- A number of other variants have been added, like Bleed, Burn, and Armour

Config
- Cleaned up a number of now-unused functions relating to the old Programming system. Added a config item for choosing the movement algorithm, which used to be handled by that programming system individually per Object

CreatorsOverlay
- Updated to accommodate path lines for the new programming system by looking at MovementKeyFrames instead of previous Programs
- Tweaked rendering of lines to show them as much thicker, and make the selected RLObject stand out further

CreatorsPlugin
- Added a number of hotkeys relating to the new Programming system
- SetLocation, addProgramStep/removeProgramStep/ClearProgramStep methods have been completely overhauled to accommodate the new Programming system

### DIFF
--- a/plugins/creators-kit
+++ b/plugins/creators-kit
@@ -1,2 +1,2 @@
 repository=https://github.com/ScreteMonge/creators-kit.git
-commit=9a2a5d0235f7564497db67d55d4407afc6c7c871
+commit=75ca260a6faa3cb7c811c1aa015cb9a9851aaa1e


LlemonDuck:
Never thought I'd see the day 🥲 🎉 

You should specify a parent where possible for JOptionPanes to prevent them from rendering underneath the runelite window if Always On Top is enabled. It causes the UI to be uninteractible.

PR #8311, opened 2025-06-28
README: add section on restricted behaviors

riktenx:
This is an attempt to inform plugin authors of in-code behaviors which we've historically disallowed. Putting this above the rest of the guide may be controversial, the intent there is to prevent anyone from doing work or submitting something that won't be allowed off the cuff.


riktenx:
Sorry, nuked my fork on accident.

PR #8315, opened 2025-06-29
Update odablock plugin

DapperMickie:
I will keep adding to this PR, but can be reviewed from time to time.
I'm trying to keep every commit to the https://github.com/dappermickie/odablock-sounds repo 1 sound at a time.

I'm working together with the dev of the c-engineer plugin and we want to push the update around the time when they start doing the bronzeman mode series together. 

We'll be adding more sounds to this plugin and his plugin until (and maybe during) this 1 month period where they do the BMM series to troll either side.

Thanks RL Team <3


riktenx:
Closing because this remains in draft state, you may resubmit when it is ready.

PR #8318, opened 2025-06-29
Add Dwarven rock cake helper plugin

YuvalPeretzAtBay:


coopermor:
Guzzling a dwarven rock cake will not kill you. You can spam click it at 1HP and it will only hit you for 0. You would have to take damage from another source (like an overload) in order to die, which this plugin wouldn't really prevent.

YuvalPeretzAtBay:
@coopermor
Thank you for clarifying it!
After further investigation I can see that this is the behaviour.
I will close the PR.
Have a good one!

PR #8319, opened 2025-06-29
Update crosshair-cursor

Benjames01:
Add toggle for crosshair
Re-set the cursor on focus


Benjames01:
> @Benjames01 Could you add yourself to the authors list? And @Thomas-gonzalez could you confirm they're allowed to also push updates to the plugins going forward? Thanks!

sure is that in runelite-plugin.properties or  here [plugins/crosshair-cursor](https://github.com/runelite/plugin-hub/pull/8319/files#diff-9696bf307dcbc945b19b26757e99d643216e049425ea7be0f07bb68ea662cd5a)

pajlada:
Yup like that - now just confirmation from @Thomas-gonzalez again would be great

Thomas-gonzalez:
@pajlada Yes that is no problem

tylerwgrass (approved):

PR #8322, opened 2025-06-29
Add menu-stones-hider

Richardant:
This plugin hides the menu stones and their respective icons, the menu background (resizable classic only), and the sidebars in the menu. It also includes a toggle that you can use to unhide them. The default keybind is: CTRL+B.

This plugins attempts to mimic something that the modern resizable layout has by default, but that the classic resizable layout doesn't.


LlemonDuck:
Use KeyManager#registerKeyListener instead of adding directly to the canvas

PR #8333, opened 2025-06-30
Update entity-render-distance

Dazuzi:
Basically a rewrite, but still the same idea. Couple new customisation options and "fading players" feature.

entity-render-distance


LlemonDuck:
needs rebase

PR #8343, opened 2025-07-01
update flipping-copilot to v1.7.2

cbrewitt:
- Add tooltip showing potential profit from sell offers
- Add "Nothing" menu entry to force right click confirm if price/quantity are set incorrectly


cbrewitt:
We've removed the deprecated API usage and fixed some other minor issues

PR #8348, opened 2025-07-02
Update dry rate tracker plugin

georgebunn97:
Update dry rate tracker plugin

Updated the detection logic for each raid so it's now based on chests rather than chat messages
- This is because the chat messages were reliant on other plugins. This new logic is more robust and will work for everyone no matter what plugins they use.

Updated the UI 

Updated dry streak save logic


LlemonDuck:
can you remove the bin directory from the repo and add it to gitignore?

georgebunn97:
> can you remove the bin directory from the repo and add it to gitignore?

Sorry been meaning to do that! Just did that commit now

PR #8360, opened 2025-07-02
Create mage-animation-replacer

washer83:
Add mage-animation-replacer -- a plugin designed to swap the trident/barrage animations back to their old animations, which have a visual animation stall.


LlemonDuck:
Might be nice to allow people to change the animations of other players as well, not just themselves

PR #8362, opened 2025-07-03
Update Vitality (bugfix)

Natesky9:
removed debug render that caused healsplats to not expire


Natesky9:
trying to update, but can't seem to find the right links... says I closed?

PR #8363, opened 2025-07-03
Create Vanced Barrows

Inzomniakk:
Hides the barrows brothers and displays an image of JD Vance instead.

### DIFF
--- a/plugins/vanced-barrows
+++ b/plugins/vanced-barrows
@@ -1,2 +1,2 @@
 repository=https://github.com/Inzomniakk/VancedBarrows.git
-commit=5622f603093e11e4cb23ce1c1cc0eddae17ea35a
+commit=77fea912a318172e600d6899bd677c4664558811


Felanbird:
Do you have permission from JD Vance to use his image in this plugin?

Inzomniakk:
No, I wouldn't think I need permission considering one could argue it isn't even him. It's also satirical and being used not for profit, and in no other way defames a person. If anything I could call it fat guy barrows, cause then unless you're familiar with the image you really don't know who it is.

Not trying to come off rude, but I honestly don't know either.

LlemonDuck:
The rights we'd need to clear is whoever created the meme, and also that whoever created the meme had the corresponding rights from the copyright holder of the original image. Considering how widespread the meme is I'd be surprised if we could track any of it down. It's possible the original image might be public domain though! Good luck hunting, if you care to.

Inzomniakk:
Well, according to what I read about what public domain is. And considering where the image came from, which is a tweet, which the history of can be found here [https://knowyourmeme.com/memes/jd-vance-edited-face-photoshops](url).

I like to think the image is in the public domain and covered under fair use. I guess you want me to find the part where it says "hey yeah you can use this image for anything you want"?

LlemonDuck:
You'd need to contact the original creator on Twitter, `@SperglerAcolyte`, and also get them to share where the original image came from, then make sure that is derivable.

Or you can just make the plugin take an arbitrary user-provided image.

iProdigy:
Yes, that image is in the public domain https://commons.m.wikimedia.org/wiki/File:VancePortrait.jpg

Inzomniakk:
My own images based on the portrait have been uploaded. Any further edits to the images will be based on the two new ones included.

iProdigy:
> Okay great! Next question, after I make my own image. What 'proof' do I need that I made it other than being honest. Serious question in case I need to do something more.

When you add new images you'll want to assert you made them & provide links to the source material in the public domain - we may do a reverse image search to verify your claim

LlemonDuck:
if the config is unused you should remove it to prevent a settings icon appearing next to the plugin

Inzomniakk:
I have planned settings for the plugin that will be added later.

LlemonDuck:
Don't add a setting that does nothing, just remove `@Provides` and add it back when you are ready to add options. You can keep the rest of the code

PR #8401, opened 2025-07-07
Create ore-highlighter

BuzzfeedBlue:
Creation of ore-highlighter plugin. Adds image to interactable paydirt in motherlode mine and amethyst crystals. Image can be changed locally, default is the map mining area circle with pickaxe,

### DIFF
--- a/plugins/ore-highlighter
+++ b/plugins/ore-highlighter
@@ -1,2 +1,2 @@
 repository=https://github.com/BuzzfeedBlue/ore-highlighter.git
-commit=469fd44e4e2da9b78e0659e2270c3ac4cd2031a4
+commit=6fea73b6794dcba3ad1934c414fcca2c024fb3b4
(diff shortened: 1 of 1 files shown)


DominickCobb-rs:
Check the workflow error. You're missing some essential files from the pluginhub sample plugin

cdfisher:
Can you not basically accomplish the same thing already with the existing "Object Markers" plugin, other than that just being a highlight rather than an image overlay?

BuzzfeedBlue:
> Can you not basically accomplish the same thing already with the existing "Object Markers" plugin, other than that just being a highlight rather than an image overlay?

It's all about choice. The object marker plugin only does the outline glow. And the other motherlode style plugins all use timers. There is the runelite base icon to mark these things as well, but it uses the HUGE mining skill icon and it cant be changed.

PR #8404, opened 2025-07-07
update visual-metronome to V1.4.8

vincent0955:
- Added Tick Counter Follows Mouse


LlemonDuck:
use client.getMouseCanvasPosition directly in your overlay instead of registering a canvas listener

riktenx (approved):

PR #8411, opened 2025-07-07
Create scrollboxcounter

thelaakes:
A RuneLite plugin that displays the maximum number of clue scroll boxes you can hold for each tier. With options to mark the stack red if the maximum quantity is reached (including banked scroll boxes). Also able to display the number of clue scroll boxes currently banked. 

Interference with clue scrolls and how it behaves if the scroll boxes are in a looting bag/other storage has to be tested.


jgresham:
+1 if the count could be put in chat too... example "holding 2/3 scroll boxes (hard)"

PR #8429, opened 2025-07-09
Update acoldone-clogging

OmbudRov:
- New Sky GIFs
- Refactored the code for better readability ~threw code into sub java classes and called it a day~


pajlada:
[OmbudRov/AColdOne-Clogging@`04787001227a1083cd882854a44583810289e694...OmbudRov:0a18509618bb4f49934b35de031d0210bb4e20ad?w=1`#diff-c586325362](https://github.com/OmbudRov/AColdOne-Clogging/compare/04787001227a1083cd882854a44583810289e694...OmbudRov:0a18509618bb4f49934b35de031d0210bb4e20ad?w=1#diff-c5863253620d810874a6a70cebf4b82329c6106b9ebf3b8d681e7be2e62b7939R119) You shouldn't change config keys without good reason or migration, since it will reset your current users' settings

If you're already refactoring your plugin, please make use of the runelite-provided `AudioPlayer`

OmbudRov:
> You shouldn't change config keys without good reason or migration, since it will reset your current users' settings

Ah ok, i'll keep that in mind from now on

pajlada:
You can change the keyname back now though - you don't have to revert the variable name. just the `keyName` field

OmbudRov:
I think I'll just let it be this time,
- I've changed the config to be for each type of GIF, instead of individual ones
- Accidentally typing the command for the gif is highly unlikely

PR #8430, opened 2025-07-09
Add 'Book of the Dead Renamer' plugin

Robadob:
This plugin renames the 5 teleport options within Book of the Dead/Kharedst's Memoirs to their in-game locations to avoid wasting charges.

First attempt at a RuneLite plugin, so keeping it super simple.

### DIFF
--- a/plugins/book-of-the-dead-renamer
+++ b/plugins/book-of-the-dead-renamer
@@ -1,0 +1,2 @@
+repository=https://github.com/Robadob/rl_botd.git
+commit=7077197e04c16039196c6a6562444dec99408c85


ldavid432:
Doesn't easy teleports already do this? https://runelite.net/plugin-hub/show/easy-pharaoh-sceptre

Robadob:
Apparently so, but neither I nor people in my clan could find that looking for this specific thing.

Robadob:
Not super familiar with runelite api, but perhaps worth them adding the specific affected item names to their plugin tags?

riktenx:
If this plugin does not add any novel behavior over the existing one mentioned above then we'd prefer to close it.

If the existing plugin is missing functionality you want, it should be contributed there.

Robadob:
Yeah happy to close this, just querying how to improve discoverability of said plugin.

Robadob:
image

Nightfirecat:
It probably needs `of,the` added to the tags, I guess. Alternatively we could consider stripping those from searches to avoid this exact problem

PR #8431, opened 2025-07-09
Add MobLocator plugin to Plugin Hub

elisadamico:
This plugin is a MOB Locator (requested by my clan!). It should allow you to pick a monster form the menu and then it will print its locations, how many are there, and highlight overlay on the mini map!


pajlada:
See https://github.com/runelite/plugin-hub/blob/master/README.md for how to submit a plugin to the plugin hub - you're currently not using the right format.

LlemonDuck:

PR #8432, opened 2025-07-09
Update global-chat v2

RusseII:
Summary

  - Info Panel: Created new sidebar panel showing plugin status, current limits, and community links
  - Cost Optimizations: Implemented pre-publish spam filtering and connection leak fixes to reduce Ably billing costs
  - UI Enhancements: Updated plugin icon and improved overall user experience
  - Patreon Integration: Integrates with Patreon to see current supporters
  - Bug Fixes: Fixed deprecated method usage and improved error handling

  Technical Changes

  - New SupporterManager class for Patreon API integration
  - GlobalChatInfoPanel with GridBagLayout for consistent UI
  - Enhanced AblyManager with spam filtering and connection optimizations
  - Fixed deprecated method calls and improved connection management

### DIFF
--- a/plugins/global-chat
+++ b/plugins/global-chat
@@ -1,3 +1,3 @@
 repository=https://github.com/RusseII/region-chat.git
-commit=8a97f22fcc22d3ff6a960483781d09d489b73f83
+commit=53f7640d079ed6a16ccf83e5b90f5c9171b3ff57
 warning=This plugin submits all sent chat messages and your IP address to a server not controlled or verified by the RuneLite developers.
--- a/plugins/global-chat
+++ b/plugins/global-chat
@@ -1,3 +1,3 @@
 repository=https://github.com/RusseII/region-chat.git
-commit=ab68b604d76d00bf813844c0fe9de54c9477498f
+commit=02e8df03b1f9cba485a2a6bbcfda973b624bf4c1
 warning=This plugin submits all sent chat messages and your IP address to a server not controlled or verified by the RuneLite developers.
--- a/plugins/global-chat
+++ b/plugins/global-chat
@@ -1,3 +1,3 @@
 repository=https://github.com/RusseII/region-chat.git
-commit=44e9d8dc74c4a0b0259535b14cd6251a714cafe2
+commit=8a97f22fcc22d3ff6a960483781d09d489b73f83
 warning=This plugin submits all sent chat messages and your IP address to a server not controlled or verified by the RuneLite developers.
--- a/plugins/global-chat
+++ b/plugins/global-chat
@@ -1,3 +1,3 @@
 repository=https://github.com/RusseII/region-chat.git
-commit=44e9d8dc74c4a0b0259535b14cd6251a714cafe2
+commit=d992f6b80a9fe5d7154250d21a7eb0d5d893c34a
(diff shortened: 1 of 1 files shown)


RusseII:
> Use of java.lang.System.err is not allowed: system i/o, use a logger instead.
> Use of java.lang.System.out is not allowed: system i/o, use a logger instead.
GM 000317

I think this can be ignored? I don't see any references to java.lang.System.err or java.lang.System.out in the code - please let me know!

LlemonDuck:
They are probably from ably. If you don't see it in your code you can ignore

LlemonDuck:
Never sleep the client thread https://github.com/RusseII/region-chat/compare/f2afe9813b9cd887809a5011e30a4ccda1178d36...RusseII:44e9d8dc74c4a0b0259535b14cd6251a714cafe2#diff-309b74bab8042a57c9bd2b2211a1cfc437563c3dffa0750a4396577688cf0fddR457

RusseII:
Thanks for the feedback @LlemonDuck.  I've addressed all your points:

  ✅ Replaced HttpUrlConnection with injected OkHttpClient - Now
  using RuneLite's existing OkHttpClient binding in
  SupporterManager instead of creating our own provider

  ✅ Downgraded log.info to log.debug - Updated across
  AblyManager, GlobalChatPlugin, and SupporterManager to reduce
  log verbosity

  ✅ Removed unnecessary ClientThread#invokeLater usage - Removed
  from showUpdateNotification since it doesn't need client data
  access

  ⚠️ Note on AblyManager connection methods - I kept
  clientThread.invokeLater for connection startup because
  AblyManager#startConnection actually does need client thread
  access - it calls client.getLocalPlayer().getName() in
  setupAblyInstances() to set the clientId for Ably
  authentication. The delay also helps prevent race conditions
  during connection cleanup on world changes.

  The timing is now properly used only where client data access is
   required, not for arbitrary delays.

raiyni:
It's not helpful if you are plugging requests into an LLM and having it reply to/fix feedback.

You need to do whatever you need to do on client thread, then start your connection after all of your conditions are satisfied.

RusseII:
Sorry - @raiyni and thanks for the comment. I think I fixed it. I was a bit confused on how the client thread / non client thread stuff works. Can you point me to somewhere to learn and understand at a deeper level how it works? Or another example plugin thats doing it all well. 

I think I correctly fixed it now with the scheduler.execute to move it off the client thread, but please let me know if further changes are needed. 

```
  private void onLoggedInGameState() {
      clientThread.invokeLater(() -> {
          if (client.getGameState() != GameState.LOGGED_IN) return true;

          final Player player = client.getLocalPlayer();
          if (player == null) return false;

          final String name = player.getName();
          if (name == null) return false;

          scheduler.execute(() -> {
              ablyManager.startConnection(name);
          });

          return true;
      });
  }
  ```

LlemonDuck:
^ This sort of behaviour looks correct, although I'm not actively reviewing the whole plugin to say it's used correctly everywhere.

Generally speaking, collect the client data you need on client thread, then move to another thread (with e.g. `scheduler.execute` as above) to work with the data. Don't hold on to references like NPC or Player across the thread boundary, copy out POJO types like player name (String), position (LocalPoint), etc.

PR #8437, opened 2025-07-10
Create pet-spell-blocker

Sacca-1:
Created a very simple plugin that removes spell casting menu entries on NPCs that are your follower to prevent misclicks.


Sacca-1:
Thank you both! That should do it

PR #8441, opened 2025-07-11
Add Chat Combiner Plugin

FKiisa:
Good evening!

I was doing some cooking and as you do while playing runescape, sometimes you zone out.. The only thing visible in the chatbox is the same message "You accidentally burn the karambwan" 15 times in a row.. I was puzzled, confused and worried, when will this ever end, only to notice the AFK Timer plugin tell me that I've been AFK for 240 seconds already.

That's it, it's time to figure something out to make this chat feel less **_spammy_**!

This here is where I would like to introduce Chat Combiner!

**Chat Combiner** is a lightweight RuneLite plugin that automatically combines repeated in-game chat messages into a single line with a counter - helping reduce spam while preserving readability.

Ideal for repetitive messages like:
- "You clean the Grimy guam leaf"
- "You catch a fish."
- "You examine the object..."

---

## 🔧 Features

- ✅ Combines repeated identical chat messages
- ✅ Optional counter position (before or after the message)
- ✅ Customize counter color
- ✅ Configurable minimum repeats before combining starts
- ✅ Auto-resets repeat tracking after inactivity

---

## 🛠️ Configuration Options

All options are available in the RuneLite Plugin Panel under **Chat Combiner**:

| Setting | Description |
|--------|-------------|
| **Counter Color** | Choose the color of the repeat counter `(xN)` |
| **Counter Position** | Show the `(xN)` counter before or after the message (e.g. `(4x) Message` or `Message (4x)`) |
| **Examines & Info** | Combine messages like item/npc/object examine, welcome/info tips |
| **Game Messages** | Combine common game messages like skill actions or feedback |
| **Minimum Repeat Count** | How many identical messages must appear before collapsing |
| **Reset After (ms)** | Time in milliseconds after which the repeat tracking resets |

image

---

## 💬 Example Output

Given `Minimum Repeat Count = 4`:

```text
You clean the Grimy guam leaf
You clean the Grimy guam leaf
You clean the Grimy guam leaf
(4x) You clean the Grimy guam leaf
```

Instead of having 27 lines of spam for a single inventory of herbs, now (based on the plugin configuration) you can actually have time to understand what is going on!

image

This is my first plugin ever created, so if I'm doing something incorrectly, please let me know and I will do my best to fix it!

### DIFF
--- a/plugins/chat-combiner
+++ b/plugins/chat-combiner
@@ -1,0 +1,2 @@
+repository=https://github.com/FKiisa/chat-combiner.git
+commit=2c63f0179b73c8614d764a899135e9f5c10a14a9


FKiisa:
I'm going to be completely honest here, I was not aware of the chat filter plugin at all.. That being said I just had a look at it and while yes it does do it, there's sadly no configurability in a sense that you could set it to reset after x amount of time, have the possibility to change colors - new gen will love that, and have overall better accessibility configurations available for the user, based on preference.

If this is still something that perhaps would stop from this going out, no worries at all, I tried my best and learned a lot while doing so!

tylerwgrass:
In cases where functionality is not distinctly different from an existing plugin, we would instead recommend you make a PR or open an issue to the existing plugin.
This helps reduce the amount of similar plugins on the plugin hub that provide different/partial sets of features and allows the changes you make available to a larger user base over a new plugin.

FKiisa:
Completely agreed! Again, sorry for all the troubles, I'm new in this plugin development side of the world. One thing I didn't find is a link to the repository for the chat filter plugin inside plugin-hub? That's the last thing I would love to get some support from your side if possible, after that I will close off my pull request.

cdfisher:
Chat filter is part of the core client so it's included in that repo.

FKiisa:
Thanks a lot everyone for your help! Glad to see such an active community behind RuneLite! <3

PR #8446, opened 2025-07-12
update action-logger to v0.1.5

pajlada:
This version adds two commands to the action logger plugin:

`::ActionLogger restart` - closes the current file and starts writing to a new file. Useful if you're wanting to separate logs from multiple activities in the same session

`::ActionLogger dump` - dumps information (NPCs, ground items, and objects) about your surrounding to the current file. Lots of data, useful for scenarios where you run into a quest where it doesn't highlight an item due to the ID or coordinates being wrong.


pajlada:
image

iProdigy (approved):

PR #8448, opened 2025-07-12
Update Inventory Setups to v1.21.0

dillydill123:
Fuzzy fixes to make it behave more like native bank tag layouts
Add right click remove setup option for all setup panels
Fix bug where tags were not hidden after modifying a setup
Allow sections to be sorted
Move to gameval API


pajlada:
@dillydill123 your build currently fails - see https://github.com/runelite/plugin-hub/actions/runs/16241589920/job/45858761622?pr=8448

dillydill123:
Thanks @pajlada.

Seems like it's working now. I also added a commit to move to the new gameval API so there are no more deprecated interfaces being used in Inventory Setups

PR #8456, opened 2025-07-13
Update fog-of-war

Dazuzi:
- Added exceptions for TzHaar Fight Cave, Inferno and Hallowed Sepulchre.
- Adjusted fading player tile logic.
- Other minor tweaks/optimisations.


riktenx (approved):
FYI the addition of `FogOfWarMinimapOverlay::boundaryPoints` doesn't seem necessary (it doesn't appear to ever be used as an instance member) and may lead to leaks in the future if misused.

riktenx (approved):

PR #8466, opened 2025-07-14
Create MoaningJewelry

L4mb0:
I regret nothing


riktenx:
https://github.com/runelite/runelite/wiki/Rejected-or-Rolled-Back-Features#not-currently-being-considered

> Plugins which are of an adult or overtly sexual nature.

PR #8467, opened 2025-07-14
Update sound-alerts-expanded

lewislarsen:
Adds a sound alert trigger when a Chambers of Xeric Raid has completed.


iProdigy:
you probably want to check `ChatMessage#getType` so player-sent messages can't trigger the sound alert

lewislarsen:
> you probably want to check `ChatMessage#getType` so player-sent messages can't trigger the sound alert

I completely agree with you!

I've made the changes here:

https://github.com/lewislarsen/sound-alerts-expanded/commit/63d7c0e5a2f4692006dea29175925914faa463c9

Please let me know if there's any other changes to be made! :) 

Thanks!

PR #8469, opened 2025-07-14
Create gewriter

ghost:
Adds the GE Transaction Writer plugin — a passive logger that writes Grand Exchange transactions to a local JSON file.

- No UI interaction, automation, or networking
- Fully Plugin Hub–compliant
- MIT licensed
- Writes to `~/.runelite/getransactionwriter/transactions-YYYY-MM-DD.json`
- Useful for tracking flips, analyzing margins, or building external tools

Repo: https://github.com/Mreedon/ge-transaction-writer

This is a simple first version. Later updates may include a config panel and session ID tagging, but will stay fully compliant.

### DIFF
--- a/plugins/plugins/gewriter
+++ b/plugins/plugins/gewriter
@@ -1,0 +1,2 @@
+repository=https://github.com/Mreedon/ge-transaction-writer.git
+commit=2899188b953ef083f4c688902569a82ace591b38


Mreedon:
Appreciate the heads up — you're right, exchange-logger already covers a lot of ground.

I started this one because I wanted something that could be further customized — JSON-only for now, daily logs, no config yet. It’s basic at the moment, but I’ve got a few features planned to help it stand out a bit more. For example, it’s set up to support logging when the player hits “Collect All” in the GE, and from there I’m planning to add:

Options to include or exclude specific fields (like timestamp, offer state, account name)

Metadata like slot number and display name

A setting to change the output folder

A version tag in the logs to keep things compatible as the format evolves

Custom timestamp formatting


Totally understand if there’s some overlap, but just hoping this offers a more structured and customizable option for users who want predictable, clean logs.

Mreedon:
Appreciate the clarification — I wasn’t aware how closely this would overlap when I started, but totally understand the desire to avoid redundancy.

I’ll close this out and take a look at whether I can contribute to the existing plugin instead. Thanks again for the feedback!

PR #8483, opened 2025-07-15
Add Burner Alarm Plugin

AltarOSRS:
**Plugin Description:**
Sends a notification and plays a sound when your POH burners are about to go out.

**Features:**
- Provides a configurable pre-notification alert when burners are close to reaching their random burnout timer
- Plays an audible alarm sound when burners fully reach their random burnout timer
- Designed to notify only once per type for a "batch" of burners lit together
- Timer scales with firemaking level
- Automatically clears active burner timers when leaving a Player-Owned House (POH) or other game state transitions
- Toggle alerts, change volume, or adjust stage 1 alert lead time in config


LlemonDuck:
where is the alarm sound sourced?

PR #8487, opened 2025-07-16
Add Quick Obtain plugin

davebrouwer1:
This plugin shows multiple options for obtaining an item when searching for it, for example "Pot" or "Cheese" It pulls data from the wiki, and shows you the different spawn and/or store locations, when you click on an option it uses the "ShortestPath" plugin code (packaged within this plugin) to route it out.

### DIFF
--- a/plugins/quick-obtain
+++ b/plugins/quick-obtain
@@ -1,0 +1,2 @@
+repository=https://github.com/davebrouwer1/quick-obtain.git
+commit=ef2f48cadcf2c738d619e60943583b3164c8943b


davebrouwer1:
Ah I'm not sure why it doesn't work on your end, I'll try to find the issue, and your idea of using shortest path sounds like a much better idea than what I was using, I will implement that, thanks!

riktenx:
> doesn't seem like your plugin does anything

To clarify -- all of your source files at the commit you have pointed to here are blank (or, they're the main plugin file which just has some example logs).

e.g. this is the code for the panel you have submitted.

image

Do you perhaps have some local changes that you haven't committed and pushed?

LlemonDuck:

PR #8488, opened 2025-07-16
Attack history

sethks:
UI element that tracks the most recent attacks the player did with weapons and damage numbers. The player can have it show as much information as they want. Does not track player attacks directly as the player can track only the most recent up to the 5 most recent.


riktenx:
Your plugin repository appears to be private.

riktenx:
Check the CI build results - your plugin is using terminally deprecated APIs.

Also, you don't need to close and open PRs. Just update the commit hash.

PR #8495, opened 2025-07-17
Add Item Marker Plugin by chatgptandcopilot

chatgptandcopilot:
The Item Marker plugin highlights specified ground items within RuneLite using a customizable overlay. Users can define a target item list, and the plugin renders clear visual indicators for those items on the game world. It’s lightweight, precise, and designed for speed-focused looting or item tracking.


tylerwgrass:
There are multiple plugins that do this already, is there a reason one of those can't be used instead?

You should also be following the guide listed in the [README](https://github.com/runelite/plugin-hub/blob/master/README.md) for creating new plugins.

PR #8504, opened 2025-07-18
Update one-tick-flick to v1.0.2

SuperNerdEric:
- Add config options to recolor overlay elements
- Add config option to detect prayer book clicks
- Change default greenStart to 0ms into the tick.


tylerwgrass:
Can you cache the config values in your plugin file instead of fetching them in the overlay render function to reduce thrash?

SuperNerdEric:
> Can you cache the config values in your plugin file instead of fetching them in the overlay render function to reduce thrash?

All of the config values used in the overlay should be cached in the overlay and only updated on ConfigChanged events now. Thanks!

riktenx (approved):

tylerwgrass (approved):

PR #8506, opened 2025-07-18
Add Yama reminder plugin

jdockerty:
Adds a simple panel which acts as a reminder for what prayer to use on the phases between Yama.

This is deterministic only after user input (clicking the button). This plugin is entirely for removing the need to write reminders to duo partners, instead we can click a button and save spamming chat.


LlemonDuck:
remove the unused config

PR #8507, opened 2025-07-18
RuneProfile v2.1.0

ReinhardtR:
- Cleaned up main RuneProfilePlugin class by moving features/logic into their own class.
- Added AutoSync
    - Syncs profile every 15 minutes in the background
    - Method to start a rapid sync, that syncs after 3 seconds
- Collection Notification Subscriber
    - Listens to collection chat message or popup notification (based on dinkplugin implementation) and starts a rapid sync on new item obtained.
- Collection Log Widget Subscriber
    - Subscribes to the collection logs setup script that allows the plugin to get the player’s collection log items.
    - The plugin now fires this script every time the collection log is opened, to populate the `clogItems` (in the PlayerDataService) that will be sent in the auto sync.
- Moved the RuneProfile button in the Collection Log window (that was based on WikiSync implementation) to the burger menu, since it will ideally not be used as often (or at all) and avoids conflicts with WikiSync.
- Removed Default Collection Log page selection from side panel. Replaced by “RuneProfile Default” right click menu option on Collection Log pages.
- Added a loot subscriber, that submits valuable drops to your RuneProfile.

### DIFF
--- a/plugins/runeprofile
+++ b/plugins/runeprofile
@@ -1,3 +1,3 @@
 repository=https://github.com/ReinhardtR/runeprofile-plugin.git
-commit=a076e0810a294204dacacc688084b7aabe183c99
+commit=2ac0d7b6552f1830f4e4bc00a4d45903921d6904
 warning=This plugin submits your player data and IP address to a server not controlled or verified by the RuneLite developers.
--- a/plugins/runeprofile
+++ b/plugins/runeprofile
@@ -1,3 +1,3 @@
 repository=https://github.com/ReinhardtR/runeprofile-plugin.git
-commit=7a83138ac1aecbd40b39ba8999c366641596737c
+commit=3fced102e49ec2b35b156aae773cd5244c9d24bf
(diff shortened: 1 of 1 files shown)


ReinhardtR:
Thanks for the review <3

1. Update on logout config
I now keep the old config key `update_on_logout` to avoid unwanted behaviour for existing users.

2. `getPlayerModelDataAsync` completeExceptionally
Early returns have been added in those cases.

3. Credits
Added inline credits to code that has been copied from other plugins. *I hope it didn't come across as malicious <3*

4. Collection Log Menu
I see there was an issue where the menu widget was resetting its height, so the RuneProfile button was not visible. This should be fixed now.

And yes that button will create a profile if you don't have one. 

To be clear this is the menu widget:
image

And this is the menu entry, for setting the default collection log page on your profile.
image

iProdigy:
>    Added inline credits to code that has been copied from other plugins. _I hope it didn't come across as malicious <3_

nah it's all good - we did appreciate the header in `ItemSearcher` <3

PR #8520, opened 2025-07-20
Add auto party plugin

neilrush:
A simple plugin that rejoins the last party on login. Requires the party plugin to be enabled.

Enable the plugin and then join a party. The plugin will rejoin your last used party at login until you leave the party or disable the plugin.

Upon joining a party at login, a chat message is displayed to notify the user.

### DIFF
--- a/plugins/auto-party
+++ b/plugins/auto-party
@@ -1,0 +1,2 @@
+repository=https://github.com/neilrush/Auto-Party.git
+commit=563283a3a09f9bae0e3fcb39d1f0f2972bcfb7b4


coopermor:
Please see https://github.com/runelite/runelite/wiki/Rejected-or-Rolled-Back-Features

> Auto-rejoin for parties: Would put undue strain on the party servers.

neilrush:
> Please see https://github.com/runelite/runelite/wiki/Rejected-or-Rolled-Back-Features
> 
> > Auto-rejoin for parties: Would put undue strain on the party servers.

Dang I looked for something like that. Slipped past me. Sorry to waste your time.

PR #8526, opened 2025-07-20
Create bank-value-changes

Rooner-Plugins:
This plugin allows you to track the changing value of specific items you have in your bank through simply opening up your bank interface. Items rising in value are highlighted in green and items decreasing in value are highlighted with red, both topped off with the percentage price change of the unit item.


Rooner-Plugins:
Thanks. Updated to use injected Gson instead. Also increased sampling resolution, updated the minimum data viewing window to be longer and fixed a logic bug.

LlemonDuck:
None of these are blocking but you may want to consider changing them in an update:
* `GSON.newBuilder().create()` is redundant
* ClientShutdown is not guaranteed to fire (if the client crashes, or is killed, etc.)
* move the mkdirs to startUp

PR #8527, opened 2025-07-20
Update fog-of-war

Dazuzi:
Added an option to prevent fog from being drawn over entities.

fog exclusion


riktenx (approved):
A setting with an implied performance impact like this should really be off by default.

tylerwgrass (approved):

PR #8533, opened 2025-07-20
Create afk-overlay

zmertz:
This plugin creates an overlay that is not attached to the Runelite window. This overlay simply shows current hp, prayer, inventory space, and if the player is performing an action (Active vs Idle)


LlemonDuck:
remove this https://github.com/zmertz/AFK-Overlay/blob/7ceb05b0f549195007005e44deb73f2bd4dfaa43/build.gradle#L5-L9

use ImageUtil#loadImageResource instead of ImageIO

zmertz:
Removed JavaLanguageVersion and replaced ImageIO with ImageUtil

LlemonDuck:
You should gitignore test-classes

PR #8534, opened 2025-07-20
Cox mega scale

FadedAshen:
Cox Mega Scale plugin helps with running a mega scale with friends by providing utility and information.

Provides:

Scouting assistance by selecting whitelisted rooms and order.
Points gained and lost.
Drop chance with amount of rolls.
How many overloads/golpar/fish needed.

NOT INTENDED FOR BOOSTING IRONS.


LlemonDuck:
downgrade a bunch of these log.infos to log.debugs

LlemonDuck:
Remove your try/catch around the overlay renders, we have our own in OverlayRenderer which deduplicates as well

FadedAshen:
Thanks for the comment, I implemented and fixed everything you mentioned!
I also synced my fork by accident which lead to that non-author plugin change label. Hopefully I didn't mess anything up with that.

Thanks!

PR #8535, opened 2025-07-21
Added Idle Fishing Notifications plugin

Salcrainia:
### Description
Enables better idle notifications when fishing.
Covers all cases encountered so far - primarily tested at Tempoross.


Felanbird:
Enabling `Idle interaction notifications` in the `Idle Notifier` plugin gives you fishing notifications.

PR #8536, opened 2025-07-21
Update Exchange Logger to latest commit

ghost:
This updates Exchange Logger to include co-maintainer attribution and use the latest commit hash.

@istid  — tagging you here to confirm approval as discussed. Thanks again!

### DIFF
--- a/plugins/exchangelogger.json
+++ b/plugins/exchangelogger.json
@@ -1,0 +1,11 @@
+{
+  "internalName": "exchangelogger",
+  "displayName": "Exchange Logger",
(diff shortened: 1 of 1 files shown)


istid:
> This updates Exchange Logger to include co-maintainer attribution and use the latest commit hash.
> 
> @istid — tagging you here to confirm approval as discussed. Thanks again!

I am okay with this author change.

Context:
I have basically abandoned this plugin and @Mreedon offered to take over it. I would have no problems with transferring the plugin, but it seems like adding them as a collaborator is the preferable way.

PR #8537, opened 2025-07-21
Add raid-spell-checker plugin

tyler-arnett:
This plugin checks that you have the correct spellbook and required spells/runes before starting COX, TOB, or TOA.
If you're missing the required spellbook or runes, it alerts you with a configurable popup and sound.

Features:
Custom sound support (.wav files in resources).
Supports rune pouch, divine rune pouch, and staves.
Configurable per-raid and per-spellbook.

NOTE: Plugin must be configured before use. The player selects their preferred spellbook and spells per raid.

### DIFF
--- a/plugins/raid-spell-checker
+++ b/plugins/raid-spell-checker
@@ -1,0 +1,2 @@
+repository=https://github.com/tyler-arnett/raid-spell-checker.git
+commit=4f08b72e5803b07a7c3b92be2a48e63e68fef842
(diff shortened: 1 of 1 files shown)


cdfisher:
There are already a couple raid-specific plugins which serve a similar purpose, including [ToB Gear Checker](https://runelite.net/plugin-hub/show/tob-gear-checker) and [ToA Spellbook Reminder](https://runelite.net/plugin-hub/show/toa-spellbook-reminder). We would prefer that you try to contribute to one of the existing plugins to help add functionality to them rather than adding another plugin to do much of the same thing to the Plugin Hub, since that leads to less work for the hub reviewers (who are volunteers), and a better experience for end users.

PR #8538, opened 2025-07-21
Update flipping-copilot to v1.7.4

cbrewitt:
- Add more advanced flip history management/analysis
- Add select all/deselect all options to item blocklist


LlemonDuck:
I don't like what you have going for ClientThreadUtil#executeInClientThread, if you admit it's a code smell then just fix the code that uses it?

It makes the plugin much harder to review overall, and anything calling it from the EDT would block the EDT.

cbrewitt:
We’ve now removed ⁠ ClientThreadUtil#executeInClientThread ⁠ completely.

PR #8539, opened 2025-07-21
Add Prayer Flick Overlay Plugin

slategraydev:
I threw together a prayer flicking overlay plugin with a few neat features. By default it tries to mimic the visual functionality of the Quickprayer orb, and has a secondary setting where it mimics the prayer bar of the StatusBar plugin.

It also has various settings like a vertical mode, different animation styles for the tick bar, color changing/resizing etc.
I know someone else recently released a prayer tick plugin, so I've kept out their core functionality of showing and tracking the player clicks.

If it's too similar to theirs I'll just kill this. I'm also pretty novice to Java development, so any feedback is appreciated.


riktenx:
This does not appear to differ significantly from the recently-added plugin pajlada mentioned, so we would prefer you just make contributions to that one instead.

slategraydev:
Will do.

PR #8546, opened 2025-07-22
Create scroll-box-info

IEarnSolo:
This plugin tracks various clue-related information and displays them through customizable tooltips, item overlays, infoboxes, chat messages, and menu options.

Features include bank counts, total counts, current stack limits, how many clues until next stack limit increase, infoboxes indicating full stacks, etc.

### DIFF
--- a/plugins/scroll-box-info
+++ b/plugins/scroll-box-info
@@ -1,2 +1,2 @@
 repository=https://github.com/IEarnSolo/scroll-box-info.git
-commit=9d3c2b69dc3acff7f887d4675a1c2b419d00cf09
+commit=71a1a6b4d352652ce2267482e7e28a3f5be4f176
--- a/plugins/scroll-box-info
+++ b/plugins/scroll-box-info
@@ -1,0 +1,2 @@
+repository=https://github.com/IEarnSolo/scroll-box-info.git
+commit=9d3c2b69dc3acff7f887d4675a1c2b419d00cf09


coopermor:
We would prefer you contribute these features to scrollboxcounter rather than introduce plugin duplication, which is more work for plugin hub reviewers and causes confusion for users.

IEarnSolo:
> I am the maintainer of scrollboxcounter. Yours looks way more flashed out and has alot of good options. Maybe you can write me on discord and we can merge our plugins or find a different solution.

Hey hope you’re well! I’ll send you a message a little later tonight once I’m off work

riktenx:
You can both be maintainers -- it doesn't really matter who owns the repo, although changing it does require some extra steps. Fundamentally though the contributions made here should go to the existing scroll box counter plugin such that you don't have to start the userbase over from scratch.

The least number of steps would be
1. @IEarnSolo makes their contributions to @thelaakes plugin.
2. @thelaakes gives @IEarnSolo collaborator access to their repo.
3. Submit an update to the plugin hub and add the `authors` field in the plugin manifest with both of your GitHub names ([example](https://github.com/runelite/plugin-hub/blob/master/plugins/dink#L3)).

IEarnSolo:
> You can both be maintainers -- it doesn't really matter who owns the repo, although changing it does require some extra steps. Fundamentally though the contributions made here should go to the existing scroll box counter plugin such that you don't have to start the userbase over from scratch.
> 
> The least number of steps would be
> 
>     1. @IEarnSolo makes their contributions to @thelaakes plugin.
> 
>     2. @thelaakes gives @IEarnSolo collaborator access to their repo.
> 
>     3. Submit an update to the plugin hub and add the `authors` field in the plugin manifest with both of your GitHub names ([example](https://github.com/runelite/plugin-hub/blob/master/plugins/dink#L3)).

If the repo ownership was transferred to me, what would those extra steps be? If you compare the plugins, it would make more sense to import his item overlays into mine, although if that creates more work on your end then that's understandable why you wouldn't recommend that. But if it's set up in the plugin hub to point to his repo, then I could work with that if it were transferred. Would removing his plugin entirely and adding mine be a quick solution, or is that too drastic? His plugin is still fairly new and the new features would have to be reviewed regardless of any route we take

PR #8552, opened 2025-07-23
Add LimeyTileMan

ModestLime:
LimeyTileMan is a highly optimized Tileman plugin that marks tiles the player has walked on, ensuring zero lag even with extensive tile sets.

### DIFF
--- a/plugins/limeytileman
+++ b/plugins/limeytileman
@@ -1,2 +1,2 @@
 repository=https://github.com/ModestLime/LimeyTileMan.git
-commit=ef93a5ece876dabad7326c0be0467f433f77dd7d
+commit=77961f4f8eb03c23d32d23c2d3b225a5133b227c
--- a/plugins/limeytileman
+++ b/plugins/limeytileman
@@ -1,2 +1,2 @@
 repository=https://github.com/ModestLime/LimeyTileMan.git
-commit=a97ea17382b8346331628e98eaf3e68a4ab2697a
+commit=aaf58065c86f1c9a2b2b1bce3f76c2a3e3a77a58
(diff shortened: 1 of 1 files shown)


LlemonDuck:
What makes this "highly optimized"? Can you contribute these improvements to the existing plugin so existing users can benefit?

ModestLime:
As for why is it highly optimized, my plugins on gametick only takes about 1 millisecond thanks to storing tiles in 8x8 chunk regions instead of 64x64 like the base game and utilising a bunch of bitwise for division, having a highlighting of 1 clickable tiles that is O(n), saving the tileset only when the plugin is turned off/client is closed or every 10minutes to avoid the overhead of serialising thousands of world points, having fiddle around with how to render tiles faster iv combind all polys into a path2d and manual do graphics calls instead of using OverlayUtil.renderPolygon to eliminate the overhead of 1 time use objects like the stroke and taking advantage that with gpu enabled when render gets called it gets passed a empty graphics2d so we can ignore all color mixing to gain quite a bit of fps, and lots of other small stuff im made sure to make this fast since the other tileman plugin stutters hard with large amount of tiles since its extremely badly optimized since every time you move in that other plugin it will get every configuration key its ever made and then parse the string and then add the size of each array to get how many tiles you have walked on total and it does this every time you move regardless if you actualy steped on a new tile. and yea there has been 3 pull requests to fix performance in that other existing tileman plugin all from 2022 and not merged so its been a known issue and the maintainers show 0 interest in fixing it and also seem to have 0 interest in letting other people fix it via a pull hence the 3 pulls requests that are over 3 years old that havent been merged

ModestLime:
sorry i removed some code in my gradle.build that i saw you tell someone else to remove and then i kinda messed up my fork i hope i rebased properly, I dont really know how to github that great 😧

PR #8553, opened 2025-07-23
Add the new varlamore totem fletching

TheRealGuru:


LlemonDuck:
remove the overlay in shutDown

PR #8559, opened 2025-07-23
Colorblind Helper for Doom of Mokhaiotl

lilbouv:
Colorblind helper for Doom of Mokhaiotl. Gives Inferno and/or ToA animations to the Range and Mage orbs.


LlemonDuck:
change the config group to something more accurate, like "doomcolorblind"

PR #8569, opened 2025-07-24
Add Gemstone Crab Timer

ShaneeexD:
Tracks the HP of the gemstone crab boss, and sends a desktop notification when it reaches a certain threshold (2% by default), for a better afk experience (especially for one monitor users).

Also includes a DPS calculator and highlights the tunnel upon the boss' defeat.

Everything is toggleable and configurable.


LlemonDuck:
I think you could simplify the xp tracker stuff significantly by just using onHitsplatApplied and checking isMine

PR #8571, opened 2025-07-24
Bump osrs wiki crowdsourcing

andmcadams:
- Adds tracking for cooked stuffed snakes
- Refactors a lot of magic consts to gamevals


andmcadams:
ah sorry about that, was branching off an old commit

PR #8580, opened 2025-07-25
rooftopfpmog

Remcodeit:
timer for force spawning mog

### DIFF
--- a/plugins/rooftopfpmog
+++ b/plugins/rooftopfpmog
@@ -1,0 +1,2 @@
+repository=https://github.com/Remcodeit/RooftopFPMOG
+commit=b241ac20bf0585e38e836f99c3f83fc793977853b241ac20bf0585e38e836f99c3f83fc793977853


pajlada:
does https://runelite.net/plugin-hub/show/afk-marks-canafis not do what you want already?

Remcodeit:
> does https://runelite.net/plugin-hub/show/afk-marks-canafis not do what you want already?

the timer is off, its supposed to be 2:15 second timer, this my first plugin i coded and i wanted get more hands on work with coding plugins

cdfisher:
If that plugin is wrong, we would strongly prefer for you to just submit a PR to that repo to fix it rather than adding a whole new plugin to the hub.

Remcodeit:
> If that plugin is wrong, we would strongly prefer for you to just submit a PR to that repo to fix it rather than adding a whole new plugin to the hub.

alright i understand

PR #8590, opened 2025-07-25
create: Spec bar utilities

Car-Role:
There is naturally a 1 tick delay to give the user feedback on if they clicked the specbar or not, this plugin fixes that


tylerwgrass:
Please remove the changes in this PR for `distraction-reducer`

Car-Role:
> Please remove the changes in this PR for `distraction-reducer`

I reverted it back to what it was previously. I'm not sure if there is a better way to discard that accidental change

PR #8591, opened 2025-07-26
Submitting LastMan plugin

Tellik:
The Last Man on Gielinor, allows toggling visibility of NPCs with "Talk-To" interaction, as well as adding a white/black list to target specific NPCs visibility status.

### DIFF
--- a/plugins/lastman
+++ b/plugins/lastman
@@ -1,0 +1,2 @@
+repository=https://github.com/Tellik/plugin-hub.git
+commit=0bcd9bbec1af7b3b75eae8acc9d9e957cce7ae9b


plugins/lastman
```diff
@@ -0,0 +1,2 @@
+repository=https://github.com/Tellik/plugin-hub.git
```
tylerwgrass, line 1:
```suggestion
repository=https://github.com/Tellik/last-man.git
```

tylerwgrass: Plugins with manual ID configuration are not allowed due to their difficulty to moderate.
https://github.com/Tellik/last-man/blob/master/src/main/java/com/lastman/LastManConfig.java#L82-L104

tylerwgrass (changes requested):

plugins/lastman
Tellik: Once it failed, I realized I had copied and pasted messing up the repository/commit hashes. Was working on fixing that now.

As for manual ID configuration, does that mean that everything needs to be hardcoded and can't be configured by within the plugin itself?

I see where this is mentioned:

ID based plugins: Plugins that use player provided IDs for the entirety of their functionality can cause moderation issues and outcomes that break Jagex's plugin rules. Due to this fact we will not be accepting any new ID based plugins. Plugins that use a specific set of IDs but do not allow user input will still be accepted. (e.g. plugins like Vardorvis Projectiles which only allows you to change a specific projectile to a set list of projectiles.)

I was hoping since it doesn't use player provided IDs for the entirety of their functionality would be able to satisfy this requirement. The idea being that the plugin will be able to toggle visibility on a large group of NPCs based on the interaction of 'talk-to', but there are a lot of other NPCs that would need to be added on the fly since there isn't a way to target them in code specifically. I do understand the reasoning to a degree.

Tellik:
I'm going to ahead and close this for now, until I can get the functionality that I want while still meeting the requirements set forth. It'll probably entail hard coding everything, which there are quite a few NPC IDs to parse through manually.

PR #8594, opened 2025-07-26
Spec bar utilities

Car-Role:
Spec Bar Utilities
A RuneLite plugin that enhances the usability of the special attack bar in Old School RuneScape by providing clear visual feedback for hover and click states. Knowing exactly when you clicked the spec bar, without a next tick delay.

There is naturally a 1 tick delay to give the user feedback on if they clicked the specbar or not, this plugin fixes that

Configuration
The plugin provides several customization options in the RuneLite settings panel:

Visual Effects
Enable Hover Effect: Toggle hover state visual feedback
Enable Click Effect: Toggle click state visual feedback
Effect Opacity: Adjust transparency (10-255)
Colors
Hover Color: Customize the color shown when hovering (default: Yellow)
Click Color: Customize the color shown when clicking (default: Green)


Car-Role:
I am unsure why distraction reducer updates are being picked up in this pr

PR #8613, opened 2025-07-27
update log-master to v1.1.1

rmobis:
This update includes:

- Big refactor with considerable perf improvements on the save data/task list loading
- Fix for a RuneProfile incompatibility on the collection log hamburger menu
- New deprecation warning for parts of the UI that will be removed in the future


rmobis:
Looks like I broke something... let me take a look

rmobis:
Looks like the issue is with the dependency I added. Are we not allowed dependencies or is it something with this specific library?

rmobis:
Dependency removed, I rolled my own implementation of a very very simple debouncer.

PR #8618, opened 2025-07-28
Update pet-spell-blocker

Sacca-1:
Extended the plugin with two more features, and therefore renamed the plugin to "Wildy QoL". By default, the new features are disabled, so the plugin will behave the same for existing users.


LlemonDuck:
This PR still has both changes

Sacca-1:
I misunderstood then, that should do it

iProdigy:
> Change the repo URL in one PR but keep the commit hash the same

the commit hash has still changed

Sacca-1:
> Change the repo URL in one PR but keep the commit hash the same, then submit another PR to change the commit hash.

I interpreted this as creating two new PRs, but I see you meant keep the commit hash the same in this PR, which is what I've now done.

iProdigy (approved):

riktenx (approved):

PR #8619, opened 2025-07-28
Add projectile-override v1.0.0

Loze-Put:
This plugin replaces boss projectiles with alternatives that are easier to distinguish in color, shape, or contrast. It aims to improve accessibility for colorblind and visually impaired players by reducing ambiguity in high-pressure PvM encounters. All projectiles are only replaced by the projectiles of other bosses to comply with the Jagex third party plugin statement. Some other similar plugins exist and I don't want to create a new plugin for each boss.

### DIFF
--- a/plugins/projectile-override
+++ b/plugins/projectile-override
@@ -1,2 +1,2 @@
 repository=https://github.com/Loze-Put/projectile-override.git
-commit=0a34d42f3f66e7d83749cbd221cc8e88eadfdc0d
+commit=2ddf793140997940bd3ec71a315056fb9634aae3
--- a/plugins/projectile-override
+++ b/plugins/projectile-override
@@ -1,0 +1,2 @@
+repository=https://github.com/Loze-Put/projectile-override.git
+commit=0a34d42f3f66e7d83749cbd221cc8e88eadfdc0d


tylerwgrass:
[This plugin](https://runelite.net/plugin-hub/show/doom-of-mokhaiotl-colorblind-helper) exists to fix the experience for doom of mokhaiotl, perhaps you two could collaborate to bring the experience to a single plugin, and expand to include what you have done?

Loze-Put:
Currently the following plugins with similar features exist:

- [doom-of-mokhaiotl-colorblind-helper](https://runelite.net/plugin-hub/show/doom-of-mokhaiotl-colorblind-helper)
- [vardorvis-projectiles](https://runelite.net/plugin-hub/show/vardorvis-projectiles)
- [hunllef-projectiles](https://runelite.net/plugin-hub/show/hunllef-projectiles)

I'm not opposed to collaborate with other people to combine features into a single plugin. How do you suggest approaching this? I could contact the doom plugin creator, but that would still leave the other two out of the loop. I think the best result would see all of these combined with the bosses I added in my plugin.

LlemonDuck:
If your plugin already handles all the cases the other plugins collectively do then you're probably best off just seeing if they want to direct their users to your plugin / submit any new changes to your plugin. We can't merge plugin entries but there's a few ways that users could be notified to migrate.

Loze-Put:
Thank you for your input. I will contact the plugins` owners to see what their opinion is. In the case of a redirection of users, what do you think is the best approach? I could think of adding a chat message when a player walks into the area that the plugin operates in.

If you could review this now, that would be great.

LlemonDuck:
>  I could think of adding a chat message when a player walks into the area that the plugin operates in.
this sounds reasonable, or if the other authors decide to merge fully, either a chat message on startUp or a plugin disabling with message could work

LlemonDuck:
https://github.com/Loze-Put/projectile-override/blob/0a34d42f3f66e7d83749cbd221cc8e88eadfdc0d/src/main/java/com/projectileoverride/ProjectileOverridePlugin.java#L49-L52

ConfigChanged fires for all configs not just your plugins'. Make sure you guard by checking the config group.

Loze-Put:
Thank you, I have updated the method.

PR #8633, opened 2025-07-29
Update demonic-larva-tracker to v1.2.1

marknewan:
Fixes a small bug, not likely to be encountered/noticed, due to a typo.


riktenx (approved):
Consider just downgrading all your logs to `log.debug` instead of doing `info` behind a debug flag.

iProdigy (approved):

PR #8637, opened 2025-07-29
Update Wildy QoL to 1.1.0

Sacca-1:
Extended the plugin with two more features, by default, the new features are disabled, so the plugin will behave the same for existing users. (second PR with the new repo and new commit hash)


riktenx:
> ### NPC Spell Blocker
> * Removes "Cast" menu entries on NPCs while in dangerous areas (wilderness, PvP worlds, Deadman worlds)
> * **Warning**: This removes the option to cast spells on NPCs altogether, so make sure you only enable this if you're sure you don't want to cast spells on NPCs

This is not allowed per this 3p client guideline:

> Makes it easier to target 3D entities with a spell by removing some options

Please review https://oldschool.runescape.wiki/w/Update:Third_Party_Client_Guidelines

Sacca-1:
Thank you for the info, I've removed the feature and will review the guidelines more thoroughly.

tylerwgrass:
Please exclude or remove the `bin/*` files

PR #8639, opened 2025-07-29
Create keybind-button-overlay

LeeT96:
This is different to a similarly-named plugin, whereby this plugin displays hotkey/keybinds on the icons/tabs themselves, whereas the other plugin shows them in an infobox.

### DIFF
--- a/plugins/keybind-button-overlay
+++ b/plugins/keybind-button-overlay
@@ -1,0 +1,2 @@
+repository=https://github.com/LeeT96/keybind-button-overlay.git
+commit=e66746d586818aaae0e67cbdbd02bf7a80c18e4f


riktenx:
+1 to @cdfisher's comment. If the author of said plugin is unresponsive, see https://github.com/runelite/runelite/wiki/Plugin-takeover-policy.

cdfisher:
> I would argue that it is a different plugin, however, seeing as someone published a plugin called "Rogue Silencer" which is very very similar to the plugin I created but was published as its own plugin.

The approach we've taken towards plugins which offer similar functionality to other existing plugins has shifted over time and we now strongly encourage collaboration to add more functionality to existing plugins over adding new plugins which do largely the same thing. This helps reduce the amount of work that the (volunteer) hub reviewers need to do, and helps improve the experience for end users as well.

HydrogenHydroxide:
Hello!

looking at new pull requests as I await my review and it's a key overlay.

I definitely do not want my shortcut keys over my icons... having that said I have no experience on how to code, and the plugin works as imagined in developer mode.

There is no plugin for overlay/info boxes for 'hot keys' or 'shortcut keys' or 'f keys'...

The plugin I submitted (KeyOverlay), was done painstakingly over 8 hours which came after 3 hours of going through every single plugin on the hub **at the very least 3 times**.

I would be extremely happy to assist in combining the development of our plugins to be configurable to a large audience.

P.S. In regards to Thrall Helper AKA Spell Reminder or whatever it's called. I created a plugin to remind myself to reanimate ensouled dragon heads. 

However some criticism as I had to find Thrall Helper under it's new name which is a pain in the ass, and was reminiscent of going through every plugin (yes, all 1106 or whatever it is). Is there anyway we can preserve names, or archive the out-date /name changes / removed ones so people know when their plugin has been depreciated/renamed?

HydrogenHydroxide:
Hey @LeeT96 are you still trying to publish a key overlay?

PR #8646, opened 2025-07-29
Create doom-projectile-replacer

RuneAaron:
This plugin is to replace the Doom of Mokhaiotl / Osto-Ayak projectiles with alternative projectiles from TOA, TOB or CoX. 

Created to support colourblind players struggling with identifying the default projectiles.


riktenx:
(and if this submission includes features that the existing plugin does not, you should seek to contribute to that plugin instead)

RuneAaron:
Sods law it wasn't there when I checked before :P thanks guys i'll raise a pull there.

PR #8651, opened 2025-07-30
Create PetTheAlpaca

nruchs:
A fun and interactive RuneLite plugin that allows you to "pet" adorable alpacas in-game and receive humorous chat messages. Keep track of how many alpacas you've petted!


LlemonDuck:
Use `client.getMenu().createMenuEntry` instead of the deprecated version

FYI configs support setters as well, just use the same keyName, a parameter, and no body:
```java
    @ConfigItem(
            keyName = "petCount",
            name = "Pet Count",
            description = "Total number of times an alpaca has been petted.",
            hidden = true
    )
    void int setPetCount(int value);
```

nruchs:
Hey! Thank you so much for the review and the suggestions, LlemonDuck!

I have implemented the changes by:
- Replacing the deprecated methods for the menu, NPC list, and player animation.
- Adding a variety of new messages to make the plugin more fun.
- Implementing the suggested setters in the config for a cleaner approach.

Let me know if there's anything else that needs to be addressed. :)

PR #8660, opened 2025-07-31
Custom preset colour support

DenchyRS:
made code less jank
added ability to custom colour presets


LlemonDuck:
>Feedback:
> *    Use of java.lang.System.out is not allowed: system i/o, use a logger instead.

PR #8662, opened 2025-07-31
Add "UI Toggle" Plugin

ItsThatNova:
This plugin allows players to toggle the visibility of key UI elements (e.g. chatbox, inventory, minimap orbs, equipment tab, etc.) via both individual configuration toggles and a global hotkey. This aims to let players set what they see based on what task(s) they're doing, or clear the screen for screenshots or immersion.

### **Features**
Configurable checkboxes for showing/hiding individual interface elements.

A single hotkey to toggle all visible elements at once (and restore them).

Automatic state memory to restore UI to its previous state.

Fully respects game-driven UI changes (such as the logout tab appearing during combat or skilling).

### **Compliance Notes**
This plugin does not override or re-show UI elements hidden by the game itself.

If an interface is hidden by the game while toggled on by the user, the plugin disables automatic restoring and adjusts config accordingly.

Widgets will only be re-shown if they were user-hidden prior to the game hiding them.

### **Additional Info**
Built against the latest client and jshell versions from RuneLite’s Maven.

All visibility logic is implemented using Client.getWidget(interfaceId) and Widget.setHidden(boolean) safely.


LlemonDuck:
Still wrong. Also have you seen [Hide Widgets](https://runelite.net/plugin-hub/show/hide-widgets)?

ItsThatNova:
> Still wrong. Also have you seen [Hide Widgets](https://runelite.net/plugin-hub/show/hide-widgets)?

Still wrong, eh? Forgive me, as this is (obviously) my first time doing anything of the sort here on Github. Can I get an elaboration of what it is I'm missing in that submission document?

As for Hide Widgets, I was under the impression it may not work for RuneLite widgets. I wanted to get a base working version going and expand on its capabilities in the future. If it's truly identical, and that plugin is still being maintained then I suppose I could retire the idea.

LlemonDuck:
I'm not sure where you even got your instructions but tyler already gave you the link to the correct ones. You need to remove the json file and the jar, and flatten your manifest (move it out of being in a folder)

We'd also much prefer if you collaborate with the existing plugin to add what you want there instead of submitting a completely new one

ItsThatNova:
Lol holy swing and a miss. This is why I should stick to scripting hahahaha.

Appreciate the elaboration. I'll try to find more than two brain cells to
do it properly and will look at working with the existing one instead of
one from scratch.


Cheers!

PR #8667, opened 2025-08-01
Add land-surveyor

kilgore-vanhaus:
Displays elevation data of surrounding tiles in debug-like format or visually via colored tiles, the colors of which may be optionally customized by the user.
Includes Hypsometric Mode, an elevation-based color gradient applicable to the floating text, underlying tile, or both simultaneously, allowing for the creation of detailed topographical maps.

### DIFF
--- a/plugins/land-surveyor
+++ b/plugins/land-surveyor
@@ -1,0 +1,2 @@
+repository=https://github.com/kilgore-vanhaus/LandSurveyor.git
+commit=19d61195dc2602319d06941434b16a1e477b2dea


kilgore-vanhaus:
> What is this used for?

I personally needed precise elevation data for a pet project (mapping a scale replica of Gielinor), and there were no tools I could find that would provide me with the elevation info I required. While I can't imagine the plugin would help in accomplishing many (if any) in-game goals, there are plenty of niche scenarios similar to my own where the height/elevation information provided by the plugin is needed, yet unobtainable through other means.

LlemonDuck:
Does this need to go on the hub then? I can't imagine any other users would find this useful and you are more than free to continue using it yourself through a dev build.

kilgore-vanhaus:
> Does this need to go on the hub then? I can't imagine any other users would find this useful

I fail to see your point. The plugin is not malicious, doesn't break Jagex's rules, and is not one of RuneLite's rejected features. Those are the only prerequisites I see listed for plugin hub submission.

With no offense directed to either of these developers, would you consider many users finding ["The Guns' Gains"](https://runelite.net/plugin-hub/show/guns-gains) "useful"? Would you consider ["Bonk Messages"](https://runelite.net/plugin-hub/show/bonk-messages) a "useful" plugin used by a lot of users? I would say no, yet they are both on the plugin hub. It seems clear to me that not every plugin needs to be a tool to explicitly assist with in-game progression, and not every plugin needs to see widespread adoption.

I know at least one other person who would stand to benefit from using my plugin, and that says to me that there are others across the internet who will likewise find my plugin useful if publicly available. If there is any further documentation regarding plugin submission guidelines that I may have missed, please let me know.

coopermor:
Llemon's point was not that it would be prohibited from the hub, but rather if you are asking for it to be added to the hub you are asking for reviewer's time to review it. If it was something that was only useful to you then you could just use it yourself without putting it on the hub burdening reviewers.

PR #8671, opened 2025-08-01
Update Wildy QoL to 1.2.0

Sacca-1:
Added a new feature that caches friends chat members to prevent them from showing up as enemies on player indicators while the friends chat loads on world hops.

### DIFF
--- a/plugins/pet-spell-blocker
+++ b/plugins/pet-spell-blocker
@@ -1,2 +1,2 @@
 repository=https://github.com/Sacca-1/Wildy-QoL.git
-commit=f9956cc022a4e6c93b83eab421c0e2d851f9b76a
+commit=22e089208a091aec418b6b593523bafee2a7afcc


tylerwgrass:
This looks like it will continue to indicate a user who was cached from being in the friends chat, even if they choose to appear offline, which would not be allowed.

Sacca-1:
While that is not the intended use case, I can see how it could be used that way and understand the privacy concern. To mitigate this, I could adjust the following:

- Limit the drawing of indicators to PvP areas
- Limit the maximum duration players can be cached for

This way the data is short-lived and only used where misidentifying targets has gameplay impact.

In my opinion, this makes it a very different feature to the previously rejected https://github.com/runelite/runelite/pull/951, which could be used to highlight users who never showed any intent to interact with you indefinitely by adding them to the friends list.

Hope you can agree, thank you for the consideration.

LlemonDuck:
This gets too near to the following _two_ rules for me to feel comfortable merging it (emphasis mine):
> Offers additional information about other players, _**_with the purpose of scouting PvP targets_**_

> Which gives summary information about a group of players, _**such as how many players are**_ attackable and _**not in your own CC**_, what prayers they are mostly using, _**etc.**_

In particular the second. 

Although rules have generally relaxed in PvM over the last few years, PvP hasn't seen the same level of change and so is still held to a quite stringent bar.

LlemonDuck:
I understood what the scope was. As you mentioned, there are intended workarounds for this.

PR #8674, opened 2025-08-01
tagPlayersPlugin

AGGnomedev:
A simple plugin that allows users to add a tag/nickname to a player that is saved to a text file. The tag/nickname is visible while a customisable hotkey is either held down or toggled depending on the settings. Useful for keeping track of players in places like FFA team bossing where you often see the same players and want to remember aspects about someone.


tylerwgrass:
This can be used to circumvent players who choose to appear private, which we choose not to support.

Please see: https://github.com/runelite/runelite/wiki/Rejected-or-Rolled-Back-Features

PR #8676, opened 2025-08-01
Adding Pushcut plugin for push notifications on iOS

usa-usa-usa-usa:
A plugin that sends push notifications to iOS using PushCut. It’s way more flexible for iPhone than other options like ntfy or pushbullet—lettingyou trigger shortcuts or open URLs just by tapping the notification.

Main reason I built it: I AFK gemstone crabs on OSRS using a remote desktop app. OSRS Mobile disconnects when it’s not in the foreground on iOS (like 30 seconds max), but with RDP I can keep it going even when the app’s closed. So now, when a crab burrows and triggers a chat message, the plugin sends me a PushCut notification. I’ve set it so tapping the alert launches my RDP app right away.


riktenx:
You can't add a dependency as a jar like this. See https://github.com/runelite/plugin-hub?tab=readme-ov-file#third-party-dependencies.

As stated in the above link, adding a new 3p dependency like this adds significant time to the review process. It is unlikely that this plugin will be reviewed or accepted within the next few months, if ever.

In general this plugin shouldn't be necessary. Take a look at Dink, https://runelite.net/plugin-hub/show/dink, which can post to a Discord webhook on chat message. You should be able to simply set up push notifications on your device for whatever channel it sends to. I understand that you won't get the one-click functionality from PushCut but as it stands I don't see this plugin being added to the hub.

usa-usa-usa-usa:
I was not aware, I will add the dependencies.

There are two other plugins that integrate with push notification services (Push Notifications and ntfy). Both of which I had used but wanted the further capability to trigger Apple shortcuts for the purpose of automatically bringing my remote app to the foreground automatically. I think that functionality sets it apart from the other plugins I mentioned.

I will update this PR, and see what happens. In the mean time, I will reach out to @dekvall, and see if there would be any interest in me adding Pushcut to the Push Notifications plugin.

TheRealGuru:
If you want to enable push notification like features, you are better off doing it via a Rest API call rather than a library. If whatever you're using hasn't got a rest api, then you can always build your own and then trigger the notifications. Naturally both of these will require some sort of api key authentication to ensure you  don't get screwed. But also note that I believe plugins that use Rest API calls need to declare that in a warning to the users.

PR #8677, opened 2025-08-01
Add Image Bankai

DeadRobotDev:
The plugin allows you to display a custom image on the client.

One use case is to display next to Status Bars to replicate WoW Unit Frames.

This is the example I was working off. Image is from OBS.
image

This is the result. Image is from the plugin.
image


LlemonDuck:
don't load the image every frame; only load it if you haven't previously

PR #8680, opened 2025-08-01
Updated Clan Spam Filter

andreasx23:
1. Fixed a bug where if the player optained a pet and it got into their backpack it wouldn't be filtered by "filterPets()" function
2. Updated the internal collection log map to contain the new bosses


riktenx (approved):
You might want to refresh your plugin-hub fork: https://github.com/runelite/plugin-hub?tab=readme-ov-file#updating-a-plugin

tylerwgrass (approved):

PR #8682, opened 2025-08-01
Add blackjack glance alert plugin

haneri98:
Plays a sound when the player fails to put a bandit to sleep while blackjacking.

### DIFF
--- a/plugins/blackjack-glance-alert
+++ b/plugins/blackjack-glance-alert
@@ -1,0 +1,2 @@
+repository=https://github.com/haneri98/BlackjackAlert.git
+commit=ba591ea38169ee1e1243192c7a977368ca80ad8c


cdfisher:
Can you not just use Watchdog to accomplish basically the same thing?

haneri98:
@cdfisher Chat notifications does not accomplish the same thing unfortunately. Haven't used watchdog, but the config looks quite confusing so I would rather use my plugin since it works.

Felanbird:
> @cdfisher Chat notifications does not accomplish the same thing unfortunately. Haven't used watchdog, but the config looks quite confusing so I would rather use my plugin since it works.

We're not really interested in adding a plugin that does one single thing, that another plugin that does a metric ton of things, does. You should put at least some form of effort in and see if it works (it probably does)

Felanbird:
> That's fair. I was wondering why there were no useful plugins for blackjacking

Probably because the "useful" plugins for blackjacking that you are thinking of, are explicitly prohibited by Jagex. 

From the README:

## Reviewing
We will review your plugin to ensure it isn't malicious, doesn't [break Jagex's rules](https://secure.runescape.com/m=news/third-party-client-guidelines?oldschool=1), 
or isn't one of our previously [Rejected/Rolledback features](https://github.com/runelite/runelite/wiki/Rejected-or-Rolled-Back-Features).  
__If it is difficult for us to ensure the plugin isn't against the rules we will not merge it__. 

> Any feature which...
Modifies menu options for blackjacking, such as 'Pickpocket' and 'Knock-out'

PR #8691, opened 2025-08-02
Update flipping-copilot to v1.7.5

cbrewitt:
Bugfix: huge array allocations
Add jptrs93 as plugin author

### DIFF
--- a/plugins/flipping-copilot
+++ b/plugins/flipping-copilot
@@ -1,3 +1,4 @@
 repository=https://github.com/cbrewitt/flipping-copilot.git
-commit=c432d868e4fc66b22380e5704c11d7e2b2eadb1e
+commit=f074ae0adec295a6d18889f920a2f1265dd6f088
+authors=cbrewitt,jptrs93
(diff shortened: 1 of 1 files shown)


jptrs93:
> You should consider smaller patches if you more timely reviews. You should also consider resizing your image down to a more meaningful size.

@raiyni this is actually a very small change but the github compare is weird due to our commit graph. Sorry this is our fault. We need to adjust our branching workflow. Only the second commit is new. Using github double dot rather than triple dot compare shows the correct diff:

https://github.com/cbrewitt/flipping-copilot/compare/c432d868e4fc66b22380e5704c11d7e2b2eadb1e..cbrewitt:f074ae0adec295a6d18889f920a2f1265dd6f088

raiyni:
Please use the correct hash that's in the tree then please.  Looks like `8dd4886f013c63adc18438e648133349ebc723f1`

jptrs93:
> Please use the correct hash that's in the tree then please. Looks like `8dd4886f013c63adc18438e648133349ebc723f1`

Updated to the correct commit now.

cbrewitt:
It looks like I may need to comment to remove the "waiting for author" tag.

PR #8696, opened 2025-08-02
Update hot-cold-helper v1.1.3

lalochazia:
Minor Changes:

- Small fix to incorrect coords
- Added new Pendant of Ates teleport locations
- Added toggle for Master Scroll Book
- Exposed final clue location for use by other plugins


iProdigy:
hub plugins do not share the same classloader; for inter-plugin communication you want to post [PluginMessage](https://static.runelite.net/runelite-client/apidocs/net/runelite/client/events/PluginMessage.html)

iProdigy (dismissed):
the `solvedLocation` field doesn't seem to be queried (and is always null)

iProdigy:
are you expecting to be able to call `getSolvedLocation` from another plugin? (that won't be possible)

PR #8706, opened 2025-08-04
Adding "Hide Prayer Descriptions" plugin

SamuelDev:
This is a plugin that will hide different parts of the prayer description tooltip on hover (or the entire thing).

### Examples
No tooltip (all config options selected to hide):
image

Only show name:
image

Show level and drain rate:
image


riktenx:
still 404

PR #8741, opened 2025-08-06
Update vanced-barrows

Inzomniakk:
Added 2 images and a cycler. Yes these are my images.


Inzomniakk:
That doesn't make sense to me. The file I uploaded is nowhere near that size.

Inzomniakk (commented):
Shrank dimensions.

Felanbird:
Java will blow up an image to it's max size when it loads it into memory. (width * height * 4)

riktenx (approved):

PR #8751, opened 2025-08-08
Update poison-moo to v1.1.0

remindful7560:
Added some options for antipoisons, cure ticks, and the ability to play a sound effect.

Thank you for your work reviewing.


remindful7560:
Improved some calculations, and added some safety clamps just in case I got my math wrong.

PR #8759, opened 2025-08-09
Update Vitality

Natesky9:
bugfix, changing the offset from true to perspective to fix the artifact with different zoom levels


tylerwgrass (approved):
If you want to have debug code, you can inject `developerMode` instead of bloating your reviews with garbage by commented out code

Natesky9:
some of that debug code is fallback in case the changes don't work. I know I could just look at previous versions, but it's easier to compare adjacent. I do actually have dev mode on

PR #8764, opened 2025-08-09
Update Metrics Tracker to v1.0.4

touchtouch105:
Added ability to pin metrics to canvas


touchtouch105:
Only feature being implemented is ability to add/remove items to canvas, as seen below

image

image

image

riktenx (changes requested):
You need to use the `OverlayManager` to exhaustively remove any overlays you have added in your plugin's `shutDown()`. Right now it appears you're only removing them as part of UI actions taken by the user, if the user just turns off your plugin with overlays added they won't be removed.

touchtouch105:
> You need to use the `OverlayManager` to exhaustively remove any overlays you have added in your plugin's `shutDown()`. Right now it appears you're only removing them as part of UI actions taken by the user, if the user just turns off your plugin with overlays added they won't be removed.

Addressed

PR #8766, opened 2025-08-10
Update max-hit-calculator to v2.1.1

j-cob44:
v2.1.1:
- Add Varlamore pt 3 weapons and specs
- Add Keris Partisan of Amascut calc
- Fix Twinflame calcs
- Fix some spec weapon calcs
- Fix Giant Mole weakness type

### DIFF
--- a/plugins/max-hit-calculator
+++ b/plugins/max-hit-calculator
@@ -1,2 +1,2 @@
 repository=https://github.com/j-cob44/max-hit-calc.git
-commit=b747687c65f2ce63f046193ddd58367c6bd3d2d4
+commit=9a6751185b1238b3a3bbd2884320ecf39b61ed22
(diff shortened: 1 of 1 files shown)


iProdigy (approved):

riktenx (approved):
You can gate [stuff like this](https://github.com/j-cob44/max-hit-calc/compare/b747687c65f2ce63f046193ddd58367c6bd3d2d4...j-cob44:ad2da04f6b562cf2037b98bfbd72521e54c25ceb#diff-da91be7960c9fd4a93bff58d3e2413846aa823904658ea78357d9e9abe70de31R143) behind the injectable developer mode flag. It will automatically be set if you run the client with `--developer-mode`.

[example](https://github.com/riktenx/loot-filters/blob/main/src/main/java/com/lootfilters/LootFiltersPlugin.java#L119)

PR #8769, opened 2025-08-10
Create KeyOverlay

HydrogenHydroxide:
KeyOverlay displays 'Key' Presses. An overlay that lights up for feedback on key press.

Intended for use with Key Remapping so you can see which Shortcut Key you recently hit (F Keys, Hotkeys, ect.)


HydrogenHydroxide:
Update names so there isn't an issue with capital letters when building.

PR #8777, opened 2025-08-12
Update Creator's Kit to v2.0.13

ScreteMonge:
v2.0.13
Re-enabled overlays and included a blacklist for TOA puzzle


ScreteMonge:
I will reach out to Jagex and bring up this scenario though. Might be something they should fix on their end

PR #8781, opened 2025-08-12
Create Custom Client Background

eatmybrick:
Draw a custom image to the left of the RuneLite sidebar (the column with the minimap & inventory). Supports Stretch / Fit / Fill / Tile, opacity, and fine X/Y offsets (negative allowed). Image is loaded from an absolute file path; if it can’t be read, a solid, user defined fallback color is used.

Intended for use alongside Fixed Resizable Hybrid plugin.


riktenx:
Closing as above comments have not been addressed. You may resubmit in the future if you do so.

PR #8785, opened 2025-08-13
Create petreminder

HerbRuns:
A small plugin that simply reminds the player that they have a follower/pet out. Adds a small UI element when pet is out, and hides when it is not.


HerbRuns:
> Have you looked at the plugin [Missing Pet Notifier](https://github.com/tylerwgrass/missing-pet-notifier)?
> 
> It seems to be similar to what you are trying to do.

Ah nice find! Looks very similar to what I was going for, thanks for highlighting.

The experience of writing and submitting a first plugin was worth it either way :D appreciate you!

PR #8787, opened 2025-08-13
Create scorching-bow-freeze-timer

adnapPanda:
Created a plugin that shows the time until the scorching bow freeze ends. Useful at bosses like Kril where you want to keep it immobilized at all times without wasting special attack energy


adnapPanda:
> Pretty sure this is against the rules:
> 
> https://github.com/runelite/runelite/wiki/Rejected-or-Rolled-Back-Features
> 
> > Opponent freeze timers: [see Jagex statement](https://secure.runescape.com/m=news/third-party-client-guidelines?oldschool=1) Indicates how long an opponent is frozen for

Ah ok. I thought this was for PVP only

PR #8789, opened 2025-08-13
Doom tracker

zichsw:
Tracks cumulative floors and loot chance at the Doom of Mokhaiotl and displays the information in a small overlay. Similar to Pyramid Plunder Counter. 

Data persists through sessions. You can toggle to see rates for individual uniques.


LlemonDuck:
Remove printStackTrace, use a logger

But also since you're storing not much data, you might want to consider just storing it in config instead of a hard file. Then it would also be cloud-synced for players with that enabled

coopermor:
You can store a string in the config and use an injected gson to parse it. Here is an example of doing it in my plugin https://github.com/coopermor/prayer-markers/blob/69c3733d332d884cddc27d9b7e95fd4735b3cc53/src/main/java/com/coopermor/prayermarkers/PrayerMarkersPlugin.java#L141C1-L141C2

zichsw:
Awesome, I'll take a look and make those changes. Thanks for the guidance.

riktenx:
Closing because the above comments have not been addressed. You may resubmit in the future if you manage to do so.

PR #8791, opened 2025-08-13
Create DoomLootLostTracker

oopsimiron:
Tracks loot lost due to dying to the Doom boss.


aHooder (changes requested):
Before this plugin can be reviewed, you need to clean this up significantly, since the plugin is very unnecessarily convoluted and contains a bunch of harder to review code which you're *completely* not using (e.g. you don't need migration code here, you're creating a new plugin!). You can accept this PR (https://github.com/oopsimiron/DoomLootLost/pull/1) and do any other cleanup you want afterwards, then push a commit to update the commit hash here.

PR #8800, opened 2025-08-14
Update petreminder

HerbRuns:
2nd PR in a day - sorry maintainers :3

Revamped the death message to trigger on respawn if a pet was previously following and is no longer following.

Refactored follower checks to use varplayer values instead as pets can be following but not rendering, causing UI and tracking vars confusion.

### DIFF
--- a/plugins/petreminder
+++ b/plugins/petreminder
@@ -1,2 +1,2 @@
 repository=https://github.com/HerbRuns/petreminder.git
-commit=cca23b1c1b5a64de203a80e0a4a54b11d8fb9992
+commit=6c791bd09995b928dbd5d2d56a508f4d3f9d1312


pajlada:
You can make use of the `VarbitChanged` event to listen for changes to the varp you're intested in instead
https://static.runelite.net/runelite-api/apidocs/net/runelite/api/events/VarbitChanged.html

Instead of hardcoding the constant yourself, you can use `VarPlayerID.FOLLOWER_NPC` from the `net.runelite.api.gameval` package

The `VarPlayerID.FOLLOWER_NPC` varplayerid can actually include NPCs that follow you too, e.g. during The Curse of Arrav ou might have Elias following you. I'm not 100% sure if there's an easy way to differentiate that, but it might be worth testing.

Are these things you want to change/test first?

HerbRuns:
> You can make use of the `VarbitChanged` event to listen for changes to the varp you're intested in instead https://static.runelite.net/runelite-api/apidocs/net/runelite/api/events/VarbitChanged.html
> 
> Instead of hardcoding the constant yourself, you can use `VarPlayerID.FOLLOWER_NPC` from the `net.runelite.api.gameval` package
> 
> The `VarPlayerID.FOLLOWER_NPC` varplayerid can actually include NPCs that follow you too, e.g. during The Curse of Arrav ou might have Elias following you. I'm not 100% sure if there's an easy way to differentiate that, but it might be worth testing.
> 
> Are these things you want to change/test first?

Thank you for this! I'm happy to merge as if that's cool? I've rewrote the current logic to account for a bug i missed in the previous one. When I get some more time this week I'll try revamp into the onchange event too! Thanks again

pajlada (approved):

PR #8797, opened 2025-08-14
update custom-emoji

LouisGameDev:
> fixed issue: Other users typing !emojifolder in clan chat opens my emoji folder https://github.com/LouisGameDev/runelite-custom-emoji/issues/1
> merged PR by [cman85](https://github.com/cman85) https://github.com/LouisGameDev/runelite-custom-emoji/pull/2

my github profile renamed from TheLouisHong -> LouisGameDev

PR #8802, opened 2025-08-14
update custom-emoji

LouisGameDev:
> fixed issue: Other users typing !emojifolder in clan chat opens my emoji folder https://github.com/LouisGameDev/runelite-custom-emoji/issues/1
> merged PR by [cman85](https://github.com/cman85) https://github.com/LouisGameDev/runelite-custom-emoji/pull/2

### DIFF
--- a/plugins/custom-emoji
+++ b/plugins/custom-emoji
@@ -1,3 +1,3 @@
 repository=https://github.com/LouisGameDev/runelite-custom-emoji.git
-commit=8be2460a77bf4480db16c2c4dc1f7f8fc0b7c50f
+commit=89533682faeea5e0202cb823d7b028f2a2557462
 authors=LouisHong


LouisGameDev:
follow up on PR from https://github.com/runelite/plugin-hub/pull/8797

pajlada:
If your command is client side, you should use the https://static.runelite.net/runelite-api/apidocs/net/runelite/api/events/CommandExecuted.html event instead. Users would type `::emojifolder` instead of `!emojifolder`.

`::`-prefixed commands are meant for client-only commands like yours, whereas `!`-prefixed commands are meant for commands that are should be sent to all players around them, and then resolved by each client (like `!pets`)

For now, I'll merge this in, but I would appreciate it if you would follow up with a PR that migrates to use a `::` command instead.

PR #8803, opened 2025-08-14
Add Lending Tracker plugin

Guess34:
Initial submission; tracks clan lending/borrowing.

### DIFF
--- a/plugins/lendingtracker
+++ b/plugins/lendingtracker
@@ -1,2 +1,2 @@
 repository=https://github.com/Guess34/lendingtracker.git
-commit=c954161b53c66a1d47b04b5d3862cdf558f715df
+commit=d9f2e8e11ea42e0dfb1f844a49f89a3a92bfee63


cdfisher:
Your plugin repo needs to be public.

riktenx:
Comment here when you make your repo public (or push another commit) to remove the waiting-for-author tag.

Guess34:
its public now

Nightfirecat:
What reason is there to add this plugin which doesn't do anything when you could instead get it to a working state first before submitting it?

Guess34:
> What reason is there to add this plugin which doesn't do anything when you could instead get it to a working state first before submitting it?

honest answer is I'm still in the finishing stages of the code and was trying to get the plugin hosted in hub to prepare for when its finished as this is my first time submitting a plugin so if i had any issues i can have it worked out and just update when I'm finished. i know most people would think that's backwards thinking but that's how my brain works tbf

PR #8811, opened 2025-08-15
Delve calculator

Speaax:
Plugin that calculated how many expected items a player has based on floor completion. Check scoreboard to sync to player stats, and killing boss increments the tracked floor completions and expected drops.


aHooder:
You just changed your review time from probably weeks to probably days with these changes :tada:

Speaax:
That's good to hear!

Speaax:
Changed the logging to log.debug.

aHooder (approved):
The warnings and errors could've been kept, but at least this doesn't spam the log now :slightly_smiling_face:

PR #8824, opened 2025-08-16
feat: add spec-sound-replacer

VerchaSEL:


riktenx:
Your repository is only a license file?

riktenx:
Additionally-- there's already the generic sound swapper plugin on the hub - https://runelite.net/plugin-hub/show/sound-swapper - which should already cover this use case.

VerchaSEL:
I know Sound Swapper is on the Hub, but this plugin is trying to solve a narrower problem with a different approach. Sound Swapper replaces specific sound IDs from the wiki list; you add IDs and provide matching WAVs. Mine doesn’t touch sound IDs at all - it just plays one custom WAV whenever spec energy drops, so it works for every weapon with zero per-sound setup.

It’s also a simpler UX: in-client file picker, optional bundled WAV, volume slider, and a test button - no folders or file naming.

Is there a chance to work towards an addition of this?

riktenx:
> It’s also a simpler UX: in-client file picker, optional bundled WAV, volume slider, and a test button - no folders or file naming.

It sounds like these are just UX improvements that could be made to the existing plugin.

VerchaSEL:
I get the concern about the duplicates. But the goal here isn't another general sound replacer, but single-purpose plugin with different trigger and zero setup.

Purpose is to play one WAV on any special attack energy drop, no sound id's, no per weapon mapping

UX is user friendly, file picker, optional already bundled sound, nothing much to configure for the end user.

Audience, are the players who only want a "sound on spec" without learning or searching for the sound ID's. Install, turn it on, and it's already there.

In my sense this would serve different audience, and it's much easier to discover.

tylerwgrass:
Have you looked at the spec jingle plugin? https://github.com/Servilabs/spec-jingle

VerchaSEL:
Thanks for the url, I took a look at it. Form the README i can deduce that it plays a jingle when spec energy regenerates to configurable threshold. My plugin is the opposite trigger. It plays one user chosen WAV when spec energy drops. My plugin is intentionally single purpose for players who want a custom sound on spec spend (it’s a common meme in PvP clips, e.g., Odablock). Therefore I think this would really find the right audience.

riktenx:
I think your argument is reasonable. But I also think that we see submissions like this _because_ the sound swapper UX is so bad, if that was improved through mechanisms such as what you seem to have added here, this wouldn't be necessary.

WRT this submission, it's ultimately going to be a question of what the Runelite maintainers want to do.

VerchaSEL:
Thanks for candid take. My standpoint is that even if the general tools get better, which would be amazing, there's always real value in focused plugins that are easy to find, setup and does one job well. If i search for "Spec sound" today I basically get nothing. That's exactly the gap we run into in our clan - we wanted simple, reliable way to play one custom WAV on spec spend since there is a hype for it after all the short videos from content creators. 

You don't really want to lease the entire office building when you only need a single desk. In the same vein, players who want ta spec spend sound shouldn't have to adopt a full sound-swapping workflow.

If there is a concern in user confusion I'm happy to adjust the naming and the description to a crystal clear ones. Example: (Special Attack Sound - plays a custom sound "WAV" on special attack energy spend).

VerchaSEL:
To avoid any copyright claims, it's just a mic recording.

pajlada:
You could get rid of some config complexity if you use a proper config class and use getter/setter functions - see https://github.com/117HD/RLHD/blob/5c8db4a8d46483668edc5d76e61fb4665d7e1696/src/main/java/rs117/hd/HdPluginConfig.java#L1076-L1081 as an example

Not a blocking change, so let me know if this is something you want to explore

riktenx:
It seems like there's no blocking requests from reviewers so I'm removing the "waiting" label.

PR #8826, opened 2025-08-17
TobMistakeTracker Added P3 Melee tracking

NickSmeding:
First time contributing and get a hold on the officiele writer of this plugin.


Felanbird:
It's mad about the dimensions, not the file size. Java will blow up an image to it's max size when it loads it into memory. (width * height * 4)

NickSmeding:
Plugin was last updated over a year ago so needed to solve some small things like deprecated or wrong image sizes, should be good now

PR #8839, opened 2025-08-19
add new plugin runepulse v1.0.0

hawolt:
RunePulse will mainly provide skilling related logs in a similar fashion to warcraftlogs which can be used to calculate a score for effieciency based on missed ticks

it additionally feature realtime data display currently including 
- experience gain
- login status

the goal of this plugin is to take skilling records to new heights and maybe even more than just skilling in the future

this version is very basic and only tracks experience gains as is with no additional metadata of location or action

### DIFF
--- a/plugins/runepulse
+++ b/plugins/runepulse
@@ -1,4 +1,4 @@
 repository=https://github.com/RunePulse/runepulse-runelite-plugin.git
-commit=a36d50cde0e5bcd48f4b4b0169996b3ceb0e5d64
+commit=c691fea8fc48124d5c5a6c0ab01088769d6b0f13
 authors=hawolt
 warning=This plugin submits your username and game related information in addition to your IP address to a server not controlled or verified by the RuneLite developers.


tylerwgrass:
What exactly are you doing with the data you're sending right now? It's not exactly clear what I'm getting if I install this plugin as-is.

And how is this different from [Wise Old Man](https://wiseoldman.net/)?

hawolt:
sorry I have not added anything to the README yet, the data is stored in a clickhouse database with 3 tables

in the first table
I store the username and link it with an ID that is persistent

the second table is sessions, whenever someone logs in a new session is created and stored with the user_id for the person, the timestamp and the gametick value of the client

the third table stores every single xp drop
associative the session_id, skill_id, xp, tick, timestamp

data in the sessions and experience table is automatically deleted after 36 hours but the frontend (once complete) will allow to make the logs of sessions persistent with a simple click


as a world of warcraft player and the influx of world of warcraft players I thought something actually competitive for skilling (warcraftlogs is logs about ability usage etc. during raids and at the end you get a parse score 0-100) would be something that is wanted.

the main domain is https://runepulse.net/ currently there is only data associated with my own account (hawolt) and you can see previous sessions of me playing by clicking the session values that pop up when you hit search, this displays exactly when which XP was gained and allows a competitive drive by having tick perfect sessions of skilling activities

players also have profile pages: e.g. https://runepulse.net/user/hawolt
if I am logged in and have the plugin enabled and gain xp it will show up here live as I gain it (basically a live session) similar to warcraftlogs live logging

the difference to wise old man is to have detailed insight on how the xp was gained and allows calculating a score how efficient the skilling session was (e.g. score of 100 if no ticks were missed in a session) this is not something wise old man can do

I have also thought about the fact that data can be tampered with easily as its almost impossible to verify, my approach is going to be to load the highscores once a session is started to know the XP of a player, then when the connection dies check the highscores again to know the ending XP then the log can be verified and if the gained xp adds up it counts as valid.

this would still allow tampering e.g. stacking up xp at blast mine and then forging a perfect 3t mining cycle but the times and xp still have to add up - it would be possible to tamper but its not easy and just like templeosrs has manual video review for high end 6 hour sessions I would do the same if I see that it becomes an issue

another factor of tampering would be creating a session for a username I do not own, this will be able to be avoided in a future version as I would consider this a working alpha - verification will look something like gain y XP in skill z and logout, if the XP was gained within the given timeframe I will assume the person was on the account and is the owner or had access and the system will hand a key which then will drop any data from someone that does not specify that key

PR #8849, opened 2025-08-20
Update cox-qol

MoreBuchus:
Updates Cox Qol to 1.1.7. 

Adds shared instance timer for starting raid. Thanks to Okke


riktenx:
Closing because the above comments have not been addressed. You may resubmit in the future if you manage to do so.

PR #8854, opened 2025-08-20
Create afk-dimmer

JoeFoote:


pajlada:
Hi, your plugin currently fails to build.

aHooder (dismissed):
Your [plugin description](https://github.com/JoeFoote/afk-dimmer/blob/857c3e1aa82356a6883043f111f0b5d964765594/src/main/java/com/afkdimmer/AfkDimmerPlugin.java#L19) is a bit weird, and your [displayed author name doesn't match your github name](https://github.com/JoeFoote/afk-dimmer/blob/857c3e1aa82356a6883043f111f0b5d964765594/runelite-plugin.properties#L2), but neither of those are blockers. Looks alright to me.

Perhaps you should also set a default dimmer strength other than zero, so the plugin seems to work when activated for the first time.

JoeFoote:
Oh shoot! I forgot to update the description while doing my final checks. For my author name I used my RSN. Is that weird? And your suggestion sounds like a good idea, so I'll tweak that as well!

riktenx:
The name in that field can be whatever really.

PR #8855, opened 2025-08-20
Added No-Stalactites

TripStank:
A Simple plugin to show scenery object IDs and hide them in the game world, Stalactites added by default.


raiyni:

```ID based plugins: Plugins that use player provided IDs for the entirety of their functionality can cause moderation issues and outcomes that break Jagex's plugin rules. Due to this fact we will not be accepting any new ID based plugins. Plugins that use a specific set of IDs but do not allow user input will still be accepted. (e.g. plugins like Vardorvis Projectiles which only allows you to change a specific projectile to a set list of projectiles.)```

PR #8856, opened 2025-08-21
No stalactites plugin

TripStank:
# Features
- Hide stalactites and cave columns.
- Simple on/off toggles. No user-provided IDs.

### DIFF
--- a/plugins/no-stalactites
+++ b/plugins/no-stalactites
@@ -1,0 +1,3 @@
+repository=https://github.com/TripStank/NoStalactites.git
+commit=c781a91d0cf6a08eb69dad27334bd1272ea6c21e
+author=Stank
--- a/plugins/no-stalactites
+++ b/plugins/no-stalactites
@@ -1,3 +1,3 @@
 repository=https://github.com/TripStank/NoStalactites.git
-commit=c781a91d0cf6a08eb69dad27334bd1272ea6c21e
+commit=9b4a92defb3d2a1f72549f8272ea17edc4f3c5ed
 author=Stank


aHooder (dismissed):
You could clean your `modelCache` now also on `shutDown`, but it's not a big deal.

In your `onConfigChanged` [here](https://github.com/TripStank/NoStalactites/blob/da12eb3d84aeb9bfa0fe8272f12c8ddecf5cb876/src/main/java/com/nostalactites/NoStalactitesPlugin.java#L170-L174), when the client is not in the `LOGGED_IN` game state, you don't need to apply any hiding, as the client will either be logged out or in the middle of loading a new scene basically. This is also not really a big problem though.

aHooder (changes requested):
This is where we kinda start running into problems with vibe coding unfortunately :sweat_smile: You've accidentally deleted the other lines of code which cleared other stuff here (see the highlighted lines): https://github.com/TripStank/NoStalactites/compare/da12eb3d84aeb9bfa0fe8272f12c8ddecf5cb876...70e8928d469eb72c37817abc54ed73954b89d732#diff-5bdfd823280d2eed0282f95fce1d1ba565fa4a82a1b3fdf033c3161026dc3630L122-L124

Can you, as a human, ensure that your plugin cleans up after itself, both by thinking through the code and by testing in-game? :slightly_smiling_face: Ideally `columnAnchors`, `stalagmiteObjects` and `modelCache` should be cleared on `shutDown` (remember `stalagmiteObjects` need to be despawned before clearing also!), *and* the set of `columnAnchors` shouldn't be allowed to grow indefinitely. Either you can put a cap on it (say max 200), or you can clear it every time a new scene is loaded, which can be detected by subscribing to `onGameState` change events (when it changes from `LOADING` to `LOGGED_IN` should be right I think). Atm you're also deleting and respawning *all* replacement objects every time a single new stalagmite spawns, though that's not a big problem if you want to keep it that way for simplicity.

TripStank:
is something missing? I feel like i remember seeing "1 approval, 1 awaiting" before I did the ai adjustments, but now I'm just seeing "1 approval". Sorry for being a noob >.<

PR #8875, opened 2025-08-23
translate game messages with local translation files

YS-jack:
game messages were only translated with deepl, but now it can also be done with local translation files

also some minor fixes to prevent characters from being replaced with their non-accent variants,
added option to output log of untranslated texts to local file


YS-jack:
made a very small change (1 "false" to "true") wont be making anymore changes till its merged

PR #8876, opened 2025-08-23
Add System Clock plugin

soda3x:
My plugin (https://github.com/soda3x/runelite-system-clock) adds a system clock to the game. You can place the clock anywhere and it supports many time formats.


aHooder (dismissed):
Not a blocker, but the position config should probably be removed, as it only controls the default position. If the overlay is moved by dragging while holding alt, the config will no longer match the overlay's actual position, which is a bit confusing.

aHooder (approved):

soda3x:
Just waiting on one more approval :)

PR #8882, opened 2025-08-23
Takeover: mistake tracker to allow further development 

NickSmeding:
until now no response next to that plugin wasn't touched for a year. Hopefully someone in plugin hub can get hold on him because I don't like to take over the plugin without getting approval from previous owner. Preferable I will just be added as contributor


pajlada:
@QuestingPet Hi! NickSmeding here is looking to take over development of your `tob-mistake-tracker` plugin. Do you have any objections to that?
Thanks!

pajlada:
@QuestingPet No worries at all, this is why we reach out to double check!
Thanks for the clarification

NickSmeding:
@QuestingPet messaged me 🔥

PR #8886, opened 2025-08-23
GIM Tracker Sync

JoshuaALawrence:
Hello, this plugin checks your items in your inventory/bank/collection logs/drops and will add/remove that item from your party on my self-hosted (self-made) 'Group Ironman Tracker'. This helps my team keep track of who already owns which items. I've attached images of the website.

The settings require the user to add the host, api key and check which boxes they want to sync; nothing goes to a server unless they directly add the location.

481329059-aae84ba3-ab34-432c-a686-ca1bfd5a561a
481329053-bc9bb9da-eefd-4618-b082-cbab456f868b
481329036-5febc5ac-78bc-45a5-95e3-ce9a0ed4aae6
481329063-e8ecb6f4-420d-4e5f-ae06-5d1fbd3bd359

### DIFF
--- a/plugins/gim-tracker-sync
+++ b/plugins/gim-tracker-sync
@@ -1,2 +1,2 @@
 repository=https://github.com/JoshuaALawrence/GIM-Tracker-Sync.git
-commit=9506e74af295c3c030eb8a44464845d53dbb5e9c
+commit=2b543a16157415ab084faf113877723c9b50898f


aHooder (changes requested):
You should use the `build.gradle` found in the [example repo](https://github.com/runelite/example-plugin/blob/master/build.gradle) with minimal modification to make review easier. E.g. remove [this](https://github.com/JoshuaALawrence/GIM-Tracker-Sync/blob/2b543a16157415ab084faf113877723c9b50898f/build.gradle#L12-L14) and [this](https://github.com/JoshuaALawrence/GIM-Tracker-Sync/blob/2b543a16157415ab084faf113877723c9b50898f/build.gradle#L46-L53). It's easiest if it's nearly identical.

You should delete [this file](https://github.com/JoshuaALawrence/GIM-Tracker-Sync/blob/2b543a16157415ab084faf113877723c9b50898f/gradle/verification-metadata.xml), just in case it can interfere with RuneLite's build process. Additionally, your `gradle` folder doesn't actually contain a wrapper jar (not that it needs to), and the `wrapper` folder in your project root is non-standard as well (should be deleted).

You should be injecting `OkHttpClient` and using that instead of creating a `HttpClient` [here](https://github.com/JoshuaALawrence/GIM-Tracker-Sync/blob/2b543a16157415ab084faf113877723c9b50898f/src/main/java/com/gimtracker/GimTrackerApiClient.java#L28).

You should change your code to make it as easy as possible to keep track of which URL your plugin is making HTTP calls to at all times. Currently it's a little bit harder to follow than it has to be, since your `hostUrl` is updated via this public `updateConfig` method [here](https://github.com/JoshuaALawrence/GIM-Tracker-Sync/blob/2b543a16157415ab084faf113877723c9b50898f/src/main/java/com/gimtracker/GimTrackerApiClient.java#L34), which *could* be called from anywhere with any input. This isn't *that* hard to track, but anything to ease review is a plus. Perhaps you could move the `updateApiClientConfig` function to the `GimTrackerApiClient` class, and inject the config there and access it directly?

If this list of template items is large, it might not be a good idea to store this in a config [here](https://github.com/JoshuaALawrence/GIM-Tracker-Sync/blob/2b543a16157415ab084faf113877723c9b50898f/src/main/java/com/gimtracker/TemplateCache.java#L215).

Regarding your comment [here](https://github.com/JoshuaALawrence/GIM-Tracker-Sync/blob/2b543a16157415ab084faf113877723c9b50898f/src/main/java/com/gimtracker/GimTrackerApiClient.java#L51), you're free to use `Gson` by injecting an instance of it. If you need to alter the options, use `gson.newBuilder()`. This is already part of RuneLite's dependencies, and would clean up your serialization & deserialization code a bit.

After resolving these issues, you'll have to wait for a maintainer to review the plugin.

PR #8888, opened 2025-08-23
Disable barbarian-assault-pbs

coopermor:


SkylerPIlot:
@coopermor

Felanbird:
When the backend request fails the plugin will output raw HTML, filling the chatbox and confusing users.
image

raiyni:
We disable plugins if breaking features are bad enough. Someone can submit an update with this removed.

PR #8889, opened 2025-08-24
Adds the brimhaven-agility plugin

kagof:
# Brimhaven Agility Arena plugin

Adds a new plugin to aid with the Brimhaven agility arena. Specifically, it currently computes the shortest path the the active ticket dispenser, and displays that path on the screen for the user. This takes into account whether the user has the agility level for a given obstacle, and is computed using the A* pathfinding algorithm.

The weightings of obstacles, line colour, and whether or not to actually draw the line are all configurable:

## Future expansions

These are not yet features in this plugin, but I may decide to expand it to include:

* removing the hint arrow when the ticket has been claimed
* highlighting the correct plank to use of the 3 options

### DIFF
--- a/plugins/brimhaven-agility
+++ b/plugins/brimhaven-agility
@@ -1,0 +1,2 @@
+repository=https://github.com/kagof/rl-plugin-brimhaven-agility.git
+commit=b164efa5195d926fc7b5d63aaf7089b1b30677b6


kagof:
I see the build failure due to a line longer than 120 chars in my build.gradle, I'm updating to fix this and will address @cdfisher's feedback at the same time.

> If you're using significant portions of code from Quest Helper you should be copying their license headers into the files that use that code.

I thought I had already done so, see [WorldLines.java](https://github.com/kagof/rl-plugin-brimhaven-agility/blob/master/src/main/java/com/kagof/runelite/plugins/brimhavenagility/questhelper/WorldLines.java), but it looks like I missed `QuestPerspective.java`. Will add the headers to this file too now.

> If all the plugin currently does is draw that path, that probably shouldn't be a configurable option since the user wouldn't currently be using it to do anything else so they could just turn the plugin off entirely.

This is a fair critique, I will remove this configuration option for the time being.

> Similarly, having the obstacle weights as config items is probably more likely to confuse users than it is to be actually helpful.

I disagree on this front, as for some users it is possible they may want to tune it themselves. For example, users may want to avoid the dart obstacle, as failing it lowers your agility level by 2, in which case giving it a much higher weighting could be useful. Another example is if someone is very AFK and does not want to look to figure out which plank of the 3 to use. I do admit though that the likelihood someone actually cares enough to do this is fairly low.

Let me know if you still feel strongly that this weighting configuration should be removed and I can do so. Another option would be to replace the weighting configuration with a checkbox for `avoid x obstacle`, which, if checked, sets the weighting arbitrarily high so the obstacle will never be used. I'd be happy to replace it with this instead.

cdfisher:
> Another option would be to replace the weighting configuration with a checkbox for avoid x obstacle, which, if checked, sets the weighting arbitrarily high so the obstacle will never be used. I'd be happy to replace it with this instead.

This is probably a lot less confusing to users so I would do that

aHooder:
I hate to be that guy, but this line *could* also cause an infinite loop if the `cameFrom` map contains a mapping from a location to itself: https://github.com/kagof/rl-plugin-brimhaven-agility/blob/1c0c137f469da33484dcfaf28df8a0fdf5f870df/src/main/java/com/kagof/runelite/plugins/brimhavenagility/BrimhavenAgilityPathFinder.java#L133

Solid changes though :+1:

kagof:
Fair enough @aHooder, could also cause an infinite loop if we somehow got a cycle larger than 1. Doesn't hurt to add an extra safety check there; I've done so now.

riktenx (approved):

PR #8890, opened 2025-08-24
Create deathscoffer

NSammut:
A RuneLite plugin that helps you keep track of your Death's Coffer value by displaying it in a convenient side panel and responding to chat commands.

### DIFF
--- a/plugins/deathscoffer
+++ b/plugins/deathscoffer
@@ -1,3 +1,3 @@
 repository=https://github.com/NSammut/Death-s-Coffer.git
-commit=0ad2237ef37404bc45693ad9f8d2c3d77e8ac898
+commit=5cb447065c18a9c070ce00fd06725742dbc99e68
 warning=This plugin submits your username, death's coffer value, and your IP address to a 3rd party website not controlled or verified by the RuneLite Developers. This is needed to track coffer value and update chat messages.


aHooder (changes requested):
Your plugin should inject `OkHttpClient` and use that instead of the `HttpClient` here: https://github.com/NSammut/Death-s-Coffer/blob/0719471d86b82081798aa79655da0f2dcbd59594/src/main/java/com/nicholassammut/DeathsCofferService.java#L21

Also, you will need to add a `warning=` to your ~~`runelite-plugin.properties`~~ (oops sorry!) `plugins/deathscoffer` file due to your plugin connecting to an external service not controlled by RuneLite. This warning must include that their IP is being sent, and should also detail what else you're intending to send. See this message from one of the RuneLite admins: https://discord.com/channels/301497432909414422/419891709883973642/1409410895783657492

aHooder (dismissed):
This looks okay to me now, but you will have to wait for a maintainer to review this.

Some smaller issues remaining:
- This doesn't actually link to anything, so should probably be removed: https://github.com/NSammut/Death-s-Coffer/blob/9e53cd974fc799ad9337a6f668b12214552f7957/src/main/java/com/nicholassammut/DeathsCofferPanel.java#L118
- [This](https://github.com/NSammut/Death-s-Coffer/blob/9e53cd974fc799ad9337a6f668b12214552f7957/src/main/java/com/nicholassammut/DeathsCofferPlugin.java#L153) is a tiny bit slow, so it would perhaps be good to either do the varp checks before the NPC check, or gate the whole thing behind a check to see whether you're in Death's Domain (which is in region ID 12633).
- The link in your `README.md` here is just an example link: https://github.com/NSammut/Death-s-Coffer/blob/9e53cd974fc799ad9337a6f668b12214552f7957/README.md?plain=1#L50

NSammut:
> This looks okay to me now, but you will have to wait for a maintainer to review this.
> 
> Some smaller issues remaining:
> 
> * This doesn't actually link to anything, so should probably be removed: https://github.com/NSammut/Death-s-Coffer/blob/9e53cd974fc799ad9337a6f668b12214552f7957/src/main/java/com/nicholassammut/DeathsCofferPanel.java#L118
> * [This](https://github.com/NSammut/Death-s-Coffer/blob/9e53cd974fc799ad9337a6f668b12214552f7957/src/main/java/com/nicholassammut/DeathsCofferPlugin.java#L153) is a tiny bit slow, so it would perhaps be good to either do the varp checks before the NPC check, or gate the whole thing behind a check to see whether you're in Death's Domain (which is in region ID 12633).
> * The link in your `README.md` here is just an example link: https://github.com/NSammut/Death-s-Coffer/blob/9e53cd974fc799ad9337a6f668b12214552f7957/README.md?plain=1#L50

Thank you! I tried .getRegionId() to do something similar previously and I kept getting random IDs.. so the NPC check was my next best thought. I'm assuming this is due to it being instanced, I just tried using getMapRegions() and that seems to line up with the ID you mentioned. I'll work on getting it working that way instead.

pajlada:
`ChatMessageManager` is unused in `DeathsCofferPlugin.java`

The `icon.png` file is currently used for the sidepanel - it should be square.
If you want a plugin icon, you need to put an icon called `icon.png` in the root of your repository.

Remove the startup/shutdown log.info calls

Since you're making web requests as a result of your chat commands, you should register them as sync (so `registerCommandAsync`).
Not required, but since you're registering aliases, I would prefer if you call the register & unregister functions 3 times manually instead of looping through the commands. This will make it easier to know what plugin uses what command in the future.

We have gamevals that describe what things do, see `InterfaceID.java` where you could use `DEATH_OFFICE`/`DEATH_COFFER`/`DEATH_COFFER_SIDE` instead

In your `onGameTick`, you can run `client.getVarpValue` just once and use its value

Since you don't have any configs, remove the config file

Instead of using `Desktop.getDesktop().browse`, use runelite's `LinkBrowser.browse`

In `DeathsCofferService` you don't need to store `rsn` as a class member

PR #8891, opened 2025-08-24
add runelogs

hawolt:
this plugin simply writes a tailored log file for a sister project which is still work in progress

its the refactored version of the last PR that I closed due to using a socket and binary logic which I was told will limit the reviewer pool but also overcomplicate everything for everyone

this more minimalistic approach simply just writes a log which can be handled by an external application outside of the RuneLite eco system which makes life easier for both parties, the maintainers and me, easier

additionally this setup should be future proof for the official jagex client as all the logic is outsourced of the plugin
hopefully this is more acceptable than my last iteration

### DIFF
--- a/plugins/runelogs
+++ b/plugins/runelogs
@@ -1,3 +1,3 @@
 repository=https://github.com/RuneLogs/runelogs-plugin.git
-commit=b2c672421d9ca54bfd4a4a077ceaa6b43cbbc921
+commit=0f40ead297cdb519f2fc73c8528dce839d5db684
 authors=hawolt


cdfisher:
Have you considered contributing to the existing "Action Logs" plugin?

hawolt:
I have talked with pajlada earlier, I was told it should be fine to create a new logger given the specific amount of data I need
additionally his project logs json which would add parsing overhead on my end

hawolt:
> Kind of minor thing, but you need to remove or change these `log.info`'s to prevent spamming the client log: e.g. https://github.com/RuneLogs/runelogs-plugin/blob/b2c672421d9ca54bfd4a4a077ceaa6b43cbbc921/src/main/java/net/runelogs/tracker/method/hitsplat/impl/HitsplatMethod.java#L28

all gone, forgot to remove those

PR #8896, opened 2025-08-24
Update barbarian-assault-pbs

SkylerPIlot:
Fixes issues brought up by https://github.com/runelite/plugin-hub/pull/8888#issuecomment-3218186299

### DIFF
--- a/plugins/barbarian-assault-pbs
+++ b/plugins/barbarian-assault-pbs
@@ -1,2 +1,2 @@
 repository=https://github.com/SkylerPIlot/Barbarians-Assault-Personal-Bests.git
-commit=f8a332e89d09a27e7239a5515c9672f36c676e7a
+commit=be70885563c25539201a3d54c7f995a0e79135b1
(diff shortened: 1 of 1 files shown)


iProdigy:
```
barbarian-assault-pbs: plugin uses terminally deprecated APIs:
ChatMessageManager.update is a no-op
```

SkylerPIlot:
To be honest, I'll get this fixed but it is annoying to have it disabled -> I fix the bug causing it to disabled which bothered no actual users of the plugin -> having to update it causing further downtime to said users @iProdigy. Who reported said bug/caused any investigation? As the community that uses it doesn't touch that feature + has no complaints

coopermor:
Multiple users in the runelite discord were confused why their chat was full of random HTML. So the community did have complaints.

SkylerPIlot:
> Multiple users in the runelite discord were confused why their chat was full of random HTML. So the community did have complaints.

I have no access to said discord @coopermor , no bug reports were brought to the github the community that mainly uses this plugin is CBA or anyone in the barbarian assault speedrunning playing it for fun past level 5.

Felanbird:
http://runelite.net/discord

PR #8901, opened 2025-08-25
Goal Tracker v2

AhDoozy:
This update fully reimagines the Goal Tracker plugin, originally created by dillydill123, with a refreshed interface, improved task management, and expanded detection features. Players can now create and share goals more easily with export/import support, use curated presets with automatic prerequisite expansion, and benefit from reliable auto-detection for quests and items.


DjilanoS:
@AhDoozy Thanks for the work, love to see the takeover policy in action. Original collaborators don't reply on anything anymore so it renders the original plugin useless as it's broken...

PR #8905, opened 2025-08-26
ironman-ground-items Initial Review

justin-mercer:
A RuneLite plugin that removes items you don't own from right-click menus. Perfect for Ironman accounts to avoid clicking on other players' drops that can't be taken.

### DIFF
--- a/plugins/ironman-ground-items
+++ b/plugins/ironman-ground-items
@@ -1,2 +1,2 @@
 repository=https://github.com/justin-mercer/ironman-ground-items-plugin.git
-commit=ea111ede736b8fc75e3475d2096118012ce38bfa
+commit=208f46d2b2f52d8f2e727c2d2ce514becd2a62db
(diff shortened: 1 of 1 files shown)


riktenx:
We have an existing plugin on the hub called loot-filters that lets you do all sorts of things with ground item display / menu entries. Rather than adding a new plugin for just this functionality it would be better if it were added in some form to the existing one. Please follow https://runelite.net/plugin-hub/show/loot-filters to the plugin's Discord and start a conversation with its maintainers about adding this feature.

justin-mercer:
The only problem I see is they are a filter-first plugin, where the main focus is filtering out drops based on conditions. Mine just prevents you from picking up drops that don't belong to you (Ironman), especially in conditions where you have lots of stacking items and are just trying to find your own (arrows at crabs where everyone's arrows are in the same drop location). Someone who simply doesn't want to have the option to pick up drops they don't own is a bit different then having a full featured plugin to filter out drops.

riktenx:
We discussed in the loot-filters Discord and it sounds like they mostly wanted left-click take removed for ownership-filtered items, which the plugin _can_ do, so it sounds like the submitter's use case is at least solved by existing plugins.

riktenx:
It looks like the author is just using loot-filters, closing.

PR #8909, opened 2025-08-26
Add Block Teleport While Alching plugin

bielie993-ui:
# Alch Teleport Block
A simple plugin which blocks teleports in your regular spellbook to prevent accidental teleports while alching. The block can be overruled by pressing a modifier key. If the user wants it to only be activated if an alch was cast x seconds ago thats also possible, to make it more dynamic.
- - -
### Features
- Choose from all regular spellbook teleports (default: Kourend Castle, Falador).
- Enable a "block window" which only adds the modifier for x seconds after alching. 
  - Detects high and low alchs.
  - Change block duration (default is 30s).
- Change the modifier key (default is CTRL).
- Right click casting is still instant (can be disabled in config).

image

---

image

### DIFF
--- a/plugins/alch-tp-blocker
+++ b/plugins/alch-tp-blocker
@@ -1,2 +1,2 @@
 repository=https://github.com/bielie993-ui/runelite-alch-tp-blocker.git
-commit=e7c8b28335595f44af22fb8010b0f31918eb8415
+commit=e72a3d609000872961a1d6491dfed403692f973a
--- a/plugins/alch-tp-blocker
+++ b/plugins/alch-tp-blocker
@@ -1,0 +1,2 @@
+repository=https://github.com/bielie993-ui/runelite-alch-tp-blocker.git
+commit=f3b9bd10431ca2b643baf54fe4be06d778ef4c47
--- a/plugins/alch-tp-blocker
+++ b/plugins/alch-tp-blocker
@@ -1,2 +1,2 @@
 repository=https://github.com/bielie993-ui/runelite-alch-tp-blocker.git
-commit=e72a3d609000872961a1d6491dfed403692f973a
+commit=47369287731f3f982bd1a3533021fca788e41733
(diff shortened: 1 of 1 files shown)


aHooder (dismissed):
Neither of these are blockers, but:
- Right-click teleporting to Seers' ignores the delay since the option doesn't include the string "cast", which kinda goes against the "allow right-click casting" option:
  image
- This may interact a little bit strangely with shift-click remapping on teleport spells.

For other reviewers, `runelite-plugin.properties` is parsed with https://docs.oracle.com/javase/8/docs/api/java/util/Properties.html#load-java.io.Reader- so the empty line is okay.

aHooder (changes requested):
You must reduce your icon file size to below 256 KiB.

bielie993-ui:
> Neither of these are blockers, but:
> 
> * Right-click teleporting to Seers' ignores the delay since the option doesn't include the string "cast", which kinda goes against the "allow right-click casting" option:
>   image
> * This may interact a little bit strangely with shift-click remapping on teleport spells.
> 
> For other reviewers, `runelite-plugin.properties` is parsed with https://docs.oracle.com/javase/8/docs/api/java/util/Properties.html#load-java.io.Reader- so the empty line is okay.

Good catch, ill change that, i also need to add the Grand Exchange i noticed.

bielie993-ui:
Added the missing teleports and improved the detection of them. Reduced icon size. Should be all good.

@aHooder Not sure what the failed pipeline means? Any tips?

bielie993-ui:
> This looks good to me now, but you'll have to wait for a maintainer to review your plugin before it can be merged.

Sorry for the hassle, but had to remove the repo and recreate it. I pushed with the wrong user and didnt want those details to be visible.

aHooder (approved):
LGTM, but you'll have to wait for a maintainer to review.

bielie993-ui:
Going to work on the same idea, but the plugin will have a different name.

aHooder:
You should've perhaps mentioned that you wanted to expand the idea to be a more general purpose accidental teleport blocker, rather than one centered around alching :eyes:

PR #8912, opened 2025-08-27
Create hopper

ChunkyAtlas:
Adds an overlay that tracks your remaining world hops (~390-400) and their regeneration (~93 per hour).
-  Shows hops left, missing to full, and time until full.
- Color-coded bar and optional icon.
- Configurable warnings and desktop notifications when low.
- Persists across sessions and can reset manually or on login.


cubeee:
Maybe integrate into https://runelite.net/plugin-hub/show/worldhop-counter?

PR #8920, opened 2025-08-27
Update Tick Beats

HarperDevLab:
Implemented 2 requested features:

-Volume control for tick sounds
Implemented using javax.sound.sampled (same approach as music playback). The Gain parameter in RuneLite’s AudioPlayer did not affect volume, so custom handling was required. Audio clips are properly disposed of to avoid memory leaks.

Playlist support (with Shuffle)
Users can define a playlist, and songs in the playlist can be shuffled during playback.

I also did a bit of refactoring and code comment adjustments.


riktenx:
@aHooder this author was previously given express permission to use the Java sound system directly.

aHooder:
Ah, my bad then!

PR #8926, opened 2025-08-28
Update petreminder

HerbRuns:
Updated petreminder:
- icon now adapts and renders the icon of the correct pet
- icon flashes (with customisation options) when pet is following but not rendering on screen/get's stuck
- death message now true by default


Alexsuperfly:
please dont spam the chat every tick
https://github.com/HerbRuns/petreminder/blob/e01c4abda02010d38e1870653203b3d81495042a/src/main/java/com/herbruns/petreminder/PetReminderPlugin.java#L98-L102

HerbRuns:
Whoops!! This was meant to have been removed - sorry. Thanks for spotting this. Will close and re-open a new PR

PR #8932, opened 2025-08-28
Gemstallhelper plugin

Pepijn-V:


riktenx:
You don't have to close and create new PRs, any commits you push will automatically get added to your current one.

plugins/gemstallhelper
```diff
@@ -0,0 +1,3 @@
+repository=https://github.com/Pepijn-V/Ardy-Gem-Stall.git
+commit=e640352a2d61a3b82e23c80d7860594876a70ab4
+authors=PepijnV
```
aHooder, line 3: This needs to match your actual GitHub handle (i.e. `Pepijn-V`, though it may be lowercase if you want).

aHooder (changes requested):
This file can be deleted (only the one in the project root is required): https://github.com/Pepijn-V/Ardy-Gem-Stall/blob/e640352a2d61a3b82e23c80d7860594876a70ab4/src/runelite-plugin.properties

Otherwise commit `e640352a2d61a3b82e23c80d7860594876a70ab4` looks good to me.

aHooder (approved):

riktenx (changes requested):
* Change `log.info` to `log.debug`
* Your README needs to have actual content and not just be the template placeholder stuff

Pepijn-V:
> Where are you seeing that the plugin has been added?

Capture

Pepijn-V:
Okay my bad haha

PR #8938, opened 2025-08-29
RuneProfile v2.2.1

ReinhardtR:
- Fix conflicts with other chat related plugins
- Improved AutoSyncScheduler logic to avoid various issues.


aHooder:
Not really a blocker, but you should synchronize [this `ImageIO.write` call](https://github.com/ReinhardtR/runeprofile-plugin/blob/c3cb9d72095c0c99e7bd57cc6c5be1bb3bb94438/src/main/java/com/runeprofile/utils/DevTools.java#L78) like so:
```java
synchronized (ImageIO.class) {
    ImageIO.write(sprite, "png", baos);
}
```

PR #8941, opened 2025-08-29
Slayer Task Sorter

MJHylkema:
Add plugin to sort Slayer task lists by weight. Allows you to also reverse the order and revert back to alphabetical.

image

image

### DIFF
--- a/plugins/slayer-task-sorter
+++ b/plugins/slayer-task-sorter
@@ -1,3 +1,3 @@
 repository=https://github.com/MJHylkema/slayer-task-sorter.git
-commit=21a338a48441c48a856d84651508597fc3dd5174
+commit=485110c20203671a4efaa5b1611b25e628bf32fb
 authors=mjhylkema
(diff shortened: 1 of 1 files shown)


aHooder (changes requested):
Your plugin name should be something other than example: https://github.com/MJHylkema/slayer-task-sorter/blob/485110c20203671a4efaa5b1611b25e628bf32fb/src/main/java/com/mjhylkema/SlayerTaskSorter/SlayerTaskSorterPlugin.java#L25

You should sort the list again whenever the sorting mode config is changed. You can do this by subscribing to `onConfigChanged`.

Otherwise, LGTM :+1:

MJHylkema:
> Your plugin name should be something other than example: https://github.com/MJHylkema/slayer-task-sorter/blob/485110c20203671a4efaa5b1611b25e628bf32fb/src/main/java/com/mjhylkema/SlayerTaskSorter/SlayerTaskSorterPlugin.java#L25
> 
> You should sort the list again whenever the sorting mode config is changed. You can do this by subscribing to `onConfigChanged`.
> 
> Otherwise, LGTM :+1: 

Thanks for the prompt review. Not sure how I missed the name, have updated now.

Good shout out about updating the interface on config change. I will implement that in the morning.

Cheers.

aHooder:
I'll go ahead and add the waiting-for-author label until you can get that resolved, since the plugin feels just slightly broken without it :slightly_smiling_face:

geheur (approved):

aHooder (approved):
I'll approve this, but you should submit another PR to reset the sort order when the plugin shuts down also.

PR #8942, opened 2025-08-29
Kitten Tracker 2

jonasmorel:
A simple and convenient plugin that helps you keep track of your pet kitten's needs, ensuring it grows into a healthy cat without running away.


riktenx:
I'm sure you're aware that https://runelite.net/plugin-hub/show/kitten-tracker is already a plugin on the hub with ~180k installs. We don't allow dupes / forks / "v2" submissions of this kind. Either contribute enhancements you want to see to that plugin, or if the author is unresponsive, follow the takeover policy: https://github.com/runelite/runelite/wiki/Plugin-takeover-policy

PR #8944, opened 2025-08-29
Update petreminder

HerbRuns:
Update to add:
- icon now shows the following pet
- some customisation for icon settings
- icon flashes when pet is stuck/off-screen

Removed previous onTick chat spam. Caught by review (thanks!)


riktenx:
Nonblocking but you can just subscribe to `onVarbitChanged` and update your follower info there instead of having to re-check every game tick in the first place.

PR #8947, opened 2025-08-29
Update location-display

trinhc2:
Added configurations for underline, italics, prefix/suffix, sound effects and fixed bug with unknown area names displaying.


aHooder (changes requested):
Can you accept this PR and include it to ease the review process? https://github.com/trinhc2/Location-Display/pull/13

PR #8959, opened 2025-08-31
Crabs for dummies

LuxOG:
Create plugin that swaps left click of crabs in cox to smash if you have a DWH or maul equipped


aHooder (changes requested):
You should remove your config, since the plugin doesn't need one, and your config group is very generic: https://github.com/LuxOG/crabs-for-dummies/blob/9df7675a5e54a1ae8f957585e0572805d558c111/src/main/java/com/crabsfordummies/CrabsForDummiesConfig.java#L6

aHooder (approved):

PR #8968, opened 2025-08-31
Add Elemental Weakness

jonathanvanduppen:
The first time a monster is spawned it looks up the elemental weakness. This is then cached in memory until the client is shut down utilizing https://vanduppenllc.com/runelite/api/elemental-weakness 
IE
https://vanduppenllc.com/runelite/api/elemental-weakness/imp 

image


riktenx:
We already have a plugin that does this and more: https://runelite.net/plugin-hub/show/monsterstats.

If that doesn't do exactly what you want, you'll need to make contributions to that plugin instead.

If its maintainers are unresponsive, you can follow the takeover policy: https://github.com/runelite/runelite/wiki/Plugin-takeover-policy

PR #8969, opened 2025-08-31
add in catracker plugin info

ehubbartt:
This plugin is an all-in-one tracker for Combat Achievements. This plugin allows you to view all the current CA's in the game and syncs with your account progress automatically when you log in. It also allows you to choose certain CA's to track in order to manage progress on your goals. Another great feature is a searchable boss tab that quickly shows you the list of all bosses that have CA's and gives easy access to the CA's for those bosses. This all syncs with your client, so you don't have to redo all your tracking on each load. You can also hover on a task for more info and right-click to pull up the wiki page for that task.
The readme at https://github.com/ehubbartt/combat-achievements-tracker provides some more detailed info and images!

### DIFF
--- a/plugins/combat-achievements-tracker
+++ b/plugins/combat-achievements-tracker
@@ -1,2 +1,2 @@
 repository=https://github.com/ehubbartt/combat-achievements-tracker.git
-commit=7f79264b84c1bff35f88524ec6cfef5164418e8e
+commit=96ff78772490dcac9f5df54a8d6c0efd807cd573


pajlada:
Howdy! Quick review

You're overusing `log.info` - please downgrade your logging entries to `log.debug` unless very necessary

Why are you sleeping on shutdown?

Use LinkBrowser instead of Desktop

You don't need to handle null-checking if ImageUtil.loadImageResource fails - it will not suddenly fail

aHooder (changes requested):
It looks like you've accidentally deleted your `shutDown` method. It still needs to remove the navigation button from the toolbar.

Not necessarily a blocker, but I assume [this line](https://github.com/ehubbartt/combat-achievements-tracker/blob/2503b4a3c547f64093b2ce67410afa6032d929a8/src/main/java/com/catracker/CombatAchievementsPlugin.java#L122) is supposed to load the CAs only upon login, but it looks like this will also reload them every time a new scene gets loaded (since it transitions from `LOADING` to `LOGGED_IN`). It's quite jarring when the panel resets when you move to a different area.

Also not a blocker, but perhaps these hard-coded point thresholds are possible to load from the client as well? https://github.com/ehubbartt/combat-achievements-tracker/blob/2503b4a3c547f64093b2ce67410afa6032d929a8/src/main/java/com/catracker/util/ChatMessageUtil.java#L103-L181

Are you intentionally tying tracked achievements to the RuneLite config profile [here](https://github.com/ehubbartt/combat-achievements-tracker/blob/2503b4a3c547f64093b2ce67410afa6032d929a8/src/main/java/com/catracker/ui/CombatAchievementsPanel.java#L626-L630)? Judging by your `getRSProfileKey` check, it looks like you meant to pass that in as well to associate it with a specific RS profile.

[This Gson field](https://github.com/ehubbartt/combat-achievements-tracker/blob/2503b4a3c547f64093b2ce67410afa6032d929a8/src/main/java/com/catracker/ui/CombatAchievementsPanel.java#L65-L66) is always null, since you instantiate the panel with `new` instead of via injection.

I think it would make sense to reload achievements also when the in-game CA interface is opened, since it can automatically complete some tasks when new ones are added.

Finally, the use of `<html>` (e.g. [here](https://github.com/ehubbartt/combat-achievements-tracker/blob/2503b4a3c547f64093b2ce67410afa6032d929a8/src/main/java/com/catracker/ui/CombatAchievementPanel.java#L87)) looks non-malicious, but it requires more careful review. If you're able to avoid it, that's ideal :slightly_smiling_face: 

Other than the above, looks good to me :+1:

PR #8977, opened 2025-09-01
Create MidiSync

coilerer:
Synchronizes MIDIs (such as the OSRS OST) to the ticks of the game, with adjustable BPM and sub tick beats/quantization.


cdfisher:
This sounds pretty similar to the existing "Tick Beats Metronome" plugin. We would prefer for you to contribute any features not already covered by that to the existing plugin, rather than submitting an entirely new one to the hub.

coilerer:
Anyone else is free to adapt my code if they think they can modify it to meet these requirements.

PR #8981, opened 2025-09-01
Create loot-ledger

ChunkyAtlas:
Show NPC drop tables in the Music tab with progress bar, hover tooltips, search, and visibility toggles.

Track obtained items per account or per NPC, dimming icons, supporting manual toggles, and updating live as you loot.

Cache drop tables to disk with weekly refresh and bundled Items.json for resolving item names.

### DIFF
--- a/plugins/loot-ledger
+++ b/plugins/loot-ledger
@@ -1,0 +1,2 @@
+repository=https://github.com/ChunkyAtlas/loot-ledger.git
+commit=a502c783fad55d889ceca21197e99eacadbe8a8c


LlemonDuck:
don't awaitTermination on your executors during shutDown, and don't interrupt the current thread. Just call shutdownNow and let it finish asynchronously

LlemonDuck:
A few things but not blocking so I'm going to merge:
* consider using ConfigManager#getRSProfileKey or what [ImageCapture](https://github.com/runelite/runelite/blob/6dcbb05a40a937cfd864cb2efbb70b3a01395803/runelite-client/src/main/java/net/runelite/client/util/ImageCapture.java#L189) does to partition data by world type as well as account, to not conflate leagues/deadman/main data 
* you can use MenuEntry#onClick in your menu entry creator to use a callback style
* it would be _awesome_ if you could move all the data fetching so that it runs once for everyone and pushes the results to github, where players fetch from there. then you can remove the jsoup dependency, shrink the plugin codebase by a lot, and get quicker reviews
* you should also look at using Bucket to query data from the wiki in a structural and programmatic way, instead of parsing the pages themselves

PR #8983, opened 2025-09-02
add wine-of-zamorak-helper-plugin

sepcol:


coopermor:
Are you aware item respawn times vary with the number of players on the server? It appears your plugin does not account for this.

sepcol:
> Are you aware item respawn times vary with the number of players on the server? It appears your plugin does not account for this.

Good catch! I was aware but completely forgot about it.
I'll provide a better implementation.

tylerwgrass:
Please also clear `despawnTimes` and `MONKS_OF_ZAMORAK_NPCS` on shutdown:
https://github.com/sepcol/wine-of-zamorak-helper/blob/9bdbfefb931b61de39bb9edbaae756e0a8ebf4e7/src/main/java/com/wineofzamorakhelper/WineOfZamorakHelperPlugin.java#L45

tylerwgrass (approved):

PR #8999, opened 2025-09-04
Add Bank Organizer plugin

Ideonomy-web:


riktenx:
so there's this awesome core plugin called bank tags

Ideonomy-web:
The purpose isn't a simple bank-tag-esque, it's a full-scale bank clean-up.

raiyni:
Maybe consider contributing to the wasted bank space plugin if that's your goal?

raiyni:
You are describing tag tabs without the tabs? There is also a plug-in that turns wiki item categories into bank tags.

Ideonomy-web:
I understand, I have used and seen bank tank generation before, but it's still not quite the same. I don't know how else to distinguish it from that besides what I have written in the original description. It just functions as a quick organizational tool in such a way that if you were to take any account, any level, and completely collapse their bank (default bank - no plugins), it can help provide a relative sense of organization within minutes with pre-set categories. E.g., it instantly picks out any potions/herbs/secondaries and lets you drag that into a desired tab. A lot of people have, for example, magic/range/melee tab, resource tab, pots tab/farm tab/loot tab. Etc. It uses that as "soft templates" while allowing mix-and-match to preference. The idea originated especially after having seen several of the WOW (& other) streamers going "ah, my bank is such a mess".

PR #9000, opened 2025-09-04
Update Fixed Resizable Hybrid

Lapask:
- Updated to using InterfaceID from deprecated ComponentID
- Fixed skill info widget alignment
- Added configurable warning to tell users when their inventory is transparent


Lapask:
- Added failsafe trigger for plugin initialization, fixing bug affecting world hopping from instances and hopping to PvP arena worlds
- Decoupled transparency warning from the gap borders config

PR #9001, opened 2025-09-05
update regex-bank-search

majiru:
Update to replace client.getStringStack() with the new client.getObjectStack().
Also updated gradle build scripts to what is now in the example-plugin repo.


iProdigy (dismissed):
optional: https://github.com/majiru/regex-bank-search/pull/2

iProdigy (approved):

PR #9002, opened 2025-09-05
Update vanced-barrows

Inzomniakk:
Added checks for minigame so that it doesn't show elsewhere.


iProdigy (approved):
have you tested whether you're getting the expected region id in the instanced area?

friendly reference https://github.com/pajlads/DinkPlugin/blob/075b32f12e223eab8b67116c856dec43701b0209/src/main/java/dinkplugin/util/WorldUtils.java#L65

PR #9005, opened 2025-09-05
Create no-prayer-fight-cave-guide

DylanGroenewoud:
The plugin I've developed is to help during NO PRAYER fight cave runs. Currently I (and probably very few other people who do no prayer fight cave runs, i mean.. who even does this.. but anyway) always have to check the Wiki rotations page on where to run and when to request a log out. This plugin shows the somewhat optimal path to run so checking the Wiki isn't needed anymore. For certain waves it also shows helpful information, like preferred amount of hitpoints for the upcoming wave and some other info.


DylanGroenewoud:
Thanks for your reply.
I guess it wasn’t the smartest move on my part to build the plugin without checking the third-party client guidelines first. I was kind of hoping this could be an exception since it’s specifically for no-prayer Fight Cave runs. These runs are fixed (rotation 2), and the spots you stand on every wave are always the same, so I figured showing that in-game instead of checking the wiki each time would be useful.
If this still isn’t allowed then that’s unfortunate, but at least I had fun making it.

PR #9008, opened 2025-09-05
Add addfriend plugin

petittriley-coder:
add drop down to add friend right click

### DIFF
--- a/plugins/addfriend
+++ b/plugins/addfriend
@@ -1,0 +1,2 @@
+repository=https://github.com/petittriley-coder/runelite-addfriend.git
+commit=0c9c0c4ce3fc27ec44da59da3138496b664deba6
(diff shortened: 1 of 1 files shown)


petittriley-coder:
also i am new to this but i think it should no longer bve blank

aHooder:
The rules seem clear-cut on this. I assume you meant to add an "add friend" option when right-clicking players? Either way, you should at the very least make sure your plugin works on your end before submitting it for review here. Like cdfisher said, your plugin doesn't do anything, and there are unfortunately a lot of issues with your `build.gradle` and with random files which shouldn't be present in the repository. I'll go ahead and close this, sorry! If you implement a plugin which works, feel free to submit another PR after reading through this page carefully: https://github.com/runelite/plugin-hub?tab=readme-ov-file#creating-new-plugins

PR #9015, opened 2025-09-06
Update bank-highlighter

Evamist:
Fix as option to add new highlights is no longer there.


Evamist:
Sending first update, so apologises if sent wrong!

Evamist:
Reposted thought seperate pull request.

PR #9019, opened 2025-09-07
Add monster-lookup plugin

robertjnic:
Adds Monster Lookup to the Plugin Hub. 

Search monsters in RuneLite from the in‑game menu or a simple search panel to see basic stats. Uses OSRS Reboxed DB, cached locally (gzipped) with a ~30‑day refresh

### DIFF
--- a/plugins/monster-lookup
+++ b/plugins/monster-lookup
@@ -1,0 +1,2 @@
+repository=https://github.com/robertjnic/monster-lookup.git
+commit=702c233e61cbeadc6119ee6201cc586a1886c60d


robertjnic:
> There's already a plugin that does this: https://runelite.net/plugin-hub/show/monsterstats
> 
> If that plugin is missing features you want or has issues, then contribute enhancements/fixes to it instead.
> 
> If the maintainers of that plugin are unresponsive, follow the takeover policy: https://github.com/runelite/runelite/wiki/Plugin-takeover-policy.

Hello, thanks for looking at my request so promptly. The reason I even started this idea was because it seemed like there were dozens of addons that did similar things but varied slightly than were already on the plugin hub. I'm having a hard time understanding why my addon can't be independent? What is the criteria around too similar of a plugin?

cdfisher:
>  it seemed like there were dozens of addons that did similar things but varied slightly than were already on the plugin hub.

This is a situation we are trying to address, which led to some changes in the way hub plugin submissions were approached. Existing plugins were allowed to remain but we strongly encourage collaboration where possible now because it makes things easier for reviewers and also leads to a better experience for end users.

PR #9032, opened 2025-09-08
Update land-surveyor

kilgore-vanhaus:
Added .csv export functionality & corrected default config settings


LlemonDuck:
don't use printStackTrace, use a logger

PR #9034, opened 2025-09-09
Create pbsync

HerbRuns:
Added new plugin 'pbsync'

ReadMe has a little more info, however it's designed to sync and send all current personal best times to a Discord channel via webhook by a chatcommand (!pbsync). In addition, clan chat is monitored for new PBs and sent to a channel via a second webhook.

Was designed for use by the Mirage clan and Discord but has been written in a way to allow all clans to pick up and use if they wish.

### DIFF
--- a/plugins/pbsync
+++ b/plugins/pbsync
@@ -1,2 +1,2 @@
 repository=https://github.com/HerbRuns/pbsync.git
-commit=f1aa36ed694e0c6341221b303fabd11cda08ab23
+commit=906e1e4a8dc5898af205eb4fb2da78e8e83856a2


HerbRuns:
I think I caught all the build step errors that I have control over! Let me know if anything else jumps out that needs addressing, thanks!

HerbRuns:
> Did you check out the Dink plugin, and if so, what did you feel was missing?
> 
> When you want to run client-side commands, use the [CommandExecuted](https://static.runelite.net/runelite-api/apidocs/net/runelite/api/events/CommandExecuted.html) event instead

Thanks for raising this. I did have a loot at Dink but it didn't look like it performed a full sync of all existing times for boss kcs. This is primarily what my clan cc needs for Discord side competitions/notifications etc.

The Dink plugin definitely offers a wealth of notifications though - just not exactly what we needed with regards to all current pbs (albeit able to notify us of new ones).

As for chat command, I missed this in the docs, my bad! I'll check this out. Thanks!

pajlada:
We would be fine with adding a command for pushing all PBs in a webhook in Dink if that would help, it seems like a small enough change.

pajlada:
That's unfortunate - please open an issue for the feature in Dink's repository if you change your mind!

For now then I'm closing this since it falls under plugins "Not currently being considered" (see https://github.com/runelite/runelite/wiki/Rejected-or-Rolled-Back-Features#not-currently-being-considered)

PR #9041, opened 2025-09-10
Clan Broadcasts plugin

2Lica:
Adds Clan Broadcasts plugin to track and send embed Discord messages via Webhooks for all in-game clan broadcasts.

### DIFF
--- a/plugins/clan-broadcasts
+++ b/plugins/clan-broadcasts
@@ -1,0 +1,2 @@
+repository=https://github.com/2Lica/Clan-Broadcasts.git
+commit=825bb84fa9f8b974c8779c82718ce6448f355257


Alexsuperfly:
is this usecase not already covered by the Dink plugin?

2Lica:
> is this usecase not already covered by the Dink plugin?

Dink is more focused on tracking/posting your own drops,levels etc.. This one is transferring all broadcasts, not only yours, but all you witness, and players with the plugin active can broadcast and cover the announcements.

2Lica:
> Wouldn't that also be supported via [their chat notifiers](https://github.com/pajlads/DinkPlugin#chat)?

I don't know if they plan to implement such thing, but from my experience with their plugin, they don't have this option/logic.
Actually that's the reason i have created this code, to give a simple tracking solution for all kind of broadcast, which can be recorded in Discord. So for my situation, clan coffer can be properly monitored, not missing broadcast which awards points to our clan members for their ranks

iProdigy:
> So for my situation, clan coffer can be properly monitored, not missing broadcast which awards points to our clan members for their ranks

See https://github.com/pajlads/DinkPlugin/issues/771#issuecomment-3065009706 for how to achieve that with Dink's chat notifier

PR #9057, opened 2025-09-13
Add Prestige Mode plugin

nick-michael:
<p align="center">
  
</p>
<h1 align="center">
  Prestige Mode
</h1>
<h3 align="center">
  A plugin for <a href="https://runelite.net/">RuneLite</a>
</h3>
Visual only plugin! Adds a 'Prestige Mode' to the skills tab. Resets skills to 1 once you reach 99, and each time you reach 99 a higher tier prestige banner is displayed.

<br />
<br />

### Config options include
- Combat Usability: Only replace the denominator for combat skills to avoid replacing current HP, prayer points, or stat boots/drains.
- Show prestige banners: Toggle rendering of prestige banners. Disabling will still replace level numbers, but won't show any indication of current prestige level.
- Individual on/off toggle for each skill

### Prestige Banners
Prestige banners are overlaid onto individual skills on the skills tab to represent current prestige level. Each time you reach 99, a higher tier prestige banner is unlocked. Since the max XP is 200m, you can reach 99 in each skill 15 times, therefore you can have 15 different prestige banners.

<p align="center">
  
  
  
  
  
  
  
  
  
  
  
  
  
  
  
</p>

### DIFF
--- a/plugins/prestige-mode
+++ b/plugins/prestige-mode
@@ -1,0 +1,2 @@
+repository=https://github.com/nick-michael/rl-prestige-mode.git
+commit=6057dc95e93656662add8bd69c8beb59a0c729c2


nick-michael:
Just to add, my knowledge of Java is not great, and my knowledge of the RuneLite plugin apis are not great, so it's quite likely that improvements can be made to the implementation of this plugin. All feedback is appreciated, and any contributions to the code are also welcome!

Source Code: https://github.com/nick-michael/rl-prestige-mode

nick-michael:
Honestly, I looked at that plugin but there's quite a significant difference in the behaviour it's trying to achieve vs the behaviour this is trying to achieve. If you are adamant this plugin concept should be merged with that one I can give it a go, but I do think there's quite a distinction between the two concepts

riktenx:
The author has previously suggested that this should exist as its own plugin so we will review as-is.

PR #9058, opened 2025-09-14
Update players who have max xp

hessrs:


iProdigy:
> src/main/java/com/calcusourceupdater/CalcusourceUpdaterPlugin.java:96: error: integer number too large

you'll want to append `L` to that constant to make it a long

PR #9059, opened 2025-09-14
Update funnybossnames

Pfannaa:
Fixed hunllef naming and added mokha


pajlada:
You should keep the misspelled config key names, otherwise people who have previously configured your plugin might have their settings changed. You can keep the function name change

Pfannaa:
Thanks for bringing this up. Just updating the key again is enough?

pajlada:
> Thanks for bringing this up. Just updating the key again is enough?

Yup! You've fixed it now so it will respect user's old configs.
Thanks!

Pfannaa:
Cool, thanks :)

iProdigy (approved):

riktenx (approved):

PR #9064, opened 2025-09-14
Update bingosrs plugin

mistereman22:
- Added panel for plugin to display bingo data.
- Added support for new Custom tile type.
- Modified data fetch logic to avoid some API calls when not necessary.

### DIFF
--- a/plugins/bingosrs
+++ b/plugins/bingosrs
@@ -1,4 +1,4 @@
 repository=https://github.com/mistereman22/bingosrs-runelite-plugin.git
-commit=2cea20088c8c2738cc927aeed7a5e4fb38c888f4
+commit=6517e1202eb7afc5beb4627a34eb462aff26952b
 authors=mistereman22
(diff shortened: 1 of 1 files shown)


LlemonDuck:
Client#getItemDefinition needs to be on client thread

if you can submit a patch for just that, everything else lgtm

mistereman22:
> Client#getItemDefinition needs to be on client thread
> 
> if you can submit a patch for just that, everything else lgtm

Is this wrapping of the entire update method not sufficient for that? https://github.com/mistereman22/bingosrs-runelite-plugin/compare/9228326f3e489c5a8ad5d9c4920069a6c29a9631..mistereman22:2cea20088c8c2738cc927aeed7a5e4fb38c888f4#diff-5f70c0cd6c43a2d5e48480c60ad408a00705518a9bcc051ec152809b49a68d50R90

I remember running into some issues with trying to inject and use the client thread at a lower level, but haven't looked in a while so if more is needed then I may need an example plugin to reference.

PR #9083, opened 2025-09-17
Add Random Event Helper to the Plugin Hub

Infinitay:
This plugin helps solve various random events by offering the solutions to them via highlighting the proper widgets or entities to interact with.

### DIFF
--- a/plugins/random-event-helper
+++ b/plugins/random-event-helper
@@ -1,0 +1,2 @@
+repository=https://github.com/Infinitay/Random-Event-Helper.git
+commit=7027f42551ba94235eb1b10d21d7e882574380e8


Infinitay:
Recent commit fixed build issue resulting due to incorrect class within plugin properties file. Also changed again was the plugin properties file but with additional tags.

Infinitay:
I realize now I probably shouldn't have force pushed... whoops. Anyway, I have made the changes you requested. Thanks for the suggestion!

Additional commits between the last two commits:

### chore: Add plugin icon

### fix(supriseexam): Fix hocus pocus and food relations
- Add hocus pocus hint tag for magic relation
- Rename food relations to be more clear
- Update plugin to v2.5.2

### refactor(maze): Remove auto enabled Shortest Path, Update chat message
- Removed the feature that auto-enabled Shorest Path plugin when installed but disabled for the maze helper
- Updated the chat message to instruct the user they have to manually enable it
- Update plugin to v2.5.3

PR #9050, opened 2025-09-12
Update pvp-performance-tracker to v1.7.1

Matsyir:
Most of the functional changes: Support for god spells, special attacks for arkan blade, burning claws & dark bow. Double deaths now tracked. New statistic labels & improved tooltips. Various calculation & detection improvements. Renamed Deserved damage to Expected damage. Thanks to sacca for many PRs here

Many refactors to help with maintainability in the future, regarding UI generation. Here are some custom diff links to hopefully help review. Let me know if you have any questions, or if we'd rather split this into multiple PRs, but I think with these seperate diff links it should be okay. Either way, I understand if it takes a while to review.

[0]->[1]: Most of the functional changes, basically the old/current hash up to right before the UI reworks and deservedDmg renames: https://github.com/Matsyir/pvp-performance-tracker/compare/43799b39409d9dfd8e20c72ba3f90d4e20e501dd...Matsyir:cd22da8bcda50a06cf9f5ecc10fe88a43fb8ba8c [12files, +230, -77]

[1]->[2]: Most of the UI reworks, before deservedDmg rename (excluding initial functional changes): https://github.com/Matsyir/pvp-performance-tracker/compare/cd22da8bcda50a06cf9f5ecc10fe88a43fb8ba8c...Matsyir:6041bf5f323483bf72cdb0cfdbae3855445b224b [12files, +1,462, -561] (this part is way bigger than I expected, surprised its not removing more than adding; but pretty straightforward to understand i think)

[2]->[3]: Renaming Deserved Damage -> Expected Damage, plus the few small remaining commits of fixes/tooltip adjustments (excluding the last 2 batches): https://github.com/Matsyir/pvp-performance-tracker/compare/6041bf5f323483bf72cdb0cfdbae3855445b224b...Matsyir:bc4bfaf7fd90894eeeba4fd924fcc8f303cea47a [15files, +234, -219]

Commit hashes used:
[0]: `43799b39409d9dfd8e20c72ba3f90d4e20e501dd` (this is the old/currently live hash)
[1]: `cd22da8bcda50a06cf9f5ecc10fe88a43fb8ba8c` (before ui reworks)
[2]: `6041bf5f323483bf72cdb0cfdbae3855445b224b` (after ui reworks/before deservedDamage rename)
[3]: `bc4bfaf7fd90894eeeba4fd924fcc8f303cea47a` (this is the new hash being commit here)

[...]

PR #9084, opened 2025-09-18
Update pvp-performance-tracker to v1.7.1

Matsyir:
Copy of #9050, apparently when I force push while it's closed you can't re-open (ty again to sacca for spotting an issue with my ui migration)

Most of the functional changes: Support for god spells, special attacks for arkan blade, burning claws & dark bow. Double deaths now tracked. New statistic labels & improved tooltips. Various calculation & detection improvements. Renamed Deserved damage to Expected damage. Thanks to sacca for many PRs here

Many refactors to help with maintainability in the future, regarding UI generation. Here are some custom diff links to hopefully help review. Let me know if you have any questions, or if we'd rather split this into multiple PRs, but I think with these seperate diff links it should be okay. Either way, I understand if it takes a while to review.

[0]->[1]: Most of the functional changes, basically the old/current hash up to right before the UI reworks and deservedDmg renames: https://github.com/Matsyir/pvp-performance-tracker/compare/43799b39409d9dfd8e20c72ba3f90d4e20e501dd...Matsyir:cd22da8bcda50a06cf9f5ecc10fe88a43fb8ba8c [12files, +230, -77]

[1]->[2]: Most of the UI reworks, before deservedDmg rename (excluding initial functional changes): https://github.com/Matsyir/pvp-performance-tracker/compare/cd22da8bcda50a06cf9f5ecc10fe88a43fb8ba8c...Matsyir:6041bf5f323483bf72cdb0cfdbae3855445b224b [12files, +1,462, -561] (this part is way bigger than I expected, surprised its not removing more than adding; but pretty straightforward to understand i think)

[2]->[3]: Renaming Deserved Damage -> Expected Damage, plus the few small remaining commits of fixes/tooltip adjustments (excluding the last 2 batches): https://github.com/Matsyir/pvp-performance-tracker/compare/6041bf5f323483bf72cdb0cfdbae3855445b224b...Matsyir:69ac80a72cb7797d0bdb802bdca0d960cd3033f6 [15files, +253, -231]

Commit hashes used:
[0]: `43799b39409d9dfd8e20c72ba3f90d4e20e501dd` (this is the old/currently live hash)
[1]: `cd22da8bcda50a06cf9f5ecc10fe88a43fb8ba8c` (before ui reworks)
[2]: `6041bf5f323483bf72cdb0cfdbae3855445b224b` (after ui reworks/before deservedDamage rename)
[3]: `69ac80a72cb7797d0bdb802bdca0d960cd3033f6` (this is the new hash being commit here)

Here is a screenshot of the new UI label (not the biggest addition for how big of a refactor this was, but it will be easier to maintain). For the overlay, I properly took into consideration what happens when the left labels overlap the middle label, it's fine and not too instrusive due to the way it's layered, and they have low opacity. For the panel, it just doesn't really happen in practice, but it does also handle it relatively gracefully.
image

### DIFF
--- a/plugins/pvp-performance-tracker
+++ b/plugins/pvp-performance-tracker
@@ -1,2 +1,2 @@
 repository=https://github.com/Matsyir/pvp-performance-tracker.git
-commit=43799b39409d9dfd8e20c72ba3f90d4e20e501dd
+commit=69ac80a72cb7797d0bdb802bdca0d960cd3033f6


LlemonDuck:
This would have been so much more legible and reviewed so much sooner if you used things like String.format instead of 15 concatenations in a row and abstract classes instead of 6 unnamed lambdas right after each other. for next time, if you're interested

PR #9096, opened 2025-09-19
seed vault memory plugin

matt-finney:
A plugin for [RuneLite] (https://github.com/runelite/runelite) which remembers the contents of your seed vault and displays that in a searchable interface. Based on LazyFaith's "Bank Memory plugin" (https://runelite.net/plugin-hub/LazyFaith)


cdfisher:
We would prefer if you tried to contribute this functionality to Bank Memory rather than submitting an entirely new plugin

PR #9101, opened 2025-09-20
Add pvp leaderboard

findarian:
PvP Leaderboard plugin.

Overall it sends POST request after you fight someone to my AWS API gateway. Backend does logic and keeps this data in a dynamoDB to process matches and be able to get this information to my website and back to the plugin.

The plugin pulls ranks of people around them in game, this information is gotten from my website https://devsecopsautomated.com/ and the API gateway endpoints depending on which information is requested. Then that information is pushed back into client to display people's ranks. 

It loads ranks of people around you that are also on the leaderboard and works with different styles of combat (multi/dmm/nh/pvp/veng). 

I appreciate all of your time in reviewing this, you guys keep this community great <3

### DIFF
--- a/plugins/pvp-leaderboard
+++ b/plugins/pvp-leaderboard
@@ -1,3 +1,3 @@
 repository=https://github.com/findarian/pvp-leaderboard.git
 commit=f60d732babfea7b9353e558e961ca61d24f928cf
-warning=This plugin connects to external servers by default to fetch and submit PvP data. Your IP address and in-game username may be transmitted to the external service.
+warning=This submits your IP address, in-game username, and PvP fight statistics to a 3rd-party service not controlled or verified by RuneLite developers.


pajlada:
you've made changes to another plugin

findarian:
Sorry about that, this should be fixed now.

pajlada:
Some minor feedback, haven't done a full review

You should probably do your development on a single branch - you're now pushing commit hashes from your `add-pvp-leaderboard` branch

The `plugins` directory in your repository serves no purpose, you only need to update the plugin hub manifest file in the `plugin-hub` fork that you PR to this repo.

You should not need to make the dependency changes you've made in your `build.gradle` file (line 20 & 29) - other plugins use gson and lombok file without specifying those. Did you run into any issues when you didn't make those changes?

pajlada:
for JSON things, inject & use gson (see https://github.com/pajlads/DinkPlugin/blob/075b32f12e223eab8b67116c856dec43701b0209/src/main/java/dinkplugin/message/DiscordMessageHandler.java#L205 as an example for parsing a json body from a network request, and see https://github.com/pajlads/DinkPlugin/blob/075b32f12e223eab8b67116c856dec43701b0209/src/main/java/dinkplugin/message/DiscordMessageHandler.java#L325 for how to send a json payload in a request)
for network requests, inject & use OkHttpClient and make async requests with enqueue (see https://github.com/pajlads/DinkPlugin/blob/075b32f12e223eab8b67116c856dec43701b0209/src/main/java/dinkplugin/message/DiscordMessageHandler.java#L186-L227 as an example)
for executor stuff, inject & use our ScheduledExecutor (unless you have really long-lived/heavy-to-perform tasks)

findarian:
These changes have been applied.

pajlada:
You're still using some random ways of parsing json & network requests despite what I told you, and you telling me you've fixed it.

Your auth seems overly complex - could you solve this without involving a web server?

You're still doing really wonky things with the executor, despite what I told you, and you telling me you've fixed it.

You've removed the warning from the plugin hub manifest - it still needs to be there.


------------------------------


I get the feeling that you're vibing too hard - please look at core plugins (i.e. plugins inside the `runelite/runelite` repository) and see how they do things and use those as reference.

findarian:
You're still using some random ways of parsing json & network requests despite what I told you, and you telling me you've fixed it.
-I've fixed these now. Sorry about before, thank you for pointing this out.

Your auth seems overly complex - could you solve this without involving a web server?
-Short answer is yes, but I don't think that this is a good idea.

So the current implementation of Authorization Code + PKCE with loopback was chosen for security mainly. I won't have the secret in the plugin. There's no passwords handling in the plugin as well. Ideally I don't want to know the passwords at all and this is how I set up cognito.
https://docs.aws.amazon.com/cognito/latest/developerguide/what-is-amazon-cognito.html
I also wanted to minimize how hard this was for the user to login (which isn't even required, it just shows more stats for those that want it). 
If you have another method that I haven't considered that would follow best practice I'm open to it, but plugins like https://github.com/cbrewitt/flipping-copilot have a different setup, but it still uses a web server regardless



You're still doing really wonky things with the executor, despite what I told you, and you telling me you've fixed it.
-I still have executor for a few things, what would you like to see changed and why? 

PvPLeaderboardPlugin
Injected ScheduledExecutorService named scheduler:
scheduleAtFixedRate: 1s GC for fight state.
schedule: 1500 ms finalize guard and 2000 ms watchdog.
schedule: 15 s delayed API fallback for rank overlay when shard miss occurs.

RankOverlay
Injected ScheduledExecutorService used as executor for lookup futures. No custom thread pools. Backpressure via per-tick cap and in-flight guards. OK.

CognitoAuthService
Injected ScheduledExecutorService for token refresh scheduling (next refresh, retry). CompletableFuture for login uses default (test path), but operational flow is bound to injected scheduler for scheduled work.

DashboardPanel
One injector fetch of the shared scheduler for auth service setup only.

You've removed the warning from the plugin hub manifest - it still needs to be there.
-I've added this back in.


Thank you again for your time here, I am learning a lot going through this process with you.

findarian:
https://discord.com/channels/301497432909414422/419891709883973642/1404614275519221860

Discussion around a month ago when I was starting to make the plugin side of this project. Pajlada said this would be a good idea to put here for reference for whoever does the full review.

findarian:
Did some bug fixes.

findarian:
The purpose of this wasn't to scout people at all, but rather show ranks of people around you. Would it be fine to remove this in the wilderness specifically, and have your own rank still displayed? My thought was to have ranks displayed with a purpose in the future to have matchmaking based off of this. 

Before I started working on this I wanted to make sure that this functionality was ok as it was a large portion of the project, and it seemed like it was flagged as fine here https://discord.com/channels/301497432909414422/419891709883973642/1404614275519221860 

For the panel, when you say you must not render information about nearby players, I have the option to right click -> pvp lookup which would get the same information in the panel currently that would be available on the website. This would include their rank for NH/Veng/Multi/DMM/Overall based off the matchesa submitted, their last 100 game performance vs various tiers of players, win rate history, and match history along with the MMR gain loss, opponent name, which style, their rank vs their opponent's rank, change in MMR, and time that the match was submitted.

The purpose of this plugin isn't for scouting, it is to see where you stack up against everyone else in a trueskill environment in different styles. Let me know if this makes any difference for what is needed to be changed about the plugin. pvp performance tracker for instance has way more detailed information about people you've fought before down to the exact prayers and everything that they use per tick in fights against you and has a lookup history that can be used to see not just what damage you did, but deserved to do and could be used in a much better way to scout people, especially with being able to import matches (I previously did this for tens of thousands of matches when I was testing out the MMR system without any match history before I had beta testers).

The Last Man Standing plugin also has scores displayed from official ranks overhead 
image


As far as I could tell though, when I messaged back in August in Discord this overlay feature wouldn't be an issue because it's not based off of any official data sources and it's generated on it's own. This is what it looks like in the panel.


image

riktenx:
Not interested in going back and forth on this or frankly anything else in this PR with you. Not really interested in anything other than me saying "fix this / remove this" and you doing it. You can update a plugin after it's released, you can literally just do that and try to reintroduce these things later so they can be discussed without blocking the review of your massive plugin.

If that does not work for you you are welcome to wait for someone else's review.

riktenx:
Your nav panel logic in your startUp is broken/unstable, please [merge this PR](https://github.com/findarian/pvp-leaderboard/pull/4).

plugins/pvp-leaderboard
```diff
@@ -0,0 +1,3 @@
+repository=https://github.com/findarian/pvp-leaderboard.git
+commit=f60d732babfea7b9353e558e961ca61d24f928cf
+warning=This plugin connects to external servers by default to fetch and submit PvP data. Your IP address and in-game username may be transmitted to the external service.
```
riktenx:
```suggestion
warning=This submits your IP address, in-game username, and PvP fight statistics to a 3rd-party service not controlled or verified by RuneLite developers.
```

PR #9104, opened 2025-09-20
Update phoenix-necklace-jingle

Thedyolf:
Updated plugin to support custom sounds and different sound IDs. It makes a placeholder empty .wav file on launch for user to replace if they want to. Sound ID can be edited for it to play a different sound rather than the GE jingle.

### DIFF
--- a/plugins/phoenix-necklace-jingle
+++ b/plugins/phoenix-necklace-jingle
@@ -1,2 +1,2 @@
 repository=https://github.com/Thedyolf/phoenix-necklace-jingle.git
-commit=b77e18a556b1ea3a3fa5d71368e93286de6342be
+commit=d053ae53da2635f1201925a3e2bf7e7fe30f6767


LlemonDuck:
what's the point of writeEmptyPcmWav?

Thedyolf:
I want the user to be able to play a custom wav sound they insert in the dedicated folder. That way they can use whatever sound they like. Initially I was going to pack a .wav file in the jar and extract it. But seeing that its only job was to create a 'placeholder' file for the user to replace and I did not want to deal with sound licensing, I opted for creating an empty .wav file instead. 
No file could be created, but it'd create more ambiguity for the user to where exactly to place the file and what to name it, so creating an empty file reduces user vagueness.
Obviously I could just create an empty wav file without any encoding, but if the user would double click that in their explorer it would give a corrupted file message as it would not be properly encoded to have wav features. Hence I decided to create an empty but valid wav file.

I am open for any suggestions on how to do this more elegantly if you would take a different approach! Let me know!

PR #9105, opened 2025-09-20
Create custom-xp-orbs plugin

Sieder16:
This takes the current XP Globes plugin and makes it so you can customize more options such as

Display Level (Default Yellow) & (Shifts Icon Upwards), Display Level Colour, Maximum Orbs, Maximum Orbs Per Row, Blacklist Per Orbs, Force Per Orbs


tylerwgrass (approved):

LlemonDuck (changes requested):
don't use the net.runelite package namespace

you need to preserve the license headers from the core plugin

LlemonDuck:
I copy pasted it because you didn't do the second item of licenses and I wanted to block other mergers until that was done. And yes, that's what you need to do

PR #9109, opened 2025-09-21
Create Minimap Cleanser (updated)

Mattb2291:
Added a de-clutterer for the minimap. This is a similar concept to the Ground Object Hider plugin, but specifically allows cleansing of the minimap icons displayed.


Mattb2291:
Adding World Map cleaning options too so will re-release when ready

PR #9116, opened 2025-09-22
Create tob-party-sync

LeonidasTech:
Using runelite's Party function is very important when doing ToB. It is quite hard to get people from the free for all world, w416, to join a party hub. This way, instead of having to manually join / have people not join, they will join automatically which will be benefitial to all raid participants.

This is also useful for we do raids, and raiding in general. It cuts out the effort of creating a party, as all you have to do is join the raid team and one is created for you.


raiyni:
We intentionally don't want to support auto-joining parties. 

PR #9125, opened 2025-09-25
Update escape-crystal-notify (v1.13)

Xylot:
Bug fixes:
- Ensure manually excluded region IDs take priority over the `Enable Everywhere` option
- Lower entrance deprioritization priority to ensure it's always on top

Add/Change:
- 6-hour log entrance warning overlay for Leviathan and Doom entrances
- Simplify Leviathan and Doom safeguard configuration settings

### DIFF
--- a/plugins/escape-crystal-notify
+++ b/plugins/escape-crystal-notify
@@ -1,2 +1,2 @@
 repository=https://github.com/Xylot/escape-crystal-notify.git
-commit=6b5322a046ffacba344b68af63611fa48136a683
+commit=809e65ac49246dedea83848b997171179a85beae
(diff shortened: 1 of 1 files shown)


riktenx (approved):
Note that you're going to effectively be resetting some of these configs for people by giving them new keys (e.g. inverting the hide option under a new key to show). You might consider doing a config migration.

tylerwgrass (approved):
You could just map the old "On" setting to "All Accounts"??

If you still don't care we can merge anyways

Xylot:
I am okay with merging this now! The new default will be HC_ONLY and I would assume most people who have it enabled would prefer that option over having it defaulted to irons and mains as well

PR #9127, opened 2025-09-25
Create rolreminder

lurkzies:
I've made a basic plugin to notify/remind you to equip your ring of life if it isn't equipped. Can be useful for HCIMs.

### DIFF
--- a/plugins/rolreminder
+++ b/plugins/rolreminder
@@ -1,0 +1,2 @@
+repository=https://github.com/lurkzies/Ring-of-Life-Reminder.git
+commit=46994fa105ec128a42ac7bd566dbf281c9a0b210


pajlada:
You have a build issue - see the build action above

There's this plugin https://runelite.net/plugin-hub/show/life-saving-jewellery that seems kind of similar - did you look at that before making your plugin?

lurkzies:
> You have a build issue - see the build action
> 
> There's this plugin https://runelite.net/plugin-hub/show/life-saving-jewellery that seems kind of similar - did you look at that before making your plugin?

Looking into issues, thanks. I did find that plugin, but it provides an infobox if it is equipped. I prefer not having the clutter when it is equipped, and having a reminder when it isn't.

PR #9140, opened 2025-09-26
Create fishing-spot-timer

ArnautS:
This plugin shows a timer above each fishing spot, starting when it last moved. Since fishing spots don't have a set move time, the length of the timer is configurable in the settings. The intended use for this plugin is when afk fishing and your spot just moved, you want to start fishing on the spot that moved last, allowing maximum afk time.


coopermor:
Does this not function the same as [NPC Idle Timer](https://runelite.net/plugin-hub/show/npc-idle-timer-plugin)?

PR #9141, opened 2025-09-27
Add Nearby Player Logger plugin

rickandreae:
Small plug-in to show names of players currently visible.


cdfisher:
Instructions on how to properly develop and submit hub plugins can be found [here](https://github.com/runelite/plugin-hub/blob/master/README.md).

PR #9158, opened 2025-09-28
Update reword-manager to 1.5.3

ejaz-karim:
Compatibility with Menu Entry Swapper's left click teleport. Before this version, it would break.


ejaz-karim:
I've removed that list, and I am instead using onPostClientTick

riktenx (approved):

tylerwgrass (approved):

PR #9160, opened 2025-09-28
added alch warning plugin

SugoiAF:
This plugin allows players to block or allow casting alchemy on items based on configurable item lists and the items GE value. 

Blacklist / Whitelist: Prevent or always allow alchemy on specific items.

Max Alch Value Filter: Blocks items with a GE value above a configurable limit.

Chat Warnings: Alerts the player when an alch is blocked.

The plugin does not change the default OSRS alchemy warning settings and does not work with the explorers ring yet.

This is also my first plugin so apologies in advance for beginner mistakes...

### DIFF
--- a/plugins/alchemy-warning-filter
+++ b/plugins/alchemy-warning-filter
@@ -1,0 +1,2 @@
+repository=https://github.com/SugoiAF/Alch-Warning-Filter.git
+commit=41e70540ed58235854249ce07c6511e2c3bbd875


riktenx:
https://github.com/runelite/plugin-hub?tab=readme-ov-file#contribute-to-existing-plugins

There are already several plugins on the hub for alch blocking. See:
* https://runelite.net/plugin-hub/show/alch-blocker
* https://runelite.net/plugin-hub/show/no-bad-alchs

If those plugins are lacking features that you want, then you should seek to contribute fixes/enhancements to them instead of creating a new plugin. If the maintainer(s) of those plugins are unresponsive, see https://github.com/runelite/runelite/wiki/Plugin-takeover-policy.

PR #9162, opened 2025-09-28
added alch warning plugin

SugoiAF:
There are similar plugins to mine but none of them are upkept and none of them have a whitelist to allow alchs to go through. They all suffer from the same problem that the actual osrs high alch settinga have and thats if you set a ge value threshold nothing is allowed through. My plugin seeks to remedy that. I also intend to keep updating it.

This plugin allows players to block or allow casting alchemy on items based on configurable item lists and the items GE value.

Blacklist / Whitelist: Prevent or always allow alchemy on specific items.

Max Alch Value Filter: Blocks items with a GE value above a configurable limit.

Chat Warnings: Alerts the player when an alch is blocked.

The plugin does not change the default OSRS alchemy warning settings and does not work with the explorers ring yet.

### DIFF
--- a/plugins/alchemy-warning-filter
+++ b/plugins/alchemy-warning-filter
@@ -1,0 +1,2 @@
+repository=https://github.com/SugoiAF/Alch-Warning-Filter.git
+commit=41e70540ed58235854249ce07c6511e2c3bbd875


raiyni:
We would rather you contribute to an already existing plugin.

cdfisher:
We would significantly prefer for you to contribute to one of the similar existing plugins to improve them rather than adding a new plugin to the hub.

iProdigy:
> There are similar plugins to mine but none of them are upkept

[no-bad-alchs](https://github.com/CreativeTechGuy/no-bad-alchs) was literally updated earlier this month

> none of them have a whitelist to allow alchs to go through

no-bad-alchs also has a `Allowed items` config option

&nbsp;

re: https://github.com/runelite/plugin-hub/pull/9160#issuecomment-3343077472

SugoiAF:
I guess i didnt understand how no bad alchs worked.  When i looked at it i didnt think it was very user friendly. If you dont want it you dont want it though. Figured id at least explain where i was coming from.

PR #9166, opened 2025-09-29
Create membership-alert

tbloody:
This plugin shows a little overlay to alert you to monitor when your memberships is about to run out.


raiyni:
You need to remove your overlay on shutdown and  remove your `log.info` calls or change them to debug.

tylerwgrass (approved):
LGTM, but I would recommend adding an option to not show the overlay if the remaining days is not in the danger/warning amounts.

PR #9168, opened 2025-09-29
add colosseum hp tracker

Supalosa:
This is a version of the [Tzhaar HP Tracker plugin](https://github.com/MoreBuchus/buchus-plugins/tree/tzhaar-hp-tracker) but adapted for Colosseum. (I have attempted to contact the author of that plugin since this could be folded into it, but to no avail)

Other than adding details for Colosseum NPCs and regions, this plugin also predicts the target that a Venator bow will hit, and includes an optional indicator to show players which NPCs will be hit.

Some clips of the plugin in action:
https://streamable.com/8fa4q8
https://streamable.com/i1kjal
https://streamable.com/3zjgu1

### DIFF
--- a/plugins/colosseum-hp-tracker
+++ b/plugins/colosseum-hp-tracker
@@ -1,2 +1,2 @@
 repository=https://github.com/Supalosa/colo-hp-tracker.git
-commit=326737985261af3cb6b3c7ad6012fb990f245f7e
+commit=4d69bb91a86f25d3d9e80f32070dbaf636b971d5


Supalosa:
I don't really think this belongs in the Fortis Colosseum plugin. It's geared towards a much narrower set of players (essentially speedrunners) and performs a lot of calculations every tick to track damage. It's also just about as big (in terms of LOC) as Fortis Colosseum so it'd be quite a lot of maintenance moving forward.

Finally, I'm actually still hopeful this can be folded back into the Tzhaar HP Tracker when Buchus finds interest in the game again so the Venator bow logic can eventually work in Inferno - so the closer the structure stays to the original plugin, the easier that transition will be.

riktenx:
closing since the above hasn't been addressed, if/when you do so you can make a new PR. our offer to transfer the original plugin to you still stands as well.

Supalosa:
Thanks guys. Ultimately I fell victim to running out of time to maintain the plugin (maybe the same issue as the original author!). Better to not have the plugin than to have yet another unmaintained plugin. Cheers.

PR #9179, opened 2025-10-01
Add plugin: Rune Pouch Redesign

trs:
Add a new plugin to redesign the rune pouch loadout widget.

image

### DIFF
--- a/plugins/rune-pouch-redesign
+++ b/plugins/rune-pouch-redesign
@@ -1,0 +1,2 @@
+repository=https://github.com/trs/runelite-rune-pouch-redesign.git
+commit=8fe79f16a7b99cbaee2163566e3b099df329cda4


evilgenious448:
Prior to the Jagex rework of the Runepouch, @DapperMickie had a plug-in that allowed users to add custom loadout names.

https://github.com/DapperMickie/runepouch-loadout-names

As of the most recent update his plug-in is broken. Your plug-in addresses a lot of the feedback players have towards the way the Runepouch UI looks.

Perhaps rather than submitting a whole new plug-in you can PR your code into @DapperMickie’S which already has nearly 4000 users. Would be better to have 1 dedicated plug-in than 2 that effectively aim to do the same thing.

Although I can’t speak for @DapperMickie, I know he has collaborated on other plugins before and I have no doubt he would be open to making you co-author and collaborator considering the amazing work youve done here. 

P.S love that you’ve made this, can’t wait to use it!

PR #9183, opened 2025-10-01
Add item-locker plugin

Dari-K:
Item locker plugin I'm used for my quest locked ironman. I've had a few people request it, so I thought I'd upload it.

Plugin adds a red overlay to all items in various UI locations (bank, inventory, etc) to indicate locked status. Default left click option is change to drop/destroy. Items are unlocked by holding shift and right clicking, which will add a lock/unlock option to the right click menu.

### DIFF
--- a/plugins/item-locker
+++ b/plugins/item-locker
@@ -1,0 +1,2 @@
+repository=https://github.com/Dari-K/item-locker.git
+commit=afbf86f13797ca63d67fb14f427e5ba5a7f4f650


tylerwgrass:
Not a full review.

> Default left click option is changed to drop/destroy.

This poses an unnecessary risk that users will drop very expensive items and cause disruptive gameplay. I would recommend instead just consuming the click and performing a no-op instead if you really want this sort of functionality.

PR #9197, opened 2025-10-03
Update phoenix-necklace-jingle

Thedyolf:
Removed creating sound from code, instead shipping empty wav in JAR now. 

From current version in pluginhub: Added option to play any ingame sound in game by sound ID, alternatively play any wav file instead that can be placed in the .Runelite folder by the user.


Thedyolf:
Accidentally closed and reopened a new PR instead of editing the old one, apologies. Here's the old PR for reference: https://github.com/runelite/plugin-hub/pull/9104

PR #9210, opened 2025-10-05
Create lockedchunkoverlay

rauvagol:
Overlay that works with the region-locker plugin to display locked chunks on the world map with any gpu.

Includes optional warnings and penalties for time spent in a locked chunk (disabled by default except for increasing (non-opaque) darkness)

First plugin, so I apologize in advance if I messed up any steps, and I will correct any issues ASAP

### DIFF
--- a/plugins/lockedchunkoverlay
+++ b/plugins/lockedchunkoverlay
@@ -1,0 +1,2 @@
+repository=https://github.com/rauvagol/lockedchunkoverlay.git
+commit=65843e2794b4a35a2069c8a41b038632bdf1513c


riktenx:
FYI if your networking happens behind an opt-in config, you can get rid of the manifest-level disclaimer and instead put it in `@ConfigItem::warning` (https://static.runelite.net/runelite-client/apidocs/net/runelite/client/config/ConfigItem.html#warning())

rauvagol:
alright, done

i also moved the "this uses a site" warning to a checkbox that is required to actually use the chunkpicker site dropdown

i figured that makes far more sense than spamming the user with a meaningless warning if they want to swap from manual input to region locker plugin input (since neither of those actually use a site)

post-change the only site access is performed if the dropdown is on "chunk picker site" and "Enable Site Access" is checked

PR #9208, opened 2025-10-05
Exact Path Preview Plugin

blindtcg:
Exact path preview is a plugin that displays the tile path your character will take to reach the tile you are mouse-hovering over, with many configuration options

PR #9211, opened 2025-10-06
Exact Path Preview

blindtcg:
Exact Path Preview is a plugin that displays the tile path your character will take to reach the tile your mouse is hovering over, with multiple configuration options.


cdfisher:
Please don't open multiple PRs for the same submission. Just push any changes to the branch that you're using for #9208

PR #9212, opened 2025-10-06
Revert rune pouch layout

vatsake:
https://github.com/vatsake/revert-rune-pouch-layout

image


Felanbird:
The plugin that originally added loadout names https://runelite.net/plugin-hub/show/runepouch-loadouts has been updated to also do the opposite, and hide names. Does this update suffice for what you wish to do with your plugin?

PR #9355, opened 2025-10-27
Update DropTracker to v5.2.2 - Fix creation of Color object

joelhalen:
Sorry for the quick succession updates today.. just got home a bit ago, and the plugin was still creating (non-breaking) NullPointer errors inside our new warning methods. Fixed this by using fromHex instead of fromString to create red.

I will do my best to be better about testing changes myself prior to publishing from here on, apologies for the troubles :'(


iProdigy (approved):
btw `Color.RED` exists

PR #9522, opened 2025-11-16
update membership-alert

tbloody:
added option to not show overlay when not in warning or danger zone as
proposed by riktenx + attempt to load properly after install without
need to login

### DIFF
--- a/plugins/membership-alert
+++ b/plugins/membership-alert
@@ -1,2 +1,2 @@
 repository=https://github.com/tbloody/osrs-membership-alert.git
-commit=953d9cad520a6ef5dfeb9777bb23bf43eea02949
+commit=05ca10a36f3878d1b3f64fe0f837d2884dcdc9d1


iProdigy (dismissed):

iProdigy:
in `startUp`, you should check whether the player is logged in before calling `getVarpValue` (and do that on the client thread)

jarromie:
I have requested collaborator access via creating an issue on @tbloody's repository, so that I may make the needed changes and open a new PR.

tbloody:
> in `startUp`, you should check whether the player is logged in before calling `getVarpValue` (and do that on the client thread)

I am picking this up again, I don't know what you mean here with doing it on the client thread. Do you simply mean  adding 
`if(client.getLocalPlayer() == null) return; `
 before calling getVarpValue?
I'll make a new pull request with that added.

PR #9642, opened 2025-11-26
Update monsterstats

Koitere:
Updating database for sailing monsters, added flat armour stat for viewing, added stricter checking for NPCs without stats available to no longer show right click options for NPCs without available stats.


Koitere:
Fwiw the xl change size is from my database changing, is there anyway (other than dynamically building the database which im working on for the future) that I can reduce the size of this change when each line needs to be changed in my database?

LlemonDuck:
in a case like this where it's egregiously off (functional size here is like 50) feel free to ping me personally

PR #10498, opened 2026-02-07
Update clogman-mode to v1.2.0

mozjay:
- Added support for alternative recipe sources. Items with multiple recipe paths (e.g. Armadylean plate -> Masori fortified variants) now unlock correctly.
- Updated clog_restrictions.json to support alternative recipes + added some alternatives for problematic items
- Updated chat message on restricted item check to find closest recipe to unlock

Release: https://github.com/mozjay/clogman-mode/releases/tag/v1.2.0


mozjay:
I realise most of the actual changes here are to the clog_restrictions.json file the plugin includes. Is there a better/recommended way to do this? The .json is already generated from a helper repo - is it allowed for me to pull that from github in the plugin or something rather than submitting a PR every time the .json gets updated? Thanks.

PR #10806, opened 2026-03-01
feat: initial runealytics plugin

mrtms:


cshunton:
> Question from a bystander; doesn't this require a warning like [wom-utils](https://github.com/runelite/plugin-hub/blob/master/plugins/wom-utils) and such?

We can definitely add that

cshunton (commented):
This is an example of the data the plugin uses to show off your drops/XP. 


image

@mhollink

LlemonDuck (changes requested):

LlemonDuck (changes requested):

mrtms:
> typo in my own suggestion, sorry

All good! Committed the update!

LlemonDuck:
do you intend for the default url to be a testing url? https://github.com/Runescape-Tracking/RuneAlytics-Plugin/blob/f4e2657ad857ac785eb4c79083133ebb4bfefb42/src/main/java/com/runealytics/RunealyticsConfig.java#L67

mrtms:
> do you intend for the default url to be a testing url? https://github.com/Runescape-Tracking/RuneAlytics-Plugin/blob/f4e2657ad857ac785eb4c79083133ebb4bfefb42/src/main/java/com/runealytics/RunealyticsConfig.java#L67

Nice catch and probably on the safer side if we do the main server rather than the testing one since that's where we deploy testing code to. Thanks for that - appreciate the thorough code review!

000744210:
I notice you could make your bank stats public. Does that go against this?
> Hiscores for PBs and other information not on the hiscores: Same as the GE Tracker, the data could easily be spoofed, rendering it useless

LlemonDuck:
remove runelite-plugin.properties from /src/main/resources

gitignore /out while you're at it

plugins/runealytics
```diff
@@ -0,0 +1,3 @@
+repository=https://github.com/Runescape-Tracking/RuneAlytics-Plugin.git
+commit=04707a5031dc19f266798494ffde3ba02244f8db
+warning=This plugin submits your IP address, in-game name, friends list, members of your clan chat, other player metrics, and optional bank contents, to Runealytics.com which is not maintained or reviewed by RuneLite.
```
LlemonDuck, line 3:
```suggestion
warning=This plugin submits your IP address, in-game name, friends list, members of your clan chat, other player metrics, and optional bank contents, to Runealytics.com which is not maintained or reviewed by RuneLite.
authors=mrtms
```

LlemonDuck (changes requested):

plugins/runealytics
mrtms: Added the line - slightly different with the authors option. Added `cshunton` to that list.

LlemonDuck:
build fail

and you don't need to request a review from me specifically on each push

mrtms:
Fixed it. Apologies on the request a review!

mrtms:
> runelite-plugin.properties



> remove runelite-plugin.properties from /src/main/resources
> 
> gitignore /out while you're at it

Didn't notice this till now - finished this as well.

LlemonDuck:
change log.info to log.debug across the whole codebase

everything else looks fine @ 19c51204ed969373a6a38eb2d2eb7f7148b45047. try not to change anything else to ensure a quicker review on the next pass.

mrtms:
Updated `log.info` to `log.debug`

PR #10820, opened 2026-03-02
Add trackpad camera plugin repository information

crichmond1989:
# Trackpad Camera Plugin for RuneLite

Control the Old School RuneScape camera using trackpad gestures. Developed for MacBook trackpads, but compatible with any input device that encodes gestures the same way.

## Gestures

| Gesture | Action |
|---|---|
| Two-finger swipe left/right | Rotate camera (yaw) |
| Ctrl + two-finger scroll up/down | Tilt camera (pitch) |
| Two-finger swipe up/down | Zoom camera (OSRS native) |

## Configuration

Each gesture can be individually enabled, inverted, and tuned:

| Section | Options |
|---|---|
| Zoom | Invert scroll direction |
| Rotation | Enable/disable, invert, sensitivity (1–20) |
| Tilt | Enable/disable, invert, sensitivity (1–20) |

The plugin intercepts scroll events by modifier key — it does not detect the input device. Mouse users who scroll while holding SHIFT or CTRL will also trigger rotation and tilt. Zoom inversion applies to all plain vertical scroll input.

## Requirements

- Java 11
- RuneLite client
- A trackpad that encodes horizontal swipes as `SHIFT+scroll` and ctrl+scroll as `CTRL+scroll` (standard on macOS)

## Development

```bash
./gradlew build    # compile and lint
./gradlew test     # run unit tests
./gradlew run      # launch RuneLite with the plugin loaded
```

### DIFF
--- a/plugins/trackpadcamera
+++ b/plugins/trackpadcamera
@@ -1,2 +1,2 @@
-repository=https://github.com/crichmond1989/rl-trackpad-camera-plugin
+repository=https://github.com/crichmond1989/rl-trackpad-camera-plugin.git
 commit=aeaae079e7df55a1df098c7c2ea2f7cfdf5c1cf5
--- a/plugins/trackpadcamera
+++ b/plugins/trackpadcamera
@@ -1,2 +1,2 @@
 repository=https://github.com/crichmond1989/rl-trackpad-camera-plugin.git
-commit=aeaae079e7df55a1df098c7c2ea2f7cfdf5c1cf5
+commit=2a127f9b713fa98f8c6da89e58ec37e73626186e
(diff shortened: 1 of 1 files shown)


crichmond1989:
I see the build failure due to the gradle settings I've been using locally. Pushing up a fix - one moment.

crichmond1989:
Sweet. Build success, but my icon is too detailed 😄 I'll make the adjustment

crichmond1989:
That's the right size, but too big on the resolution. Moving to 48x48 next.

jgaver08:
This is great.  I tried to create something similar to this about a year ago and eventually gave up, so bravo for actually making this.  I have a request, is it possible to swap the control function?  Ideally tilt and rotate would be active the entire time.  Zoom is less frequent.

PR #10862, opened 2026-03-04
Carters Prayer Alert

CarterDurham100:
adds a sound and textbox alert when you are below 10 prayer. my first plugin, so the names of files are weird.


riktenx:
closing because this submission is still broken. if you fix it you can make a new pr

PR #10911, opened 2026-03-06
Enemy attack highlighter

SigurdSmeby:
Highlights NPCs you damage using your own hitsplats (including cannon/0 hits), with optional whitelist, multicombat-only, timer, and range-based untagging. It is hitsplat-driven and does not track or infer who an NPC is targeting.


riktenx:

PR #10919, opened 2026-03-07
Add better-notes

TzJake-Jad:
A notepad plugin with per-line checkboxes, strikethrough on completion, paste support, and a Clear Done button.


TzJake-Jad:
this ever going to be looked at?

Felanbird:
> this ever going to be looked at?

The queue is _extremely_ filled with AI submissions so the time to review new plugins is not great right now

Felanbird:
nit: we don't accept "better" named plugins anymore since they are rarely better due to the constant bugs people have with them, it will need a real name

riktenx:
per above comment this plugin should be renamed

riktenx:
nevermind, it appears you have done so in the actual plugin

PR #10928, opened 2026-03-07
Add osrs-tracker-bridge

tytoowns:
Adds Screen of Knowledge Bridge to the RuneLite Plugin Hub.


tytoowns:
updated the UI for the plugin.

Added a side panel so users can quickly access actions for the plugin.
Before these actions were done in the config/plugin settings section and it was clunky.
Now theres actual buttons in the side panel.

For example "Target currently logged in player" button will switch the device to track the currently logged in player and display their skills/stats.
This is useful if you are switching from your main to another account like an iron.

Also added visible status feedback for things like ping and config push in the side panel so you know if it worked.
added a side panel icon

PR #10962, opened 2026-03-09
Add 420 kc plugin

420kc:
Get a special message on your 420th boss kill.

https://github.com/420kc/420kcp


tylerwgrass:
please add a README

PR #10989, opened 2026-03-10
Update commit reference in gauntlet-map

StickySerum:
Refactor code to use varbits and fix known issues. Added minor settings adjustments


riktenx:
closing because the above has not been addressed. if you fix it you can make a new pr

PR #10990, opened 2026-03-10
Add Grouping Customizer Plugin

skillyourself:
Lets you customize the Grouping tab by hiding activities, changing text colors, and setting a default activity selection.


tylerwgrass (commented):
Can you add a proper README so people know what your plugin does?

skillyourself:
> Can you add a proper README so people know what your plugin does?

Done

PR #10997, opened 2026-03-11
Add custom-object-hider

LuxOG:
Allows the user to right click objects to hide them.


riktenx:

PR #11012, opened 2026-03-12
Add Chat Editor plugin

samstorment:
Edit chat messages with an input that has copy/paste, text selection, cursor movement, etc.


riktenx:
thanks for the submission but I don't see any way this could be implemented without violating the chat rules so I am going to close this

PR #11023, opened 2026-03-12
Add MorphPlugin

MakingStan:
Allows you to replace animation IDS and Menu Texts


riktenx:

MakingStan:
Hey this is unfortunate

PR #11034, opened 2026-03-13
Add custom death sounds plugin repository

Tuomar2:
Plugin adds custom sound effects for player and NPC deaths.
Displays custom text messages when a player or NPC dies.
Enhances death events with more noticeable and entertaining feedback.


riktenx:
you will need to provide attribution for the sound files here such that we can verify they are licenseable

riktenx:
closing because the above has not been addressed

PR #11076, opened 2026-03-15
Add AHKusations plugin

csvernon:
This plugin analyzes pvp fights in an attempt to estimate if either player is cheating/ahking. It does not send the data anywhere or automatically report anyone. It is purely for ease of mind in knowing how likely it is your opponent is cheating and to show others that you are not cheating.

https://github.com/csvernon/ahkusations

image


riktenx:
thanks for the submission but we are not interested in hosting any more heuristic-based "cheating" detectors due to concerns of player harassment

PR #11105, opened 2026-03-16
Update commit hash in bank-tab-names - Bank Tab Names v2: Multi-icon system, overlay rendering, icon editor

Psyda:
This replaces the single-icon sprite editor with a multi-icon system and overhauls how tab overlays are rendered.

Why: The old editor had a persistent issue where sprite IDs shuffled whenever the icon list updated, breaking user configs. The rendering was also clipped to the tab bar boundaries, limiting what users could do with custom icons and offsets. Both problems required a ground-up rethink rather than incremental patches .I held off on shipping incremental fixes because most of the reported problems shared root causes in the rendering and sprite ID systems. Fixing them properly meant reworking both at the same time.

What changed:
Tabs now support multiple icons (items, skills, game sprites, custom PNGs) with per-icon size, offset, z-index, and aspect-ratio-locked resizing. Icons and text render on Bankmain.INFINITE instead of Bankmain.TABS, removing the clipping limitation. Game sprite dimensions are resolved from SpritePixels.getMaxWidth/getMaxHeight rather than the trimmed BufferedImage from SpriteManager.getSprite(), which was causing sprites to render at wrong sizes.

The side panel got a scrollable icon editor, curated sprite category browser, fit toggle for stacking/overlay designs, config presets with import/export (versioned JSON), and panel visibility options.

See the updated README for full feature documentation.

Testing: Tested across all three tab display modes (item, digit, roman numeral), with bank settings page open/closed, with 1-8+ icons per tab, with custom user PNGs, and with various game sprites including non-square and padded sprites. Config migration from old single-icon format verified.

image

image

image

### DIFF
--- a/plugins/bank-tab-names
+++ b/plugins/bank-tab-names
@@ -1,2 +1,2 @@
 repository=https://github.com/Psyda/bank-tab-names.git
-commit=c21369dd1f55e90f6316e25b95b30aa8637dd42a
+commit=f9d94c35ae6078dc1df9e2028926d81501dccaf1
--- a/plugins/bank-tab-names
+++ b/plugins/bank-tab-names
@@ -1,2 +1,2 @@
 repository=https://github.com/Psyda/bank-tab-names.git
-commit=b83653b9a20f381954b73e56aa74887656954bf5
+commit=c21369dd1f55e90f6316e25b95b30aa8637dd42a


LlemonDuck:
use runelite's LinkBrowser instead of awt Desktop

code looks fine (f9d94c35ae6078dc1df9e2028926d81501dccaf1) but you'll need to prove ownership/licensing of any images bundled in with your plugin. it is probably easier to just allow the custom image loading that you already have and don't have any bundled in by default.

riktenx:
closing since the above is unaddressed. if you fix it you can make a new pr

PR #11126, opened 2026-03-18
Add xray

1Defence:
- Mark npcs while hiding their model, similar to enabling blindfold while having npc indicators on, but without having to hide all entities.

https://github.com/user-attachments/assets/cdf2db5c-8f80-441b-b1b4-a8b234b27001


riktenx:

PR #11128, opened 2026-03-18
Add Wildy Slayer World Picker plugin

samjhill:
Ranks OSRS worlds by estimated revenant cave risk for Wilderness Slayer. Heuristic recommendation only; no auto-hop or menu injection. Users can report empty/players/PKers and see a best world plus backups.


tylerwgrass:
There are unresolved build errors that need to be fixed for this to be reviewed

PR #11147, opened 2026-03-19
Add Roster Plugin

Mylosis:
Roster — Account management plugin for organizing multiple OSRS accounts into color-coded categories with drag-and-drop, grid/list views, clipboard import/export, privacy controls, and more


riktenx:
Don't specify gson in build.gradle, just use what's transitively available through runelite.

Mylosis:
> Don't specify gson in build.gradle, just use what's transitively available through runelite.

Hey, this should be fixed now!

PR #11155, opened 2026-03-19
Add fc-stat-checker plugin

WildSwab:
A plugin that checks if members of your current Friends Chat meet a minimum skill level requirement. Only displays members who fail the requirement.


riktenx:
Do not put the manifest file in a subdirectory under plugins.

WildSwab:
> Do not put the manifest file in a subdirectory under plugins.

Fixed, moved the plugin file out of the subdirectory.

riktenx:
You also need to add `build=standard` in runelite-plugin.properties.

WildSwab:
> You also need to add `build=standard` in runelite-plugin.properties.

Gotcha, should be added now! :)

riktenx:
don't use Thread sleep

WildSwab:
> 

should be sorted!

riktenx:
closing since this is still broken, if you fix it make a new pr

PR #11206, opened 2026-03-25
Add BossBank Sync plugin

bornthiswayttv-cpu:
Adds BossBank Sync plugin which allows users to sync their bank, inventory, equipment, and stats to an external service for gear optimization and boss recommendations.


riktenx:
doesn't build

riktenx:
closing because this submission is still broken. if you fix it you can make a new pr

PR #11241, opened 2026-03-27
Create sounds-of-san-andreas

donusgits:
My plugin is heavily based on Odablock sounds, but instead I've changed sounds for Carl Johnson of GTA: San Andreas. All it does it add or change some sounds, such as accepting/declining trades, shooing stray dogs, or getting hit. 

I didn't see if mature language/cussing was against ToS in any of the documentation, but if that's the case, I can easily swap the offending sounds out for one that doesn't have swearing in it, God knows I have plenty to choose from.

Thank you for considering my plugin and I hope everything's all tidy and clean and ready to go.


riktenx:
we cannot host a plugin that includes copyrighted audio from another game

PR #11244, opened 2026-03-27
Add Adventurers Log plugin

runitebar:
I've created an Adventurer's Log plugin that tracks Levels, Quests, Boss KCs and Collection Log entries in a dedicated, OSRS-themed sidebar panel. The plugin is purely local and follows all RuneLite 3rd party client rules. I tried as hard as possible to replicate RS3's adventurer's log, obviously it has some limitations but I think it's about as close as we could get.


riktenx:

PR #11246, opened 2026-03-27
Update commit reference in object-hider plugin

umer-rs:
Re-enable map loader thread calls from gpu callback.


riktenx:
you can resubmit this when it is ready

PR #11249, opened 2026-03-27
Add burot plugin repository information

D4nders:
Broadcasts drops, levels, deaths, and PBs to Discord with a generated chatbox image and custom audio.


riktenx:
From https://github.com/runelite/runelite/wiki/Rejected-or-Rolled-Back-Features:

> Plugins which are of an adult or overtly sexual nature.

* https://github.com/D4nders/burot-runelite/blob/8019f94c3aec0654f14f54ad953cfe03d62aef3f/src/main/java/com/burot/event/QuestEventProcessor.java#L44
* the audio files

PR #11285, opened 2026-03-30
Hide Minimap Hotkey Plugin

jarromie:
Plugin allows you to hide the minimap widgets using a hotkey; depends on the default `Minimap` plugin being enabled. 

Short, simple and sweet; a nice QoL upgrade for those who use that function of the `Minimap` plugin.


riktenx:
we don't really want to accept new plugins that interfere with core plugin configs like this

jarromie:
> we don't really want to accept new plugins that interfere with core plugin configs like this

Even if I make it crystal clear that the plugin works by changing the Minimap plugins config? It seems redundant to duplicate that functionality if not necessary

riktenx:
> Even if I make it crystal clear that the plugin works by changing the Minimap plugins config?

yes (unfortunately the userbase tends to not read and then they come into the discord with issues anyway)

also, when you update your plugin, you will need to add `build=standard` in runelite-plugin.properties otherwise it won't build

riktenx:

PR #11289, opened 2026-03-30
Add contrast plugin.

Motokuva:
Adds a contrast plugin that applies an S-curve filter to the game client's rendered frame, allowing users to increase or decrease screen contrast with a slider (1-200, default 100)


riktenx:

PR #11333, opened 2026-04-02
Add placeholder-prices

Maximuis94:
Adds a tooltip to unowned bank placeholders that shows the item name, GE- and HA prices.
The plugin uses the same configurations as the item prices plugin, effectively extending it to unowned placeholders.
tooltip-example


riktenx:

PR #11337, opened 2026-04-02
Add boss-respawn-notifier plugin repository info

ASong5:
Sends a notification when a boss is about to respawn. Configure which bosses should send notifications and specify a global and/or per-boss lead time that dictates how many seconds before the boss respawns to send the notification


riktenx:

PR #11351, opened 2026-04-03
Add clan-management plugin

WoodyCode23:
Clan management plugin for tracking drops, hiscores, and XP across a clan using Google Sheets as a backend.

  Features:
  - Drop logging with Discord webhook notifications
  - Clan hiscores (boss completion times)
  - XP tracking
  - Admin tools for clan management

### DIFF
--- a/plugins/clan-management
+++ b/plugins/clan-management
@@ -1,2 +1,2 @@
 repository=https://github.com/WoodyCode23/clan-management-plugin.git
-commit=0ebb60afc297379a8c2379b2926592138c7ad308
+commit=27fa0a60fe6f9f8f8c4ef9c0d893bafd854d51d2
--- a/plugins/clan-management
+++ b/plugins/clan-management
@@ -1,2 +1,2 @@
 repository=https://github.com/WoodyCode23/clan-management-plugin.git
-commit=4303b15896c581d37ba9d31d118ba10d154dad76
+commit=df712fccaa26032a7ff771f627ced5491dfcc0fb
--- a/plugins/clan-management
+++ b/plugins/clan-management
@@ -1,2 +1,2 @@
 repository=https://github.com/WoodyCode23/clan-management-plugin.git
-commit=8fdf94a5b415972fbc98449a3eb960ff954befb6
+commit=7e2bea6d33bdac6313047cdec8c71c525d7ad549
--- a/plugins/clan-management
+++ b/plugins/clan-management
@@ -1,2 +1,2 @@
 repository=https://github.com/WoodyCode23/clan-management-plugin.git
-commit=27fa0a60fe6f9f8f8c4ef9c0d893bafd854d51d2
+commit=b29903e7668b00085c2fe0ead8a8c7c5b000a1ed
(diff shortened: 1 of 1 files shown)


riktenx:
you have one or more invalid png files in your repository, please correct them

WoodyCode23:
Fixed — `panel_icon.png` was a JPEG with a `.png` extension; re-encoded it as a valid PNG. Also added `build=standard` to the plugin properties (newer requirement). CI build is passing now. Ready for re-review, thanks!

riktenx:
thread sleep is not allowed

WoodyCode23:
Done — removed the `Thread.sleep`; that path now uses `executor.schedule(..., 500, MILLISECONDS)` on the existing ScheduledExecutorService instead. CI build is passing. Thanks for the reviews!

riktenx:
we are not comfortable hosting a plugin that makes a network call to get a list of URLs to make requests to, as this has significant [security implications](https://portswigger.net/web-security/ssrf)

i understand that you are likely doing this for user convenience sake, since they basically just have to paste some config string in and it works, but this is not a risk we can accept

if a plugin is making network calls, it needs to either be a fixed endpoint or set by the user in all cases. you should be able to rewrite your code to do that

WoodyCode23:
hey, changed it like you asked. the plugin doesn't pull the url from a pasted code anymore. it just hits one fixed url thats baked into the plugin (https://api.solusosrs.com), so it never makes a request to figure out where to send data. the only thing a user types in now is their api key, and thats not a url. let me know if anything else needs changing, thanks

plugins/clan-management
```diff
@@ -0,0 +1,2 @@
+repository=https://github.com/WoodyCode23/clan-management-plugin.git
+commit=58de87c596df3b4964f3108961e972936ee31111
```
riktenx:
```suggestion
commit=58de87c596df3b4964f3108961e972936ee31111
warning=This plugin submits your IP address, RSN, in-game location, and various other account data to a 3rd-party server not controlled or verified by Runelite developers.
```

riktenx:
all of your file i/o needs to be in a plugin-specific subdir within .runelite/, right now you're just putting json files directly in there

riktenx:
? i said this wasn't allowed https://github.com/WoodyCode23/clan-management-plugin/blob/e3f236b5f594a30466159cf69b9d0e9b1137cb3a/src/main/java/com/droplogger/ClanManagementPlugin.java#L351, re-audit your codebase for this please

WoodyCode23:
You're right, sorry I missed that. It's fixed now. The plugin no longer fetches the Discord webhook URL or posts to it; all Discord notifications are sent server-side instead. While re-auditing the whole codebase like you asked, I also pulled out an old dead Google Sheets fallback and a disabled bingo feature that still had third-party calls and a configurable URL in it. The plugin now only ever makes requests to the single hardcoded API URL, nothing fetched and nothing configurable.

One small thing on the warning: the plugin doesn't actually send IP address or in-game location. No IP is logged on the server, and location is only read locally to credit the right team at raid/boss start, never transmitted. I left your warning line exactly as you wrote it, but I'm happy to adjust the wording if you'd like.

riktenx:
it exposes the client's ip address by virtue of making the network request, that's how networking works and it's unavoidable. we require that warning on every plugin that does that. you can remove the location part

PR #11355, opened 2026-04-03
Update compact-orbs to v1.7.5

its-cue:
- overhaul overlay orb creation & handling
- add functional Hp, Prayer, Run & Special orbs to the minimap overlay

bugfixes:
- limit children hiding in setHidden to the Compass menu options

preview:
compact-orbs-mapoverlay

### DIFF
--- a/plugins/compact-orbs
+++ b/plugins/compact-orbs
@@ -1,2 +1,2 @@
 repository=https://github.com/its-cue/compact-orbs.git
-commit=7d9c4bcefa37ff88e2c8e614d6583a3dfbfa7b52
+commit=200a521fa6d91fef0ed6842b7e5753f3de293647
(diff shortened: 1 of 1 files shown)


its-cue:
thanks for the review 😢, for clarity

 1. is the use of `menuAction` in general not allowed, or just the way it was used here (for the data orbs, or all orbs including xp/logout/worldmap)?
 2. would this be acceptable? ```widget.setOnOpListener(Script.TOPLEVEL_SIDEBUTTON_OP, OPINDEX0, 1131, 10), //logout x, set side panel tab to logout```

Scoliox:
> use of [client.menuAction](https://github.com/its-cue/compact-orbs/blob/7d9c4bcefa37ff88e2c8e614d6583a3dfbfa7b52/src/main/java/com/compactorbs/widget/WidgetManager.java#L315) is not allowed, this feature will need to be removed

Seriously lame that this feature isn't allowed when in actual reality all it did was allow for the plugin maker to attach the XP orb and the minimap orb to the minimap. Makes genuinely zero sense.

PR #11372, opened 2026-04-04
Update Glamourer to 1.7.0

jhughes:


jhughes:
* Party Sync is disabled by default.
* The first time a user joins a party, Glamourer will ask if they want to enable it or not, after which it's configurable in the settings.
* Data sent through party service uses a binary format so it's relatively small
* Joining a party / doing a large glamour change / hitting the full resync button all trigger a full glamour update which on average should be a few hundred bytes
* After a full sync, it uses incremental updates to send patches for individual item changes which are under 100 bytes

PR #11373, opened 2026-04-04
Add DPS Threshold Alerts plugin

sonnypb:
Adds a plugin that alerts when damage exceeds a configurable threshold. Useful when alting bosses to help to prevent alts from sniping kills.


riktenx:

PR #11398, opened 2026-04-06
Creating a plugin to display your total level and the maximum total level in the skill tab 

psikoo:
I made a plugin to display your total level out of the max amount of total levels.

Default view:
image

Plugin enabled:
image

Compatible with virtual levels:
image

I dont own an account with levels over 99 to test the compatibility with virtual levels (and the "Fake Stats" plugin is completely broken since sailing released), but it should be compatible with virtual levels as I set the priority on the @Subscribe to -1. 🐈


riktenx:

PR #11399, opened 2026-04-06
Add Expanded bank plugin

VolkezXO:
Allows the bank to be expanded when the chatbox is closed. (Only in Resizable)


iProdigy:
Please move this file to `plugins/expanded-bank`

riktenx:

PR #11401, opened 2026-04-06
Create hotkey-toggle-sidepanels

bopsec:
Allows for settings a hotkey to close sidepanels on fixed modern, ignores the ingame "Side panels can be closed by hotkey" setting

example


ASong5:
> > > > Not a reviewer but I'm curious what purpose this serves? You can already toggle side panels by hitting hotkey of the panel that is already open. Is it just to have a singular toggle button?
> > > 
> > > 
> > > I personally am not a fan of the fact that it closes when I hit the same key twice, this is something seperate for if you don't want that. F-keys will still _only_ open them, so you won't accidentally close them by doubletapping for example.
> > > This was requested by someone else, and I've asked around, and it seems like this is something people would like.
> > 
> > 
> > This is already achievable with this setting within game settings: image
> 
> That is indeed the setting that does the exact thing I said that I didn't want.

You said you did not like that the panel closes when you double press the key. This setting allows you to disable that...

PR #11404, opened 2026-04-07
Updated BA PBs

CCatRS:
Fixed stale executor issue. When the plugin shutdown was being called, the executor was shutdown. But a new one was never created on startup. This ensures a valid executor is always available.


iProdigy:
why not just inject runelite's executor? better yet, you can avoid blocking `Call#execute` calls via `enqueue` + CompletableFuture

iProdigy (approved):

PR #11424, opened 2026-04-08
Add Rune Vault plugin

EhhForge:
Rune Vault is a portfolio tracker for OSRS. The plugin syncs bank contents, GE trades, and item pickups/drops to the Rune Vault app in real time. Auth is via a 6-character link code generated in the app. All data is sent over HTTPS to our own Supabase backend.

Note: the app is currently in pre-launch and not yet on the App Store or Google Play. Happy to provide early access to the team for testing.


riktenx:
* use of Thread sleep is not allowed
* use of Thread interrupt is not allowed

EhhForge:
Thanks for the review @riktenx , both addressed.

`Thread.sleep` and `Thread.currentThread().interrupt()` have been removed from `src/main` (verified: 0 occurrences of each). All retry backoff now goes through a daemon `ScheduledExecutorService` (`retryScheduler.schedule(...)`), which means the runtime no longer parks a thread on failure and no interrupt restoration is needed.

Fix: [370ad41] [https://github.com/EhhForge/RuneVault-on-RuneLite/commit/370ad41]

Plugin-hub commit= bumped to 28b954a465a75c617452ffcc90bb4a6f6bb50ae9. 
Ready for another look.

ldavid432:
You should not be adding an icon in this repository. An `icon.png` can be placed at the root of your plugin's repository

See step 10 on https://github.com/runelite/plugin-hub/blob/master/README.md#creating-new-plugins

EhhForge:
Thanks @ldavid432 , this has been addressed.
The `icon.png` has been moved to the root of the plugin repo.
The icons/ directory has been removed from this PR, so it now only touches plugins/rune-vault.
Plugin-hub `commit=` bumped to `4bc5d547fdb986c298beff1c07649ad388d384d8`. Ready for another look.

plugins/rune-vault
```diff
@@ -0,0 +1,3 @@
+repository=https://github.com/EhhForge/RuneVault-on-RuneLite.git
+commit=05715d5c745fac20a4444ceaf38390846191f4d4
+
```
riktenx:
```suggestion
warning=This plugin submits your IP address, RSN, Grand Exchange trades, inventory pickups and drops, worn equipment, bank contents, and wealth to a 3rd-party server not controlled or verified by Runelite developers.
```

PR #11442, opened 2026-04-09
Add personalized friend sounds

astroferrous:
This plugin lets you specify sounds and overhead text when friends show up near you!

Configuration syntax is 1 line per username in the format: username=sound.wav|overheadText

Here is an example: Astroferrous=huehue.wav|hue hue

The default sound shipped with the plugin is "huehue.wav"; if you want to add custom sounds please add them to the personalized-friend-sounds directory within your .Runelite directory


riktenx:
I don't think we can allow this because it is tantamount to notifying you when a player appears on screen which collides with https://github.com/runelite/runelite/wiki/Rejected-or-Rolled-Back-Features

astroferrous:
> I don't think we can allow this because it is tantamount to notifying you when a player appears on screen which collides with https://github.com/runelite/runelite/wiki/Rejected-or-Rolled-Back-Features
> 
> > Highlighting "Offline" Friends (and generic player highlighting): Not included to respect players with their messaging mode set to Private and to avoid harassing players (see [#951](https://github.com/runelite/runelite/pull/951))

Private off should not trigger the event; should only trigger for an online, added friend. 

If this does not conform to rules then neither does this with ~5k installs:
https://runelite.net/plugin-hub/show/heres-trouble

riktenx:

As long as you're doing the friends check it is fine. Note though that the rejected list is not enforced retroactively and there will be instances of existing plugins that violate items on the list but are "grandfathered"

astroferrous:
> add `build=standard` to runelite-plugin.properties
> 
> As long as you're doing the friends check it is fine. Note though that the rejected list is not enforced retroactively and there will be instances of existing plugins that violate items on the list but are "grandfathered"

Updated; thank you for the clarification and the review. I appreciate your time

PR #11444, opened 2026-04-09
update tasks-tracker

JamesShelton140:
- Add shortest path plugin integration to route mode


pajlada:
Not blocking, but it looks like you're calling clear on shortest path more than you need to

PR #11449, opened 2026-04-10
Create PetScape

DamnLol-GIT:
Absolute beast of a project for me - tested as much as possible, but its hard to find all 70 pets :) have confirmed working for around 60~ of them.
Maya animations and PoH funkyness = tough 2 weeks of learning :)

PetScape, simply multiply your pets in the POH (x2, x3, x5, x10), capped at 350 so low ends pcs dont crash - might increase if anyone asks in the future.
Works in players house, gets center of POH them evenly distributes them to avoid clutter.
Since they are just visual, there is no click options on them - but I added an option to default all original pets in the house to have 'Walk Here' active to avoid miss clicks.
ZoomedInPet
ZoomedOutPet


DamnLol-GIT:
reworking the code to use single path instead of dual path

DamnLol-GIT:
reworked the code, seems much better then before + added option to hide cloned pets when original out of render

DamnLol-GIT:
Sorry for the spam, but just wanted to say this is my final update until post-release. I made some changes and dropped the snapshot method. A few days later and am finally pleased with the results and got rid of all my bugs + also added a family pet follower option.

riktenx:

PR #11453, opened 2026-04-10
Add OSRS MCP plugin

nickbeddows-ctrl:
## OSRS MCP

Exposes RuneLite client data via a local MCP (Model Context Protocol) server, letting any MCP-compatible AI assistant see your stats, gear, inventory and location to give context-aware in-game advice.

> Originally built with Claude AI in mind, but MCP is an open standard — any MCP-compatible assistant can connect to it.

### How it works
- Plugin starts a local HTTP MCP server on `127.0.0.1:8282`
- AI tools connect via `mcp-remote` as a stdio bridge
- Three connection modes: **Local** (same machine), **LAN** (same network), **Cloud relay** (any network, no extra software)

### Tools exposed
- `get_all` — all player data in one call
- `get_player_stats` — skill levels, XP, XP to next level, combat level
- `get_equipment` — equipped items by slot
- `get_inventory` — inventory contents and quantities
- `get_location` — world coordinates and area name

### Privacy
- Server only binds to `127.0.0.1` by default (never exposed externally)
- Per-toggle controls for stats, equipment, inventory, location, and username
- Optional auth token for LAN and cloud relay setups
- Read-only — never sends commands to the game

### Connection modes
- **Local** — same machine, default, no config needed beyond the basic setup
- **LAN** — devices on the same subnet; plugin detects and displays the local IP automatically
- **Cloud relay** — SSH tunnel via serveo.net (SSH is built into macOS and Windows 10+); guided setup flow built into the plugin panel

### Setup
See the plugin panel or README for full setup instructions including Claude Desktop, Cursor, Windsurf, and generic MCP config.

### DIFF
--- a/plugins/claude-osrs
+++ b/plugins/claude-osrs
@@ -1,3 +1,3 @@
 repository=https://github.com/nickbeddows-ctrl/claude-osrs-plugin.git
-commit=8589ee382143057f391f9cc0f401641c559c5117
+commit=be431daea89db5d4d8a504fb214ae28e183d2ea1
 authors=nickbeddows-ctrl
--- a/plugins/claude-osrs
+++ b/plugins/claude-osrs
@@ -1,0 +1,3 @@
+repository=https://github.com/nickbeddows-ctrl/claude-osrs-plugin.git
+commit=79fda2215b9c05a0e044a5caeab0f5152b1ceaf5
+authors=nickbeddows-ctrl
--- a/plugins/claude-osrs
+++ b/plugins/claude-osrs
@@ -1,3 +1,3 @@
 repository=https://github.com/nickbeddows-ctrl/claude-osrs-plugin.git
-commit=ffe1cfcf0ca0777d98887416ded2e556f3e3bc7c
+commit=8589ee382143057f391f9cc0f401641c559c5117
 authors=nickbeddows-ctrl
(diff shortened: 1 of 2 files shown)


raiyni:
What is the point of this cloud relay? I don't think it's a good idea to be allowing plugins to establish their own ssh tunnels and exposing clients to an unknown service like this.

nickbeddows-ctrl:
> What is the point of this cloud relay? I don't think it's a good idea to be allowing plugins to establish their own ssh tunnels and exposing clients to an unknown service like this.

Hey, fair concern, it's entirely opt-in and off by default, users have to explicitly switch to Cloud Relay mode to activate it. It's also clearly labelled in the UI with a guided setup flow that explains it tunnels through serveo.net before the user does anything.

That said, happy to add an explicit warning dialog when the mode is selected so it's crystal clear what's happening. Would that address it?

raiyni:
That doesn't answer my question.

nickbeddows-ctrl:
The use case I built this for: my gaming PC is on ethernet and my Mac is on wifi - same router, but they couldn't talk to each other over LAN. Cloud Relay was the workaround for that, but Tailscale is already implemented as the proper solution for cross device setups (it's in the current commit, covered in the README). Cloud Relay is just there as a no install fallback.

raiyni:
We generally don't allow process execution and generating ssh keys and tunnels for users isn't going to be allowed I'd imagine.

nickbeddows-ctrl:
Fair enough, I'll remove it. Tailscale covers the use case

raiyni:
I meant all uses of process builder. If a user needs to use tailscale, they can just run tailscale.

nickbeddows-ctrl:
Cloud relay is fully removed and the ProcessBuilder fallback in TailscaleService is gone too. Tailscale detection now uses pure Java network interface scanning only. No process execution anywhere in the codebase.

riktenx:
thanks for the submission but we are going to reject this per the following to prevent abuse:

PR #11460, opened 2026-04-11
add extra-wiki-links

iskela45:
Adds right-click wiki links on skills in the skills tab. Link options as of now are level-up table, temporary boosts, quests, and training guides for members, F2P, ironman, and UIM accounts.  

Each link type is individually toggleable in the plugin's settings.  

Kuvakaappaus_20260410_172535


riktenx:

PR #11467, opened 2026-04-11
Add plugin: Dialogue Fonts

theOranguzang:
image

image

Replaces the OSRS bitmap dialogue font (Quill 8) with a configurable system TrueType font via Graphics2D overlay. Covers NPC dialogue, player dialogue, option menus, and sprite/item dialogue.

Configurable font family, size, bold, anti-aliasing, and per-dialogue-type toggles.

Addresses runelite/runelite#11729 (open since 2020).


riktenx:
we don't really allow new plugins to just be named "better x" "improved x" etc., please choose an alternative name that is more descriptive.

theOranguzang:
> we don't really allow new plugins to just be named "better x" "improved x" etc., please choose an alternative name that is more descriptive.

@riktenx Thanks for the review, I've renamed to "Dialogue Fonts", keeping it open for future work if people want to use their own loaded fonts, if they don't like the default Java ones.

PR #11473, opened 2026-04-12
Update deepseatrawling

CarelessEsper:
I would like to take over development on this plugin as the author has not been responsive for longer than 1 month.


CarelessEsper:
Adding for reference: 
- [Collaborator access request](https://github.com/Maewi/deepSeaTrawling/issues/27) (1 week old)
- [Pull request with plugin changes](https://github.com/Maewi/deepSeaTrawling/pull/23) (no activity from plugin author for >1 month)

The latter isn't required by the plugin takeover policy; just sharing it for more context on why this is a transfer request rather than the preferred option of being added as a collaborator.

PR #11475, opened 2026-04-12
Update tasks tracker

JamesShelton140:
## New User Experience
- Add route mode info dialog
- Handle route import/export from global buttons below task list

## Route Mode
- Add "collapse/expand all" sections to route manager
- Add "collapse all except" to section header panels popup menu
- Add premade route browser (fetched from github)
- Add keybind to complete current custom task
- Improve custom item panel look and feel
- Change routes and sections to use `ID` for reference rather than `name`

## New Feature
- Add random task picker button and keybind

### DIFF
--- a/plugins/tasks-tracker
+++ b/plugins/tasks-tracker
@@ -1,3 +1,3 @@
 repository=https://github.com/osrs-reldo/tasks-tracker-plugin.git
-commit=982c78eab7212c5e7c521435e32714588b6bc502
+commit=6d13289c913cf3a8e18e339a6489342d1ba8439d
 authors=tylerthardy,JamesShelton140


pajlada:
@JamesShelton140 
Your `readPremadeRoute` function reads filenames from another API call, could you add a check for allowed characters in the filename in the next change? e.g. `a-zA-Z0-9-_`

You could use the async api (enqueue) on your HTTP requests instead of spinning up a thread when using them. Makes it a bit easier to review the network -> usage flow. It also makes handling your nested requests easier.

The "browse premade routes" currently doesn't work, but I imagine that's a server issue

perterter:
> @JamesShelton140 Your `readPremadeRoute` function reads filenames from another API call, could you add a check for allowed characters in the filename in the next change? e.g. `a-zA-Z0-9-_`
> 
> You could use the async api (enqueue) on your HTTP requests instead of spinning up a thread when using them. Makes it a bit easier to review the network -> usage flow. It also makes handling your nested requests easier.
> 
> The "browse premade routes" currently doesn't work, but I imagine that's a server issue

Thanks for the feedback. Logged it in our issues.

Also fixed the backend issue by merging the data on that side. All good now. Thanks for pointing it out!

PR #11494, opened 2026-04-14
Add Kill Clips plugin

csvernon:
## Kill Clips

Automatically records PvP kills and deaths as video clips.

**Features:**
- Records kills (players you fight who die) and your own deaths
- Combat tracking ensures only fights you're involved in are recorded
- Optional auto-upload to Streamable.com
- Sidebar panel with clickable links to recent uploads
- Configurable resolution (480p/720p/1080p), FPS (15/20/25/30), duration (10-60s)
- Sensitive content protection (login screens, bank PIN blurred)

**Requires:** GPU plugin enabled

**Repository:** https://github.com/csvernon/killClips


riktenx:
The new dependency here basically means that this submission will never be reviewed, as there is nobody around with the time to review new dependencies these days.

csvernon:
> The new dependency here basically means that this submission will never be reviewed, as there is nobody around with the time to review new dependencies these days.

Would it be any better if I reduce JCodec down to 40~50 essential classes directly in my plugin source?

raiyni:
it would be better to use OBS clips

csvernon:
> it would be better to use OBS clips

I've used OBS for years, it's easy to forget to turn it on, even with replay buffer. Not to mention that OBS for me picks the wrong monitor to record half the time.

riktenx:
> > The new dependency here basically means that this submission will never be reviewed, as there is nobody around with the time to review new dependencies these days.
> 
> Would it be any better if I reduce JCodec down to 40~50 essential classes directly in my plugin source?

no

csvernon:
> > > The new dependency here basically means that this submission will never be reviewed, as there is nobody around with the time to review new dependencies these days.
> > 
> > 
> > Would it be any better if I reduce JCodec down to 40~50 essential classes directly in my plugin source?
> 
> no

Alright JCodec removed

plugins/kill-clips
```diff
@@ -0,0 +1,2 @@
+repository=https://github.com/csvernon/killClips.git
+commit=246025c266688c9bbbbd6ba6eb60fbee27358b2c
```
riktenx:
```suggestion
commit=246025c266688c9bbbbd6ba6eb60fbee27358b2c
warning=This plugin submits your IP address and game footage to a 3rd-party server not controlled or verified by Runelite developers.
```

riktenx:
Please add `build=standard` to runelite-plugin.properties.

riktenx:
image

csvernon:
> image

Should be good to go now

PR #11497, opened 2026-04-14
Add bonfire-alarm

Ohaq2108:
Adds a plugin that triggers an alarm when all logs in inventory are burned without waiting for the ending animation.


riktenx:
closing because the above has not been addressed, if you fix it you can make a new pr

PR #11502, opened 2026-04-14
Soft dim rule limit & import/export functionality

EIKOOT:
- add soft dim rule limit
- add a way to import/export dim presets


riktenx:

LlemonDuck:
do not create menu entries that send actions to the server:
> Any addition of new menu entries which cause actions to be sent to the server

however, I think you can just move around, remove, and change the existing actions instead of your current approach

riktenx:
closing because the above is unaddressed. if you fix it you can make a new pr

PR #11514, opened 2026-04-15
Add HeroHud Plugin

Jdogcrane:
Minimalist Hero-like in-game visual indicators, with some QOL to support a subtle, immersive experience.

This is a purely UI/HUD plugin. The goal is to have less intrusive UI with context aware, simple, and satisfying energy management while traveling/questing. _(Health/Prayer/Special attack will be updated/added more down the road)_

There is also an optional setting to display "Effective Stamina," which adds a stamina wheel that scales/layers depending on how light you are and the buffs/armor you wear. This helps players feel a sense of weight and progression as they level agility by visually showing the progress made.

HeroHud

java_hBs4mU07Bw


riktenx:

PR #11535, opened 2026-04-15
Voicescape Update

derfurkan:
I have completly rewritten VoiceScape and want to rerelease it.

Description:
Voicescape is a proximity voicechat plugin for OSRS
Players can enable the plugin and open the Plugin Panel on the side to connect to a Server or change configuration.
By default there is a Server hosted by me for Players to connect to.

The Client sends the Server, over a TCP Socket, a list of player-hashes which are nearby. 
The Server routes audio from the Player to others who are also nearby the Player (Based on the hashlist every Client sends to the Server). The Client double checks that the Audio being received is coming from a Player who is actually nearby.
By default the Client mutes all Speakers and does not connect to a Server.


derfurkan:
I make use of `javax.sound` to capture the Microphone for the voicechat

plugins/voicescape
```diff
@@ -1,3 +1,2 @@
 repository=https://github.com/derfurkan/VoiceScape.git
-commit=d8f1aec188c05157e41b5d9b5372e4876d0c4f5f
-disabled=deleted by author
\ No newline at end of file
+commit=334f0f27242aee111314fd3a452ca3652c5b26e4
```
LlemonDuck:
```suggestion
commit=334f0f27242aee111314fd3a452ca3652c5b26e4
warning=This plugin sends your IP address, player name, and audio from your microphone to a server not controlled or verified by RuneLite.
```

also remove the following from your README. Whether you store it or not can't be verified by us.
> The server does not receive your name, your account details, or your IP beyond what it needs to route audio

LlemonDuck (dismissed):

LlemonDuck:
also this plugin is old enough, has been gone long enough, and since it requires a warning, it would be better if you can submit this under a new glob. Feel free to reuse the VoiceScape name publicly, if you wish, just use a different name under the `plugins/<something>` path, even just something like `plugins/voice-scape-new`.

plugins/voicescape
derfurkan: 👍

derfurkan:
> also this plugin is old enough, has been gone long enough, and since it requires a warning, it would be better if you can submit this under a new glob. Feel free to reuse the VoiceScape name publicly, if you wish, just use a different name under the `plugins/<something>` path, even just something like `plugins/voice-scape-new`.

Alright ive re-named the file to `voice-scape-new` and removed the Readme line, should be fine now i hope

LlemonDuck:
don't [delete the existing voicescape file](https://github.com/runelite/plugin-hub/pull/11535/changes#diff-cb5ba076439bff80a880dde5e5aa5b547ba18f2cb6e82bf4ee02d33d5dcdedc2)

derfurkan:
> don't [delete the existing voicescape file](https://github.com/runelite/plugin-hub/pull/11535/changes#diff-cb5ba076439bff80a880dde5e5aa5b547ba18f2cb6e82bf4ee02d33d5dcdedc2)

Alright
Ive added it back

riktenx:
* use of Thread sleep is not allowed
* use of Thread interrupt is not allowed

derfurkan:
> * use of Thread sleep is not allowed
> 
>     * use of Thread interrupt is not allowed

I switched to scheduled executors which i hope is fine @riktenx

derfurkan:
Bot reviewing 💀

PR #11548, opened 2026-04-16
Update demonic-pacts-highlighter

SolutionsCMD:
Updated plugin with fixes from review feedback:

- Fixed stacked ground item tooltips — now scans all menu entries on a tile instead of only the top entry
- Added per-account task completion tracking using RSProfile configuration
- Added auto-detection of task completions via game chat messages
- Added right-click "Mark Task Complete" / "Unmark Task Complete" menu entries on highlighted NPCs and items
- Added config section for completion: Hide Completed Tasks, Auto-Detect Completion, Show Completed In Tooltip
- Completion data loads on login and clears on logout, so multiple accounts are tracked independently


LlemonDuck:
lower log.infos to log.debug

SolutionsCMD:
> lower log.infos to log.debug

Fixed

PR #11550, opened 2026-04-16
Custom UI anchors

Car-Role:
## Custom UI Anchors
 
Adds user-defined anchor regions to RuneLite: create, resize, and position "boxes" pinned to any corner, edge, or center of the game window. Drag overlays into a box to assign them; they stack (vertical, horizontal, or wrapping) and stay put when the window is resized.
 
Configurable per region: name, constraint (where it pins), alignment (how contents sit inside), stacking direction, position, and size. Sidebar panel lists all regions with full property editing. In-game overlay (hold Alt or open the panel) shows each box with drag/resize handles.
 
### Plugin Hub compliance
 
- No reflection into client internals. `java.lang.reflect.Type` is Gson plumbing against this plugin's own config strings only.
- No deobfuscation, bytecode manipulation, or mixins.
- Only public API used (`Client`, `Overlay`, `OverlayManager`, `ConfigManager`, `ClientThread`, `MouseManager`).


riktenx:
don't use Thread interrupt

Car-Role:
> don't use Thread interrupt

Fixed, thank you for all the reviews you've done!

PR #11551, opened 2026-04-16
Create map-of-alacrity

Sacca-1:
New plugin which adds markers of the available teleports of the map of alacrity to the world map, and triggers the matching menu actions when the markers are clicked.


Felanbird:
use of `client.menuAction` is not allowed

Sacca-1:
> use of `client.menuAction` is not allowed

Understood, I will see if I can still make a useful display-only approach.

riktenx:
closing because this submission remains broken, if you fix it you can make a new pr

PR #11562, opened 2026-04-18
Update tasks-tracker (revert unix fix which actually broke)

perterter:
Sorry about the rapid fire revert. We recently pushed a fix to help visibility of tooltips on linux & mac, but it broke the usability worse.

Going to get a proper fix and test it thoroughly before sending again. Just need to unbreak unix users...

### DIFF
--- a/plugins/tasks-tracker
+++ b/plugins/tasks-tracker
@@ -1,3 +1,3 @@
 repository=https://github.com/osrs-reldo/tasks-tracker-plugin.git
-commit=5b8d99946e4ad4b47e3900171e7b93023d0927e2
-authors=tylerthardy,perterter,JamesShelton140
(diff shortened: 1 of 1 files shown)


plugins/tasks-tracker
```diff
@@ -1,3 +1,3 @@
 repository=https://github.com/osrs-reldo/tasks-tracker-plugin.git
-commit=5b8d99946e4ad4b47e3900171e7b93023d0927e2
-authors=tylerthardy,perterter,JamesShelton140
```
iProdigy, line 3: intentional author removal?

Felanbird: #11531

perterter: Yes, as stated above

PR #11565, opened 2026-04-18
update osrs-tracker

dennisdevulder:
Bumps the pinned commit to ship Vulkan Video Encode (H.264) for client-side clip encoding. When the user's GPU supports it the clip is encoded locally instead of sending raw MJPEG frames to our backend for server-side transcoding, which saves bandwidth and drops a large chunk of server cost.

Frame capture keeps working with the GPU plugin off: if no GL context is available the plugin reads pixels straight from `Client.getBufferProvider()` and feeds them into the same encoder path.

Dependency-wise this needs `org.lwjgl:lwjgl-vulkan:3.3.6`, which RuneLite doesn't ship, so the same PR adds it to the `thirdParty` configuration with its SHA256 pinned. No other plugin currently uses it but it sits next to the existing `lwjgl-opengl` entries and is available for anyone who wants to try hardware video.

Co-authored with Claude Opus 4.7.


dennisdevulder:
Superseeded by: https://github.com/runelite/plugin-hub/pull/12885#issuecomment-4793724434 closing.

PR #11580, opened 2026-04-19
Update bankstanding

mhollink:
- Removed the initial 5 minute delay before marking players as AFK
- Fixed the timer displayed on the debug overlay
- Added a developer-mode chat command to reset the Bankstanding level
- Replaced the `!lvl` command to not overlap/break the existing chat commands.


riktenx:

PR #11581, opened 2026-04-20
Update Flip Smart

mablack01:
## Summary
This release includes five improvements to GE session tracking, bank wealth capture, and flip suggestion accuracy:

- **Stuck sell queue**: Sell offers that completed but were not properly cleared could block subsequent sells in the same session. The queue is now correctly drained on offer completion.
- **Stale GE slot timers**: Slot timers were not being reset when a slot was emptied, causing incorrect time-remaining displays on fresh offers placed in reused slots.
- **Stale flip cleanup race condition**: A race condition during plugin restart could cause active flips to be incorrectly cleaned up while the GE history was still loading.
- **Bank gear value tracking**: Bank snapshots now capture equipped gear and item-level inventory separately, allowing the backend to price these server-side for accurate total wealth calculation.
- **Offline sell duplication**: Sell offers completed while offline were being incorrectly marked as estimates when synced, causing duplicate entries in flip tracking.
- **GE-only flip suggestions**: Flip suggestions are now only displayed when the player has the Grand Exchange interface open, preventing suggestions from appearing in irrelevant contexts.

### DIFF
--- a/plugins/flip-smart
+++ b/plugins/flip-smart
@@ -1,4 +1,4 @@
 repository=https://github.com/Flip-Smart/flip-smart-runelite-plugin.git
-commit=eeb5286b06b965059cd2c5d13231db8de7f7db1f
+commit=112c61addf20fb9c89045a0f096d3ea0f6dde10c
 authors=mablack01
 warning=This plugin submits your IP address, account name, grand exchange transactions, and information about your account wealth to a 3rd party server that is not controlled or verified by RuneLite developers.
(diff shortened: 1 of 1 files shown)


mablack01:
Added a few extra commits to this which should still keep it `size-m`. There were some bugs that needed to get in and some changes to our API requests to better handle scaling of our API requests.

LlemonDuck:
lower log.info to log.debug

mablack01:
> lower log.info to log.debug

@LlemonDuck just updated the new (and old) `log.info` logs to `log.debug`

PR #11597, opened 2026-04-21
leaguessync: update commit — skill/quest sync, seasonal world guard

RPBTwisted:
Updates the `leaguessync` plugin to the latest commit.

Changes in this release:
- Plugin now only syncs when logged into a seasonal (leagues) world — prevents main-game profile data being submitted when hopping to a non-leagues world
- Sync payload now includes real skill levels and quest completion states alongside league tasks, to support a "Requirements Met" filter on the tracker website


RPBTwisted:
Should just be minor. 

Grabs player levels and quest completions to allow better filtering on the website the plugin is linked to

iProdigy (approved):

PR #11604, opened 2026-04-21
Add Divine Dialogue

GarnetDivine:
Adds Divine Dialogue, a plugin that mirrors NPC and player dialogue into a configurable overlay window separate from the OSRS chatbox. Intended primarily for Let's Play recordings and streams where the chatbox is typically obscured, but also useful as an accessibility aid for larger, repositionable dialogue text.

Read-only: does not modify game state, automate gameplay, inject menu entries, intercept input, or make network calls. Uses only standard RuneLite overlay and widget APIs.

Second entry in the Divine family, following Divine: Skies.

### DIFF
--- a/plugins/divine-dialogue
+++ b/plugins/divine-dialogue
@@ -1,0 +1,2 @@
+repository=https://github.com/GarnetDivine/divine-dialogue.git
+commit=7a50e689d9e380d92ebf110641874b7a4e31e006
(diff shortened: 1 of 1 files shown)


riktenx:

LlemonDuck:
you have an extra plugin file being added here

PR #11608, opened 2026-04-22
Create veng-timer

Divine-Dream:
this plugin displays a visual countdown of player's veng timers.
I made sure to disable this for pvp worlds/areas

### DIFF
--- a/plugins/veng-timer
+++ b/plugins/veng-timer
@@ -1,0 +1,2 @@
+repository=https://github.com/Divine-Dream/veng-timer
+commit=9d24d9ee7590ca62ea95484e52e3c572072eaef8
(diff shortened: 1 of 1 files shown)


riktenx:
~~an "is in pvp" check like this is not really enforceable and I don't think this would cover everything as written. (e.g. safe minigames w/ pvp).~~

riktenx:

Divine-Dream:
I just noticed this comment, was waiting a bit and forgot I even made this.

Would I need to make this plugin fail in areas like pvp minigames and stuff? tbh I  only added those checks because I assumed it would avoid a revision in case it was against the rules.

Let me know what you need from me! I'd love to have this plugin

Divine-Dream:
do you have any advice on how I can just make this exclusive to party members only? that way the pvp checks aren't even a real concern, since you'd have to be in a party to even see the timer anyway?

riktenx:
This should be totally fine if it's only showing for party members, please make that change.

riktenx:
you can open a new pr if you get around to fixing this

PR #11612, opened 2026-04-22
Updated trade-highlighter to v1.0.2

asundr:
- Added time-out to detect stalled load operations and retry them
- Executor pools are created and shutdown along with plugin activation


riktenx:

asundr:
Ah neat, I'll give that shot!

PR #11614, opened 2026-04-23
Add Persistent Bank

rsvaultpudgy-create:
Adds Persistent Bank — a RuneLite plugin that snapshots bank / inventory / equipment / seed vault / Grand Exchange state to per-account JSON files on disk with a rate-limited cadence (15-minute cooldown by default).

### Features
- Side panel with a running wealth total across every Jagex account logged in on this machine, sorted by value. Values are precomputed from ItemManager GE prices at snapshot time and cached into the JSON, so the panel has no live cost on open.
- Per-container toggles (bank / inventory / equipment / seed vault / GE).
- Configurable snapshot cooldown, flush-on-logout, and snapshot folder path.

### Compliance
Strictly observational. No input synthesis, no packet modification, no modifications to client state, no network I/O. The plugin reads from `Client` and writes JSON to disk; that's the full surface area.

### Notes
The plugin's primary UI is the wealth panel — it works standalone and doesn't depend on any external tool. The snapshot JSON is a separate capability: its schema is documented in the README so anyone can write a reader. I'm building one such reader myself (RSVault) for aggregated cross-account views, but that's just one consumer among many possible ones.

Repo: https://github.com/rsvaultpudgy-create/persistent-bank-plugin


rsvaultpudgy-create:
Pushed d5dc03d to rename the manifest from persistent-bank.yml → persistent-bank so it matches the plugin-hub naming convention. Should resolve the build failure from the first CI run.

riktenx:
your manifest file is invalid, https://github.com/runelite/plugin-hub#submitting-a-plugin

plugins/persistent-bank
```diff
@@ -0,0 +1,2 @@
+repository= https://github.com/rsvaultpudgy-create/persistent-bank-plugin.git
+commit= 9de14e2e7181055c4058ab5078a34976f6adacc7
```
riktenx:
```suggestion
repository=https://github.com/rsvaultpudgy-create/persistent-bank-plugin.git
commit=9de14e2e7181055c4058ab5078a34976f6adacc7
```

riktenx:
you need to set `build=standard` in runelite-plugin.properties

PR #11621, opened 2026-04-23
Update Shortest Path to 1.19.0

FIrgolitsch:
The second of the large releases. Large rework of internal workings to fix some major problems.

All changes are listed under the 1.19.0 milestone:
https://github.com/Skretzo/shortest-path/milestone/1?closed=1

A large part of the changes is in the test package due to some testing validation tools we built (https://skretzo.github.io/shortest-path/). It shouldn't pollute the final plugin package.

After this one is approved, we will shortly push 1.19.1 as well, which will be just a reformat of all code files to align with the RuneLite code conventions and fix final linting issues.

### DIFF
--- a/plugins/shortest-path
+++ b/plugins/shortest-path
@@ -1,3 +1,3 @@
 repository=https://github.com/Skretzo/shortest-path.git
-commit=9435629cb73afcf1dc0b9c88d25b7859219b0114
+commit=0fa115ea81ebfa708528a96f2a401dfaf2d1f111
 authors=Skretzo,FIrgolitsch,wvanderp
--- a/plugins/shortest-path
+++ b/plugins/shortest-path
@@ -1,3 +1,3 @@
 repository=https://github.com/Skretzo/shortest-path.git
-commit=a52c69e72b03373b44765b35c8d424213cc57395
+commit=e5607c1c078124ff8b5313252f89dd99694322c0
 authors=Skretzo,FIrgolitsch,wvanderp
--- a/plugins/shortest-path
+++ b/plugins/shortest-path
@@ -1,3 +1,3 @@
 repository=https://github.com/Skretzo/shortest-path.git
-commit=0c0f800a04de02e23db086ce857200e8147c6a6f
+commit=34b02ff1ebf5fc9f5c508e19f75c45b8d9423282
 authors=Skretzo,FIrgolitsch,wvanderp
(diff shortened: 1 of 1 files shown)


iProdigy:
>  PluginBuildException: shortest-path: All gradle files must wrap at 120 characters or less

FIrgolitsch:
>PluginBuildException: shortest-path: All gradle files must wrap at 120 characters or less

Got it! Fixed the Gradle files, also split them out a bit to not pollute the main one as much with our own development utilities. For future reference, is this a check that came from your side, or did the checkstyle already get enforced? I wanted initially to push it without any of those checks failing the build. The commit I pointed to didn't have it included yet.

pajlada:
@FIrgolitsch still not fixed

FIrgolitsch:
My bad, the comment lines weren't caught. Sorry about that, it should be good now. I'm seeing it fails now on the me.champeau.jmh plugin. We use it for performance profiling of our pathfinding algorithm. Is it not allowed? It's a test-only dependency.

pajlada:
not a gradle expert, but i think you'd need to load it in in a different way than you're doing now
preferably your gradle files doesn't really contain a bunch of things that's not relevant for the runelite build, but i'm not sure if there's a reasonable way to split these things up

FIrgolitsch:
Thanks for the response. I'm no gradle expert either (hence the issues...), but I'll see if I can't get it split up somehow.

FIrgolitsch:
Dropped the dependency altogether. Seemed like a cleaner way to do things

pajlada:
this now looks reviewable :+1:

FIrgolitsch:
Awesome! Hopefully the next checks will go a bit smoother haha

riktenx:

FIrgolitsch:
I don't think we use anything out of the ordinary. Added the directive to our plugin.

LlemonDuck:
> I don't think we use anything out of the ordinary

you have hundreds of lines of custom gradle stuff, this is _very out of the ordinary_.

Even though these are only intended to be run manually, it would be a lot better if you just implemented them as tests under a different run configuration or something and use system properties instead of gradle properties. We'd very much prefer if you keep the gradle stuff as simple as possible.

Take a look at how I did something related in the Sailing plugin: [create a new source set](https://github.com/LlemonDuck/sailing/blob/main/build.gradle#L20-L25), and then [place the files in there](https://github.com/LlemonDuck/sailing/blob/main/src/generateSeaChartTasks/java/com/duckblade/osrs/sailing/GenerateSeaChartTasks.java). This also excludes them from the release jar and the normal test path, so you don't need to do anything special to exclude it from those.

PR #11625, opened 2026-04-24
Add GerWar PvP plugin

ascmm:
All-in-one PvP companion for the GerWar clan.

- Fight tracking with off-pray %, expected damage, KO chance (from pvp-performance-tracker)
- K/D balance with loot-key validated PK kills (PENDING ? VALIDATED ? UNCLAIMED)
- Group loot splits with chat detection and settlement (from split-session-tracker)
- Auto-handoff: validated loot-key GP automatically added to active split session

License: BSD-2-Clause (upstream attributions in LICENSE + NOTICE at repo root).


riktenx:
use of reflection is not allowed

ascmm:
Thank you for your Feedback I will work on it and come back!

riktenx:

PR #11634, opened 2026-04-25
Update Visibility Plugin

Endrit-alt:
- Fixed performance bug with isActive() being called by shouldDraw() way too many times.
- Fixed performance bug reallocating HashMaps every frame


riktenx:

iProdigy (approved):

PR #11635, opened 2026-04-25
Add Flask Fervour Cooldown

ParadoxGods:
Adds Flask Fervour Cooldown, a Leagues Flask of Fervour cooldown overlay/highlight plugin.

Plugin repository:
https://github.com/ParadoxGods/flask-fervour-cooldown

Plugin commit:
db0a2c2da7290dcaec048b1907e9db7ce67f08db

The plugin uses standard build metadata, includes a BSD 2-Clause license, and has a 48×48 icon.png generated from the provided new_icon.png asset.

### DIFF
--- a/plugins/flask-fervour-cooldown
+++ b/plugins/flask-fervour-cooldown
@@ -1,2 +1,2 @@
 repository=https://github.com/ParadoxGods/flask-fervour-cooldown.git
-commit=d954af3318a3e942c8f1a82e439e8b3aa4f044d1
+commit=ca954e45639faec1b5d51ccfc05182a7039b8e85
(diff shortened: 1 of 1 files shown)


bopsec:
you could just use the cooldown varbit
and this functionality is already in the base client

ParadoxGods:
> you could just use the cooldown varbit and this functionality is already in the base client

yeah the overlay is in seconds though and also above the chat and can't be moved, it's really the only reason I started this, I died at a boss because I thought the cooldown was up but I just didn't see it lol

I will look at the cooldown varbit though, when I last grabbed it it was reporting improperly and I think I had the wrong one, to be fair I have no clue what im doing

PR #11655, opened 2026-04-27
Mining Stats

alisendjsc-crypto:
Mining Stats: replaces the standard Mining plugin's drifting per-hour rate with one computed over a rolling window of active mining time only. AFK gaps are excised from the denominator; per-ore breakdown shows the player's current activity.

The standard plugin uses `total_ores / total_session_time`, so the rate sags as the session lengthens, even at constant pace; this fixes that.

Source: https://github.com/alisendjsc-crypto/osrs-mining-stats (tag v0.1.0, commit 1f273acf59b5971c006fb068dccbd94b33651671)

Pure passive overlay — no automation, no input simulation, no clicks.


riktenx:
remove the blank line at the end of runelite-plugin.properties

PR #11665, opened 2026-04-27
Update camera-points

JoeBosch:


pajlada:
@JoeBosch You must set the `build` parameter in the `runelite-plugin.properties` file now - see https://github.com/runelite/plugin-hub/blob/master/templateplugin/runelite-plugin.properties#L7 from the template plugin.

pajlada (approved):

PR #11678, opened 2026-04-28
Update custom-object-hider

LuxOG:
Disallow hiding bloat room pillars and chamber


Alexsuperfly:
please add the Sotetseg maze tiles to the banned list as well

Alexsuperfly:
you have to block the hard mode tiles too.  A region check works too, like [how the ground object hider has theirs](https://github.com/dey0/pluginhub-plugins/blob/6a0925e46612535768eb862975278287f4203ffe/src/main/java/com/objecthider/ObjectHiderRenderCallback.java#L35)

LuxOG:
Added HMT tiles. I'd like to avoid banning the region altogether as I want to preserve the ability to hide other things as well, such as the walls behind sote

PR #11681, opened 2026-04-28
Update toa-invocation-display to 2351fc49

noahnoah711:
Adds a Side panel display mode option (alternative to the existing overlay) and fixes a bug where some party members can't see invocations.


noahnoah711:
Just as a heads up, I have build=standard in my runelite-plugin.properties file.

PR #11709, opened 2026-04-29
Update Twitch Live Loadout to v2.3.4

pepijnverburg:
fix: fixed minor issue where Leagues VI pact node IDs are being collected indefinitely when re-opening the interface. Missed a proper data cleanup.


pajlada:
@pepijnverburg You need to specify the build method in your `runelite-plugin.properties` (e.g. `build=standard`) - and you can specify your version in that file too. See https://github.com/runelite/plugin-hub/blob/master/templateplugin/runelite-plugin.properties from the template, and https://github.com/pajlads/DinkPlugin/blob/master/runelite-plugin.properties from how we applied it in Dink as an example

PR #11728, opened 2026-05-01
update pvmhud plugin

mayvqt:
Recloned the base example plugin properly and replaced with my plugin


tylerwgrass:
Curious why you did this?

mayvqt:
when i was doing prs i was having issues with the commit id for some reason, this seems to have fixed it, thought better to clone properly now than later

edit: my original repo i just downloaded the example repo instead of forking

PR #11738, opened 2026-05-02
MintsSpellWarning

EvilNegi:
A Simple Mod that, checks if your holding a weapon with magic attack, if you are, it checks if its a power staff, then if you do not have an autocast spell selected It flashes on the screen in a moveable box (This was created for a rare case for my husband when dying with gravestones)


pajlada:
The plugin manifest file (`plugins/MintsSpellWarning`) may not contain uppercase characters - rename it to something like `mints-spell-warning` or `mintsspellwarning`

PR #11740, opened 2026-05-02
Adds `Fiskmans` as contrib

Nerdpuff:


Fiskmans:
Closes https://github.com/Nerdpuff/max-skill-trim/issues/29

pajlada:
Thanks for the update - @Nerdpuff or @Fiskmans you'll need to update your `runelite-plugin.properties` file to include the `build=standard` field, then update the commit hash in this PR for things to build.

PR #11749, opened 2026-05-03
Update Tob Utilities to 1.0.7

NCG-RS:
Changed way of hiding floor to make it work regardless of use of gpu/gpu legacy
Added option for hiding other player 2d
Removed the forced perspective change when going through the door from Xarpus to Verzik
Fixed use of -1 to denote top level worldview to use the correct WorldView.TOPLEVEL constant


LlemonDuck:
just fyi, there's a much easier way to test with injects instead of writing your own reflection stuff, e.g. https://github.com/LlemonDuck/tombs-of-amascut/blob/e3e09bff0db9e39173271e6fa17efa5174160083/src/test/java/com/duckblade/osrs/toa/features/pointstracker/PointsTrackerTest.java#L37

PR #11750, opened 2026-05-03
Update Flip Smart

mablack01:
## Changes

- Resolved problems with active offers getting stuck when you haven't logged on for a few days.
- Improvements with offline syncing and preparation for the future Grand Exchange history update.
- A few fixes with trades being dismissed too quickly, letting the backend handle that.
- Added a message to display if there are no available trades for a given cash stack amount so that it is more clear, rather than delivering an error.
- Offline sells are now recorded by using the Grand Exchange history, you will be prompted to open the interface when there were offline trades recorded.


mablack01:
Bumped the commit to include a QOL feature for backfilling offline sells using the Grand Exchange history interface

PR #11767, opened 2026-05-05
Daily agility daily laps marks logging

DrieVeertien:
## Daily Agility

Track daily rooftop agility lap goals and Marks of Grace collection. Includes a lap timer, ETA, rolling averages, and a persistent session log with calendar view.

### Features

- **Daily Lap Goal** — Set a target number of rooftop laps per day. Counter resets automatically at midnight based on system clock.
- **Daily Marks of Grace Goal** — Set a target number of Marks of Grace to collect per day. Pickups are detected from inventory changes.
- **Lap Timer** — Tracks last lap duration and a rolling average over the last 100 laps per course, persisted across sessions.
- **ETA** — Estimates time remaining to complete the daily lap goal based on average lap time.
- **Overlay** — Configurable in-game overlay showing laps left, marks left, last lap time, average lap time, and ETA. Each line is individually toggleable. Supports four display modes: always, when goal is incomplete, when actively running rooftops, or never.
- **Session Log Panel** — Sidebar panel with a calendar view. Days with recorded data are highlighted. Clicking a day filters the log to that date. Log is stored indefinitely per course per day using ConfigManager.

### Technical notes

- Lap detection uses the game message `Your <Course> Rooftop lap count is: <count>`, which fires on all standard rooftop courses.
- Mark of Grace pickup detection ignores drops by the player via `MenuOptionClicked` to prevent false counts.
- Last known course is persisted across sessions so marks picked up before the first lap of a session are correctly attributed.
- Inventory baseline is established via a retrying `clientThread.invokeLater` call to ensure marks present on login are not double-counted.

### DIFF
--- a/plugins/daily-agility
+++ b/plugins/daily-agility
@@ -1,2 +1,2 @@
 repository=https://github.com/DrieVeertien/DailyAgility.git
-commit=b21d319d28d8411214dcb86113fd04304bf3dae5
+commit=31b40cb27a07e2e308f68ad025676a73fc0a44c3
--- a/plugins/daily-agility
+++ b/plugins/daily-agility
@@ -1,2 +1,2 @@
 repository=https://github.com/DrieVeertien/DailyAgility.git
-commit=31b40cb27a07e2e308f68ad025676a73fc0a44c3
+commit=eeccb618555b7480eb0589e6f200e70575a6ebef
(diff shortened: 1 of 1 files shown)


DrieVeertien:
The file size should be: 30.255 bytes

I dont understand where the 5,678,400 bytes comes from after the fix

Felanbird:
Java blows up an image to it's max size when it loads it into memory, (length * width * 4)

DrieVeertien:
> Java blows up an image to it's max size when it loads it into memory, (length * width * 4)

Thanks, I got a smaller image. Now it passed.

@Felanbird

PR #11769, opened 2026-05-05
Add `Random Timer` plugin

matthewxmar10:
A RuneLite plugin that fires a hidden alarm at a random time between two intervals you set. You set the range, but never know when it will go off. 

Useful as an awareness check, an anti-AFK prompt, or a general activity reminder. I am personally planning to use this for a Leagues-based series when available.

When you start the timer, the plugin picks a random moment between your minimum and maximum time settings. When that moment arrives:
- A sound plays - either your own custom .wav file or the built-in alarm beep
- A message appears in your chatbox
- The timer resets and waits for you to start it again

The sidebar panel shows how long the timer has been running, and a small HUD infobox displays the elapsed time on screen, but the target time is always hidden.


riktenx:
use of Thread.sleep is not allowed

riktenx:

PR #11770, opened 2026-05-05
notable-players

pairofcrocs:
Adds Notable Players, a plugin that highlights notable players (curated content creators and unique accounts, plus auto-detected Jagex Mods and a user-managed custom list) when they appear near you in-game. Each match gets a configurable model outline, tile, label, minimap text, and an info-panel entry with the reason they're notable.

Repo: https://github.com/pairofcrocs/notable-players-osrs

The bundled list is fetched from the plugin's GitHub repository on startup and falls back to the JSON shipped in the JAR if the request fails. Users can disable the network fetch in the plugin config.

### DIFF
--- a/plugins/notable-players
+++ b/plugins/notable-players
@@ -1,0 +1,2 @@
+repository=https://github.com/pairofcrocs/notable-players-osrs.git
+commit=79a2ceb775488b0de09e2858c18b9cc893a42195


tylerwgrass:
Generic player highlighting like this is not allowed:

Felanbird:
Hi, we're not going to be interested in hosting this plugin mainly because we have no way of verifying that these people want to basically have an `important person ------>` sign applied to them.

While I'm sure many of them are used to being recognized, that natural interaction is very different than allowing a plugin to indicate that they are noteworthy.

Thanks

PR #11776, opened 2026-05-05
Update Creator's Kit to v2.1.19

ScreteMonge:
v2.1.19 - Lighting Code Cleanup
- CustomModelComp no longer keeps ids
- Lighting code has been cleaned up. CustomModelComp no longer saves it. Lighting of models will now be determined solely by CustomLighting instead of LightingStyle, with translation between the two as required
- Bugfix: it should no longer be possible to set orientation above 2048 manually


pajlada:
FYI: Your plugin will show up with the commit hash rather than your version - for your version to show up in the sidepanel you need to set the `version=...` option in your `runelite-plugin.properties` file.

PR #11791, opened 2026-05-06
Update Bone Mining Notifier commit hash

staticvoid07:
Updates the commit hash to fix the author name in runelite-plugin.properties (was real name, now GitHub username).


staticvoid07:
Forgot to change the author name when I initially made the plugin, will be correct if/when this is merged.

PR #11792, opened 2026-05-06
Add cosmetic-cannon plugin

FigmentsOSRS:
Replaces dwarf multicannon cannonball projectiles with custom visuals, configurable per firing direction.

- Supports standard cannonball (ID 53), granite cannonball (ID 1443), and cannon reskin variant (ID 2018)
- - Four swap modes: Random, Rainbow Cycle, Custom (per direction preset), and Experimental (raw ID entry)
- - Direction detection via source/target WorldPoint vector math (atan2), 8 compass directions
- - 70+ preset projectile visuals sourced from boss projectiles and other in-game graphics
- - Swap technique based on the Chinbompa method

### DIFF
--- a/plugins/cosmetic-cannon
+++ b/plugins/cosmetic-cannon
@@ -1,0 +1,2 @@
+repository=https://github.com/FigmentsOSRS/Cosmetic-Cannon-Plugin.git
+commit=be198b540bde46f84e16e0fedeec4ff5c1d864c1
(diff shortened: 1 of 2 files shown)


riktenx:
arbitrary projectile IDs on the cannon is not allowed, from https://github.com/runelite/runelite/wiki/Rejected-or-Rolled-Back-Features

presets are perfectly fine

FigmentsOSRS:
Experimental capability to manually input IDs removed. Only presets available.

PR #11803, opened 2026-05-07
Update to xp-to-level-notification

R-Dson:
This update fixes an annoying bug that was spamming players with "0 XP left"
notifications when hitting 200m XP. I also cleaned up the backend logic, updated the config settings and bumped the gradle build version.


LlemonDuck:
if you have renamed your github account, you should also update the repository pointer in the manifest file

PR #11822, opened 2026-05-08
Create local-data-exporter

GoblinTek:


riktenx:
doesn't build

riktenx:
your file i/o needs to happen within .runelite, ideally within a plugin-specific subdirectory, rather than being used-configured as you have written it

GoblinTek:
@riktenx Removed the user-configurable export path and changed exports to always use `.runelite/local-data-exporter`. Also updated the README accordingly.

PR #11836, opened 2026-05-08
Update skillingloottracker

RyUkY-Hub:


RyUkY-Hub:
Added a plugin icon and resolved a bug in the fishing tracker. The tracker now correctly identifies and displays individual fish types and their respective counts, rather than grouping all fish as a single type.

PR #11841, opened 2026-05-09
update attacktimer

Lexer747:
Sorry for the large diff but this should make all future diffs smaller, the only change is https://github.com/ngraves95/attacktimer/commit/506d859c86c3a3348bf9ce0c86bfcf77c787255c which enables linting and a format checker for the plugin. There's no functional changes to the plugin at all.


pajlada (approved):
Because it was only formatting / variable renaming, this PR was easy to go through. Diff size was fine :+1:

PR #11857, opened 2026-05-10
Add Fire at will Timer plugin

theshane0314:
Adds the **Fire at will Timer** plugin.

**What it does:** Tracks the in-game `Fire at will!` crew call during boat combat (Sailing). Starts a 10 minute countdown when the call is issued, displays it as a small tile or large tile, color-shifts at configurable warning / alert thresholds, and notifies via RuneLite's standard Notifier on warning, alert, or expiry. Detects manual cancellation via `Attack my targets!` / `Await further orders!` calls.

**Repo:** https://github.com/theshane0314/fire-at-will-plugin
**Author:** Shalysa

### DIFF
--- a/plugins/fire-at-will-timer
+++ b/plugins/fire-at-will-timer
@@ -1,3 +1,3 @@
 repository=https://github.com/theshane0314/fire-at-will-plugin.git
-commit=643753bd7721693a63dd5d42a99c542922a8675b
+commit=7167bccc17e64d5729935c12aed48edb6c3d1991
 authors=Shalysa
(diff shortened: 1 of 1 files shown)


theshane0314:
Pushed a fix and re-pinned the manifest: repackaged from `net.runelite.client.plugins.fireatwill` to `com.fireatwill` to satisfy the "use of net.runelite package namespace is not allowed" packager check. New commit: theshane0314/fire-at-will-plugin@ebfd0d6fb306e97b98e03baa522352a783444e45. Workflow is waiting on maintainer approval to rerun.

theshane0314:
@tylerwgrass thanks for the re-run, but that one ran against the old head SHA `ea9d281` which still has the `net.runelite` packaging issue. The fix is on the newer head `63a773a` (run [25933948088](https://github.com/runelite/plugin-hub/actions/runs/25933948088)), which is sitting in `action_required` and was never approved. Could you approve that run instead? It points at theshane0314/fire-at-will-plugin@ebfd0d6 which has the plugin repackaged to `com.fireatwill`.

PR #11864, opened 2026-05-11
Add Wiki Chat

flowercodes:
Adds **Wiki Chat**: an OSRS wiki chatbot in a RuneLite side panel.

## What it does

- Side panel where users can ask any OSRS question and get streamed, cited answers from the OSRS wiki
- Right-click "Ask AI" on items, NPCs, or objects with Examine — pre-fills a question about the target
- Hybrid retrieval: vector search over ~2,000 indexed wiki pages + live MediaWiki search on every query, so coverage extends to the long tail
- Bring-your-own-key for Gemini / OpenAI / Anthropic (no shared quotas)

## Privacy

Data submitted to the 3rd-party server (Cloudflare Worker): user's IP and the text of typed questions. No game state, no account info. Privacy disclosure is in the README, a warning is configured on the backend URL config item, and the manifest includes the standard warning string.

## Links

- Repo: https://github.com/flowercodes/osrs-wiki-chat-plugin
- Commit: 3fe7bffc23fba0ef1a7e6946f38ed417d0ebe2af

### DIFF
--- a/plugins/wiki-chat
+++ b/plugins/wiki-chat
@@ -1,3 +1,3 @@
 repository=https://github.com/flowercodes/osrs-wiki-chat-plugin.git
-commit=3fe7bffc23fba0ef1a7e6946f38ed417d0ebe2af
+commit=798cbea9a7f3e071218162b50960f8010aadee69
 warning=This plugin submits your IP address and the text of your typed questions to a 3rd-party server not controlled or verified by RuneLite developers.
--- a/plugins/wiki-chat
+++ b/plugins/wiki-chat
@@ -1,0 +1,3 @@
+repository=https://github.com/flowercodes/osrs-wiki-chat-plugin.git
+commit=3fe7bffc23fba0ef1a7e6946f38ed417d0ebe2af
+warning=This plugin submits your IP address and the text of your typed questions to a 3rd-party server not controlled or verified by RuneLite developers.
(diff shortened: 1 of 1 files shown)


flowercodes:
Bumped manifest to commit `798cbea`. Changes since the initial submission:

**Architectural change — API keys no longer transit the backend:**

- Worker reduced to a RAG-only `/context` endpoint that returns wiki chunks as JSON (`{chunks: [{title, url, historyUrl, text}, ...]}`).
- LLM calls moved from the Worker to the plugin. Three Java clients (`GeminiClient`, `OpenAIClient`, `AnthropicClient`) call the provider APIs directly using OkHttp with SSE streaming, with the user's API key flowing only between the user's machine and their chosen LLM provider.
- `WikiChatClient` is now a two-stage flow: fetch context from the backend, then call the LLM directly.

**OSRS Wiki Generative AI policy compliance:**

- Each cited source now includes a `(hist)` link to `?action=history` per the policy's stated example of acceptable single-page attribution.
- A `CC BY-NC-SA 3.0 · OSRS Wiki contributors` footer appears under every AI response, making share-alike visible on each derivative output rather than only in docs.
- Welcome message and README now explicitly declare CC BY-NC-SA 3.0 inheritance, non-commercial use, and link to the wiki's generative AI policy.
- README clarifies that the plugin retrieves content at query time and never trains on wiki content.
- `User-Agent` strings on every MediaWiki API call updated to point to the real plugin repo.
- Plugin description corrected (no longer claims the assistant is "trained on the wiki" — it isn't).

riktenx:
based on the wiki discord discussion it sounds like you are proceeding with an alternative solution so i am closing this

PR #11875, opened 2026-05-11
Update 07flip to 8917dbb

UserD40:
Update 07flip plugin to commit 8917dbb. Major UX overhaul: Item insights tab, My Trades redesign with progress-bar rows, guided GE buy/sell auto-fill flows, GE overlay with cached insights + 24h chart, incremental partial-fill recording with merge-by-offerInstanceId, GE tax accounting in profit calculations, and premium gating for rec prices + frozen-sell feature.

### DIFF
--- a/plugins/07flip
+++ b/plugins/07flip
@@ -1,3 +1,3 @@
 repository=https://github.com/UserD40/Runelite07Flip.git
-commit=d08c43d52d84f6e859e8f700937b6f3bed2eaf0d
+commit=8917dbb11948642c837801bf2a550cd71b8f75db
 warning=This plugin submits your IP address to a 3rd-party server not controlled or verified by RuneLite developers.
--- a/plugins/07flip
+++ b/plugins/07flip
@@ -1,3 +1,3 @@
 repository=https://github.com/UserD40/Runelite07Flip.git
-commit=43c634fe8360d2c150facd9be194594f8bac8ba2
+commit=c2689aaef9752f43f56dc0557e1fc18bc3cee19b
 warning=This plugin submits your IP address to a 3rd-party server not controlled or verified by RuneLite developers.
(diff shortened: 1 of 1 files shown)


LlemonDuck:
log.info -> log.debug

UserD40:
Thanks for the review! Demoted every `log.info` to `log.debug` across the plugin source — pushed in 3a449cf.

UserD40:
Updated to 43c634f. **My Trades correctness pass:**

- Three dedup passes in `TradeHistoryDedup` for the duplicate-trade-row patterns that were inflating Today/Worst figures (server-fill subsumption, stale-partial twins, stuck-observation re-records).
- `healBackdatedTimestamps` rewrites buy timestamps that the old back-dating heuristic pushed before unrelated trades, restoring FIFO order.
- Bonds moved to a lifetime `BondLedger` persisted outside the 200-row trade window so Membership cost survives heavy flipping. One-shot migration seeds from existing bond rows; manual `Adjust lifetime` dialog covers history evicted before the ledger existed.
- New Filter pill on My Trades (Daily/Weekly/Monthly/All time) scopes stats and list rows; Recent gets a Profit/ROI/Quantity sort; Margin sort moves under the stats card.
- Frozen sell prices now timestamped; refresh against current 07flip rec_sell after configurable staleness threshold (default 3h). Cache pre-warms on buy fills so the GE setup auto-fill arms instantly.
- Server-side post path simplified: one terminal-state row per offer instead of per-fill deltas (eliminates the sync echo that caused the original duplication).
- Test count: 42 (up from 23).

UserD40:
Updated to c2689aa. Shortens the My Trades period filter labels (`Today / Week / Month / All`) so the Margin sub-tab stays on the visible header row. Empty-state copy and tooltips use a longer-form phrase helper so prose still reads naturally.

PR #11881, opened 2026-05-12
Kick history final

itsdukey:
a plugin that logs friends chat kicks into a side panel, and optional discord webhook to help with clan moderation.

tested webhook and side panel functionality 

Screenshot 2026-05-11 205044

Screenshot 2026-05-11 205159

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,3 @@
+[submodule "plugins/kick-history"]
+	path = plugins/kick-history
+	url = https://github.com/itsdukey/Kick-History.git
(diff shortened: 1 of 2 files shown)


itsdukey:
Added support for 'Confirm kick' dialogs (like in Chat Channels). It now resets the timer when you confirm the kick so the log doesn't expire.
Bumped the timeout window to 30s from 5s to give time to "review" a kick prior to clicking "confirm kick"

sorry for all the pushes, should be the last c:

PR #11885, opened 2026-05-12
Global Chat: Bug Fixes

RusseII:
* Fixes bugs with crashing on leagues / quest worlds
* Reduce requests to save on costs


RusseII:
> Feedback:
> java.lang.System.err: system i/o, use a logger instead.
> java.lang.Exception.printStackTrace: system i/o, use a logger instead.
> java.lang.System.out: system i/o, use a logger instead.

Not sure what that is about. Didn't touch any of that in my changes, let me know if I need to fix something. TYSM!

LlemonDuck:
> Not sure what that is about. Didn't touch any of that in my changes, let me know if I need to fix something. TYSM!

they're being picked up by your ably dep. you don't need to do anything

PR #11886, opened 2026-05-12
Add drop-rates-clog

pairofcrocs:
Adds the **Clog Drop Rates** plugin.

Shows drop rates in a tooltip when hovering over items in the Collection Log. Drop rate data is bundled as a JSON resource — no external service calls are made.

- Plugin repo: https://github.com/pairofcrocs/drop-rates-clog

### DIFF
--- a/plugins/drop-rates-clog
+++ b/plugins/drop-rates-clog
@@ -1,2 +1,2 @@
 repository=https://github.com/pairofcrocs/drop-rates-clog.git
-commit=325ad677d688376da1c57169170affb0781a12b0
+commit=5ba27a0fc99fe492ba18f240c4e9531dd9c1089d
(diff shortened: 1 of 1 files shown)


riktenx:
your plugin is able to and should use `build=standard` (not `gradle`) if you want this to be automatically reviewed

pairofcrocs:
Going forward on updating this plugin, what would you suggest my best course of action?

The plugin pulls the json list to determine drop rates for the overlay. Obviously this changes frequently with items begin added, removed, or changed.

My first thought was to run a GH action checking the API for changes, merging those into the json, and automatically open a PR.

Will that be too much, or is there a better approach that you'd prefer?

riktenx:
your plugin is reviewable by the bot, which is generally rate limited, so any sort of procedural update you do like that should be fine

your automated action stuff makes sense, just when you do that make sure you keep your "scripting" code to update your jsons out of the main/ source set so the bot doesn't have to review it. you could do that in the form of like a general bash script, or if you wanted to write it in java you could just make a new source set.

i do something just like the latter in my plugin if you want an example, see https://github.com/riktenx/loot-filters/blob/main/src/script/java/com/lootfilters/UpdateDefaultFilters.java and the associated config in build.gradle (you will still be able to use build=standard)

PR #11887, opened 2026-05-12
Add account-history plugin

josephmr:
The account history plugin tracks notable events for a given account and sends those events to the https://maxcape.net server. The goal is to provide a historical account of an accounts progression so players can look back at what levels, drops, boss kills they have gotten in a timeline view.

This is similar to tools like WiseOldMan and TempleOSRS, but neither of them let you keep track of **when** special milestones were achieved on your account.

The logic in this plugin is **very** similar to the logic in the [Discord Dink](https://github.com/pajlads/DinkPlugin) plugin. Both just track specific events and send HTTP requests to the server with that metadata. There is a config setting that must be enabled before any requests are sent.

Here is a small example of what the resulting view looks like on https://maxcape.net:

image

### DIFF
--- a/plugins/account-history
+++ b/plugins/account-history
@@ -1,3 +1,3 @@
 repository=https://github.com/josephmr/account-history.git
 commit=148e9f2276918c0037100f94e467716e61b069bb
-warning=This plugin submits your RSN and account progression events (level ups, collection logs, boss kc, diary completions, etc.) to a 3rd-party server not controlled or verified by Runelite developers.
+warning=This plugin submits your IP Address, RSN, and account progression events (level ups, collection logs, boss kc, diary completions, etc.) to a 3rd-party server not controlled or verified by Runelite developers.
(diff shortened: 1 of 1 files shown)


JZomDev:
You need to add a warning to your plugin file, something like

warning=This plugin submits your IP address, RSN, and data about your raids to a 3rd-party server not controlled or verified by Runelite developers. 

or what ever data you actually are sending to your site

josephmr:
@JZomDev Updated to include a warning: https://github.com/josephmr/account-history/commit/295b69585d83a706a3d8fbee78aefb5b0d0b37cb

Felanbird:
> @JZomDev Updated to include a warning: [josephmr/account-history@295b695](https://github.com/josephmr/account-history/commit/295b69585d83a706a3d8fbee78aefb5b0d0b37cb)

It needs to be added to the plugin file in this repo, not your project.

josephmr:
@JZomDev whoops, fixed -- added to plugin hub file

PR #11888, opened 2026-05-12
Update Odablock Sounds Plugin

DapperMickie:
Fixes issue with Odablock Sounds Live Notifications, also adds 1 sounds for live notification.
Thanks @LogicalSoIutions

### DIFF
--- a/plugins/odablock
+++ b/plugins/odablock
@@ -1,2 +1,2 @@
 repository=https://github.com/DapperMickie/odablock-sounds.git
-commit=6ed6dc26efd4cee6d35632b0aaf226fa2a716f59
+commit=130a785360dbb833a5fd4ce4abe0fc90ba85556b
--- a/plugins/odablock
+++ b/plugins/odablock
@@ -1,2 +1,2 @@
 repository=https://github.com/DapperMickie/odablock-sounds.git
-commit=6c1173e9e365aac4101f1ab5efccf6271086149e
+commit=6ed6dc26efd4cee6d35632b0aaf226fa2a716f59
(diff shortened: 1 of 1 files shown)


LlemonDuck:

DapperMickie:
Done, now also includes a small config fix

pajlada:
[DapperMickie/odablock-sounds@`77e3b66914dc6c6725378887aa6f6b9e6898d226..DapperMickie:6ed6dc26efd4cee6d35632b0aaf226fa2a716f59`#diff-c259c6cab6](https://github.com/DapperMickie/odablock-sounds/compare/77e3b66914dc6c6725378887aa6f6b9e6898d226..DapperMickie:6ed6dc26efd4cee6d35632b0aaf226fa2a716f59#diff-c259c6cab6db4f0a560232db240a72ac260957ff09cf06af28fb61d10029bb6cR91) this change is incompatible with how your plugin was previously configured.
You need to revert this back to a `raw.githubusercontent.com` url, or make live stream checking opt-in with a warning annotation.

LogicalSoIutions:
What would the opt-in warning have to say? There is no data being sent to the servers, only incoming so I assumed I didn't need a warning, shame on me for assuming. (I was the one who made the changes hence why I am asking)

coopermor:
> warning = "This feature will submit your IP address to a 3rd-party server not controlled or verified by Runelite developers."

pajlada:
Yep, what coopermor shared above is correct.
Keep in mind that adding that alone would not make the feature opt-in - users who have previously installed the plugin would still have the live stream feature installed & sending requests to a non-verified host without any recourse. You would need to reset the user's live-stream setting (i.e. by changing the key name) _and_ changing the default value to false.

LogicalSoIutions:
Just going to go the GH route. Got the URL updated, just have to wait on Dapper to merge PR and update. 

Thank you!

DapperMickie:
Has been updated and reverted back to gh url

PR #11889, opened 2026-05-12
New Plugin: Shortest Grave

FIrgolitsch:
Releasing Shortest Grave! A small plugin that integrates with Shortest Path to plan a route to the players' gravestone after death.

- Keeps track of the location the player died and plans a path to that location after respawn.
- Overrides Shortest Path path colour if desired
- Checks if a gravestones has spawned and plans conditionally.
    - Can be overridden if desired.

### DIFF
--- a/plugins/shortest-grave
+++ b/plugins/shortest-grave
@@ -1,3 +1,3 @@
 repository=https://github.com/FIrgolitsch/shortest-grave.git
-commit=dd71b8af646e8ab8f2b9ed62f697fd8ec5db9c88
+commit=db2c1ce1cb5fcf48d7aac13e130e3cf079617534
 authors=FIrgolitsch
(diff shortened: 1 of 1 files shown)


FIrgolitsch:
Was that error on my end or does it have something to do with the build process with you?

```
> Could not resolve net.runelite.arn:http-api:1.2.22. 
  Required by:
  root project : > net.runelite:client:1.12.26.3
```

LlemonDuck:
that error is transient, but you have a new one:
```
net.runelite.pluginhub.packager.PluginBuildException: shortest-grave: Plugin class "shortest-grave.ShortestGrave" is missing from the output jar
```

LlemonDuck:
`plugins=shortest-grave.ShortestGrave` should be `shortestgrave.ShortestGravePlugin` to match the plugin class' fqdn

FIrgolitsch:
Got it! Thanks!

PR #11908, opened 2026-05-13
disable sailing-reorderer

marknewan:


coopermor:
If your plugin still functions I would recommend keeping it around as long as you don't mind supporting it. Jagex only allows for reordering the panel while docked and doesn't allow you reorder the auto-repair. This plugin still has some features that the base game does not.

marknewan:
It seems the base game doesn't allow moving the steering nor the repair facilities; I'll keep this plugin up until they make the base options more flexible.

PR #11936, opened 2026-05-14
Add Camera Snapshots plugin

BPM-OSRS:
Adds the Camera Snapshots plugin.

Saves and loads exact camera angle and zoom presets (yaw, pitch, zoom) per account with hotkeys. Unlike existing compass plugins this saves your precise angle, not just cardinal directions. Supports 5 slots.


ldavid432:
There is a reason current plugins do not do that. It is against the rules as per: https://github.com/runelite/runelite/wiki/Rejected-or-Rolled-Back-Features

> Presets for Camera yaw/pitch/position

PR #11955, opened 2026-05-16
Update RuneDiary to 1.1.0 (seasonal world handling)

CalciumG:
Update runediary to 1.1.0

Skips plugin sync on seasonal worlds (Leagues / DMM / beta) so leagues data doesn't overwrite the main account's plugin tables in the hub DB.

Plugin changes: https://github.com/CalciumG/runediary-plugin/compare/114b5a86...1e5ac293


CalciumG:
Thanks in advance for any reviews!

PR #11961, opened 2026-05-16
RuneProfile v2.5.0

ReinhardtR:
- Now retrieves CA varps from the client and sends to the RuneProfile API.
- Now triggers autosyncs when plugin detects activity for: Skills, Quests, Diaries and CAs.


riktenx:
address the build failure

riktenx:

ReinhardtR:
Made new PR to fix the crash error, without use of menuAction. 
https://github.com/runelite/plugin-hub/pull/11987

PR #11963, opened 2026-05-16
Watchdog v4.0.0

adamk33n3r:
The big boi is here!

Big things to note for reviewers:
- Advanced alerts are opt-in, turned on within the config, and marked as Beta.
- I have added more banned region checks earlier in the process so it should be more stable and not "do things" while within a banned area.
- I am now compressing the alert json with gzip for better storage.
- I've done some work on performance especially with regards to object spawning.

Let me know if there are any questions!


Bram91:
This PR needs rebase; it has conflicts currently.

adamk33n3r:
Sorry I missed your comment here. It's probably because I've been patching v3 for blocking boss regions and once they merged in it probably messed things up. I have another boss patch coming in soon so I'll update this PR once that one's merged in and i can make sure there's no conflicts.

PR #11969, opened 2026-05-16
Update RuneDiary to 1.1.1 (bank sync trigger fix)

CalciumG:
Update RuneDiary to 1.1.1

Fixes excessive hub traffic from the bank value tracker. Previously the tracker fired a full profile-sync HTTP POST on every ItemContainerChanged event, so every deposit/withdraw/sort produced a network round-trip. The sync now fires only when the bank UI closes (WidgetClosed for group 12), which keeps the in-memory snapshot fresh as before but emits at most one call per banking session.


CalciumG:
2nd of the day sorry! Initial bugfixes 😓

Thank you in advance again!

PR #11971, opened 2026-05-16
Update Ultimate Stat Tracker With Performance Mode Enabled by default

cabrilliant:
some users have reported some performance issues during movement heavy scenarios. I have added a performance mode default enabled to disbale all game tick and widget load based tracking for now which should reduce performance impact greatly. Most stats still work without this anyways.


cabrilliant:
im not sure why its showing like 8 commits. really its just the last commit that is new

cabrilliant:
i dont know where this merge thing came from im just gonna re create this one

PR #11988, opened 2026-05-17
Add collection-log-commands

unk-user-shade:
Adds Collection Log Commands, a RuneLite plugin that lets users share cached collection log progress in chat with !log commands, including missing-item views, aliases, summary output, and verbose/condensed formatting modes.

### DIFF
--- a/plugins/collection-log-commands
+++ b/plugins/collection-log-commands
@@ -1,2 +1,2 @@
 repository=https://github.com/unk-user-shade/collection-log-commands.git
-commit=6321c293b5983fbd6279212873a7c5c5371591ba
+commit=2e334b8dcaaf20e768d6e9b995dcf8de108e338b
--- a/plugins/collection-log-commands
+++ b/plugins/collection-log-commands
@@ -1,0 +1,2 @@
+repository=https://github.com/unk-user-shade/collection-log-commands.git
+commit=6321c293b5983fbd6279212873a7c5c5371591ba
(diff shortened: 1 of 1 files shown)


ldavid432:
Just FYI runeprofile already uses the `!log` command. You're not required to change it but it might be confusing to users if they have both plugins installed

pajlada:
also, this does not work how you've explained it in the readme. when someone else types `!log zulrah` it will show you _your_ zulrah log, not the log of the user who typed it.

unk-user-shade:
> also, this does not work how you've explained it in the readme. when someone else types `!log zulrah` it will show you _your_ zulrah log, not the log of the user who typed it.

Intended but I did not I guess do the description best, is it only intended as a local/for yourself quick way of doing collection log commands.



> Just FYI runeprofile already uses the `!log` command. You're not required to change it but it might be confusing to users if they have both plugins installed

I made this because this used to be its own separate thing. We did not realize it was likely absorbed or turned into a runeprofile feature. Checking with the people to see if there is anything to make it a separate OR if it's worth just using runeprofile.

Felanbird:
If it's not your intention to show the information to other people, then you should be using [CommandExecuted](https://static.runelite.net/runelite-api/apidocs/net/runelite/api/events/CommandExecuted.html)

unk-user-shade:
Made changes for using [CommandExecuted](https://static.runelite.net/runelite-api/apidocs/net/runelite/api/events/CommandExecuted.html) instead.

Still interested as instead of the entire Runeprofile setup this is purely for logs and local.

PR #11992, opened 2026-05-17
Update odablock

DapperMickie:
first changes - add config option to play sound for going live + stream offline message

add lvl 99 sound, add full override custom system, format config

Thanks @LogicalSoIutions :)


LogicalSoIutions:
The GSON fix has been applied, pending PR Merge

PR #11998, opened 2026-05-17
update tile-overlay-indicators

Resh404:
player true tile metronome and tick counters added


Resh404:
A lot of the changes are from formatting cleanup

PR #12008, opened 2026-05-17
Update womclanplugin

ShakyPizza:
## Summary
- Add a local JSON cache for the last successful clan sync, loaded on startup
- Add a `Name Changes` tab in the expanded GUI showing recent clan name changes
- Show clan summary info in the sidebar above search
- Space WOM API requests with a short delay
- Change auto-refresh interval from 30 minutes to 1 hour
- Update README with user-facing setup, cache, and refresh behavior

### DIFF
--- a/plugins/womclanplugin
+++ b/plugins/womclanplugin
@@ -1,3 +1,3 @@
 repository=https://github.com/ShakyPizza/womclanplugin.git
-commit=3f9d298791f9ea3d5044773893ea71c54301d505
+commit=156dc6cf6f2d8fe2441c979565b1342dcd67f58e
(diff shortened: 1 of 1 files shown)


riktenx:
* use of thread sleep is not allowed

ShakyPizza:
> * use of thread sleep is not allowed
> 

Thanks, I'll remove the delay and interrupt.

ShakyPizza:
@riktenx Would it be considered a good practice to use GameTick(s) for spacing out the API calls?

ShakyPizza:
Superseded by #12347. This splits out only the name-change tab and README changes; cache work will come later.

PR #12052, opened 2026-05-19
update tasks-tracker : task progress bars & fixes

perterter:
Update tasks-tracker plugin
- added progress bars to tasks (e.g. catch 500 chinchompas) - most loaded from vars
- fixed bug where plugin panel caused RuneLite size to max out infinitely, requiring restart to fix
https://github.com/osrs-reldo/tasks-tracker-plugin/issues/166
- changed the build runelite plugin property
- fixed a high memory usage class (reducing plugin memory footprint by ~40%)
https://github.com/osrs-reldo/tasks-tracker-plugin/issues/171
- added a sum mode to progress bars (think echo gauntlet + non-echo gauntlet for KC)
Screenshot 2026-05-17 185429

### DIFF
--- a/plugins/tasks-tracker
+++ b/plugins/tasks-tracker
@@ -1,3 +1,3 @@
 repository=https://github.com/osrs-reldo/tasks-tracker-plugin.git
-commit=416520eebc0fc942bada103a837997c75d73d606
+commit=c83f3663e00aef01403382e6663509d59c73d00f
 authors=perterter,JamesShelton140


riktenx:
you should use `build=standard` in runelite-plugin.properties, i am not seeing anything preventing you from doing so (you don't have any custom deps or build tasks) and it would allow your plugin to get autoreviewed

perterter:
- Changed the build runelite plugin property
- Fixed a high memory usage class (reducing plugin memory footprint by ~40%)
- Added a sum mode to progress bars (think echo gauntlet + non-echo gauntlet for KC)

perterter:
@riktenx 
> you should use `build=standard` in runelite-plugin.properties, i am not seeing anything preventing you from doing so (you don't have any custom deps or build tasks) and it would allow your plugin to get autoreviewed

Would it be better if I opened a PR with that change only? So the remainder of the changes could be autoreviewed? It seems like we still have the maintainer review gate maybe because we flipped the property?

perterter:
Going to reopen with only the properties change, then follow with feature PRs

PR #12067, opened 2026-05-20
Add Frieza Sounds For Nex

muddygms:
Has a Frieza AI that is calling out the 4 names at the start, along with telling you to switch prayers and enter melee range when needed!


riktenx:

PR #12074, opened 2026-05-20
Add will-it-land plugin repository information

EthonSmoth:
Added `build=standard` to `runelite-plugin.properties` in the plugin repo and pushed it. Could the checks be rerun when you have a moment? Thanks!


EthonSmoth:
@pajlada does this need any changes?

riktenx:
what is the file i/o for?

EthonSmoth:
@riktenx The file I/O was a fallback in NpcStatsRepository that read npc_stats.json from RuneLite.RUNELITE_DIR if the bundled jar resource failed to load. It wasn't needed — the stats file is always bundled in the jar. I've removed the filesystem fallback entirely; the plugin now only loads from the classpath resource.

PR #12077, opened 2026-05-20
Update eventtools-attendance

jeromkiller:
Fixed bug where players could not create new capture areas

Added filter options to track friends, FC members and CC members seperately


ldavid432:
You need to add `build=standard` to your `runelite-plugin.properties`

PR #12079, opened 2026-05-20
Fix hide-and-seek-tracker

jeromkiller:
Fixed bug preventing players from creating new capture areas


ldavid432:
You need to add `build=standard` to your `runelite-plugin.properties`

ldavid432:
Also while you're at it remove `internalName`, that is not a valid property

PR #12087, opened 2026-05-20
Add Raid Damage Tracker

osrschak-commits:
Adds Raid Damage Tracker, a plugin that tracks personal damage dealt to each boss across Chambers of Xeric, Theatre of Blood, and Tombs of Amascut.

Features:
- Live damage counter per boss, shown as soon as the room is entered
- Boss Split mode: per-boss DPS calculated from first hit to kill
- Raid Overall mode: running DPS across the full raid, paused between rooms

No external dependencies beyond the RuneLite API.


ldavid432:
If you want to be eligible for automatic review you should set your build type to `standard`. I don't see anything special in your build.gradle that would require `build=gradle`

PR #12090, opened 2026-05-20
update to v1.1

psenk:
fix bugs and update ui


tylerwgrass:
Please escape text you are wrapping in html tags

psenk:
So sorry about all the commits, I am still a fresh dev.  I noticed I accidentally committed more code than I desired after fixing the HTML escaping stuff requested above, and then I got super confused working with all the different branches.  I believe things should be good to go now.  Hopefully.

Thanks for your patience and for everything you guys do for RuneLite and the OSRS community.

PR #12101, opened 2026-05-21
Add Merch Lens

RealGrind:
Adds Merch Lens, a RuneLite plugin for Grand Exchange market watching, flip research, item screening, and local GE offer tracking.


riktenx:
use of thread interrupt is banned

PR #12129, opened 2026-05-22
add color-locked

unidarkshin:
## Plugin: Color Locked

Group Iron color-lock enforcement plugin.

- Blocks Eat/Drink/Equip/Wield/Wear on items restricted by the player's assigned color
- Strips Mine/Chop when carrying a restricted pickaxe or axe
- Marks restricted items with a red overlay in inventory, bank, and worn equipment
- Includes a sidebar for looking up which items each color can use
- Syncs assigned color and item rules from a hub group, or runs standalone with manual settings

**Repository:** https://github.com/unidarkshin/osrs-color-lock-runelite
**License:** BSD-2-Clause
**No third-party dependencies** — only RuneLite client APIs and transitive Gson.

### DIFF
--- a/plugins/color-locked
+++ b/plugins/color-locked
@@ -1,2 +1,2 @@
 repository=https://github.com/unidarkshin/osrs-color-lock-runelite.git
-commit=8b8887566e6e7b7c3c2faedd01e390c968bbb777
+commit=8e526323fdb9c51467eb7e8c5dfeafc0824f82e1
--- a/plugins/color-locked
+++ b/plugins/color-locked
@@ -1,2 +1,2 @@
 repository=https://github.com/unidarkshin/osrs-color-lock-runelite.git
-commit=c4467d7586a245c6fbbcb12cfb2d8ec880c28d58
+commit=403bde8b4a0e6d5186f39721ecf13181f5742f86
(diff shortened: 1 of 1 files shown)


riktenx:
use of thread interrupt is not allowed
use of javanet httpclient is not allowed, inject and use runelite's OkHttpClient instead

devinlematty:
@riktenx I fixed those two issues - thank you.

riktenx:
use of httpurlconnection is not allowed, all networking should be through okhttp

devinlematty:
@riktenx Got it, I've removed all httpurlconnection references in favor of okhttp.

riktenx:
thread sleep and thread interrupt are not allowed eihter

devinlematty:
@riktenx I've removed all thread interrupt and sleep in the latest commit.

plugins/color-locked
```diff
@@ -0,0 +1,2 @@
+repository=https://github.com/unidarkshin/osrs-color-lock-runelite.git
+commit=403bde8b4a0e6d5186f39721ecf13181f5742f86
```
riktenx, line 2:
```suggestion
commit=403bde8b4a0e6d5186f39721ecf13181f5742f86
warning=This plugin submits your IP address to a 3rd-party server not controlled or verified by Runelite developers.
```

PR #12146, opened 2026-05-23
Update and rename stop-misclicking-orbs to orb-clickthrough

jtclowner:
Updates and renames `stop-misclicking-orbs` to `orb-clickthrough`.

### Features

- Added clickthrough support for the compass and logout orb.
- Added support for plugins that hide or remove the minimap.

### Fixes and maintenance

- Refactored widget transformation logic so widget state changes are tracked and restored more safely.
- Improved orb menu-entry filtering so it is scoped to the relevant widgets.
- Reduced unnecessary churn by avoiding full per-tick orb widget syncing.


cdfisher:
You can update the display name of your plugin but you can't change the name of the manifest file.

PR #12157, opened 2026-05-23
gear name typos, added void, blowpipe setting, deposit bug + void helm fixes

MikeN3:
…m fixes

### DIFF
--- a/plugins/slayer-loadout
+++ b/plugins/slayer-loadout
@@ -1,2 +1,2 @@
 repository=https://github.com/MikeN3/slayer-loadout.git
-commit=deb1b7d30236201a1020ed519788ca2ec02c2486
+commit=deb52cad46faf05736e4092f85275aeb6994fc2e


MikeN3:
Gear name fixes — corrected an item-name mismatch (Archers ring → Archer ring), made Slayer-helmet colour/imbued variants match the head slot, and added Lightbearer as a ring option.
Bug fix: recommendations no recommend lower tier items after depositing worn/inventory items
Bug fix: a Void melee helm no longer leaks into the Ranged/Magic head slots (void helms are style-locked).
New setting (default off): "Prefer blowpipe" — overrides the ranged weapon with a blowpipe + best darts when owned.
Smarter ammo: boss tasks prefer enchanted dragon bolts; regular slayer keeps broad bolts.
DPS calc: Void and Elite Void set bonuses are now modelled (melee/ranged +10%, elite ranged +12.5% dmg, magic +45% acc, elite magic +5% dmg).

PR #12160, opened 2026-05-24
Roster: v1.1.1 (critical fix for v1.1.0 data-loss bug)

Mylosis:
Update Roster to **v1.1.1** — critical bug fix release.

## ⚠️ Why this is urgent

v1.1.0 (just merged) shipped a data-loss bug for users upgrading from v1.0. The migration from the old file-based storage to `ConfigManager` renamed `roster.json` to a backup file immediately after calling `setConfiguration()`. Because `ConfigManager` debounces its writes to disk, any user whose RuneLite session ended (force-close, crash, restart, fast quit) before that flush hit `profiles2/<profile>/roster.properties` lost their in-memory write — and the file backup had been renamed away, leaving the data unreachable on the next launch.

The data isn't actually destroyed (it's still on disk as `roster.json.pre-v1.1.bak`), but the average user has no way to know that. They reinstall the plugin, see an empty panel, and assume their accounts are gone.

## The fix

`AccountStorage.load()` is now **idempotent**:

- `roster.json` is never renamed. It stays as a permanent secondary store, written by every `save()` alongside the `ConfigManager` write.
- A future `ConfigManager` flush failure simply re-seeds from the same file on the next launch — no orphaning.
- The load path now tries `roster.json`, then `roster.backup.json`, then the legacy `roster.json.pre-v1.1.bak` (the migration sentinel from v1.1.0), so users who already hit the v1.1.0 bug recover their data automatically the moment they install v1.1.1.

## Also in v1.1.1

- **OSRS sprite badges for account types** — replaced the letter-in-circle rendering with actual sprites from the wiki: official chat badges for the six Ironman variants, Abyssal whip for Main, the in-game Skills tab icon for Skiller, Granite maul for Pure. ~2 KB of resources total.
- Badge sizes bumped (16 → 20 px list view, 12 → 16 px grid view) for legibility.
- More breathing room between display name and the type badge.
- Footer menu labels normalized to match the per-account menu wording ("Import from Clipboard" / "Export to Clipboard").
- Deep architecture & hot-path cleanup (no functional change): config-change echo loop broken, search debounced, lazy id indexes on `AccountData`, group-bucket pass in `AccountListBuilder`, cached sorted-groups view, incremental card refresh on selection/online change, memoized `Font.deriveFont`, drag-overlay stroke cache, snapshot-based drop-target lookup, coalesced file-mirror writes, and a shutdown flush.

## Verification

- Built locally with `./gradlew build`; no test failures.
- The dual-write strategy (ConfigManager + `roster.json`) is preserved; only the rename was removed.

Source: https://github.com/Mylosis/roster
Commit: https://github.com/Mylosis/roster/commit/dce968937216b3bbba73bf63c233cbb812f360b1


riktenx:
* if you have an "urgent" bugfix that you want reviewed ASAP then you will want to submit the absolute minimum amount of code for said fix, rather than that fix + ~1000 lines of unrelated things
* we typically do not consider bugs urgent unless they are freezing/crashing clients
* as a reminder this queue is staffed by volunteers and there is no guarantee that people will always be around to manually review your code. mistakes happen but we recommend you test your plugin thoroughly

PR #12161, opened 2026-05-24
Required gear (Lizardmen Shayzien), Salve variants, ammo-aware ranged…

MikeN3:
Required gear (Lizardmen Shayzien), Salve variants, ammo-aware ranged DPS, bank-only recompute, weapon/ammo fixes

### DIFF
--- a/plugins/slayer-loadout
+++ b/plugins/slayer-loadout
@@ -1,2 +1,2 @@
 repository=https://github.com/MikeN3/slayer-loadout.git
-commit=deb52cad46faf05736e4092f85275aeb6994fc2e
+commit=42c1917bb40bc873e568be4acf870f014cf6ceaa


MikeN3:
- New "Required gear" callout — surfaces mandatory/protective gear for tasks that need it, with owned/missing state. First use: Lizardmen (full tier-5 Shayzien to block shaman poison).
- Salve amulet is now modelled correctly per variant and style vs undead (base/(e) melee-only; (i)/(ei) cover ranged & magic), and doesn't stack with the slayer helm.
- Ammo-aware ranged DPS — crossbows are now judged with the bolts they'd fire instead of at zero ranged strength, and a crossbow is never paired with bolts it can't use (e.g. Dorgeshuun ≠ broad bolts).
- Recommendations only recompute at the bank, so swapping gear during normal play no longer changes them.
- Vorkath gear updated from the wiki strategies.
- Fixes: Lightbearer is only suggested with a special-attack weapon; the Skull sceptre is no longer treated as a powered staff; non-imbued Salve isn't suggested for ranged/magic.

PR #12163, opened 2026-05-24
Update slayer-loadout

MikeN3:
Add Amulet of rancour + Avernic treads; magic autocast-staff/spell consistency

### DIFF
--- a/plugins/slayer-loadout
+++ b/plugins/slayer-loadout
@@ -1,2 +1,2 @@
 repository=https://github.com/MikeN3/slayer-loadout.git
-commit=42c1917bb40bc873e568be4acf870f014cf6ceaa
+commit=0279bb6a2ae907b296dc70b66ef67c3a2096074b


MikeN3:
Added

Amulet of rancour (melee amulet) and Avernic treads (melee boots) added to the melee gear lists.

Changed / Fixed

Magic loadout is now consistent with the recommended spell. When a monster has a recommended spell (e.g. Ice Barrage) and the player owns an autocast-capable staff, the plugin now recommends that staff (Master wand, Kodai wand, Ancient sceptre, Staff of the dead, etc.) so the spell can actually be cast — previously it could recommend a powered staff (Trident/Sanguinesti/Shadow) next to a spell it can't autocast.
The spell row now matches the weapon: a powered staff shows "no spell needed," and a spell only appears beside a staff that can cast it.
The Ancient sceptre is no longer mis-classified as a powered staff (it autocasts Ancient spells).

PR #12168, opened 2026-05-24
update flipping copilot to v1.10.2

cbrewitt:
- Add 1 day padding to flip visualisation time window
- Simplify portfolio remove menu label
- Display keybind in offer editor
- Add keybinds for skipping suggestions and opening graphs
- Add ROI and cost tooltips to suggestions


riktenx:
don't use getenv, you can use a jvm sysprop for your custom api host instead

PR #12173, opened 2026-05-24
update chat-fade

internetter:
Adds horizontal and vertical position offset sliders to the Display section. Both default to 0 so existing behavior is preserved.

Requested in https://github.com/internetter/chat-fade/issues/13.


ldavid432:
If you would like to be eligible for automated review you can use `build=standard`. I don't see anything special in your build.gradle that would warrant you using `build=gradle`

PR #12174, opened 2026-05-24
Add Combat Indicator plugin

66lexel999:
Adds a Combat Indicator plugin that shows whether the player 
is idle, walking, or in combat via a colored overlay.
Visual only - no automation.


riktenx:
closing since this is still broken, if you fix it make a new pr

PR #12179, opened 2026-05-24
1.4.0: Fossil Island wyverns (BiS + defence stats + required shield),…

MikeN3:


MikeN3:
Added

Fossil Island wyverns task, with Melee/Ranged/Magic BiS from the OSRS Wiki and the monster's defence stats so the DPS estimate works. The mandatory anti-wyvern shield (any one of Ancient wyvern shield, Dragonfire ward/shield, Mind or Elemental shield — an Anti-dragon shield won't work) is surfaced as a "Required gear" callout for the icy breath. Draconic, so Dragon hunter weapons + dragon bolts apply.
The installed plugin version is now shown in the side-panel footer (e.g. "Slayer Loadout v1.4.0").

Changed

Enchanted dragon bolts (Ruby/Diamond dragon bolts (e)) are now preferred on any draconic task, not just bosses, so crossbows on dragons/wyverns use dragon bolts over broad bolts when owned.

PR #12180, opened 2026-05-24
TerpinheimerCC Update

FadedOSRS:
Updated some item's that where not working and also fixed the Warning as it spanned multiple screen's for some reason.


FadedOSRS:
If I did this incorrectly please let me know. This is my 1st time updating the plugin.

PR #12185, opened 2026-05-25
Add potion-quantity-labels

theldurin:
This plugin adds a quantity label to each potion in the Potion Storage interface, allowing you to better visualize just how many potions are actually stored.


riktenx:

PR #12188, opened 2026-05-25
Update grand-flip-out

Tunatroll:
Bumps Grand Flip Out to the latest commit:
- old: 468d45ed59b1716dedcb40caf3ec32b84cd0a6bf
- new: ac3a6a2452f30351312bb2a6f6d2c7557ca8e815

This update adds:
- Local NDJSON trade log with flip completion events
- Trade log reader and history rendering
- Wealth snapshot (coin/inventory/bank/total) captured on flip completion
- Session wealth delta tracking
- CSV export with wealth columns
- Optional server intelligence advisor (default off, read-only)


Tunatroll:
sorry its such a big update, i can trim it

ldavid432:
Doesn't build
> Do not create fresh Gson instances, always @Inject the client's Gson.
> You can customize it by calling .newBuilder() on it

Tunatroll:
Thank you man 🙌

Tunatroll:
Sorry it's so big I didn't know if I should do a lot more small updates or just do a sweep of pretty much everything the plugin needs so that I don't have to do bulky updates ever again, let me know if you want me to trim down some stuff for this version

Tunatroll:
Should I drop some features for a separate update and make this one smaller?

Tunatroll:
Following up on my question above — I'd like to make this easier to review. The size-xl label is fair (the target is `81a2eeb`, a 13-commit delta).

Concrete proposal: split into two manifest bumps.
- **PR-A** — just the GE History importer: client-side only, no network calls, has its own tests (`GeHistoryImporterTest`, 9 cases) and builds green on JDK 11. Reviewable in minutes.
- **#12188** — I'd re-point this PR to the remainder (Advisor panel, price charts, recipes, and the opt-in grandflipout.com intelligence features, default-off). The `warning=` line for the 3rd-party data is already in this PR's diff and stays.

This keeps #12188 and its review history — I'd open PR-A as a small new PR, then re-point this one after it merges. Happy to do it whichever way you prefer, or leave it as one PR if that's easier to review. Just let me know.

Adam-:
#12188 is stuck in some sort of error state with the bot, I'll try and rerun it.

Adam-:
It seems since you are adding new network calls to grandflipout.com we need to have someone look at it, especially since the warnings can't be retroactively added.

Tunatroll:
Thank you man I really appreciate it

Tunatroll:
Should I update the description really quickly to match? The warnings in
the plugin are good, but I just noticed that one of the descriptions is
outdated, or should I leave it as it for review?

Tunatroll:
Hi @Adam- not trying to pester or be impatient, no rush, I know the network-call review takes time and I'm happy to wait. Just following up on my earlier question: I went ahead and aligned the plugin description with the opt-in networking disclosure, so it's consistent now 
  
  The opt-in install warning is in place per your note. And I'm glad to answer anything about what's sent to grandflipout.com and when: the only non-opt-in data is the IP from the request itself; GE offers/transactions/coin total are sent only if the user opts in. Thanks for taking a look whenever you get the chance!

riktenx:
the warning can't go in the manifest here because existing installs won't have had a chance to see it

any http requests to grandflipout.com MUST be opt-in only, and therefore MUST be gated behind an "enable grandflipout.com functionality" boolean config (which defaults to false) whose `@ConfigItem` has its `warning` set to `"This feature submits your IP address to a 3rd-party server not controlled or verified by Runelite developers."`

Tunatroll:
Thank you man, been working so hard on this and the website that I
completely missed that in my fixes yesterday

Tunatroll:
Ok that's it sorted now. I added a master "Enable grandflipout.com functionality" toggle that defaults to off, with the `@ConfigItem` warning set exactly to what you gave me:

> This feature submits your IP address to a 3rd-party server not controlled or verified by Runelite developers.

Every grandflipout.com call is gated behind it now — the advisor, the server intelligence signals, the trade contribution and the account check all bail out unless it's switched on, so with it off the plugin just runs on the Wiki prices and doesn't touch grandflipout.com at all. The individual feature toggles underneath stay off by default too, they just pick what actually gets sent once the master one is on.

And no worries on the retroactive part — there aren't any real installs of this yet, so nobody's had the chance to turn networking on; first time anyone does they'll get the confirm dialog from the warning. I also flipped the GE price-fill helper to default-off while I was in there, so a fresh install isn't doing anything network- or input-related until it's opted into.

Manifest's bumped to the new commit. Happy to change the wording or the structure however you'd like it — thanks for bearing with me on this.

Tunatroll:
Pushed another round and bumped the manifest — swapped the deprecated JsonParser calls for the injected Gson, expanded the opt-in warning to spell out the IP + GE data it sends, moved the keybind handlers onto the client thread (clipboard to the EDT), and switched to LinkBrowser. Thanks for bearing with me — happy to change anything else.

plugins/grand-flip-out
```diff
@@ -1,2 +1,3 @@
 repository=https://github.com/Tunatroll/grand-flip-out.git
-commit=468d45ed59b1716dedcb40caf3ec32b84cd0a6bf
+commit=3707cb7bf771cba89add2b7a541617aad132284a
+warning=This plugin can submit your IP address and, if you opt in, your Grand Exchange offers, transactions, and approximate coin total to a 3rd party server not controlled or verified by the RuneLite developers.
```
riktenx:
```suggestion
```

Tunatroll:
sorry i tried using claude for one simple repo scan to see if there were any left over offences and he completely went off the rails on everything 😭

Tunatroll:
Thanks @riktenx — applied your manifest suggestion, and all grandflipout.com calls are gated behind a default-off toggle with the IP-warning string you specified. Settled at 7242b90, build's green — ready for re-review when you have a moment.

riktenx:
* you have this opt-in split through like half a dozen configs, you need one global toggle that turns everything off, e.g. it looks like requestSuggestion() can still touch the server without the user opting in. you should really just put an "if !enabled() throw exception" in front of everything so a request can never leak through
* do not do any file i/o outside of your .runelite/flipping folder, looks like you're reading from .runelite/profiles

Tunatroll:
Thank you, I thought I did add a master switch but you're right that adding individual toggle per features does make things more convoluted, not more transparent. I'll fix it now, I appreciate the help and patience a ton

Sent from my Verizon, Samsung Galaxy smartphone
Get Outlook for Android<https://aka.ms/AAb9ysg>
________________________________

riktenx:
`runFlippingUtilitiesImport()` appears to still be looking around generally everywhere within `.runelite/`

Tunatroll:
I can't tell you how much I appreciate the help and quick checks, this was
a bigger first solo project than I anticipated when I started in like
February, but I am so excited about how far it's come and how much I've
learned - so thank you for holding with me through these rookie mistakes

PR #12204, opened 2026-05-25
Update Flip Smart

mablack01:
## Summary
- Reset the price adjustment timer when a user relists an item at a new price, so the timer reflects the actual relist time rather than the original listing time
- Added contextual trade data to the Grand Exchange buy/sell window, showing live flip information inline when viewing an item in the GE interface
Screenshot 2026-05-25 at 3 43 41 PM
Screenshot 2026-05-25 at 3 44 06 PM

### DIFF
--- a/plugins/flip-smart
+++ b/plugins/flip-smart
@@ -1,4 +1,4 @@
 repository=https://github.com/Flip-Smart/flip-smart-runelite-plugin.git
-commit=2eb818aa3ee9c389d35e64ed84fd42e6e3168e2e
+commit=42516841e6d498d9dbb00e792e1af157c1294319
 authors=mablack01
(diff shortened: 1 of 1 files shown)


Adam-:
Please remove your use of `dispatchKeyEvent()`

mablack01:
> Please remove your use of `dispatchKeyEvent()`

Addressed and updated the commit hash

Adam-:
> > Please remove your use of `dispatchKeyEvent()`
> 
> Addressed and updated the commit hash

Remove the entire feature which relies on spoofing ops.

mablack01:
> > > Please remove your use of `dispatchKeyEvent()`
> > 
> > 
> > Addressed and updated the commit hash
> 
> Remove the entire feature which relies on spoofing ops.

@Adam- updated the plugin to no longer spoof ops for that, let me know if these changes are still up to standard. The difference now is that users will have to use built in functionality to select the search item (pressing enter or clicking the item)

PR #12206, opened 2026-05-25
Add Responsive Tooltips plugin

ZacharyHJacobson:
This plugin removes the delay after mousing over an element of the side panel display before a tooltip appears.  Without the plugin, it only instantly appears if you moused over another part of the UI recently.  With it, the tooltip always appears almost immediately.

The second feature of this plugin customizes prayer tooltips to be easily identifiable without directly reading them.  If enabled, prayer tooltips will have their text removed, size changed to user-provided dimensions, relocated to compensate, and recolored in a similar way to inventory highlights.


ldavid432:
Please choose a more descriptive name for your plugin, we no longer accept plugins named like "improved x" or "better x"

ZacharyHJacobson:
> Please choose a more descriptive name for your plugin, we no longer accept plugins named like "improved x" or "better x"

Fixed, the plugin is now named "Responsive Tooltips"

PR #12217, opened 2026-05-26
Archipelago Plugin UI Rework

digiholic:
Ports the UI rework from the Beta Members Randomizer into the base free-to-play randomizer version

### DIFF
--- a/plugins/osrs-archipelago
+++ b/plugins/osrs-archipelago
@@ -1,2 +1,2 @@
 repository=https://github.com/digiholic/osrs-archipelago.git
-commit=7868d02ac6bb25145f324b6d74e70b7594a02e4d
+commit=58983078781ab44a8f6468fc02f16d6a3a92d82a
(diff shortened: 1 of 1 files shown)


ldavid432:
You need to add the build type to your runelite-plugin.properties. I think since you are depending on the websocket library you'll need  `build=gradle`

riktenx:
are you able/willing to drop your websocket dependency and switch over to okhttp websockets, that would allow you to set build=standard and then your plugin could be automatically reviewed by the bot. this should be possible for you since okhttp3 websockets do the permessage-deflate stuff automatically afaik

digiholic:
Updated for okhttp3. Still some issues with Websocket Compression, but that's not a hard requirement yet. The problem is solved in later versions of okhttp3, maybe it'll be updated by the time I actually need to worry about it? Should work fine in the interim though until Archipelago makes the switch to force it.

riktenx:
https://github.com/runelite/plugin-hub/actions/runs/28481008772/job/84417145147?pr=12217

```
osrs-archipelago: plugin uses terminally deprecated APIs:
Do not create fresh OkHttpClients - always @Inject the client's OkHttpClient
```

You can use `.newBuilder()` on the client if you want to customize it etc.

PR #12221, opened 2026-05-26
Add Scammer Alerter plugin

vhdante:
Highlights players on your ignore list with a configurable color overlay. Supports tile highlight, hull outline, overhead name label, and minimap dot options.


tylerwgrass:
Generic player highlighting is not allowed.
See https://github.com/runelite/runelite/wiki/Rejected-or-Rolled-Back-Features

PR #12222, opened 2026-05-26
Add Slayer Task Logger

staticvoid07:
Adds the Slayer Task Logger plugin.

## What it does
- Logs slayer events to `~/.runelite/slayer-log.txt`
- Sends Dink notifications for each event type (requires Dink plugin with external plugin requests enabled)
- POSTs structured JSON to a user-configured webhook URL (optional)

## Event types
- **New task** — fires when a slayer master assigns a task (supports standard and Konar message formats)
- **Task completed** — fires on completion, logs monster, assigned amount, kill count, XP, task streak, and points
- **Task skipped** — fires when a task is cancelled, reads current task from varbits so it correctly identifies tasks swapped in via slayer cape task storage
- **Superior spawn** — fires when a superior slayer creature appears
- **Cape perk proc** — fires when a slayer master offers a favour (supports standard and Konar dialog)

### DIFF
--- a/plugins/slayer-task-logger
+++ b/plugins/slayer-task-logger
@@ -1,2 +1,3 @@
 repository=https://github.com/staticvoid07/slayer-task-logger.git
 commit=58a7a0f2bb1b3603302e2a8c42b7130e7a999ef0
+warning=This plugin can send slayer event data to an external webhook URL configured by the user. The webhook is disabled by default.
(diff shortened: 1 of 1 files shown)


staticvoid07:
Plugin should be ready for review, now that tzhaar tasks are supported I believe every edge case is handled.

riktenx:
file i/o needs to go within a plugin-specific subdirectory in .runelite/

staticvoid07:
moved i/o to ~/.runelite/slayer-task-logger/slayer-log.txt

PR #12237, opened 2026-05-27
Update 3d weather

ScreteMonge:
Fixes NPE


ldavid432:
Since you have the jaco dependency in your build.gradle you will need to use `build=gradle`

riktenx:
please merge https://github.com/ScreteMonge/3D-Weather/pull/39 and include that here, we want to get people off of depending on the "local" jar

riktenx:
actually you can just do that later but do please try to follow up on that when you are able

PR #12238, opened 2026-05-27
update bank-skill-sorting

Frailrain:
Updates pinned commit from `ba84c8e1` (audit session 64 final) to `a318e9a6` — 48 commits of feature work and fixes since the last plugin-hub release.

### Highlights

**Dynamic bank layout refresh** — items now sort into their correct position the moment you deposit or withdraw, no more close-and-reopen. Built on the bank-slot-sync (PR #10512) pattern: `ItemContainerChanged` flags a pending rebuild, `GameTick` consumes it one tick later, `tabInterface.reloadActiveTab()` triggers the full re-read-then-re-render chain.

**LLM-driven item classification** — replaced the prior heuristic regex/stat rules with Sonnet 4.6 classifications cached offline, plus a `mapping.py` override layer for surgical force-include/exclude. Item-meta.json carries per-item tier, slot, weapon class, sections, requirements, and a `setName` field used by cosmetics + crafting Gems for row-per-set layouts.

**Per-tab section redesigns** — fletching split into 11 product-type rows (Tools / Logs / Feathers / Arrows / Arrowheads / Bows / Crossbows / Bolts / Darts / Javelins / Misc), cosmetics row-per-set via a new `set_name` mechanism, crafting Gems row-per-family (uncut → cut → ring → necklace → amulet → bracelet), cooking raw-paired-with-cooked, herblore flattened to 5 sections, hunter creature products split into furs/meats/tertiaries, melee/range/mage gained a Food section that always surfaces to the loadout zone.

**Combat tab zone refinement** — top-N unique requirement levels per section (configurable 1–6, default 3) replaced the static "best 2 per slot" heuristic. Capes / Neck / Rings / Food always Zone 1.

**First-run setup wizard** — chat-message check on first login for Bank Tag Layouts installation + the "Enable layout by default" setting. Side-panel banner mirrors the same check and stays visible until the dependency is satisfied.

**Other** — new plugin icon, Teleports tab added at slot 5, 22-tab layout (Sailing included), wiki-sourced equipment requirement data, ~12k items routed.

### Verification
- `./gradlew build` clean
- Manual in-game test on the dynamic-refresh path: deposit/withdraw cycles all visually sorted to correct positions
- Setup wizard verified against `banktaglayouts` config keys (matches geheur's upstream source)

### DIFF
--- a/plugins/bank-skill-sorting
+++ b/plugins/bank-skill-sorting
@@ -1,2 +1,2 @@
 repository=https://github.com/Frailrain/Bank-Skill-Sorting.git
-commit=ba84c8e17c399136d3aea20474a5f53c618f536e
+commit=2df5b6cc89a19547b74fb8ea21823c53b821bae7


Frailrain:
Hello! Wondering if there was anything I missed that I needed to do to get this latest update approved. Thanks :)

tylerwgrass:
I didn't read the whole thing since you have a ~64000 line diff, but if you are only using file i/o to write debug files you can get rid of that and maybe the bot will review it again.
otherwise you will need to wait for human review

Frailrain:
Thanks for getting back. The file I/O was two dev-only diagnostic tools that wrote debug logs to the .runelite directory. I've removed both along with their supporting code and now-unused imports in this build, and I'll remember, (hopfully), to do so in any future builds haha. 

The plugin no longer does any filesystem I/O. The only remaining resource access is loading the bundled item-meta.json from the classpath via getResourceAsStream, which reads from inside the JAR rather than the user's disk. Build passes and I've pushed the update. let me know if there's anything else you need to see :)

PR #12240, opened 2026-05-27
Add Canoe Hotkeys

blueueberryguy:
Configurable hotkeys for choosing your canoe type at any canoe station - no more clicking between canoe types.


ldavid432:
Use of client.menuAction is prohibited

PR #12241, opened 2026-05-27
Update task-checker

bopsec:
Few new tasks from the past few weeks and moved to new repo

I'll change this to _only_ be new repo, and update it in a later PR


pajlada:
i personally strongly prefer if you'd keep your plugins in separate repositories. it makes looking through changes for reviews & submitting bug reports much easier. not a blocker, so still up to you

bopsec:
> i personally strongly prefer if you'd keep your plugins in separate repositories. it makes looking through changes for reviews & submitting bug reports much easier. not a blocker, so still up to you

Okay fair enough, 
Guess I can keep the bigger plugins in their own repos, I just didn't like having so many different runelite-plugin repositories :p

I'll move the new commits over to the old repo

PR #12243, opened 2026-05-27
Add OSRS Item Browser plugin

Vivialdi:
An item browser sidebar for OSRS, inspired by JEI from Minecraft.

Features:
- Search all ~14,000 OSRS items from the game cache
- GE price + high alch profit/loss
- Crafting recipes with clickable ingredients
- Used to craft (parsed from wiki Products section)
- Drop sources with monster locations and rates
- Shop sources with GE-relative price coloring
- Equipment bonuses (attack/defense table)
- Path to monster spawn (integrates with Shortest Path plugin)
- Recently viewed history + favorites (persisted between sessions)
- Right-click any in-game item → "JEI Lookup"

Wiki data is fetched from the public OSRS Wiki API. All results are cached per session to minimize requests.

### DIFF
--- a/plugins/osrs-item-browser
+++ b/plugins/osrs-item-browser
@@ -1,0 +1,2 @@
+repository=https://github.com/Vivialdi/osrs-item-browser
+commit=da9e0e36db2a5d1dc19908d235d50333de96cc7f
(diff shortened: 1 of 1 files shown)


raiyni:
reflection is not allowed and you need to properly build and submit your plugin https://github.com/runelite/plugin-hub

riktenx:
use of HttpUrlConnection and/or java's HttpClient is not allowed, inject and use runelite's OkHttpClient

PR #12247, opened 2026-05-27
Add Music Visualizer plugin

ScottyOBrien:
Flashes nearby objects in time with the notes of the current OSRS music track. Pulls MIDI from the local OSRS cache via runelite-cache's TrackLoader, parses with javax.sound.midi, flashes each object in player radius on the matching note. Local-only, sends no data. Source: https://github.com/ScottyOBrien/music-visualizer


riktenx:
reflection is not allowed

PR #12259, opened 2026-05-28
Add RuneSpeak

Daranix:
# RuneSpeak

Local AI translation plugin for Old School RuneScape using ONNX models from HuggingFace.
Translates in-game text in real time — NPC dialogues, right-click menus, chat messages, overhead text, dynamic widgets, and more — all running locally on your machine with no data sent to external servers.

## Features
- **Local execution** — runs ONNX translation models (`opus-mt`) on-device via ONNX Runtime
- **Multi-source capture** — NPC dialogs, player chat, right-click menus, system messages, overhead text, widget text
- **Auto-discovery** — dynamically detects new widget types without manual registration
- **~28 language pairs** — models auto-download from HuggingFace based on selected source/target
- **Translation cache** — LRU cache with unlimited option to avoid re-translating repeated text
- **Side panel** — status, translation log, cache management

## Dependencies

- `com.microsoft.onnxruntime:onnxruntime:1.17.1` — CPU inference of ONNX models
- `ai.djl.huggingface:tokenizers:0.36.0` — HuggingFace tokenizer support (with ORT excluded, already declared above)

## Build

- `build=gradle` — requires custom `build.gradle` for fat-jar assembly with native DLLs
- Targets Java 11
- BSD-2-Clause license

Repo: https://github.com/Daranix/RuneSpeak
Commit: `6a63a7f1632aa2a605f421f8a69bf3aed60d672f`


riktenx:
we cannot approve these dependencies because they make your plugin too large

the act of adding _any_ dependency in general also guarantees that your plugin will never be reviewed, because dependency review requires an expert and nobody has time to do so these days.

if you still want to submit this you will need to do so without the dependencies. when you do please take note of https://github.com/runelite/runelite/wiki/Rejected-or-Rolled-Back-Features#forbidden-language-features because there are certain things we allow such as process execution etc.

unfortunately it may turn out that this is not possible in a hub plugin with the technical constraints we enforce

Daranix:
Hey, @riktenx  thanks for the honest answer. I totally get it.

Here's the thing though: this plugin can't really work without those deps. The whole plugin is built around running an actual ML model, onnxruntime runs the neural net, and the HuggingFace tokenizers handle the text preprocessing. There's no way to do that with just the standard library, it's not like I can swap them for a pure Java thing.

So I'm fine waiting. Whenever there's someone available to take a look at the dependencies, I'm happy to go through the review. No rush at all. If there's any extra info I can provide about the deps (where they come from, what they weigh, security stuff, whatever) that'd make things easier when the time comes, just let me know.

About the size:

I've already trimmed the dependencies quite a bit. The build strips out all native libraries (.so, .dll, .dylib, .jnilib), DJL training modules, audio stuff, CUDA utils, CLI tools, and a bunch of other bloat that's not needed at runtime for this specific use case.

For more extra context: the build.gradle has two layers of exclusions that strip out all native bloat. Both the shadowJar task and the hub packaging task exclude:

- All native shared libraries (*.so, *.dll, *.dylib, *.jnilib)
- Debug symbols (*.pdb, *.dSYM/**)
- Static libraries (*.a)
- ONNX Runtime native dirs (ai/onnxruntime/native/**, native/lib/**)
- JNA native binaries (com/sun/jna/**/*.{so,dll,dylib,...})
- DJL audio modality (unused)
- DJL training modules (unused)
- CUDA utilities (unused)
- Apache Commons CLI tools and compress CLI (unused)

After all that, the final jar that gets submitted is 3.2 MB. That's just the Java classes plus the lightweight Java API jars.

The heavy stuff is downloaded at runtime from trusted sources i'm talking about things like:

- ONNX Runtime native libraries,  downloaded on first run from Maven Central (repo1.maven.org), the standard Java package repository. Same place RuneLite itself pulls dependencies from.

- Tokenizer native libraries: also downloaded from Maven Central.

- ONNX model files (encoder_model.onnx, decoder_model.onnx, tokenizer.json, config.json) - downloaded from HuggingFace (huggingface.co/{modelId}/resolve/main/...), the industry-standard model hub.

These are cached locally after download and only fetched once. The code is completely transparent about it - you can see every URL in `NativeLibraryManager.java` and `DownloadFile.java`.

raiyni:
We are not going to allow you to download libraries at run time.

Daranix:
> We are not going to allow you to download libraries at run time.

I'm thinking about rewriting the plugin to use pure Java instead (with the idea of avoid JNI stuff, load libs in runtime etc) of the current dependencies, but I need to be sure about one thing before I start.

The AI models are about 135 MB each, way too big to include in the .jar. So the only option is to download them from HuggingFace the first time they're needed.

Would downloading the model files (the data, not code) from an external source be allowed? If not, there's no point in rewriting it.

Just want to be sure before I put the time into it. Thanks!

riktenx:
downloading data at runtime is allowed

> So I'm fine waiting. Whenever there's someone available to take a look at the dependencies,

jfyi it's not a question of when, it's **if**. we basically just don't review new dependencies anymore

Daranix:
> downloading data at runtime is allowed
> 
> > So I'm fine waiting. Whenever there's someone available to take a look at the dependencies,
> 
> jfyi it's not a question of when, it's **if**. we basically just don't review new dependencies anymore

ok knowing that i'll work around it thanks! 

Do you prefer I close this PR and open a new one when I have it or prefer to keep this one open?

riktenx:
just submit a pr when you have your revised solution

PR #12263, opened 2026-05-28
Update click-to-minimize

WaylonJ:
[Addition] Added new config option to allow window to be sent in to back of windows. This will help prevent alt-tabs immediately bringing the window back to focus.


ldavid432:

WaylonJ:
Added 

> build=standard

 to runelite properties

PR #12267, opened 2026-05-28
Create darkwintertodt

kulcris:
New Plugin
Reduces the brightness of the Wintertodt arena.


coopermor:
Reflection is not allowed

kulcris:
> Reflection is not allowed

Clobbered Reflection. 
Reworked the plugin.

PR #12277, opened 2026-05-29
Add Arachnophobia Mode

D4nders:


riktenx:
we don't allow direct id input in config lists. name-based input is fine but you appear to be doing both.

(we understand that plugins exist today that allow ID input such as "better npc highlight". the rejected list is not strictly retroactive and some plugins are grandfathered.)

D4nders:
ah that is my bad, I see that the id check which falls back to names is left from the earliest version I tried. Should be a quick fix

PR #12283, opened 2026-05-30
Add Advanced Skybox

ixl911ixl:
Adds Runesky Video Player, a RuneLite plugin for displaying user-supplied local media as a configurable overlay.

The plugin does not download, stream, host, or provide media. Users supply all files locally.


ixl911ixl:
Project temporary closed as the way I have tried to handle audio playback is not supported. I will return with a new approach.

ixl911ixl:
Update on audio implementation:

The original implementation used Java's standard audio playback classes for local media playback. During Plugin Hub validation, that approach was flagged because Plugin Hub plugins are expected to use RuneLite-approved client APIs for audio handling.

This version has been revised to use RuneLite's `net.runelite.client.audio.AudioPlayer` instead. The plugin still uses user-supplied local media only, but audio playback is now routed through RuneLite's supported audio system rather than direct Java sound APIs.

Current behavior:
- `audio.wav` is still expected inside the selected media pack folder.
- Playback starts through RuneLite's audio system when media playback begins.
- Video timing remains controlled by the plugin's playback clock.
- Advanced media-player controls such as exact audio seeking, pausing, and looping may be limited by RuneLite's public audio API.

This change was made to better align the plugin with Plugin Hub requirements while preserving local user-supplied audio support where possible.

ixl911ixl:
Hahahaha, okay. No. This audio system has failed spectacularly. 

I'll revisit this when I know what the hell I'm doing.

ixl911ixl:
Completely retooled the project to be a video-only option. Audio support on Runelite currently is a joke, you can't do anything useful and it just layers on top of itself like crazy. 

I've decided this will be better suited for trippy skybox visuals (visualizers) while you play your own music locally, or movies/tv shows that have hardcoded CC's. 

Audio isn't possible

ixl911ixl:
Pretty much a proof of concept at this point, the plugin works but without audio it feels pretty pointless.

coopermor:
> The original implementation used Java's standard audio playback classes for local media playback. During Plugin Hub validation, that approach was flagged because Plugin Hub plugins are expected to use RuneLite-approved client APIs for audio handling.

You can use javax.sound if there is a reason that the AudioPlayer API does not support the necessary customization. It is only for plugins that use a basic audio implementation that they should use AudioPlayer.

riktenx:
* as cooper said your usecase requires beyond what AudioPlayer can do so you are welcome to use the sound APIs directly
* **please remove the .class files from your repo. this submission cannot be reviewed until you have done so**

riktenx:
stale, if you end up fixing this you can make a new pr

PR #12292, opened 2026-05-30
Add Combat HUD plugin

upsetsingleevent:
Adds a compact, display-only HUD that floats above the local player showing:

- **Buff timers** (e.g. antifire, stamina, stat boosts) that flash when near expiry
- **Attack cooldown** in ticks, with automatic weapon-speed detection from the equipped item
- **Spec indicator** — a dot + energy % that lights up when the threshold is met

The plugin is strictly read-only. It reads game state (varbits, skill levels, equipment slot) and draws text/shapes via an overlay. It never injects input, modifies menus, sends clicks, or interacts with the game in any way.


riktenx:
stale, if you end up fixing this you can make a new pr

PR #12303, opened 2026-05-30
Improved loot tracker

Oufqt:
Updates Loot Table Viewer to the latest plugin commit.


riktenx:
reflection is not allowed

PR #12315, opened 2026-05-30
update tasks-tracker - use standard build property

perterter:
No plugin updates, just using the `build=standard` property for auto reviews


perterter:
@riktenx I did the thing. Thanks for calling it out. Also we appreciate all you did for us during the league to keep the updates moving. :)

PR #12333, opened 2026-06-01
Update 07flip to b80785d

UserD40:
Update 07flip plugin to commit b80785d.

### DIFF
--- a/plugins/07flip
+++ b/plugins/07flip
@@ -1,3 +1,3 @@
 repository=https://github.com/UserD40/Runelite07Flip.git
-commit=091ea1b119e16baa043176134402d039e8c64de3
+commit=b80785daf6ebedc8dfcaa03c7af24197b8801311
 warning=This plugin submits your IP address to a 3rd-party server not controlled or verified by RuneLite developers.


UserD40:
Summary of changes in this update (commit `b80785d`, since the previously listed `591f1eb`), for reviewer context:

**Optimiser (GE flip planner) — fill sync**
- In-progress GE offer fill progress is now synced to the plugin's shared session as a single consolidated entry per slot leg (running quantity + quantity-weighted average price) instead of one entry per fill tick, so progress reflects live and the terminal fill folds into the same entry without double-counting.
- Added a low-frequency background refresh (45s) of the session so an existing plan is picked up regardless of which tab is open. This is well under the endpoint rate limit (~1.3 req/min/user).

**Grand Exchange price overlay**
- "Margin / item" now shows the after-tax figure (GE 2% tax, capped at 5M/item, exempt under 100 gp) and is computed from the displayed buy/sell prices so it reconciles with the rows shown above it.
- The locked sell price now acts as a floor: when the live market rises above it, the higher live price is shown instead of pinning the lower locked value.

No new third-party dependencies, and no use of reflection, JNI, or external process execution. Data handling is unchanged in kind from the listed version — the only external calls remain read-only price fetches from 07flip.com plus the existing optimiser/trade sync endpoints (the user's own GE activity); this update only changes the granularity and representation of that existing sync.

Adam-:
There is a per-plugin review limit if about 10k tokens/hour, you are submitting plugins for review at a rate of about 32k tokens/hr, and so the bot is rate limiting your PRs. This is to keep it fair so that other people can get reviews, too. Stop opening new PRs and just wait for the previous one to go through, or update the previous one. You have three now: 
image

UserD40:
Thanks for flagging the review rate limit - understood, and apologies for the extra load on the queue. We're closing this PR and our remaining open branch now, and we'll hold off resubmitting until the per-plugin budget has reset. Going forward we'll keep a single open PR and update that one in place for each change rather than opening a new PR per update. Thanks for your patience.

PR #12342, opened 2026-06-01
Add Pump Coker Alert plugin

MeatballAWM:
Alerts when blast furnace pump is low on coke by identifying when the stove's object ID is 9085.


riktenx:
the file goes in plugins/

riktenx:
https://github.com/runelite/plugin-hub/pull/12404

do not open new prs in the future. just push updates to your current one

PR #12343, opened 2026-06-01
Lending Tracker: fix cross-machine invite codes & sync reliability

Guess34:
Patch update

- Fixed invite codes failing to join across computers.
- Cloud Sync now activates immediately when toggled on (no restart).
- Invite codes are confirmed published before sharing; clearer join error messages.
- Added a relay keepalive and catch-up sync so joiners get current group state.
- "Invite to Lending Group" now anchors on the player "Report" option.

Cloud Sync remains opt-in and off by default. No new dependencies.

### DIFF
--- a/plugins/lendingtracker
+++ b/plugins/lendingtracker
@@ -1,2 +1,2 @@
 repository=https://github.com/Guess34/Lend-Borrow_Tracker.git
-commit=af437e4ac00ca30c9e8675141125ca0978053bbb
+commit=9cc974928a970a0eaa916a7428d23ec40bd3dcea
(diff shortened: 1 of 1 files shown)


riktenx:
* thread sleep is not allowed
* thread interrupt is not allowed

riktenx:
get rid of this https://github.com/Guess34/Lend-Borrow_Tracker/blob/e52dbe4d7d6060dd199783d94e363d6fb7db3b5f/src/main/java/com/guess34/lendingtracker/services/ProofScreenshot.java#L166 and use the draw manager instead https://static.runelite.net/runelite-client/apidocs/net/runelite/client/ui/DrawManager.html#requestNextFrameListener(java.util.function.Consumer)

riktenx:
you need much stricter gating of requests to your backend behind the opt-in flag. every api call should have a check for the optin config right before it, that throws an exception / returns early if it happens to be false, so it is exhaustively clear from reading the code that you're not making requests without user consent.

PR #12357, opened 2026-06-02
Add ClanFinder

OSRSClanFinder:
Adds ClanFinder, a read-only RuneLite sidebar plugin for browsing approved OSRS clan recruitment listings from osrsclanfinder.com.

The plugin displays public listing data, listed member totals, requirements, event information, and banner thumbnails. Member totals are listing data from Wise Old Man group stats or manually entered counts, not live online clan/member status.

The plugin does not automate gameplay, inject game menu entries, send chat messages, join clans, request RuneScape credentials, submit RSNs, or scrape private clan data.


ldavid432:
Your image is still too big. You need to actually resize the image file. Adding a function to limit the size is not an acceptable solution.

OSRSClanFinder:
Thanks for this, i've now reduced the file size to 30x29, 702 bytes

PR #12366, opened 2026-06-02
Add Waypointer plugin

Mitchole:
Adds Waypointer, a plugin for saving in-game locations and pathing to them.

- Repository: https://github.com/Mitchole/waypointer
- License: BSD 2-Clause

What it does: save tiles as named waypoints, organise them into categories, search the library, and one-click path to any waypoint through the Shortest Path integration. It also ships a small curated set of preset locations and a nearest-landmark bar.

### DIFF
--- a/plugins/waypointer
+++ b/plugins/waypointer
@@ -1,3 +1,3 @@
 repository=https://github.com/Mitchole/waypointer.git
-commit=edd3e33119383748cbba493af02a815d3a89aa87
+commit=0f38c5115012d267305f5d46fc43815caf05afea
 authors=Mitchole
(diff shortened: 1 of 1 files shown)


riktenx:
* runtime shutdown hook is not allowed
* file i/o will make your plugin require manually review. if you can not use it your plugin can be automatically reviewed. i did not read your code, idk how feasible that is but it seems like if you're just saving paths it should be possible to do so in config

Mitchole:
Thanks for taking a look! Appreciate the info. I've gone ahead and made the changes:

- Killed the shutdown hook. It was only there to flush a pending save if the client got closed mid-write, but saves happen on every change now so there's nothing left to flush.
- Dropped all the file I/O. Waypoints live in the config now (per-RS-profile via setRSProfileConfiguration), routes in the config group too, same approach as ground markers / loot tracker. Import/export is just copy/paste share codes instead of reading/writing json files, and I ditched the "open data folder" menu item while I was at it.

No java.io / java.nio left in the plugin at all now, so it should be good for auto review. Let me know if anything else needs changing!

PR #12367, opened 2026-06-02
Update 07flip to b80785d

UserD40:
Update 07flip plugin to commit b80785d.


UserD40:
What's in this update:

- **GE price overlay now shows the after-tax margin.** Previously the overlay displayed the pre-tax figure, so the on-screen projected profit didn't match what actually landed after the Grand Exchange tax. It now accounts for the tax so the number reflects real take-home margin.
- **Item page can now trigger a Grand Exchange buy.** From an item's detail page you can now initiate the GE buy directly, rather than having to set it up separately.

Thanks for taking a look.

PR #12369, opened 2026-06-02
update fashionscape to 1.4.2

equirs:
so I refactored all of the core stuff in my plugin several months ago but stupidly did it all in one massive commit, which (understandably) nobody wanted to review. I still want to get my changes out somehow but have been torn on how to do that exactly. What I'm hoping to do now is introduce most of the new classes in my refactor as dead code for now, and then I'll integrate them incrementally once they're all added.


riktenx:
use of serializationutils stuff is not allowed

PR #12371, opened 2026-06-03
update `collection-log-master` to v2.0.0

rmobis:
This update is mostly to migrate from using a local save and logic to instead relying on [TaskApp](https://www.osrstaskapp.com/) as the backend. Sorry for the big diff, I tried to change as little as possible for this first release.

### DIFF
--- a/plugins/collection-log-master
+++ b/plugins/collection-log-master
@@ -1,4 +1,4 @@
 repository=https://github.com/OSRS-Taskman/collection-log-master.git
-commit=456ad3bf3f6073ae1a796f543489e931eed90fe8
+commit=d122b39b071d2ec4d1d6d205973b06f47ccc758d
 authors=ImTedious,rmobis
 warning=This plugin submits your IP address to a server not controlled or verified by the RuneLite developers.


riktenx:
* your build in runelite-plugin.properties should be standard
* ~~use of client.menuAction is not allowed~~

rmobis:
> * your build in runelite-plugin.properties should be standard

I kept it as `gradle` so I can read the version from a unified place, as per [this recommendation](https://discord.com/channels/301497432909414422/419891709883973642/1501214630746259566) I read in Discord. It was [mentioned a bit earlier](https://discord.com/channels/301497432909414422/419891709883973642/1501209672051589170) that there might be plans to expose the `runelite-plugin.properties` version field later on so I plan to migrate to `standard` once that's done. Is that ok?

More specifically, so I could keep this in `build.gradle`:

```gradle
processResources {
    filesMatching ('**/version') {
        filter {
            it.replace('%PLUGIN_VERSION%', version)
        }
    }
}
```

riktenx:
that is available now e.g. https://github.com/riktenx/loot-filters/blob/main/runelite-plugin.properties#L7, please migrate to that

rmobis:
> that is available now e.g. [riktenx/loot-filters@`main`/runelite-plugin.properties#L7](https://github.com/riktenx/loot-filters/blob/main/runelite-plugin.properties#L7), please migrate to that

The property was already available back then, what was lacking was a way to access that value in the plugin's code. Has that been implemented? Tried looking at recent commits in `runelite` repo, or other code in the the plugin repo you linked, but couldn't find anything. Can you point me in the right direction here?

I've tried `ExternalPluginManager.getDisplayData(CollectionLogMasterPlugin.class).getVersion()`, but `getDisplayData` returns `null`.

Adam-:
> ExternalPluginManager.getDisplayData(CollectionLogMasterPlugin.class).getVersion()

`ExternalPluginManager.getDisplayData(CollectionLogMasterPlugin.class).getVersion()` works and isn't new, it is from 2019. It returns null on dev builds however.

rmobis:
> > ExternalPluginManager.getDisplayData(CollectionLogMasterPlugin.class).getVersion()
> 
> `ExternalPluginManager.getDisplayData(CollectionLogMasterPlugin.class).getVersion()` works and isn't new, it is from 2019. It returns null on dev builds however.

Will it work if I package it using the `shadowJar` task? Mostly just so I can test it.

Adam-:
> > > ExternalPluginManager.getDisplayData(CollectionLogMasterPlugin.class).getVersion()
> > 
> > 
> > `ExternalPluginManager.getDisplayData(CollectionLogMasterPlugin.class).getVersion()` works and isn't new, it is from 2019. It returns null on dev builds however.
> 
> Will it work if I package it using the `shadowJar` task? Mostly just so I can test it.

I don't think so.

rmobis:
> > > > ExternalPluginManager.getDisplayData(CollectionLogMasterPlugin.class).getVersion()
> > > 
> > > 
> > > `ExternalPluginManager.getDisplayData(CollectionLogMasterPlugin.class).getVersion()` works and isn't new, it is from 2019. It returns null on dev builds however.
> > 
> > 
> > Will it work if I package it using the `shadowJar` task? Mostly just so I can test it.
> 
> I don't think so.

Ok so, since I won't be able to test this myself, can you confirm these for me please?

1. Once the plugin is built and distributed through RuneLite, `ExternalPluginManager.getDisplayData(CollectionLogMasterPlugin.class).getVersion()` will return me the `version` property value I set in `runelite-plugin.properties` file at the root of my plugin repository?
2. Is this the recommended way to get that?

riktenx:
yes

rmobis:
Had to give up on a few nice jetbrains annotations, but build has been migrated to `standard`.

PR #12377, opened 2026-06-03
Update kill-clog to 1.4.1

420kc:
Updates Kill Clog to 1.4.1 (edited from 1.4.0)

Changes:
- Adds RuneProfile collection-log lookup for public player data.
- Expands comparison mode summaries and tooltip polish.
- Adds combat achievement tier display inside PvM summaries (local and RuneProfile-synced public)
- Refreshes README screenshots and plugin metadata.
- Adds "lookup" as opt-in option for right click menu text
- adds configuration toggles for right menu locations
- moves comparison mode into default search bar instead of 2nd search bar
- adds support for Resource Packs (border sprites and fill for tooltips)
- adds GOTR rift count to skill summary

Local verification:
- .\gradlew.bat compileJava checkstyleMain checkstyleTest test --rerun-tasks

### DIFF
--- a/plugins/kill-clog
+++ b/plugins/kill-clog
@@ -1,3 +1,3 @@
 repository=https://github.com/420kc/kill-clog-plugin.git
-commit=ee0262a52e27374f808af7a2b076a061848b7b38
+commit=ae2ff40aee46d4e5524ee163345fefc46119311e
 warning=This plugin submits your IP address to a 3rd-party server not controlled or verified by Runelite developers.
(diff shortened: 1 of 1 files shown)


riktenx:
* thread interrupt is not allowed
* client.menuAction is not allowed

420kc:
> * thread interrupt is not allowed
> * client.menuAction is not allowed

Thanks! I’ll fix the thread interrupt.

For client.menuAction, does that effectively rule out bulk capture for clog plugins, or would a manual flow still be acceptable where Kill Clog prompts the user to click Search and Back themselves?

420kc:
I think this should be clean now, follows RuneProfile's search/back sync, no client.menuAction + updated readme + snuck in support for resource packs on the resubmission

PR #12379, opened 2026-06-03
Update immersive-dialogue

DjilanoS:
**TLDR**
Adds mouse interaction and repositioning to the relocated dialogue panel. All existing native keyboard controls (Esc / Spacebar / 1–5) still work — clicking via menuActions API is additive.

**Changed**

- New defaults: `backdrop padding 4, avatar backdrop on, bottom margin 150, animate head on, drag mode on`.
- Options no longer render the `[1]–[5]` number prefix / "use keys" hint (click and native number keys both still select).
- Continue prompt now reads `"Click to continue"`.
- Position sliders widened (`bottom margin 0–2160, horizontal offset ±1920`) and drag is clamped to keep the box on-screen at any resolution.

### DIFF
--- a/plugins/immersive-dialogue
+++ b/plugins/immersive-dialogue
@@ -1,2 +1,2 @@
 repository=https://github.com/DjilanoS/osrs-immersive-dialogue.git
-commit=2a10330dd6c5712b4ebea1f6b5f498cb2d232406
+commit=5d7c70897e5b97d0ca0e5c0f505c031f4b4619be
--- a/plugins/immersive-dialogue
+++ b/plugins/immersive-dialogue
@@ -1,2 +1,2 @@
 repository=https://github.com/DjilanoS/osrs-immersive-dialogue.git
-commit=a0fe42fe540b4fbc2ece8dbd53c345640237f97c
+commit=2a10330dd6c5712b4ebea1f6b5f498cb2d232406


DjilanoS:
I have a feeling Runelit hub checks failed due to the user of `menuActions`, if so here's a bit more information on how and why I use `client.menuAction(..)`:

A left‑click on that panel is forwarded to the game's own continue / option‑select action via `client.menuAction(..., MenuAction.WIDGET_CONTINUE, ...)`, targeting the native dialogue widgets (`ChatLeft/ChatRight.CONTINUE, Chatmenu.OPTIONS`) by their component IDs.

This affects only **non‑combat conversation UI**; it grants **no** information, efficiency, or capability the player doesn't already have, only the convenience of clicking where the dialogue is shown.

The `client.menuAction(..)` call results in the same server interaction the native click it mirrors would normally cause  so no extra or different packets are sent to the server, and nothing is sent unless the user clicks.

See [DialogueWidgetController.java#L441](https://github.com/DjilanoS/osrs-immersive-dialogue/blob/main/src/main/java/com/immersivedialogue/DialogueWidgetController.java#L441) for **"continue to click"** usage.

See [DialogueMouseListener.java#L166](https://github.com/DjilanoS/osrs-immersive-dialogue/blob/main/src/main/java/com/immersivedialogue/DialogueMouseListener.java#L166) for **"select options"** usage.

**Feature in action:**
Screen Recording 2026-06-03 at 12 06 25 PM

riktenx:
yeah, use of client.menuAction is not allowed for this

i understand this usage may seem benign but it's a question of accidentally tripping jagex's macro detection, which programmatic server interactions outside of the vanilla jagex client code paths will do. we cannot host features that might get users banned

DjilanoS:
> yeah, use of client.menuAction is not allowed for this
> 
> i understand this usage may seem benign but it's a question of accidentally tripping jagex's macro detection, which programmatic server interactions outside of the vanilla jagex client code paths will do. we cannot host features that might get users banned

Allright, totally understand! Was worth a try.

- Removed: `menuActions` that supported click actions all together
- Rewrote README
- Added support for System message dialogues in the custom dialogue box

PR #12380, opened 2026-06-03
Update Valiance Clan Plugin to 1.1.3

Sicryption:


riktenx:
* your `build` in runelite-plugin.properties should be `standard`
* use of client.menuAction is not allowed

Sicryption:
> * your `build` in runelite-plugin.properties should be `standard`

Will make this change

>     * use of client.menuAction is not allowed
This must be a new ruling. I'll look what other plugins are doing for taking Collection Log data now. 

Edit: Looks like temple is still using client.menuAction. Any suggestions of plugins with new behavior for this?

riktenx:
> * use of client.menuAction is not allowed

this may actually be ok, although if there is a way to accomplish whatever you're doing _without_ menuAction please let me know

Sicryption:
> > * use of client.menuAction is not allowed
> 
> this may actually be ok, although if there is a way to accomplish whatever you're doing _without_ menuAction please let me know

When opening the collection log, only the items which are on the current visible menu are able to be grabbed. Clicking the search button loads the data for all items regardless of page. So, what we are doing here is creating a fake search button click when the Collection Log opens, in order to load all of the item data.

Temple's plugin has a button you click on to sync it where as this plugin does it once the Collection log loads, without user input. We can add a button to map it to if that feels more correct. Otherwise, we'll just be asking our users to click the search button which is a little jank but we can live with.

Sicryption:
Accidentally rebased off an old master causing the above issue. Didn't need to rebase at all. Ooops

ReinhardtR:
> > * use of client.menuAction is not allowed
> 
> 
> 
> this may actually be ok, although if there is a way to accomplish whatever you're doing _without_ menuAction please let me know

This is the same case that I had for RuneProfile, I thought it was a total ban of using menuAction. But it sounds like some cases are allowed? Would love to add it back to RuneProfile if this specific case is allowed as it's a drastically better UX.

riktenx:
@ReinhardtR yes the use case of specifically doing the clog sync thing is ok, sorry

PR #12385, opened 2026-06-03
Add detect-auto-alchers plugin

FalseProfit:
Adds Detect Auto Alchers, a local-only informational plugin that highlights nearby players based on observable alchemy-like activity, fire-staff evidence, and hiscores lookup.

Review notes:
- Does not click menu options or automate gameplay.
- Does not automatically submit reports.
- Does not upload, crowdsource, or expose player information over HTTP.
- Uses RuneLite/client APIs and local files only.
- No third-party runtime dependencies.
- `build=standard`.


riktenx:
thanks for the submission but we are not interested in hosting any more heuristic-based "cheating" detectors due to concerns of player harassment

PR #12386, opened 2026-06-03
Update 07flip to 57852b7

UserD40:
Update 07flip plugin to commit 57852b7.


UserD40:
What's in this update:

- **Fixed the GE buy price-fill.** When buying an item from a search result (and now from the item insights page), the plugin was pre-filling the high/sell-side price. It now fills the low buy-side price - what a flipper actually sets their buy offer to.
- **Added a setting to disable auto price-fill.** A new "Auto-fill GE price" option (on by default) lets users stop the plugin from writing to the GE custom price input. With it off, the price overlay still shows the recommended buy/sell prices for reference, but nothing is typed automatically.

Thanks for taking a look.

PR #12388, opened 2026-06-03
update drop-rates-clog

pairofcrocs:
Automated update of drop_rates.json from the OSRS wiki.


coopermor:
If you're frequently going to change the JSON you should store it in a different branch or repo. Then have your plugin download the file on start. Then you don't need to have the plugin reviewed for every drop rate change.

pairofcrocs:
I was under the impression that this wasn't allowed.

Obviously this is a much better approach. Can I get confirmation before I make the change in the code?

Thanks!

coopermor:
Yes this is allowed. Network requests to github, discord, and the OSRS wiki do not need a warning about IP addresses. So if your json is hosted on github you don't even need to add a network warning.

pairofcrocs:
That's perfect, thank you!

I have a larger release coming soon, so I'll push the change along with that.

PR #12400, opened 2026-06-03
Rehaul ultimate-normie-mode

aaronslindner:
Removed leaderboard, added roguelike upgrade side panel


aaronslindner:
The leaderboard/data-submission feature has been fully removed in the current version - the plugin no longer makes any network calls or contacts any 3rd-party server. It only uses the RuneLite API and local config storage (wealth is calculated locally for UI display only). That's why this PR drops the `warning=` line. Happy to restore it if you'd prefer to keep it for any reason.

PR #12408, opened 2026-06-04
Create zebak-roar-helper

FHaisal:
A plugin to highlight jugs and safe tiles during Zebak's Great Roar


ldavid432:
Automatic indication of where to stand is not allowed per https://secure.runescape.com/m=news/third-party-client-guidelines?oldschool=1
> Automatically informs you where or where not to stand in a boss fight

PR #12418, opened 2026-06-05
Update 07flip to 5e5da18

UserD40:
Update 07flip plugin to commit 5e5da18.

### DIFF
--- a/plugins/07flip
+++ b/plugins/07flip
@@ -1,3 +1,3 @@
 repository=https://github.com/UserD40/Runelite07Flip.git
-commit=57852b782b03e74a7bf733f13d5900bde3a0bf4c
+commit=5e5da18cf9fba2e2c334d6fafa75c84ecc8b9dc6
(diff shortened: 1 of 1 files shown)


UserD40:
What's in this update:

- **Tier-correct GE price fill.** The auto-filled buy/sell price now follows the account tier everywhere (flips list, item search, sell setup): accounts get 07Flip's recommended prices, free accounts get the live market prices. Backed by new v2 API endpoints that return both price sets.
- **Fresh prices at click time.** The queued price is re-resolved from the item endpoint when the user initiates a buy/sell, so a stale sidebar row can no longer pre-fill an outdated price.
- **Favourites fixes.** Favourites are no longer hidden by the capital filter, and the empty-state messages now say precisely why a list is empty.
- **Cleaner onboarding banners.** The free-account upsell and the link-your-API-key prompts are now compact cards, with the key instructions pointing at the account page.

Thanks for taking a look.

PR #12256, opened 2026-05-28
Auto Split Manager - Large update

dev-elie:
## Dear maintainers, take your time with approving this pr
The current plugin does what it needs to do decently well. This pr is kindof a 1.1 version, it adds bunch of nice to have things. So, if this gets approved in 3 months I dont care, i know you guys are flooded with pr's.

## Plugin repo changed
Yea, i changed my GitHub username

## Changes

- Refactored code
- Added History management, user can now view past sessions and change what occurred during those sessions.
   - Users can now also drag/move splits, join/leave entries around in the current/past sessions 
- Added Ge tax deduction
- Added pop-out mode 
- Added Graphs and such to view statistics of current and past sessions. Graphs
   - GP/hr over time
   - Highest earnings
   - Settlement balance
- Various bugfixes, small improvements and tweaks
- Added chat detection regex to the config 
- [ ] Updated readme <-- **can you guys update the [runelite page](https://runelite.net/plugin-hub/show/auto-split-manager) : )** 
- [ ] Session data no longer saves to config but to disk instead

PR #12419, opened 2026-06-05
Auto Split Manager - repo url changed

dev-elie:
I changed my GitHub username

Also added `build=standard` changed my username here and added version number

Also see https://github.com/runelite/plugin-hub/pull/12256


dev-elie:
Oopsy, did not ment to push that here

PR #12424, opened 2026-06-05
Squeaky toy v2.3

H-Mill:
- Add two SFX triggers:
  - Player Kill (triggers when you kill another player)
  - Player death (triggers when you die)
- Add 5 more built in SFX (mostly death sfx)
- Move toggle options into collapsible section
- Add Global attack section
- Add toggles to Received Attacks section


H-Mill:
@raiyni can you specify which SFX exactly, some of these are widely used across youtube videos, etc. I've also got a branch that adds the ability for people to use their own sound clips, however it has to use some file i/o to read the .wav files. I could include that in this PR, but last time I had it in a PR, it sat here for over a week

PR #12425, opened 2026-06-05
Create zebak-roar-helper

FHaisal:
Updated the plugin to remove all tiles on the ground, it only highlights helpful jugs and does not show the player where to stand or move. As per Jagex's guidelines here: https://secure.runescape.com/m=news/third-party-client-guidelines?oldschool=1

### DIFF
--- a/plugins/zebak-roar-helper
+++ b/plugins/zebak-roar-helper
@@ -1,0 +1,2 @@
+repository=https://github.com/FHaisal/zebak-roar-helper
+commit=94755c6bb4fb5877bd9d7a490a283d349bacf35a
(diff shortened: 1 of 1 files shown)


ldavid432:
Perhaps I should have been more detailed in my response last time. To be honest I don't think you have almost any features in this plugin that will be acceptable

jug highlighting

roar countdown timer
> Attack counters

The jug true tile is fine, although you can do that already with NPC indicators.

FHaisal:
Ah okay my bad, didn't see those, happy to close this one then. I'll stop with the Zebak development, thanks.

PR #12445, opened 2026-06-06
Update iron-value plugin to 1.3 (ores, nature runes prices)

hawkins:
Adds ore prices and nature rune prices thanks to @Diapolo10.

Also updates the plugin structure with new changes to `example-plugin` since I haven't contributed this plugin in several years


hawkins:
Oh; I see I'm using deprecated APIs, will resolve and re-open

PR #12452, opened 2026-06-06
update captains-ledger

windyr12:
Reworked plugin to be more flexible


windyr12:
Updated to reduce size

PR #12453, opened 2026-06-06
Update bh-pages - fixes

Un4gott3n:
Fixed the scrolling issues on the history tab to work correctly, fixed issue where herblore boost isn't accounted for in Surge Pot checks & fixed issue where boss KC panel doesn't show on client restarts when enabled.

Just a few fixes before I work on the next version.

Thanks!


Un4gott3n:
I've made a small change to the runelite-plugin.properties to include "build=standard" as I've seen in replies to other pull requests. Let me know if anything else needs changing.

Thanks.

PR #12457, opened 2026-06-06
Update Blert to 0.9.11

frolv:
Fixes spurious spawn/death events with Verzik tornadoes and Sotetseg.

### DIFF
--- a/plugins/blert
+++ b/plugins/blert
@@ -1,4 +1,4 @@
 repository=https://github.com/blert-io/plugin.git
-commit=6cbc286b3cb9600cb3f4b094e9fb1cfe78c6c991
+commit=965d03eedd8a179c51a8b3723222ca53501372f2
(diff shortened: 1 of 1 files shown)


riktenx:
doesn't build

frolv:
It looks like git is no longer available in the builder. I use it to get the plugin build commit hash to send to my server when opening a socket; is there an alternative method (such as a builder env var)?

Adam-:
> It looks like git is no longer available in the builder. I use it to get the plugin build commit hash to send to my server when opening a socket; is there an alternative method (such as a builder env var)?

This might be due to the ubuntu 22 to 24 upgrade we did recently. We just use the stock GitHub runners+images so if it isn't available you will probably need to figure out a different solution.

frolv:
Alright, I'll look into alternatives

PR #12486, opened 2026-06-08
Add MMO HUD and Loot Chat

jaydog990:
Adds two external plugins:

- MMO HUD: MMO-style player, target, and RuneLite party HUD frames. Known issue documented in the plugin README: after hopping worlds, the player portrait orb can appear blank until the player changes equipment; HUD bars continue to update.
- Loot Chat: colored NPC drop chat messages with configurable value tiers and saved kill count context.

Both plugin repositories are public and use build=standard with no third-party dependencies.


riktenx:
submit one pr per plugin

PR #12488, opened 2026-06-08
Update Hiscore Notifications

malafel-dev:
Implements:
- Notify players when they gain boss hiscore ranks.
- More aggressively rate limit requests to the hiscore server.
- Allow players to completely disable notifications for rank ranges by setting the interval value to 0.


malafel-dev:
Sorry if someone already started reviewing this, or if my recent change complicates the merging process. 

I thought of a good way to prevent the concerns around spamming hiscore servers, and thought it's important to get that change in with the latest update. Requests to the hiscore servers will not happen for a skill or a boss until the plugin has reason to believe that the player is training or fighting the respective skill or boss.

I do not expect to make further changes to this PR without request by the plugin hub maintainers.

Thanks.

PR #12499, opened 2026-06-09
update tyrs guard clan

torlokt:
Fixed GE clan listing sync — the overlay and panel now correctly update in real time for all clan members simultaneously. Previously only one plugin client could be connected at a time, causing broadcasts to only reach the most recently logged in member. Multiple clients can now stay connected at once. The "I Listed It!" button is now disabled and labelled "Already Listed" when another member has the clan listed, and automatically re-enables for everyone when the lister logs out or hops worlds.


riktenx:
image

PR #12503, opened 2026-06-09
Update runecrafting tracker

mcscoobe:


riktenx:
you should only be updating one plugin

mcscoobe:
> you should only be updating one plugin

What the hell. I have no idea what I did but perhaps the 2am plugin fixes is not the strategy. Will figure this out asap.

PR #12504, opened 2026-06-09
Update gim hub to 1.6.0

wouterrutgers:
* Sends the timezone to the server so we can show the player's local time on the dashboard.
* Fixes an issue where data would be sent even when there were no changes.


riktenx:

wouterrutgers:
Done!

PR #12507, opened 2026-06-09
disable nh-input-tweaks

riktenx:
plugin automates input by invoking the client script that triggers tab switches


NotCoco:
@riktenx Could I get clarification on this disable reason?

My understanding is that the plugin maps one real physical F-key press to one normal tab switch. It does not generate repeated input, choose actions, click items/spells/prayers, interact with menus, or do anything without a user keypress.

The goal is input latency/queueing improvement, not gameplay automation. Is the issue categorically that it invokes the tab-switch client script, or are there constraints/guards that would make this acceptable?

PR #12510, opened 2026-06-09
Add Nightmare Prayer Shuffle Indicator

iiceJ:
Adds a plugin that indicates when The Nightmare or Phosani's Nightmare shuffles your protection prayers. Well within the Jagex rules.


coopermor:
Jagex prohibits any feature that:

This is not "Well within the Jagex rules."

PR #12524, opened 2026-06-10
Insert Tangle Crew Plugin File

ZH-Systems:
This is a plugin based on the Dink plugin, it is forked off of the most recent Dink Plugin. Essentially it expands capabilities to the plugin to allow for a JSON payload to be synced from a web server to update specific fields within the plugin during clan events, and disabled and reverted to clan standards when event ends.


ZH-Systems:
Incorrect project merged

PR #12531, opened 2026-06-10
OSRGB

tar14444:
OSRGB integrates RuneLite with OpenRGB allowing for tons of lighting effect triggers.

### DIFF
--- a/plugins/osrgb
+++ b/plugins/osrgb
@@ -1,2 +1,2 @@
 repository=https://github.com/tar14444/OSRGB.git
-commit=61050101b2664d52949693ad3d52e1cc3022ad06
+commit=a22861835ae184bc8ebe6037200384a4fdad9101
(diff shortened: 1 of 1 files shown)


tar14444:
OSRGB v2.0 released.

New features:
- Collection Log effects
- Quest/Miniquest/Diary effects
- HP value based color system
- Critical HP animation overhaul
- Flash, Pulse, Strobe, Breathing, Rainbow, Fire, and Lightning styles

Latest source available in the repository.

tar14444:
OSRGB v3.0 released.

- Replaced external helper EXE with direct OpenRGB SDK connection
- Added configurable OpenRGB host and port
- Removed Python/Flask/helper dependency
- HP colors and critical effects now run directly through Java
- Restored all RGB event effects

riktenx:
thread sleep / interrupt are not allowed

riktenx:
thanks for the submission but we are not able to host plugins that explicitly require and direct a user to download an external bespoke native binary for security reasons

PR #12537, opened 2026-06-10
Add Player Circles plugin

iiceJ:
Draws a configurable circle or ring under players.

- 3D disc rendered in-world (occluded by models naturally)
- 2D overlay ring with solid or dashed outline
- Per-type colors for self, friends, clan, team, and PVP
- True tile support for own player


ldavid432:
This is not a full review.
- your repository URL needs to end in `.git`
- you need to add `build=standard` to your runelite-plugin.properties

PR #12540, opened 2026-06-10
update clogger

ETurnz:
The plugin previously required users to manually click through every page in the collection log to sync its' data. It now uses methods similar to RuneProfile and Temple to automatically sync all collection log data when the clog is opened. Also removed some redundant code and added an option for users to hide their drops from the clogger home page


tylerwgrass (commented):
<https://github.com/ETurnz/Clogger/compare/319f50187f6209c168ffca8596e38eacf9954793..ETurnz:7789ba4a9f8924892e59025112587ffab44f14bf#diff-47d73dc233f30652ffc442105dba83b382735ead0ba633587e814a6530ada2c2R483>

ETurnz:
Do you guys have suggestions on any methods to sync the collection log without having to manually click through pages? I looked up the repositories for RuneProfile and Temple and saw that's what they were using

coopermor:
Clicking the search button will load all the items

riktenx:
You should use `RuneLite.RUNELITE_DIR` instead of the user.home sysprop when constructing filepaths, please fix that in a future submission

PR #12545, opened 2026-06-11
Re-enable sound-triggers and update to 1.0.3

tylanphear:
Hello @pajlada, I believe I have addressed the concerns you outlined in #12530. I have pushed a new plugin version with these changes. Please let me know if there are any further issues to address 🙏


pajlada:
Thanks for the update, I will take a look at this.

PR #12547, opened 2026-06-11
Update lobster-pot plugin

datlalo:
Removed dependency on clan website API.
All external communication is now read from a single JSON hosted on GitHub.


riktenx:
remove the url config item, if you're going github-only it needs to be hardcoded

datlalo:
> remove the url config item, if you're going github-only it needs to be hardcoded

@riktenx Should be sorted now! Thanks :)

PR #12553, opened 2026-06-11
Add PK Tracker

L-U-N-E:
PK Tracker is a side panel for PKers that tracks session kills, deaths, K/D and kill streak, with loot from kills, bounty crates and loot chests shown as item grids with GE values and saved to a persistent history. It also auto-looks-up opponent combat stats, BH and LMS ranks on fight start or target assignment, with manual lookup.

Kills are detected from the "You have defeated ..." game message; crate/chest loot is captured via the Loot Tracker's LootReceived event; stat lookups use the HiscoreClient.

### DIFF
--- a/plugins/pk-tracker
+++ b/plugins/pk-tracker
@@ -1,2 +1,2 @@
 repository=https://github.com/L-U-N-E/pk-tracker.git
-commit=43f438040b1faecaeb0aab6a0e6681c551627f72
+commit=e25507354d6a4b15b41509830b85b6ea142de6ea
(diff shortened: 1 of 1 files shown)


ldavid432:

L-U-N-E:
> You need to add `build=standard` to your runelite-plugin.properties

Done — added build=standard and updated the commit hash.

riktenx:
use HiscoreManager, not HiscoreClient

L-U-N-E:
> use HiscoreManager, not HiscoreClient

Thanks for the review! I've switched from HiscoreClient to HiscoreManager for the opponent stat lookups. It now calls hiscoreManager.lookup(...) on the executor thread so it benefits from the manager's caching. Pushed in the latest commit.

PR #12558, opened 2026-06-11
Update pvp-performance-tracker to v1.7.4

Matsyir:
New opt-in feature which automatically uploads all of your fights to the PvP-Hub website, where it can be publicly viewed by anyone. This is disabled by default. When 2 of the same fight is uploaded, fight data is merged for more accurate results.

This properly includes the warning attribute on the config, which clearly states: "This feature will submit your RSN, fight logs, and IP address to a 3rd-party server not controlled or verified by Runelite developers."

More discussion, including links to the website itself + its github repo can be found here: https://github.com/Matsyir/pvp-performance-tracker/pull/78

Many thanks to @LogicalSoIutions for creating this feature and hosting the website!

Edit:
1) The initial commit had a build error due to `new okHttpClient()`, it's been replaced with an Inject and re-tested
2) Then I had to add the build=standard parameter to the plugin.properties. I'm honestly not too sure if I should use standard or gradle, but I don't think I have any fancy build steps in neither build.gradle, or settings.gradle, so standard should be fine

### DIFF
--- a/plugins/pvp-performance-tracker
+++ b/plugins/pvp-performance-tracker
@@ -1,2 +1,2 @@
 repository=https://github.com/Matsyir/pvp-performance-tracker.git
-commit=ae0e3020a15a5f99ad68923d96a9ae8d22e3a4e9
+commit=ad8bf72b1907593dcda08860132c93223ebc76cf


Matsyir:
Is the printStackTrace critical to be changed? That is part of an old feature which is rarely used and likely to be removed soon anyways (Fight Analysis/Merge), since this new website essentially replicates the feature in a simpler way.

Matsyir:
Actually, I think it's best we hold off on merging this until:

1. We add a configurable randomized delay for fights to be displayed publicly by default, so this cannot be used as a 'UAV' or scouting tool by any means. I think a default of random between ~6-20 minutes would be reasonable, then a few more manual options like, 15-45 mins, 30-120mins, 2hrs, 6hrs, 12hrs, 24hrs
2. Opponent RSNs should be anonymized on the website, so only users who opted into the feature have their RSN displayed.

I don't think it would be too catastrophic to release as-is, but this update could definitely raise a lot of concerns without these changes.

@LogicalSoIutions I would be willing to start working on these changes probably Sunday, maybe Monday, but also feel free to get started if you'd like.

Matsyir:
The above changes described have been added, I believe there should no longer be any concerns in regards to this feature. Many thanks again to @LogicalSoIutions for getting it done quickly

The local "Fight analysis/merge" feature & all references has been removed as well, and I've ensured that the copyright headers are present on all source files - that's why the size went from M to L, lots of removals, and copyright header additions. Otherwise it was fairly small changes to make for these new changes.

*Edit: New commit from ~8hrs ago = minor label change on one of the new configs, to shorten the dropdown width and allow us to still read the config label a bit

*Edit2: Newest commits Jun18 => Added config option to anonymize your RSN even when you opt-in to the feature, LMS fixes related to recent changes

PR #12561, opened 2026-06-11
Add Bank Highlight Search

Reasel:
Adds Bank Highlight Search: search the bank with a hotkey and highlight matching
items (blinking/feathered outline) instead of filtering them out; Enter
switches to the ALL tab and scrolls to where the most matches are visible.

Repository: https://github.com/Reasel/Highlight-Search


riktenx:
use of client menuAction is not allowed

PR #12576, opened 2026-06-12
Add banker

randytkrx:
Banker organizes the bank into color coded categories. Items are sorted automatically by item type rules and name filters that the user configures per category, categories color the bank tab row and highlight items, and a side panel manages everything with viewing, filtering, value totals and setup sharing. It creates no widgets in the bank interface so it is compatible with Bank Tags.

Repository: https://github.com/randytkrx/banker


randytkrx:
Closing to resubmit from a clean branch.

PR #12599, opened 2026-06-13
Update scaper plugin - add icon, fix panel text, startup login detection

Deafpie:
Updates to the Scaper plugin:

- Add `scaper_icon.png` as the RuneLite sidebar navigation button icon
- Update panel description text: removed outdated Discord-only linking instructions, added clearer two-method linking guide (Scaper.icu dashboard or `/link` Discord command)
- Fix startup login detection: if player is already logged in when the plugin is enabled, `onLogin()` is now triggered immediately via `GameState` check


ldavid432:
You've somehow managed to put an invisible character in front of each line in this file. Please delete them, otherwise your build will not succeed

PR #12625, opened 2026-06-15
Update to BankTabNames

Psyda:
Reopening #11105 

It has been reviewed here: https://github.com/runelite/plugin-hub/pull/11105, but I was unavailable for the last month to provide an update. Sorry for the delay, work has been rough, and the changes were implemented.

I removed the images I didn't create, so it now only includes my work. I also swappedthe awt.Desktop usage.

---

This replaces the single-icon sprite editor with a multi-icon system and overhauls how tab overlays are rendered.

Why: The old editor had a persistent issue where sprite IDs shuffled whenever the icon list updated, breaking user configs. The rendering was also clipped to the tab bar boundaries, limiting what users could do with custom icons and offsets. Both problems required a ground-up rethink rather than incremental patches .I held off on shipping incremental fixes because most of the reported problems shared root causes in the rendering and sprite ID systems. Fixing them properly meant reworking both at the same time.

What changed:
Tabs now support multiple icons (items, skills, game sprites, custom PNGs) with per-icon size, offset, z-index, and aspect-ratio-locked resizing. Icons and text render on Bankmain.INFINITE instead of Bankmain.TABS, removing the clipping limitation. Game sprite dimensions are resolved from SpritePixels.getMaxWidth/getMaxHeight rather than the trimmed BufferedImage from SpriteManager.getSprite(), which was causing sprites to render at wrong sizes.

The side panel got a scrollable icon editor, curated sprite category browser, fit toggle for stacking/overlay designs, config presets with import/export (versioned JSON), and panel visibility options.

See the updated README for full feature documentation.

Testing: Tested across all three tab display modes (item, digit, roman numeral), with bank settings page open/closed, with 1-8+ icons per tab, with custom user PNGs, and with various game sprites including non-square and padded sprites. Config migration from old single-icon format verified.

image

image

image

### DIFF
--- a/plugins/bank-tab-names
+++ b/plugins/bank-tab-names
@@ -1,2 +1,2 @@
 repository=https://github.com/Psyda/bank-tab-names.git
-commit=c21369dd1f55e90f6316e25b95b30aa8637dd42a
+commit=f9d94c35ae6078dc1df9e2028926d81501dccaf1
--- a/plugins/bank-tab-names
+++ b/plugins/bank-tab-names
@@ -1,2 +1,2 @@
 repository=https://github.com/Psyda/bank-tab-names.git
-commit=b83653b9a20f381954b73e56aa74887656954bf5
+commit=c21369dd1f55e90f6316e25b95b30aa8637dd42a
--- a/plugins/bank-tab-names
+++ b/plugins/bank-tab-names
@@ -1,2 +1,2 @@
 repository=https://github.com/Psyda/bank-tab-names.git
-commit=b8c6ad04c2a7220035384356d24157be1e08f908
+commit=225a94164b5ce9af2463425be2a3562897b43feb
--- a/plugins/bank-tab-names
+++ b/plugins/bank-tab-names
@@ -1,2 +1,2 @@
 repository=https://github.com/Psyda/bank-tab-names.git
-commit=4e610a93dca9efb98322c1c30f29b0688cb51bbb
+commit=b8c6ad04c2a7220035384356d24157be1e08f908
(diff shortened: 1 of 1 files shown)


ldavid432:

Psyda:
> You need to add `build=standard` to your runelite-plugin.properties

I've added build=standard to the runelite-plugin.properties

riktenx:
all file i/o needs to be restricted to a plugin-specific subdir under .runelite/

Psyda:
> all file i/o needs to be restricted to a plugin-specific subdir under .runelite/

Ow. Good catch, the icon folder was already confined to .runelite/banktabnames/icons, the saved presets went through ConfigManager rather than disk, but the config import/export still used a JFileChooser that could read and write anywhere. I've confined both to a plugin subdir.

I'm not exactly sure of the best practice here, but I've made it so that the Export now writes a sanitized <name>.json directly into .runelite/banktabnames/configs(created on demand), with no file picker. Import lists only the .json files already in that folder and reads the selected one by name, so there's no arbitrary read path either. Clipboard and the ConfigManager-backed presets are unchanged. I removed the now-unused JFileChooser and FileNameExtensionFilter imports too.

Should I drop the file-based path entirely and keep clipboard only or is this okay, since that's also an option given presets already cover persistence, but I find that physical configs which are readable and sharable fit best with this plugin's design.

riktenx:
Use of JFileChooser is fine, it's possible I misinterpreted the code. Basically if you are doing file i/o WITHOUT prompting the user for the filepath it needs to be in that plugin-specific dir in .runelite. If that is true today you can disregard.

riktenx:
(or put that code back if I erroneously made you remove it)

Psyda:
I think the change is better this way, to be honest. It guides them to the plugin folder where assets are, making the learning curve less steep, too. I'll keep it with the config folder setup.

PR #12628, opened 2026-06-15
Route-Planner

BlackDeath-Osrs:
This is a tool to create custom routes. Custom Highlighting of npc's objects and the ability to have Waypoints to direct you where you need to go. you can export and import routes. while playing a route it should give you the feeling of a Quest helper assistant.

### DIFF
--- a/plugins/route-planner
+++ b/plugins/route-planner
@@ -1,2 +1,2 @@
 repository=https://github.com/BlackDeath-Osrs/route-planner.git
-commit=6916d10ba43fba32757f51e272d1cc3118f59e86
+commit=e1d634aa1072d1a7f10dd80a6d4a5d15a65d723f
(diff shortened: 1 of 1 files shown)


riktenx:
the keyboard event dispatch and focus stuff you're doing is not allowed

BlackDeath-Osrs:
> the keyboard event dispatch and focus stuff you're doing is not allowed

Thank you, Removed the global KeyEventDispatcher / KeyboardFocusManager hook; shift detection now uses client.isKeyPressed(KeyCode.KC_SHIFT). That was the only global keyboard interception in the plugin.

adamk33n3r:
I was actually just thinking of and planning a plugin like this and did a quick search to find this was only made recently! Rad :) i haven't had a chance to try it out yet, but if you're open to PRs I could probably improve the UI a bit.

Edit: just coming back here to say I took a look more into the repo and it seems like it was probably vibe-coded at least looking at the overly detailed commit messages. So I'm not sure I want to contribute to the project. No wonder the UI/UX wasn't great.

PR #12629, opened 2026-06-15
add aim-trainer

NotCoco:
﻿Adds the Aim Trainer plugin to Plugin Hub.

This is submitted as a separate `aim-trainer` entry rather than updating the disabled `nh-input-tweaks` entry. The referenced plugin commit is an Aim Trainer-only submission: it removes the old input-tweaks plugin source from the public submission branch and exposes only `com.nhtrainer.inputtweaks.NhAimTrainerPlugin` in `runelite-plugin.properties`.

Validation:
- `./gradlew.bat test` passed in the plugin repository submission worktree.

### DIFF
--- a/plugins/aim-trainer
+++ b/plugins/aim-trainer
@@ -1,0 +1,2 @@
+repository=https://github.com/NotCoco/nh-input-tweaks.git
+commit=4e2ef00dbaf84196ce054356fd236d23d54b3f97


NotCoco:
﻿Demo GIF for reviewer context:

Aim Trainer demo

This shows the aim-box overlay/training flow visually, since the plugin is easier to understand from a short demo than from the manifest text alone.

PR #12639, opened 2026-06-15
Update combat-logger to v1.5.0

SuperNerdEric:
- Fix issue where death event logged before final damage on the same tick caused the killing blow to start a new fight
- Add Runelogs.com live logging functionality

### DIFF
--- a/plugins/combat-logger
+++ b/plugins/combat-logger
@@ -1,3 +1,2 @@
 repository=https://github.com/SuperNerdEric/combat-logger.git
 commit=4284a4ad8fee52056550f32affb17ee51317c128
-warning=When live logging is enabled, this plugin sends your IP address, RSN, and detailed combat log data to a third-party server not controlled or verified by RuneLite developers.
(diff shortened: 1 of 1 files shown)


riktenx:
the warning can't go in the manifest here because then people that already installed it won't get a chance to see it

you can add the api.runelogs.com stuff but it MUST be opt-in with a boolean `@ConfigItem` with its `warning` field set to `"This feature submits your IP address to a 3rd-party server not controlled or verified by Runelite developers"`

plugins/combat-logger
```diff
@@ -1,2 +1,3 @@
 repository=https://github.com/SuperNerdEric/combat-logger.git
-commit=ef8dd1b79d68941806ee398ab41174c0d6f4ddf7
+commit=4284a4ad8fee52056550f32affb17ee51317c128
+warning=When live logging is enabled, this plugin sends your IP address, RSN, and detailed combat log data to a third-party server not controlled or verified by RuneLite developers.
```
riktenx:
```suggestion
```

SuperNerdEric:
> the warning can't go in the manifest here because then people that already installed it won't get a chance to see it
> 
> you can add the api.runelogs.com stuff but it MUST be opt-in with a boolean `@ConfigItem` with its `warning` field set to `"This feature submits your IP address to a 3rd-party server not controlled or verified by Runelite developers"`

Done.

PR #12649, opened 2026-06-16
create finished-bounty-notify

StaySleeping:
Notifies you when you finish collecting all parts for a sailing bounty task.

### DIFF
--- a/plugins/finished-bounty-notify
+++ b/plugins/finished-bounty-notify
@@ -1,0 +1,2 @@
+repository=https://github.com/StaySleeping/finished-bounty-notify.git
+commit=c3e1f54d22be6a7e9e1cdc17d74597f383331ca7


coopermor:
@StaySleeping You have code that is clearly copied from the Port Task plugin. 

https://github.com/StaySleeping/finished-bounty-notify/blob/c3e1f54d22be6a7e9e1cdc17d74597f383331ca7/src/main/java/com/finishedbountynotify/BountyTaskCatalog.java#L61-L117

https://github.com/nucleon/port-tasks/blob/9f427aafd851662fe77bc809dcca0ad71410fe8f/src/main/java/com/nucleon/porttasks/enums/BountyTaskData.java#L92-L163

https://github.com/StaySleeping/finished-bounty-notify/blob/c3e1f54d22be6a7e9e1cdc17d74597f383331ca7/src/main/java/com/finishedbountynotify/PortTaskTrigger.java#L7-L71

https://github.com/nucleon/port-tasks/blob/9f427aafd851662fe77bc809dcca0ad71410fe8f/src/main/java/com/nucleon/porttasks/enums/PortTaskTrigger.java#L33-L106

But you have stripped any attribution of this code. To be compliant with the BSD 2-Clause License that Port Tasks uses you must include the authorship of the code segments you have used.

StaySleeping:
@coopermor Thanks so much for letting me know, you're right and I misunderstood the licence. Updating now.

StaySleeping:
@coopermor I've updated those and two more files which I used code from, would you recommend updating the commit in runelite to the recent one with the attribution updates, happy to do so even without functionality changes if you recommend. Also I was planning on submitting a PR to port tasks with similar functionality but personally I wanted one which only did this separate from the port tasks plugin but thought it could use this too.

PR #12650, opened 2026-06-16
Update real-loot-key-value to 1.0.0 version

pk-enjoyer:
The plugin now makes beter use of the runelite API, the compact keys no longer overlays the loot chest UI at the top. It instead just calls the API and changes the value (It also saves the original displayed osrs value, so it can be reverted when the plugin is switched off). 

This also fixes an issue with other plugins that change the UI, our overlay would just overlay 'ugly' brown Runescape background over the no longer brown UI. 

That said, the "top right" version of the plugin still does this... not sure if ill ever make this pretty.

~- [ ] Fix the pictures in the README, as the old repo was deleted, meaning that the pictures url are now broken : (~


pk-enjoyer:
Oke, ill update the README pictures in another pr, i cba for now.

Plus it allows me to get some candid screenshots

PR #12708, opened 2026-06-18
Add wilderness-teleport-blocker plugin

Aiyasei:
Adds the Wilderness Teleport Blocker plugin.

Prevents accidental casting of wilderness teleport spells  (Ancient, Arceuus Cemetery, Lunar Ice Plateau) by removing  their menu entries while keeping the spells visible on the  spellbook. Does not rearrange or reorder spells in the spellbook to do so. Has options to toggle off and on. Useful for players who want safety against misclicks from going into the wilderness.


riktenx:
stale, you can open a new pr if you fix it

PR #12714, opened 2026-06-19
fix tab bar on hop, option to ungrow for dialog

shanktank:


riktenx:
build=standard, not gradle

PR #12730, opened 2026-06-19
Update pvp-performance tracker to v1.7.5

Matsyir:
Various fixes and clarifications, overall fairly minor
- properly apply guaranteed expected damage from bolt procs (still being averaged on every hit) (Thanks to @LogicalSoIutions for this fix, and Port Khazard for noticing it!)
- Properly detect LMS ammo everywhere it's relevant: was still forcing normal diamonds(e) for accuracy and some ui stuff i think
- Default to B ring (i) for LMS instead of normal B ring

Config clarifications/improvements
- Remove unimbued rings from config options. Since unimbued B ring is no longer used in LMS, there are no code references to them, and nobody really uses unimbued rings so we have no reason to clutter the dropdown with them.
- Shorten strongBoltChoice label so it's actually readable, update tooltip for current behavior
- lowercased & prefixed non-dragon bolts for the strongBoltChoice so it is harder to accidentally pick them. Changed default from diamond dragon bolts to opal dragons since they are more commonly used

### DIFF
--- a/plugins/pvp-performance-tracker
+++ b/plugins/pvp-performance-tracker
@@ -1,2 +1,2 @@
 repository=https://github.com/Matsyir/pvp-performance-tracker.git
-commit=ad8bf72b1907593dcda08860132c93223ebc76cf
+commit=8d9c17aabe40268097bced4aaa9eab525f2133f4


Matsyir:
oops, i forgor to update version numbers in the actual plugin, still 1.7.4 everywhere. No big deal though I'll just leave it for now

thanks for quick merge<3

One can confirm they're on 1.7.5 by seeing these lowercase bolts on DCB+ config, or the lack of unimbued rings in the ring config.
image

PR #12740, opened 2026-06-20
Add gif-login-osrs

VitaMors:
gif-login-osrs is a RuneLite plugin that plays a looping GIF as the login screen background while keeping the normal Old School RuneScape login UI visible.


riktenx:
file i/o must be limited to a plugin-specific subdir under .runelite/

PR #12749, opened 2026-06-20
added new wave tile highlighting feature

Jufinda:
Updated to include wave tile highlighting


riktenx:
this is not allowed per https://secure.runescape.com/m=news/third-party-client-guidelines?oldschool=1

> Projectile targets, or target locations, as well as impact locations

and also probably

Jufinda:
ah, apologies, i thought that i may have been ok, ill revert the code now, thanks for clarifying

PR #12751, opened 2026-06-20
Update action-progress

guillaume009:
- Support for incendiary/chainshot cannonballs
- Refactor smithing logic
- Add simple mode display
- Update smithing outfit calculation


riktenx:
`build=standard`, not gradle

riktenx:
you want lombok's NonNull not this jetbrains one. your build.gradle is pretty deviated from the current example template so i would just re-copy that in so you can be sure everything builds locally

PR #12754, opened 2026-06-20
Add object id marker

jtclowner:
Object ID Marker allows objects to be highlighted directly by object ID, without first marking them in-game.

## Features
Highlight objects using configurable object IDs
Object hull/outline highlights
Tile highlights for tiles occupied by objects
Optional tile highlight expansion radius (id:radius)

### Compliance
No automation or input modification
No external communication or data transmission
Java 11, no reflection, JNI, native libraries, external dependencies, or classloader modification

_Note: Unlike the built-in Object Markers plugin, which marks individual object instances at specific locations, Object ID Marker highlights all game, wall, decorative, and ground objects matching the configured object IDs._


Felanbird:
> Reviewing

PR #12756, opened 2026-06-21
Add osrscn (Simplified Chinese translation)

Aoldbald:
﻿Adds OSRSCN Chinese, a clean-room Simplified Chinese translation plugin (package com.osrscn).

It translates game text (menus, chat, dialogue, overhead text, interfaces) into Simplified Chinese from local lookup tables. OSRS's native UI cannot render Chinese, so the plugin rasterizes each glyph and shows it inline via ChatIconManager.

In case you are wondering about the plugin size and the AI option:

- Plugin size (~6 MB): it bundles the Source Han Sans SC font (SIL OFL 1.1), loaded via getResourceAsStream. Runtime rasterization of arbitrary characters is required because the optional AI translation can output any character, so a pre-rendered glyph atlas would not cover them. The OFL license text is in the repo and inside the JAR, with attribution in NOTICE.
- AI translation is off by default and, when enabled, calls a local Ollama instance (localhost) only. Nothing is sent to any third-party server by default. This is documented in the README and config descriptions.

License: BSD 2-Clause.

### DIFF
--- a/plugins/osrscn
+++ b/plugins/osrscn
@@ -1,2 +1,2 @@
 repository=https://github.com/Aoldbald/OSRS_CN_Runelite.git
-commit=0f59163a014e3b5dba7e7431524c82f3eb9d0fbc
+commit=56a56b70e2e7fb13a33bc45f02d919c89326b668
--- a/plugins/osrscn
+++ b/plugins/osrscn
@@ -1,2 +1,2 @@
 repository=https://github.com/Aoldbald/OSRS_CN_Runelite.git
-commit=b4283d9dc3cd466b7cb3494cdeae4bd3071291d7
+commit=4606b150367676e59b1b9dda9646ac1cba8cbf4b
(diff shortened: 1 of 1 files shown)


Aoldbald:
﻿Hi, thanks for taking the time to look at this.

There are already a few translation plugins here, so let me explain why this is a separate one. OSRSCN only does Simplified Chinese and it runs offline. It ships community-made lookup tables that already cover about 99.8% of the in-game text, so most translation is just a local lookup rather than a live API call, and there is a hotkey to flip the whole UI back to English. I am Chinese and stay close to the Chinese OSRS community, so I can fix Chinese-specific wording fast.

I went through the RuneLingual PR (#7141) and tried to deal with the common review points ahead of time. The data download now uses the injected OkHttpClient instead of HttpURLConnection. AI translation is off by default, and when it is on it only talks to a local Ollama on localhost. A user can point it at an online LLM instead, and in that case the English text being translated is sent there (player names are never sent). If you want a warning line in the manifest for that, just tell me the wording and I will add it.

Thanks again.

plugins/osrscn
```diff
@@ -0,0 +1,2 @@
+repository=https://github.com/Aoldbald/OSRS_CN_Runelite.git
+commit=56a56b70e2e7fb13a33bc45f02d919c89326b668
```
riktenx, line 2:
```suggestion
commit=56a56b70e2e7fb13a33bc45f02d919c89326b668
warning=This plugin submits your IP address to a 3rd-party server not controlled or verified by Runelite developers.
```

PR #12765, opened 2026-06-21
Add OmniPresence plugin

jonathanavis96:
Adds the **OmniPresence** plugin.

Source: https://github.com/jonathanavis96/omnipresence-runelite-plugin

**What it does:** reads the local player's in-game activity (current skill, boss/NPC, region, banking state) and POSTs a small JSON summary to the OmniPresence desktop companion app so it can show OSRS activity in Discord Rich Presence. It is conceptually the same as the built-in Discord plugin, but feeds a local companion app instead of Discord directly.

**Data / network:**
- Read-only observation of game state. No input is ever sent to the game; no new menu entries; no automation.
- Network egress is **loopback only** — the host is hard-coded to `127.0.0.1`; only the port is user-configurable, so it cannot be pointed at any remote/third-party server.
- Account name is **opt-in and off by default**; chat content is never forwarded.
- If the companion app is not running, the plugin silently no-ops.

**Dependencies:** only gson + okhttp, both transitive dependencies of `net.runelite:client` (no third-party deps requiring verification metadata).

Builds and tests pass under the standard build against `latest.release`.

### DIFF
--- a/plugins/omnipresence
+++ b/plugins/omnipresence
@@ -1,2 +1,2 @@
 repository=https://github.com/jonathanavis96/omnipresence-runelite-plugin.git
-commit=b5f2ee8cc851fc0240494159c6b00e32e0c0271e
+commit=494837b5be2b724cc7b8747e930b38311b06ba12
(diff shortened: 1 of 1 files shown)


riktenx:

jonathanavis96:
Thanks for the review! Fixed in 6ffd76e and the manifest now points at it.

The only `Thread.interrupt()` was the standard "restore the interrupt flag after catching `InterruptedException`" idiom in `ContextPublisher.shutdown()`. I've removed it — that path only runs at plugin teardown and nothing downstream re-checks the flag, so the `InterruptedException` is simply swallowed now. No other behaviour change.

riktenx:
thanks for the submission but for security reasons we do not want to host plugins that expressly direct users to download and run separate native executables.

PR #12772, opened 2026-06-21
v1.33_gif-login-osrs_Bug-Fixes

VitaMors:
fixed 3 bugs preventing login and also bad gif playback


riktenx:

PR #12800, opened 2026-06-22
Add Flip Finder Sync

betmodejoe:
Adds **Flip Finder Sync**, a companion plugin for [OSRS Flip Finder](https://osrsflipfinder.com).

### What it does
Listens for `GrandExchangeOfferChanged` and syncs completed Grand Exchange buys/sells to the user's OSRS Flip Finder trade journal over HTTPS, authenticated with a per-user API key generated on the website. Each fill is sent as a delta (newly-filled quantity + average price) with a deterministic id so re-logs never double-count. Only filled buy/sell quantities are sent — no inventory, location, or other data.

### Details
- **Source:** https://github.com/betmodejoe/osrsflipfinder.com — BSD 2-Clause licensed
- **3rd-party server:** the plugin sends data to the user's own OSRS Flip Finder account, so a `warning=` line is included in the manifest.
- **Opt-in:** Base URL and API key are user-configured; nothing is sent until a key is entered.

### DIFF
--- a/plugins/flip-finder-sync
+++ b/plugins/flip-finder-sync
@@ -1,3 +1,3 @@
 repository=https://github.com/betmodejoe/osrsflipfinder.com.git
-commit=7c2b04804c297894a1e0d5552c8bceb3e7596ba5
+commit=b275f7eeddff9e8e1f402eec486f08568b8889c0
 warning=This plugin sends your Grand Exchange transactions and IP address to OSRS Flip Finder, a 3rd-party server not controlled or verified by RuneLite developers.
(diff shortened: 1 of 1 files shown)


riktenx:
hardcode your endpoint to your actual production thing instead of letting the user configure it

betmodejoe:
> hardcode your endpoint to your actual production thing instead of letting the user configure it

Hey riktenx, good point, kinda pointless why I set it up like this.

Will make this change, thanks for your feedback!

betmodejoe:
Done — the endpoint is now hardcoded to `https://osrsflipfinder.com` in the API client and the user-configurable **Base URL** setting has been removed. Only the API key and the enable-sync toggle remain in the plugin config.

I've pushed the change to the source repo and bumped the manifest in this PR to the new commit (`1f99e46`). Thanks again for the review!

PR #12806, opened 2026-06-22
update flux 1.4.1

abristow3:
small bug fixes
Update login message type to broadcast

### DIFF
--- a/plugins/flux
+++ b/plugins/flux
@@ -1,3 +1,3 @@
 repository=https://github.com/abristow3/Flux-Runelite-Plugin.git
-commit=61152873c8a8b8199d732f9e49683adc801df9d9
+commit=a55a7a286dc41910daf7cf6c80310590576f3771
 warning=This plugin submits your IP address to wiseoldman.net and a 3rd-party server not controlled or verified by Runelite developers.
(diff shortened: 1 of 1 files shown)


riktenx:
please remove your use of thread interrupt

abristow3:
> please remove your use of thread interrupt

Hi Rikten :D 

What should I be replacing it with? I think it was you who told me to use it all those months ago ;)

riktenx:
so looking at your code i think you actually don't need those 3 threads? because their runnables just directly schedule _another_ runnable on the scheduled executor you have here, you'd get the same effect by just calling your "start task" runnables themselves

so e.g. you can just collapse

```
                leaderboardThread = new Thread(this::pollLeaderboard);
                leaderboardThread.start();
```

to

```
                pollLeaderboard(); // maybe rename this startPollLeaderboard etc.
```

then you can just delete the threads

riktenx:
and make sure you cancel the scheduling or shutdown the scheduled executor etc. on shutDown

abristow3:
> so looking at your code i think you actually don't need those 3 threads? because their runnables just directly schedule _another_ runnable on the scheduled executor you have here, you'd get the same effect by just calling your "start task" runnables themselves
> 
> so e.g. you can just collapse
> 
> ```
>                 leaderboardThread = new Thread(this::pollLeaderboard);
>                 leaderboardThread.start();
> ```
> 
> to
> 
> ```
>                 pollLeaderboard(); // maybe rename this startPollLeaderboard etc.
> ```
> 
> then you can just delete the threads

OK. I think I am understanding your explanation. I have also taken your advice and renamed my methods. Thank you again Rikten

PR #12831, opened 2026-06-23
update compass-camera-control to 1.2.5

RaazKH:
Jagex increased the camera yaw range from 2048 to 16384, this plugin needed a patch so it would work with the new range.

Users reported the issue here: https://github.com/RaazKH/CompassCameraControl/issues/31
Patch applied here: https://github.com/RaazKH/CompassCameraControl/pull/32


dabbidhan-byte:
I tried uninstalling and reinstalling the plugin but it hasn't seemed to update, anyway I can manually download the update?

riktenx:
i've just now merged this update, restart your client in ~5-10 minutes and it should be fixed.

dabbidhan-byte:
omg thank you so much to you guys both

PR #12834, opened 2026-06-23
disable sound-swapper

riktenx:
We are disabling sound-swapper because arbitrary replacement of sound effects by ID can be abused to provide additional / clearer auditory indicators of boss mechanics.

In order for this plugin to be re-enabled, it **MUST** be updated to disable the entirety of its functionality in all of the same map regions that the Watchdog plugin does.


riktenx:
We are giving the author an opportunity to make the required changes before disabling, see https://github.com/petertalbanese/SoundSwapper/issues/29.

riktenx:
closing in favor of #12866

PR #12848, opened 2026-06-23
Add 'Collection Log Hider' plugin

mattjrumble:
This a new plugin that hides obtained items and completed sections from the collection log.


riktenx:
use of client.menuAction in this way is not allowed

mattjrumble:
> use of client.menuAction in this way is not allowed

Addressed - no more `client.menuAction` calls. Thanks for the review!

PR #12850, opened 2026-06-23
Add Mouse Wheel plugin

Largemann:
Certain window sizes in combination with high mouse wheel scroll speed settings in the OS can cause the bank and chat interfaces to scroll too far with a single wheel click, sometimes further than a visible page.

This plugin allows adjustment of how many pixels an interface should scroll for each mouse wheel click.


riktenx:
use of client.menuAction in this manner is not allowed

PR #12864, opened 2026-06-24
Update 07flip to 43392f7

UserD40:
This update is a maintenance and polish pass for the 07Flip plugin. It trims unused code to keep the plugin lightweight, fixes item search so it returns all matching results instead of only the first few, and adds a small show/hide toggle to the My Trades stats box. No existing features were changed or removed.

### DIFF
--- a/plugins/07flip
+++ b/plugins/07flip
@@ -1,3 +1,3 @@
 repository=https://github.com/UserD40/Runelite07Flip.git
-commit=d5315439903e6a3172a185280bcae89f6819c838
+commit=5132f8613f38daa3c63a5c4bf96b8d015e3fb5bd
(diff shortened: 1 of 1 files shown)


UserD40:
Updated to commit `5132f86`: fixed the build by deriving the JSON parser from the client's injected Gson (via `.newBuilder()`) instead of constructing a fresh Gson instance. No behaviour change — just the packager-required fix.

riktenx:
JFYI the bot removes comments before it tokenizes, so you don't need to remove those.
It also doesn't look at any unit tests you have.

It looks like the bot can review your plugin again.

JFYI we do not really have the capacity to review a plugin of your size + diffs of this size + patches at the rate at which you submit them so if you want to keep updating this you will need to keep it within the token limit.

The limit is 200k tokens, the only thing counted is the source code in your main source set (tests and comments are excluded).

PR #12876, opened 2026-06-24
Update AFK Stats Tracker

Reasel:
Bump to the latest commit. Only the session history list scrolls now, so the Start/Stop buttons stay in place.

panel


Reasel:
For reference this was the old scroll behavior this PR changes.

image

PR #12885, opened 2026-06-24
add gpu-vulkan

dennisdevulder:
Submitting GPU (Vulkan), an experimental Vulkan renderer for RuneLite.

What makes it more than a renderer is the extension API: other plugins can hook directly into the GPU pipeline. The two features it ships with are really proofs of concept for that API. An FSR 1.0 upscaler is a plugin adding its own post-process pass, and a clip recorder encodes on the GPU through Vulkan Video Encode (H.264). The rendered frame is already in GPU memory, so the clip is encoded there and written to disk locally, with no external binary and no raw frames copied back into the JVM heap. Both could just as easily be standalone plugins on top of the API rather than built in.

One thing I'd like maintainer input on: there's no way in core for a plugin to discover or register against a renderer, so each renderer ends up defining its own handshake. A small way to register a renderer backend would let any plugin call into it cleanly (e.g. ask for the last 20 seconds as a clip) without knowing which renderer is active. Not for this PR, but this plugin is a concrete case for it and I'm happy to help design it.

Dependency-wise this needs org.lwjgl:lwjgl-vulkan:3.3.6 (for the ratified KHR Vulkan Video Encode bindings, not present in the 3.3.2 RuneLite ships) and org.lwjgl:lwjgl-jawt:3.3.2. The same PR adds them to the thirdParty configuration with SHA256 pinned, next to the existing lwjgl-opengl entries. lwjgl-vulkan ships natives only for macOS; Linux and Windows use the system Vulkan loader.

First submission targets Linux and Windows. macOS can run the renderer through MoltenVK, but MoltenVK doesn't expose Vulkan Video Encode, so there's no recorder there.

AI assisted by: Mythos preview


riktenx:
we do not have anybody with expertise to review new gpu / vulkan / whatever code so this is completely impossible to merge

PR #12889, opened 2026-06-24
Add Clan Wars Timer plugin

BloodGateRS:
Please add my clanwars plugin for  the clanwars community


riktenx:
has to be the full commit hash

riktenx:
your submission needs to only add one plugin, if you have this new plugin then it needs to be its own pr

BloodGateRS:
> your submission needs to only add one plugin, if you have this new plugin then it needs to be its own pr

@riktenx, hopefully that fixed it!

PR #12891, opened 2026-06-24
Update Inferno (Now Colo too) Tracker

evaan:
Update inferno tracker to support Colosseum attempts


evaan:
wait didnt realize it'd uninstall everything, redoing this

PR #12892, opened 2026-06-24
Update inferno tracker

evaan:
Second times the charm, un-renamed the plugin


riktenx:
use of runtime shutdown hooks is not allowed, use https://static.runelite.net/runelite-client/apidocs/net/runelite/client/events/ClientShutdown.html event instead

PR #12893, opened 2026-06-25
Add repository and commit information to ironman plugin

Nubles:


riktenx:
thread sleep is not allowed

riktenx:
your commits did not fully resolve this

riktenx:
keep your readme user-focused, you can add instructions to debug the client in a CONTRIBUTING.md or something

PR #12906, opened 2026-06-25
Add Kaisham's Exhumed Markers plugin

KaishamRL:
This plugin automatically populates tile markers for the Xarpus exhumed grid layout in the Theatre of Blood (Plane 1) using the native Object Indicators plugin configurations

### DIFF
--- a/plugins/kaishams-exhumed-markers.properties
+++ b/plugins/kaishams-exhumed-markers.properties
@@ -1,0 +1,2 @@
+repository=https://github.com/KaishamRL/kaishams-exhumed-markers.git
+commit=118b166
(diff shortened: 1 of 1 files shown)


riktenx:
we have rejected this in the past because it breaks jagex rules:

https://oldschool.runescape.wiki/w/Update:Third_Party_Client_Guidelines

KaishamRL:
> we have rejected this in the past because it breaks jagex rules:
> 
> > Anything that automatically indicates where to stand, or not to stand. This applies to only automatic indicators, and not tiles which have been manually marked.
> 
> https://oldschool.runescape.wiki/w/Update:Third_Party_Client_Guidelines

I see.. I tried to make sure that they aren't "automatically marked" but rather like importing a profile into object markers plugin, that mark all the exhumed on the specific coordinates of only the xarpus room itself. Is this still not allowed? Thanks regardless

riktenx:
it is not allowed sorry

PR #12910, opened 2026-06-25
log-adviser: bump to 17b88a2

SFranciscoSouza:
Updates Log Adviser to v1.1.8.

Pins `commit=17b88a2d6ca4cb3c64376bc6b1fa08e39bacd38b` (was `7602f97`); `repository=` unchanged.

Config-only change: adds a notice on the plugin's settings page asking users to temporarily disable the Temple OSRS plugin in order to see Log Adviser's collection-log Sync button, while a fix for the conflict between the two plugins is prepared.


SFranciscoSouza:
Closing — shipping the actual fix (which resolves the Temple-OSRS conflict) instead of this interim notice.

PR #12915, opened 2026-06-25
Update Blert to 0.9.12

frolv:
- Snapshots player gear on change events before other plugins can mutate compositions
- Stops the Verzik attack state machine once all players are dead
- Adds websocket protocol support for draining messages requesting that the client reconnect

### DIFF
--- a/plugins/blert
+++ b/plugins/blert
@@ -1,4 +1,4 @@
 repository=https://github.com/blert-io/plugin.git
-commit=965d03eedd8a179c51a8b3723222ca53501372f2
+commit=89f95df215543b36abf76a70e263827be92020ee
 authors=frolv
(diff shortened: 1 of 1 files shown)


riktenx:
is there a reason you need build=gradle

frolv:
> is there a reason you need build=gradle

I have custom build steps that generate a metadata file from git to identify the plugin when connecting to the server

riktenx:
ok i mean i will merge this one but please just do that manually and set build=standard, we really don't want to be reviewing people's code manually unless we absolutely have to and this isn't strictly necessary. setting build=standard will allow your code to be automatically reviewed and updated

frolv:
Will do, thanks

PR #12950, opened 2026-06-26
Add Camera Presets plugin

titanherow-cmd:
Aims to fix and add some lacking features from 'camera points' plugin. Adds ability to save and restore camera yaw, pitch and zoom as named presets. Supports 5 slots, each with configurable save/load hotkeys and a side panel with Save, Load and Clear buttons.


1Defence:
These are not lacking features, they are not currently allowed and thats why they aren't in the camera-points plugin.

See : https://github.com/runelite/runelite/wiki/Rejected-or-Rolled-Back-Features

- **_Presets for Camera yaw/pitch/position_**

PR #12969, opened 2026-06-26
Add silly-specs plugin repository information

Inzomniakk:
Changes the sound effect played when speccing with certain weapons. Currently supports Fang, D Baxe, and D Claws.


riktenx:
>  use net.runelite.client.audio.AudioPlayer instead

PR #12973, opened 2026-06-27
Add Prayer Sound Swapper plugin

aaronenberg:
Allows the user to replace prayer-related sound effects.

Unlike the original Sound Swapper plugin, Prayer Sound Swapper only replaces prayer sound IDs from a static allowlist of prayer sound IDs. This is to allow this plugin to be compliant when enabled in areas where the original SoundSwapper plugin cannot be due to RuneLite guidelines around swapping arbitrary sounds.

### DIFF
--- a/plugins/prayer-sound-swapper
+++ b/plugins/prayer-sound-swapper
@@ -1,0 +1,2 @@
+repository=https://github.com/aaronenberg/PrayerSoundSwapper.git
+commit=d69e607abd63dcb4d556bb9ebcf7f0270c7eeda5


riktenx:

If you want to do this you need to expose explicit configs for each prayer sound, not just an open ID list (even if you are validating it behind the scenes)

aaronenberg:
@riktenx thanks for the feedback. 

- replaced sound ID input with an explicit option for each prayer sound
- replaced `javax.sound` usage with runelite `AudioPlayer`

aaronenberg:
Plugin Hub check bot is still flagging usage of exception types from `javax.sound` for catching from `AudioPlayer.play` even though it's required to handle all checked exceptions from the method.

riktenx:
you can ignore that for now, you're using the audioplayer which is what matters there

PR #12995, opened 2026-06-27
New plugin: Bank Watcher

Khaled2143:
Reopening as a fresh PR per the previous closure.

**Fix**: All network and price-fetch work, including the Thread.sleep rate-limiting now runs on a background thread via ScheduledExecutorService. The client thread is only used to read the bank container and item compositions; results are pushed back to the Swing UI via SwingUtilities.invokeLater.

**What it does**: Bank Watcher tracks your bank's total value and shows per-item price and quantity changes between scans, with daily-limited scans and gainer/loser/top-mover filtering.


riktenx:
use of thread sleep is not allowed

Khaled2143:
Removed Thread.sleep completely. The price fetches are now spread out using executor.schedule, and the results get put together through a callback once the last batch finishes.

PR #13011, opened 2026-06-28
Add OSRSFliAdd OSRSFlipper Telemetrypper Telemetry

DeadArrow4:
Adds OSRSFlipper Telemetry, a read-only local telemetry exporter for OSRSFlipper.

The plugin does not click, buy, sell, cancel, reprice, automate trading, or send data to a remote server. It only writes local RuneLite telemetry JSON for OSRSFlipper to read.


riktenx:
File i/o needs to happen in a plugin-specific subdir under .runelite/. You can make filepaths user-configurable but they have to _default_ to that which you are currently not doing.

PR #13012, opened 2026-06-28
Add OSRS Algo (GE log)

RichardLechko:
Adds OSRS Algo (GE log). A plugin that auto-logs Grand Exchange offer state (place / partial / fill / cancel) to a user-configured URL, defaulting to http://127.0.0.1:5000 (localhost). Intended to pair with a personal local backend for flip/arbitrage tracking; the `warning=` field discloses the external POST behavior.

  Source: https://github.com/RichardLechko/osrs-algo-ge-log


riktenx:

RichardLechko:
> * use of thread sleep is not allowed; use a scheduled executor etc. instead
> 
>     * use of thread interrupt is not allowed

Fixed. BackendClient no longer uses Thread.sleep or Thread.interrupt. postEvent now makes a single HTTP attempt and queues on failure; retries happen via the existing ScheduledExecutorService drain timer in OsrsAlgoPlugin.

PR #13014, opened 2026-06-28
Sorry last time I Promise - play-music

VitaMors:
Hey,
Yeah sorry haha will make sure not to do more than one pull request thingo next plugin or update, very new to coding and github still and been struggling to navigate the site and knowing exactly how to use it.

Thanks

### DIFF
--- a/plugins/play-music
+++ b/plugins/play-music
@@ -1,0 +1,2 @@
+repository=https://github.com/VitaMors/play-music.git
+commit=26efcff121c5245fe406f73673fed2d1628ce25d


riktenx:
jfyi ignoring the fact that the build is broken, i'm looking at your code and these lavaplayer/youtube dependencies you're adding aren't allowed. you can't add additional dependencies to your plugin

VitaMors:
ohhh sorry yeah my bad thanks for letting me know I'll go and have a look at what allowed so i don't make the same mistake in the future 

cheers again :)

riktenx:
i dont think what you're attempting to do here will be possible in an rl plugin unfortunately

PR #13020, opened 2026-06-28
Update OSParty to 1.0.1

iodrareg:
- Learner/teacher raids
- Killcount requirements.
- CoX layout scouting. Layout is pushed to the api on the advertisement of the host so that it's clear for applicants which layout they'll be doing.
- Defence tracking and map pings with the in-game overlays that go with them
- A pile of smaller stuff: country flags next to host worlds, some icon updates, the API base URL is now overridable for local dev, and the heartbeat keeps the still-open roles up to date.
- Big comment cleanup generated by el claudio


riktenx:

iodrareg:
> * use of getenv is not allowed; use a jvm sysprop instead

removed!

PR #13025, opened 2026-06-28
Update Fate Locked Ironman: gear-tier + slayer warnings, notifier, richer chunk info

Nubles:
This update adds: over-tier gear warnings, locked-slayer-task warnings (complete monster coverage), optional RuneLite notifications, transport-network (teleport/spirit tree/glider) locked-destination warnings, surrounding locked-chunk shading, richer "what's here" (shops/farm patches/POIs), and an unlock-progress HUD line. It reads the system clipboard (on a button) and optionally the newest fate-locked-bundle-*.json in Downloads both local. Clipboard payloads may be gzip+base64 (FLGZ: prefix), inflated with the JDK's GZIPInputStream.


riktenx:
all file i/o needs to happen within a plugin-specific subdir under .runelite/

Nubles:
Hopefully fixed, all file I/O is now confined to a plugin subdir (RuneLite.RUNELITE_DIR/fate-locked, created on startup). Dropped the arbitrary configurable path and the Downloads scan; the plugin only reads the newest fate-locked-bundle-*.json from that folder now (clipboard import is unchanged).

Just a heads up: this commit also includes a few additional local-only features added since the last push (world-map markers, infoboxes, diary-completion via varbits, an "open tracker" button, a clipboard re-import hotkey). Happy to split those into a separate PR if you'd rather review just the file I/O fix here.

PR #13042, opened 2026-06-29
Tob Utilites bugfix

NCG-RS:
Removed unnecessary consume call of KeyEvent


riktenx:
build=standard, not gradle

PR #13048, opened 2026-06-29
Add camera-locks

Varzeki:
Allows locking the camera so it doesn't move by accident.
Allows saving presets of camera positions and provides on screen guidance to manually return the camera to that position later.


Varzeki:
As I've now noticed that specifically camera presets are on the Rejected or Rolled Back features list, I have proactively refactored the plugin to simply lock when the user manually moves to a camera position instead of having a preset to do it for them. 

Hopefully this is more inline with what is acceptable, as all movement is now completely manual.

PR #13050, opened 2026-06-29
Add idle-notifier-overlay plugin

ytjc4y9mjg-bit:
Idle Notifier Overlay is a simple plugin that displays a blinking text box above your character's head when you stop animating or when your inventory is full.
The overlay is invisible while you are active, and appears automatically when:

Your character stops animating for ~1.8 seconds
Your inventory reaches 28 items

This is useful for catching AFK moments while skilling or training.


pajlada:
seems like you're missing plugin info in this pr

riktenx:
closing since this was submitted broken and remains broken, if you fix it you can make a new pr

PR #13062, opened 2026-06-29
Add OSRS Preview

sayuop:
Adds OSRS Preview, a local-only multi-client preview and focus-switching plugin for RuneLite.

The plugin discovers other local RuneLite clients running the same plugin, displays live sidebar previews, and allows clicking a preview to focus that client. 
Communication is local-only over 127.0.0.1 and temporary heartbeat files.

### DIFF
--- a/plugins/osrs-preview
+++ b/plugins/osrs-preview
@@ -1,0 +1,2 @@
+repository=https://github.com/sayuop/osrs-preview.git
+commit=a7d38ffe9e80117cdc216dc10a7db95c6c34c3cf


riktenx:
thanks for the submission but we no longer allow plugins that manipulate the game windows themselves

sayuop:
Thanks for clarifying. Just to make sure I understand the boundary: the plugin does not move, resize, overlay, automate input, or interact with the game client. The only window-related action is requesting focus for another local RuneLite client when the user clicks its preview, similar to manually Alt-Tabbing to that window.

Is programmatic focus switching itself considered disallowed window manipulation for Plugin Hub purposes?

riktenx:
yes

PR #13067, opened 2026-06-29
update fashionscape to 1.4.6

equirs:
big LoC numbers but it's entirely from adding unit tests, and some new test-only adjustments to build.gradle. figured I would add the tests all in one swoop without making changes to the plugin


riktenx:
The bot now reviews your code so individual diff size isn't as relevant. Though make sure you keep your codebase tight as there's a 200k token limit. Only code in your main source set counts (no comments,tests,resources etc. are counted)

PR #13077, opened 2026-06-29
add arceeus cd timer

jacobgantzpolyglot:
RuneLite plugin that overlays timers on the Arceuus spellbook for:

Greater Corruption
Death Charge
Mark of Darkness


riktenx:
doesn't build

jacobgantzpolyglot:
Updated the plugin source commit and removed the local Gradle Java home from the plugin repo. The build should use the clean commit now.

PR #13083, opened 2026-06-30
Add idle-familiar plugin

mackenzie0cameron0:
The idle familiar plugin is a pop-out, always on-top, widget. It reads a players in-game state and plays a corresponding animation. This is intended to act as an idle notifier while the client is minimized. 
It's a status display only: no game input, automation, or interaction with the client. The README has previews and full feature list.


riktenx:

mackenzie0cameron0:
I've removed the reflection and replaced it with a whitelist of animation IDs. The plugin should still function the same. Thank you for your time in reviewing this.

riktenx:
programmatically manipulating the main game window, such as trying to request focus, minimize, move it, etc. are not allowed

mackenzie0cameron0:
Removed the return-to-game button & added shimmer effect to the collapsed widget for idle status

PR #13089, opened 2026-06-30
Update loot-lookup to v1.2.2 - fix broken icons

donth77:
- Fix item icons never loading: [https://github.com/donth77/loot-lookup-plugin/issues/50](https://github.com/donth77/loot-lookup-plugin/issues/50)
     - The OSRS Wiki began returning WebP for .png URLs, which Java's ImageIO can't decode - so most item icons were broken and loading animations spiked CPU. The image request now advertises only formats ImageIO supports.

### DIFF
--- a/plugins/loot-lookup
+++ b/plugins/loot-lookup
@@ -1,3 +1,3 @@
 repository=https://github.com/donth77/loot-lookup-plugin.git
-commit=f267c38c900d6bd58a53e690d8c26756a5f00030
-authors=donth77,riktenx,iProdigy
+commit=283a82dadfa2e8e68034a2fbab0353a9eeb0fcfa
+authors=donth77,riktenx


riktenx:
are you able to drop the jsoup dependency and just parse wiki pages syntax by querying them with ?action=raw?

this would allow you to set build=standard and then your plugin could be automatically reviewed and updated by the bot

donth77:
@riktenx 
I looked into this, and unfortunately querying with ?action=raw doesn't return the full drop tables, so parsing the HTML with jsoup may still be the best option.

The wiki generates some of the tables with Lua modules. For example, `{{GemDropTable}}` is a wrapper around `{{#invoke:GemDropLines|main}}` ([Module:GemDropLines](https://oldschool.runescape.wiki/w/Module:GemDropLines?action=raw)). That module computes each row from an access rate that varies per monster. The rare table works the same, and the herb and seed tables expand the same way.

PR #13091, opened 2026-06-30
osrscn: update to 0.1.2

Aoldbald:
Updates OSRSCN Chinese to 0.1.2.

- ~1,600 more translations (item names/examines, skill-unlock messages, NPC dialogue)
- more robust lookup: folds typographic quotes/dashes and non-breaking / full-width
  spaces, and falls back across translation categories, so fewer lines stay English
- new opt-in toggle (default off) that writes untranslated game text to a local file
  the user can choose to share; nothing is uploaded automatically


riktenx:

Aoldbald:
Thanks for the review. Fixed in bd9c432: replaced the plugin-created daemon thread + Thread.sleep/interrupt in MissingCollector with the injected ScheduledExecutorService (scheduleWithFixedDelay). No other Thread.sleep/interrupt in the plugin.

PR #13113, opened 2026-07-01
Update Boss Helper Bank Sync

ENiceApps:
Updates **Boss Helper Bank Sync** to a newer commit of the same repo.

Changes since the merged version:
- Simplified the config panel to a single opt-in toggle ("Save my bank & gear for the web app"). Removed two settings that weren't needed — the write throttle is now a fixed 2s, and the companion web-app URL is fixed.
- New plugin icon.
- Minor git history cleanup (no functional change).

Behavior is unchanged: read-only, **no network requests**, writes only a local file in the RuneLite directory, opt-in and disabled by default. BSD-2-Clause.


ENiceApps:
Thank you for approving and merging my app earlier. I'm making a slight change here and I hope to not have to make a change for a while. Now I'm working on adding Blood Moon Raises content to the app haha.

PR #13120, opened 2026-07-01
Update ping-grapher to 1.7.1

yuh25:
From [Ned-57/fix/brazil-worlds-ipv6-ping](https://github.com/Ned-57/Ping-Grapher/tree/fix/brazil-worlds-ipv6-ping)

"Replace the Ping.ping() call with a local pingWorld() helper that uses getAllByName() to prefer IPv4 when available. If only IPv6 is returned, falls back to a direct TCP connect on port 43594, which Socket.connect() handles natively on all platforms."


pajlada:
Hello, you must now update your plugin's `runelite-plugin.properties` file to include a `build` key. First try using `build=standard` and pushing that update.

If that doesn't work, you can try `build=gradle`.

PR #13121, opened 2026-07-01
Add Digital Compass

anonyser:
Adds Digital Compass, a small overlay that shows which way your character is facing in degrees (north 360, east 90, south 180, west 270). It draws a small dial with N/E/S/W markers and an arrow for the facing direction, next to the number.

It's display only. It reads the player's orientation and draws an overlay, nothing else, so no input and no automation. There are options for font size, bold, text colour, an outline for readability, toggling the dial and the cardinal label, and a small calibration offset.

Repo: https://github.com/anonyser/runelite-digital-compass

### DIFF
--- a/plugins/digital-compass
+++ b/plugins/digital-compass
@@ -1,0 +1,2 @@
+repository=https://github.com/anonyser/runelite-digital-compass.git
+commit=6ebc9cf1c63a0ef5b475a0f3d7ffe85901c626b2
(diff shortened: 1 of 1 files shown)


anonyser:
Updated the commit hash. The compass now follows the camera direction and reads the correct headings (verified in-game), and it is resizable with a configurable text outline.

riktenx:

Reasel:
> set `build=standard` in runelite-plugin.properties

Is there a reason this check is not added to the RuneLite Plugin Hub Checks so as not to force you and other reviewers to make this comment all the time?

cdfisher:
It's not always applicable

ldavid432:
Also it does show in in the build workflow: 

Many people just do not pay attention to the workflow

anonyser:
Good catch, added build=standard and pushed. Updated the commit above.

PR #13136, opened 2026-07-01
Add All Skills Calculator

Reasel:
A fork of the built-in Skill Calculator that works for every skill and supports custom actions and hourly rates with copiable XP totals.

All Skills Calculator panel


riktenx:
we don't allow plugins to use 'better' 'improved x' whatever in their names, you'll have to give this a different name

riktenx:
https://github.com/runelite/plugin-hub/pull/13136#issuecomment-4858896447 remains unaddressed

PR #13137, opened 2026-07-01
Update farming-profit

JaredEzz:
Set author= to the current maintainer (JaredEzz). Original authorship stays credited in the plugin's README Attribution section and LICENSE per BSD-2-Clause.


JaredEzz:
This update now also fixes a client crash in the currently-live 1.0.0: `IconStyle` was package-private while `FarmingProfitConfig` (a public interface) exposes it as a config-item return type, so RuneLite's dynamic config proxy throws `IllegalAccessError` under the real Hub external-plugin classloader (doesn't repro in a dev/gradle-run client, since that shares one classloader) — crashes the client shortly after login. Fixed by making the enum public. Would appreciate a look given it's actively affecting installed users. Thanks!

PR #13141, opened 2026-07-01
update nex-nostalgia

hex-agon:


dontirun:
@hex-agon . The plugin is very loud now. Even at `-25` (the max negative) volume it's much louder than it was before. There is also a lot of reverb in the sounds

hex-agon:
@dontirun submit an issue to the plugin repo with your ingame sound settings & plugin config.

PR #13146, opened 2026-07-01
Update All Skills Calculator to 1.0.0

Reasel:
Sets the plugin author to Reasel and adds a proper version (1.0.0) so the install page shows a version instead of a commit hash.


Reasel:
Sorry I missed this in the initial submission. I will try to check for that in the future before PR to not waste anymore time from reviewers for simple stuff like this.

PR #13148, opened 2026-07-01
Update Bank Highlight Search to 1.0.0

Reasel:
Bumping the version to 1.0.0 so it shows in the download.


coopermor:
You need to put the version in runelite-plugin.properties not in build.gradle

PR #13150, opened 2026-07-01
bank-templates: 1.5.2

TheSpryt:
Updates bank-templates from 1.5.1 to 1.5.2 (commit 7ad93a925818d0f71772aab0ba2306249f8e34fc).

- Template tabs the player's bank doesn't have yet (rendered as extra "virtual" tabs) now show their value in the bank title, matching real tabs.
- Fixed the extra tab buttons disappearing while certain numbered tabs were selected (a selected tab's duplicate background collapsed the computed tab spacing).

No new external requests or permissions.


TheSpryt:
Closing for now - folding an additional fix into the next hub update so they ship together.

PR #13151, opened 2026-07-01
update dink to v1.14.2

Felanbird:
stuff


riktenx:
the jetbrains annotation stuff isn't available in standard build, you will want gradle here i think

riktenx:
or you can swap all those with the lombok equivalents, some others have done so

PR #13159, opened 2026-07-02
update last-prayer-highlight: fix icon size, re-enable

robisa693:


coopermor:
The plugin violates the 3rd party client guidelines and will not be re-enabled.

robisa693:
> The plugin violates the 3rd party client guidelines and will not be re-enabled.

Ok, thank you for the clarity!

PR #13163, opened 2026-07-02
Submit new plugin "Beep Deep"

Cold-1:
This plugin allows players to have custom sounds play when certain things (See list bellow) happen in a raid. The plugin was made as an inside joke, none of the events should give an unfair advantage, no sounds are replaced by the plugin either, events and ids are hard coded and only the sounds/volume cannot can be changed.

Enter Path of Crondis
Leave Path of Crondis
Crocodile defeated
Crocodile attack
Het seal not one-phased
Apmeken issue not fixed
Vault opens with no rare loot
Vault opens with rare loot

### DIFF
--- a/plugins/beep-deep
+++ b/plugins/beep-deep
@@ -1,2 +1,2 @@
 repository=https://github.com/Cold-1/beep-deep.git
-commit=b98e88e1d7398c843d7762be88732f4c61a456eb
+commit=48599672720101bc2eeaf3fa3dbcc7b11dd4a9de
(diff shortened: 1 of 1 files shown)


pajlada:
> Crocodile defeated
Crocodile attack

those are the only ones i see as sus, rest seem fine to me

riktenx:
please address the above

Cold-1:
Thanks for reviewing so quickly! I removed the Crocodile defeated and Crocodile attack events.

PR #13170, opened 2026-07-02
Add maggot-king

randytkrx:
Adds Maggot King, a helper plugin for the Maggot King boss.

Features:
- Reactive callouts for the boss's telegraphed mechanics: screech (configurable text), slam and airborne larvae. All react to visible animations only, nothing is predicted.
- Attack style indicator showing the last style observed from the boss. It starts empty each kill and only ever reports what has already happened.
- Larvae highlighting with a separate color for airborne larvae, boss true tile colored by phase, and a corpse highlight labelled with the user's preferred loot interaction.
- Boss health with phase thresholds, an optional burn damage counter, and a session panel with kills, deaths, kill times, personal best per profile, loot value and egg breakdown.
- Optional cosmetic toggles to hide the post screech rock litter and the darkwood trees around the arena, off by default. Interactable objects and hazards are never hidden.

The plugin deliberately omits hazard tile marking, projectile landing tiles and any prayer or next attack indication per the third party client guidelines.

### DIFF
--- a/plugins/maggot-king
+++ b/plugins/maggot-king
@@ -1,2 +1,2 @@
 repository=https://github.com/randytkrx/maggot-king.git
-commit=8a010fd198856ed88c3220c30c6f11824df4683f
+commit=0f30d49f7128be5acb2105565a4622c0189f3eff
--- a/plugins/maggot-king
+++ b/plugins/maggot-king
@@ -1,0 +1,2 @@
+repository=https://github.com/randytkrx/maggot-king.git
+commit=3d3f682b819cf1cb2ad1b31efcbff1f30bdb31d8
--- a/plugins/maggot-king
+++ b/plugins/maggot-king
@@ -1,2 +1,2 @@
 repository=https://github.com/randytkrx/maggot-king.git
-commit=3d3f682b819cf1cb2ad1b31efcbff1f30bdb31d8
+commit=8a010fd198856ed88c3220c30c6f11824df4683f


ldavid432:
Callouts for boss mechanics are definitely not allowed. https://secure.runescape.com/m=news/third-party-client-guidelines?oldschool=1

Additionally a last prayer tracker has previously been rejected as a plugin.

EDIT: By last prayer tracker I am referring to the last attack style used

randytkrx:
> Callouts for boss mechanics are definitely not allowed. https://secure.runescape.com/m=news/third-party-client-guidelines?oldschool=1
> 
> Additionally a last prayer tracker has previously been rejected as a plugin.
> 
> EDIT: By last prayer tracker I am referring to the last attack style used

Oh i see, my bad i was basing the callouts feature on the xarpus screech in tob qol plugin where it brings up an infobox when there is a screech about to happen, Also removing the last prayer tracker. Thanks for the info!

randytkrx:
Thanks for the review. I removed the mechanic callouts (screech, slam and airborne alerts) and the attack style indicator entirely, and updated the descriptor to the new commit. What remains is larvae and boss tile highlighting, boss health with phase thresholds, a corpse loot reminder, session and loot tracking, and two cosmetic scenery toggles.

ldavid432:
Also I think you will need to remove the color changing aspect of tile highlight as that "Indicates the time where a boss mechanic may start or end"

randytkrx:
Removed the phase coloring and the phase threshold display as well. The true tile is now a single configurable color and boss health is shown as a plain number. Descriptor updated.

PR #13190, opened 2026-07-03
Blast mine tracker bugfix + warn user + delete past sessions

samkovaly:
warns user when they start BlastMineTracker while in an active session. Allows deleting of past sessions


samkovaly:
You marked as waiting for author, what is the issue?

PR #13196, opened 2026-07-03
Add last overhead prayer plugin

justinizmlg12-boop:


ldavid432:
As I already said on your previous PR this plugin will not be accepted

PR #13201, opened 2026-07-03
Add Remember Last Prayer

Shackin:
When an overhead prayer is switched or disabled, displays that prayer as an icon or InfoBox.


coopermor:
We will not be hosting plugins that track your last prayer as they are to assist with boss mechanics and are not compliant with the 3rd party client guidelines.

PR #13203, opened 2026-07-03
Ground object id highlighter

WoozOSRS:
## Description

Ground Object ID Highlighter allows users to highlight every occurrence of configurable GameObject and GroundObject IDs as they appear in the current scene.

Unlike the built-in Object Markers plugin, highlights are based on object IDs rather than fixed world coordinates, making the plugin useful for dynamically spawned PvM mechanics and temporary encounter objects.

### Features

- Highlight configurable object IDs
- Separate support for GameObjects and GroundObjects
- Configurable colors
- Adjustable opacity
- Tile fill / outline render modes
- Border width customization
- Optional pulsing highlights
- Optional region restrictions
- Automatic cleanup when objects despawn or instances change

### DIFF
--- a/plugins/ground-object-id-highlighter
+++ b/plugins/ground-object-id-highlighter
@@ -1,2 +1,2 @@
 repository=https://github.com/WoozOSRS/Ground-Object-ID-Highlight.git
-commit=5b3cbde4fa5197b9a7d4f99d1d32820b59aca027
+commit=4552bbfa236d36fcdda5b83b4ba69f1341b3fe06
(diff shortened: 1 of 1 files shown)


Bram91:
> Reviewing

> Plugins that use player provided IDs for the entirety of their functionality can cause moderation issues and outcomes that break Jagex's plugin rules. Due to this fact we will not be accepting any new ID based plugins. Plugins that use a specific set of IDs but do not allow user input will still be accepted. (e.g. plugins like Vardorvis Projectiles which only allows you to change a specific projectile to a set list of projectiles.)

PR #13208, opened 2026-07-03
update osrs-tcg

Azderi:
Collection album updates, Dink and webhook integration


riktenx:
don't read/write files outside of your plugin-specific subfolder within .runelite, it looks like this stuff you added is for debugging purposes. you can either add a separate debug plugin in the test source set (which won't ship with your plugin and won't get looked at but you can use it during development) or just remove it

the sailing plugin has an example of a debug plugin if you want to look there, they can be pretty handy

PR #13221, opened 2026-07-03
Bump version and add H2 dep

DNiemeijer:


riktenx:
this will prevent your plugin from being bot-/auto-reviewed, meaning all review going forward will have to be manual. your call but not sure if you knew/want that to be the case

iodrareg:
Would sqlite or a flatfile be included in the autoreview?

riktenx:
if you have to set build=gradle it is no longer autoreview

riktenx:
if you could clarify what you intend to do here

iodrareg:
We will just move to locally storing and reading a json file with some data to fill a list of favorites, blocked and historical data of parties. 

I assume we can just update the hash of **this** PR once we have implemented and tested it?

riktenx:
yep

iodrareg:
@riktenx i opened https://github.com/runelite/plugin-hub/pull/13277 that reverts the H2 dependency and now has a local file which is written/read to store certain values.

you can close/reject this PR if you want :)

PR #13238, opened 2026-07-04
Update afk-overlay

zmertz:
Fixes a small bug where moving the mouse while clicking on the overlay window didnt bring up the client

### DIFF
--- a/plugins/afk-overlay
+++ b/plugins/afk-overlay
@@ -1,2 +1,2 @@
 repository=https://github.com/zmertz/AFK-Overlay.git
-commit=fb7e6035a3a9ac38a567842f6df3298e8dff50c8
+commit=1eb3968087178b931babf9a521298c4f01b7ca0f
--- a/plugins/afk-overlay
+++ b/plugins/afk-overlay
@@ -1,2 +1,2 @@
 repository=https://github.com/zmertz/AFK-Overlay.git
-commit=b4866ed0b522f4e73971da5bddc1ab973a22a3f6
+commit=fb7e6035a3a9ac38a567842f6df3298e8dff50c8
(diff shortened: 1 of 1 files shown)


riktenx:
remove .class files from your repo

riktenx:
you cannot programmatically move/minimize/request focus on the runelite window. you can do whatever with separate windows you create

zmertz:
> you cannot programmatically move/minimize/request focus on the runelite window. you can do whatever with separate windows you create

I'm a little confused. I checked the provided resources to ensure this would be in line with Jagex's rules. I didn't see anything about focusing the Runelite window. This plugin has been in the game for about a year now with no issue. Clicking the window acts in the same way as clicking the Runelite button in the taskbar, so I wouldn't think that is an issue. Could you point me to a rule section of FAQ detailing this?

ldavid432:
> I'm a little confused. I checked the provided resources to ensure this would be in line with Jagex's rules. I didn't see anything about focusing the Runelite window. This plugin has been in the game for about a year now with no issue. Clicking the window acts in the same way as clicking the Runelite button in the taskbar, so I wouldn't think that is an issue. Could you point me to a rule section of FAQ detailing this?

This rule was just recently added due to the situation with Marni & click-to-minimize

https://github.com/runelite/runelite/wiki/Rejected-or-Rolled-Back-Features#removed-in-the-past

zmertz:
I understand the reasoning behind the new rule given the Marni situation, and I think it's a good idea. Looking at this plugin specifically though, I wanted to raise whether grandfathering it in might be worth considering.

This plugin has existed for awhile without issue. The window focusing is strictly a 1-1 action: the user clicks the overlay, RuneLite comes to front. It can only ever target one window, requires deliberate user input each time, and has no capacity to automate anything across ticks or accounts.

I recognize you may want a clean rule with no exceptions, but I just wanted to make the case since I don't think this specific implementation could result in exploits. Let me know your thoughts!

riktenx:
yeah no we will need you to remove this, otherwise we'll be forced to disable the plugin

jfyi the whole "1:1" thing is not real, the playerbase made that up

PR #13262, opened 2026-07-04
Add No-Misclicks-In-Bank

Welkon1:
Adds the No-Wield-No-Wear-In-Bank plugin.

Removes the Wield, Wear, and/or Drink options from items in your inventory while the bank is open, to prevent accidental clicks when depositing all. Each option is toggleable.

Repository: https://github.com/Welkon1/No-Misclicks-In-Bank


riktenx:
doesn't build

riktenx:
not really one to tell you your business but surely you want to exercise some forethought with the name here and maybe call it something more generic? "bank option hider" considering you have also separately submitted "no drink in bank"

Welkon1:
Yeah I'll combine them both into one, fairly new to this.

riktenx:
delete all .class files from your repo

PR #13277, opened 2026-07-04
osparty to 1.0.18

iodrareg:
- removes H2 dependency and now uses jsonfiles to store some local data
- update to user interface to include a history tab for local historical parties


riktenx:
build=standard, not gradle

iodrareg:
should be good now, thanks for the swift reviews!

PR #13307, opened 2026-07-05
Update upper-case-transformer: Fix !-prefixed chat commands being uppercased

blocks-dot:
## Summary
- Skip chat messages whose first visible character is `!`, so in-game command lines (e.g. `!kc Wintertodt`) are left unchanged
- Detection ignores chat color tags and leading whitespace before checking for the command prefix
- Normal all-caps/title-case messages are unaffected


blocks-dot:
fixed failing build

PR #13336, opened 2026-07-05
Update icon

DuckTown407:
Icon was mistakenly the default one, not the one I had made. Updated the icon.png


DuckTown407:
I changed the commit hash from the original submission. I am new to git and I keep making mistakes. Everything I intended to fix should be fixed now with the latest commit.

PR #13350, opened 2026-07-06
Add PianoQueen-Music Plugin

LogicalSoIutions:
Add's the PianoQueen plugin. 

Let's users set one of the recordings from PianoQueen as their sound when they enter their POH and also replaces them in the in-game Audio Player so they can pick to listen to them whenever they want. 

Big thanks to [alowaniak](https://github.com/alowaniak/music-replacer) for having easy to understand and re-use code making this take a lot less time than it should have.

### DIFF
--- a/plugins/pianoqueen-music
+++ b/plugins/pianoqueen-music
@@ -1,2 +1,2 @@
 repository=https://github.com/LogicalSoIutions/PianoQueen-Music.git
-commit=c8914e04e96b2db5268ccdef9c6356ab8017e95e
+commit=235086fd3b3df63f6c2c128d4a6729fe6b1f5b39
--- a/plugins/pianoqueen-music
+++ b/plugins/pianoqueen-music
@@ -1,0 +1,2 @@
+repository=https://github.com/LogicalSoIutions/PianoQueen-Music.git
+commit=1adbc17510edfffd8b92c62a078eaba610545113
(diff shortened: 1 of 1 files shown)


Bram91:
Do you have the creator's permission to use these sounds?

LogicalSoIutions:
Yes, I worked with her privately via Discord. Can share screenshots if
needed.

riktenx:
you should just encode these as .wav so you don't need the jacomp3 dependency and can set build=standard. this would make your plugin eligible for automatic review and you would not have to wait for someone to manually review your plugin.

LogicalSoIutions:
I originally had them as WAV files but it was ~320mb for all the sounds as WAV compared to ~36mb as MP3

riktenx:
that is not a problem on our end since you're downloading the files from github, but it's your call. no idea when this will be reviewed though, if ever, there is not that much capacity/desire for manual review these days

LogicalSoIutions:
Forgot to post my comment when I pushed the commit, but, switched back to WAV

riktenx:
we do not allow new usages of [script 9238](https://github.com/runelite/cs2-scripts/blob/0f363f0e3473f4b43927895e4e616cdd83cb9b72/scripts/%5Bproc%2Cscript9238%5D.cs2#L12) because it triggers a server action, misuse is liable to trigger anticheat bans

PR #13351, opened 2026-07-06
Update auto-entity-hider

P2GR:


riktenx:
doesnt build

P2GR:
> doesnt build

Apologies, should be good now.

PR #13356, opened 2026-07-06
Add chat-tab-hotkeys

tinokaartovuori:
Adds **Chat Tab Hotkeys** — configurable hotkeys for navigating and filtering the chat.

It only automates chat-UI actions the player can already do by mouse (same category as Chat Filter / Chat History / Toggle Chat): switch chat tabs, set a tab's filter, clear a tab's history, collapse/expand the chatbox, and set the chat input mode. No gameplay automation, no overlays, no network or file access, no third-party dependencies.

- Repo: https://github.com/tinokaartovuori/chat-tab-hotkeys
- License: BSD 2-Clause
- `runeLiteVersion = 'latest.release'`, Java 11

### DIFF
--- a/plugins/chat-tab-hotkeys
+++ b/plugins/chat-tab-hotkeys
@@ -1,0 +1,2 @@
+repository=https://github.com/tinokaartovuori/chat-tab-hotkeys.git
+commit=1edcf3a2a8b279bec1ad9a413ee2cf83e38b8ce0


riktenx:
we don't allow clientscript execution for the purpose of switching active interfaces, even when behind keypresses, because this is liable to trigger anticheat and get users banned

tinokaartovuori:
@riktenx
I posted the same message to the RuneLite Discord development channel:

The thing I'm not sure about: there's a kinda similar plugin already on the hub, Toggle Chat, and it opens the chatbox to a chosen tab with runScript(175, 1, tabIndex):
https://github.com/MoreBuchus/buchus-plugins/blob/toggle-chat/src/main/java/com/togglechat/ToggleChatPlugin.java#L152

That's basically the same call I'd want to use, and if I understood correctly it's packet free, it just sets the CHAT_VIEW varc client side (175 is the chat tab button's own onop script).

Questions:
1. Is it fine to use that same runScript(175, 1, N) on a keypress to switch the current tab, since Toggle Chat already uses it to open to a chosen tab? Or is the problem the purpose (switching tabs at all) regardless of how it's done?
2. If that script's a no go, would just setting the varc directly (setVarcIntValue(CHAT_VIEW, N)) be ok, or is switching tabs off the table entirely?
3. And separate from the tab stuff, would the rest be ok on its own: a close/open chat toggle like Toggle Chat, clear history (same thing the core Chat History plugin does), and a bind to change which channel you type in?

PR #13187, opened 2026-07-03
Update pvp-performance-tracker to v1.8.2

Matsyir:
~~*Note: Currently live is `v1.8.0`, there is another PR for `v1.8.1` (#13143)
Feel free to skip `v1.8.1` and push `v1.8.2` right away.
But if it would be longer, there's a fairly significant visual issue in `v1.8.0` that's been fixed in `v1.8.1`, so it wouldn't hurt to have that merged ASAP.~~ (thanks for merging it fairly quickly 😄)

1.8.2:
- Improved filter logic/efficiency
- Improved overall performance & responsiveness - various UI threading optimizations. It feels much smoother & snappier overall. From me & LogicalSolution's testing, there doesn't seem to be any issues, but this includes some of those potentially problematic race-condition-causing threading changes (I think, I don't fully understand swing or how this threading all works, but at this point I do generally understand what kind of actions should be going on client thread vs. swing ui thread)
- Customizable UI ~ New config options for: hide panel BG images, center vs. side-align text, and customize nearly all colors, including a dropdown for preset color styles/configurations.
- Various code/comment improvements & clarifications. Standardized 95% of Color uses into PvpColorScheme class, which also houses the preset color styles.
- Adjusted death icon (red skull) placement

Preview of new color presets, configs, and various other UI adjustments:
image

A confirmation popup appears when selecting a color preset config, to clarify that the config UI doesn't visually update when we change these configs and internally apply the newly selected color preset:
Screenshot 2026-07-02 175253

Many thanks once again to @LogicalSoIutions for helping with many of the fixes involved here, testing, as well as Pan1c + Lagunarium for constantly testing and giving valuable feedback :)

Thanks to @Sacca-1 who also made a PR for the 1.8.0->1.8.1 fix, although I took @LogicalSoIutions' since it had a smaller diff.

*edit: I force-pushed another commit about 1hr after opening this PR, there was simply a broken tooltip I forgot to fix initially. Should be fine for all the core behaviors

edit2: Force-pushed a 2nd commit, new fix by LogicalSolutions. There was a bug where clients keep retrying a failed network request every 60s on loop. https://github.com/Matsyir/pvp-performance-tracker/pull/94

*edit3: added more various UI fixes & improvements, notably converted the fight history format from just 1 big .json to .json.gz chunks, went from 17.2mb for 700 fights to 1.2mb
image

Also added this warning to clarify level/pray assumptions 
Screenshot 2026-07-04 042421

PR #13372, opened 2026-07-06
Update pvp-performance-tracker to v1.8.3

Matsyir:
- Fixed/re-applied data migration from 1.8.1->1.8.2. It only worked if you had >100 fights saved, otherwise it did nothing, and it also failed to import your last chunk of up-to 100 fights. Silly logic bug/loop fail
  *Re-applying this for people who already ran the update doesn't introduce any issues with how the previous format vs. imported format vs. new format is structured. (See #13187 for example files/details)
- Support for seeking arrows. (Thanks to @Sacca-1 for this PR + @LogicalSoIutions for reviewing [#92](https://github.com/Matsyir/pvp-performance-tracker/pull/92) [#98](https://github.com/Matsyir/pvp-performance-tracker/pull/98))
- Fixed fight deletions not persisting. Also kept previously "broken" behavior as "Hide fight (until reload)" feature.
- Fixed synced fight statistics being visually inverted. (@LogicalSoIutions [#97](https://github.com/Matsyir/pvp-performance-tracker/pull/97))
- Include buttons to wiki + discord on panel (hide via right-click or config).
- Various fixes & improvements regarding recent UI changes.
- Replaced default orange fight border with white, keep orange only for hover.

Social buttons + visibility toggling options / new hide fight option
https://github.com/user-attachments/assets/81e8ce22-0cd8-4ae0-834f-ba0914490a66

New white border:
Screenshot 2026-07-06 133740

### DIFF
--- a/plugins/pvp-performance-tracker
+++ b/plugins/pvp-performance-tracker
@@ -1,2 +1,2 @@
 repository=https://github.com/Matsyir/pvp-performance-tracker.git
-commit=df2d08578bf66e0ae7852e28ed7513f8417d34fb
+commit=960d614d0a1f0184ee20089aa3b6ca5f4dcd3cc4


Matsyir:
re-pushed a few commits - there was still 1 issue with removals, if you tried to "Delete forever" a fight from the ongoing session before it was saved, it would actually just hide it visually and save it next time we wrote data

also changed default fightHistoryRenderLimit from 200 to 30, and force-change it as well. Totally unnecessary to render 200 with how the total stats + filters apply to all your saved fights, most people probably dont even need 30 rendered. Otherwise some more random 1px adjustments and small refactors

PR #13385, opened 2026-07-07
Add Better Object Hider

JakhRS:
Adds **Better Object Hider**, a client-side scenery-decluttering plugin.

Repo: https://github.com/JakhRS/better-object-hider (BSD-2-Clause)

### What it does
Hides game objects at three scopes, each a deliberate manual right-click choice:
- **This tile** — one specific object on one tile (the core differentiator).
- **This map area** — one object ID within a single map region.
- **Everywhere** — one object ID globally (the classic behaviour).

Hides are organized into named groups you can enable/disable, rename, reorder via drag-and-drop, and import/export as clipboard codes. Tile/area hides are stored instance-safely (region-relative, like Ground Markers) and are scoped so an instance hide never leaks to the template's overworld location.

### Why a new plugin rather than a PR to custom-object-hider
Custom Object Hider is ID-only. The headline feature here is hiding a *single* object on a *single* tile (and per-region), plus instanced-area safety, groups, and sharing — a different data model and UX that outgrew a minimal upstream diff. Full credit to LuxOG's custom-object-hider in the README; the render-suppression approach builds on it.

### Compliance
- Purely cosmetic: `RenderCallback` draw suppression only. Nothing is sent to the server; menu entries are `MenuAction.RUNELITE`.
- No automation — every hide/unhide is a manual click.
- No network calls, no data collection, no new dependencies.
- Raid-mechanic objects (ToB Bloat pillar, Sotetseg maze tiles) are blocked from hiding and stripped from imports.

### DIFF
--- a/plugins/better-object-hider
+++ b/plugins/better-object-hider
@@ -1,0 +1,2 @@
+repository=https://github.com/JakhRS/better-object-hider.git
+commit=3d81c167f5b84dfe1942001b79955db88ef5a857


riktenx:
Thanks for the submission but please review https://github.com/runelite/runelite/wiki/Rejected-or-Rolled-Back-Features

Since it seems like your intent here is to explicitly be ID-only this is just not allowed and so I'm closing this.

JakhRS:
@riktenx Thanks for the quick look, and understood on the ID-based policy. Could I clarify the intent to see if a revision qualifies? The core feature is hiding a single specific object on a single tile — the player right-clicks one object and hides just that instance, stored by world location like Ground Markers/Object Indicators, not by acting on an ID. IDs only appear as fallback labels for unnamed objects. The "Hide all of ID" options are secondary; if those bulk-ID options are the blocker, I'm glad to remove them and submit a purely per-tile version. Would that be acceptable, or is per-object scenery hiding out of scope entirely now? Want to understand the line before reworking.

riktenx:
that's just user-input ID with extra steps. we've had a plugin exactly like that before and we had to disable it for moderation issues.

essentially the only ID-based hiding you can do is hardcoded. for example there is a plugin that just hides stalactites in caves, etc.

JakhRS:
@riktenx Thanks for taking the time to explain, I appreciate it.

Trying to understand the line here: custom-object-hider and object-hider are both currently on the hub and do ID-based object hiding. Is the distinction that those are grandfathered and the policy applies only to new submissions?

And on scope, the primary feature here isn't bulk ID hiding, it's hiding a single specific object on a single tile that you right-click in the world, stored by location like Ground Markers / Object Indicators. The object ID only surfaces as a label when an object has no name, so there's no typing IDs or acting on an ID as a class. I get the "user-input with extra steps" point for the hide-all-of-ID option specifically, I'm mainly trying to understand whether the single-object, single-tile case reads the same way, since that's the actual differentiator.

Either way I respect the call and appreciate the context.

JakhRS:
@riktenx I'm sorry man I'm not trying to be annoying but I just really don't think this is non-compliant! 

Hope you can give this a deeper look.

riktenx:
you can come ask about this in the discord or something, not discussing it here

PR #13390, opened 2026-07-07
create haggle-helper

gavin-lb:
- Highlight profitable and unprofitable shop sales with configurable overlays
 - Show how many items can be sold profitably (Sell 1, Sell 5, Sell 10, etc.)
 - Display potential profit and current shop prices directly on item overlays
 - Show tooltips detailing the revenue and profit of the current action
 - Display a profit summary panel with total profit and profit/hr
 - Automatically import Grand Exchange prices when adding tracked items
 - Block unprofitable transactions with configurable allowance for bulk transactions
 - Fully customizable colors, text displays, and overlay positioning

### DIFF
--- a/plugins/haggle-helper
+++ b/plugins/haggle-helper
@@ -1,2 +1,2 @@
 repository=https://github.com/gavin-lb/haggle-helper.git
-commit=9180d0675caecc02ac175f682ddad0a8cec1c205
+commit=aebf3de3b4e8d50c61ddb46282fcb06ccb876e02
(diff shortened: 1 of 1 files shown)


riktenx:
broken

riktenx:
build=standard, not gradle

gavin-lb:
> build=standard, not gradle

Could you clarify what you mean? I am using gradle build type because I have a groovy build script which consumes the wiki bucket api to generate a json data file

riktenx:
that shouldn't matter, you can have custom script stuff in your build.gradle just for running locally and still set build=standard and it should be fine. build=gradle is basically only needed if you have additional dependencies

PR #13396, opened 2026-07-07
Update Glamourer to 1.7.4

jhughes:


riktenx:

PR #13405, opened 2026-07-07
Update DropTracker to v5.4.0

joelhalen:
Adds a number of reliability enhancements and features:
- Fixes raid split tracking logic
- Reset data for various handlers on relog to prevent partial data issues
- fix EDT blocking group config fetching
- fix webhook url management to be more resilient
- move image conversion off frame listener thread to prevent client stalls
- replace boss-name switch method with a registry
- implement unit tests for future development
- support tracking deaths
- support tracking achievement diaries
- accurately determine if the player's plugin is out of date and notify them
- enhancements to submission panel to provide better insights into the lifestyle of a submission

_edit_
Also:
- Removed all video / lwgjl functionality
- Enhanced in/outbound connections for users who enable them
- Added support for events
- Fixed Maggot King tracking
- Removed the every-single-login welcome message the plugin sent its users.

Much of these changes were assisted using Claude Fable 5, just for full transparency :)

### DIFF
--- a/plugins/droptracker
+++ b/plugins/droptracker
@@ -1,3 +1,3 @@
 repository=https://github.com/joelhalen/droptracker-plugin.git
-commit=2a33535b9c677e4c13791f2a69e8066f0882396e
+commit=abb7818fb5e83722029091690b9371077e47c7ef
 authors=joelhalen,OSRSKoeppy
--- a/plugins/droptracker
+++ b/plugins/droptracker
@@ -1,3 +1,3 @@
 repository=https://github.com/joelhalen/droptracker-plugin.git
-commit=abb7818fb5e83722029091690b9371077e47c7ef
+commit=7f76641149c0d908cabb0e76f2eef80e2e8fb1b7
 authors=joelhalen,OSRSKoeppy
--- a/plugins/droptracker
+++ b/plugins/droptracker
@@ -1,3 +1,3 @@
 repository=https://github.com/joelhalen/droptracker-plugin.git
-commit=dbf916467265e445dfa6fc2b4f9a6788b1f24324
+commit=2fb52ec58f081d3e29e9433b7b678f47b036b050
 authors=joelhalen,OSRSKoeppy
--- a/plugins/droptracker
+++ b/plugins/droptracker
@@ -1,3 +1,3 @@
 repository=https://github.com/joelhalen/droptracker-plugin.git
-commit=7f76641149c0d908cabb0e76f2eef80e2e8fb1b7
+commit=dbf916467265e445dfa6fc2b4f9a6788b1f24324
 authors=joelhalen,OSRSKoeppy
--- a/plugins/droptracker
+++ b/plugins/droptracker
@@ -1,3 +1,3 @@
 repository=https://github.com/joelhalen/droptracker-plugin.git
-commit=2fb52ec58f081d3e29e9433b7b678f47b036b050
+commit=7ad6720d1fb55cdd9e039afa328cfd40157f19aa
 authors=joelhalen,OSRSKoeppy


riktenx:

riktenx:
also jfyi your code is not auto-reviewable due to use of sensitive lwjgl code. you are welcome to wait for manual review but there is nobody really doing that at the moment. if you would prefer to continue updating your plugin on a regular cadence you might want to remove lwjgl-based features for now

joelhalen:
> also jfyi your code is not auto-reviewable due to use of sensitive lwjgl code. you are welcome to wait for manual review but there is nobody really doing that at the moment. if you would prefer to continue updating your plugin on a regular cadence you might want to remove lwjgl-based features for now

Thanks for the heads up. I'll see if I can tackle this this weekend.

joelhalen:
> also jfyi your code is not auto-reviewable due to use of sensitive lwjgl code. you are welcome to wait for manual review but there is nobody really doing that at the moment. if you would prefer to continue updating your plugin on a regular cadence you might want to remove lwjgl-based features for now

After a bit of digging, there's no way around us using lwjgl code in order to provide our video cap functionality. Not that it will make any difference for auto-review, but none of the relevant lwjgl code has changed in a meaningful way since our last update was approved...

I'll just cross my fingers it doesn't take months, and be patient. Thanks!

joelhalen:
> also jfyi your code is not auto-reviewable due to use of sensitive lwjgl code. you are welcome to wait for manual review but there is nobody really doing that at the moment. if you would prefer to continue updating your plugin on a regular cadence you might want to remove lwjgl-based features for now

Sorry to bug you, just to confirm before I spend a bunch of time on it, lwgjl is the roadblock here and removing it (to another companion plugin for example, and letting that wait for review) would get us reviews more quickly in general moving forward? Considering options so that we aren't stuck in limbo every time after this too 🤔

riktenx:
yes, removing that will allow your plugin to be autoreviewed. though we're not really accepting new lwjgl/gpu code so not sure what your intent is there

joelhalen:
> yes, removing that will allow your plugin to be autoreviewed. though we're not really accepting new lwjgl/gpu code so not sure what your intent is there

Eh, the intent is just continuing to have our existing functionality which some of our groups specifically signed up for; I guess I was hoping a simpler companion plugin might be able to get checked through considering it would be a shell of the already approved code from this plugin's video functionality; letting us keep it eventually while getting our main plugin back on track for auto review.

Not the end of the world, thanks for the quick reply

joelhalen:
I autopiloted and discarded my fork's changes; and this closed on its own--apologies :(
I had most of this planned for the next update but since we were kinda forced into removing video capture, I bundled it all into one (updating the original PR with this):
- Removed all video / lwgjl functionality
- Enhanced in/outbound connections for users who enable them
- Added support for events
- Fixed Maggot King tracking
- Removed the every-single-login welcome message the plugin sent its users.

riktenx:
you cannot dynamically/programmatically retrieve URLs from other API responses and call into them, that is an [ssrf](https://en.wikipedia.org/wiki/Server-side_request_forgery) risk and it prevents us from being able to exhaustively review the domains that your plugin connects to without user input. this includes just fetching things like images as well as doing any other API requests

all URLs that the plugin talks to **MUST** either be hardcoded in the plugin, or directly input by the user, without exception

joelhalen:
> you cannot dynamically/programmatically retrieve URLs from other API responses and call into them, that is an [ssrf](https://en.wikipedia.org/wiki/Server-side_request_forgery) risk and it prevents us from being able to exhaustively review the domains that your plugin connects to without user input. this includes just fetching things like images as well as doing any other API requests
> 
> all URLs that the plugin talks to **MUST** either be hardcoded in the plugin, or directly input by the user, without exception

Got it. I assumed that this isn't stating that our existing functionality behind using `droptracker-io.github.io`-provided discord webhooks for one-way communication is no longer acceptable, but I strengthened that so it is only obtaining webhook paths rather than entire URLs, alongside any other place where our plugin could attempt to make a request externally.

PR #13406, opened 2026-07-07
disable path-timer

ldavid432:
~~Disabling as it can be abused at sotetseg to show party members the correct path~~

~~@cnnnr If you would like the plugin to be re-enabled please restrict / remove the party integration at sotetseg.~~

~~Watchdog can be used as a reference to find the regions to block: https://github.com/adamk33n3r/runelite-watchdog/blob/master/src/main/java/com/adamk33n3r/runelite/watchdog/Region.java~~


ldavid432:
Closing as this is actually probably fine since the maze and room regions are separate.

Sorry about that @cnnnr !

PR #13409, opened 2026-07-07
update osparty

iodrareg:
- fixes flickering of jpanel when new parties are created and rendered
- sync discord roles between api and plugin
- updated party history to no longer show parties you were rejected for
- updated README with most actual implementation
- update the fetching of the playername even when the player is not signed in

### DIFF
--- a/plugins/osparty
+++ b/plugins/osparty
@@ -1,4 +1,4 @@
 repository=https://github.com/osparty/osparty.git
-commit=dfa0e1d809ea6b017423a19f36f3ba4091684934
+commit=812d4cb24b0efb5eb419eafff766b73e29628035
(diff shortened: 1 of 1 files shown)


plugins/osparty
```diff
@@ -1,3 +1,3 @@
 repository=https://github.com/osparty/osparty.git
-commit=9f968792852711a8012bc91ad1d4df3917c59107
+commit=dfa0e1d809ea6b017423a19f36f3ba4091684934
```
riktenx:
```suggestion
commit=dfa0e1d809ea6b017423a19f36f3ba4091684934
authors=iodrareg
```

riktenx:
apply the above

PR #13428, opened 2026-07-07
Add vividflash-qol-aio

vividflash:
Adds **VividFlash QoL AIO** — a single plugin bundling a set of small quality-of-life and misclick-guard tweaks (No Use Player, Easy Unnote, withdraw limits, Require Tool, Pickpocket/Foundry/Motherlode safety guards, Take-before-Attack menu swap, potion misclick guards, clue held-count reminder, category sound muting, and more), so users can enable just the ones they want from one place instead of running many tiny plugins. Every feature is off-by-default or config-gated.

Repo: https://github.com/vividflash/vividflash-qol-aio

### DIFF
--- a/plugins/vividflash-qol-aio
+++ b/plugins/vividflash-qol-aio
@@ -1,0 +1,2 @@
+repository=https://github.com/vividflash/vividflash-qol-aio.git
+commit=04446abd5b38314afc593f6179413eda7a6de3e3


coopermor:
If you are using code from other plugins you need to retain their license information and not attribute all the files to yourself.

vividflash:
Thanks, will fix tomorrow and repush. Should have been in there in the list, mb. Probably same issue with other two.

riktenx:
thanks for the submission but we don't accept plugin repacks of this nature. if a user wants all of these plugins they can just install the original versions of them directly

vividflash:
Fair enough. Turning it private and creating new stripped down public version containing only new features, that existing plugins missed as a complementary plugin to the existing options with additional new features. 

Question: How is [Konar Milestone Reminder](https://runelite.net/plugin-hub/show/konar-milestone-reminder)  different from https://runelite.net/plugin-hub/show/slayer-boosting  ? 

Slayer boosting seems to just have more features compared to the Konar only version.

raiyni:
They are two different code bases

PR #13438, opened 2026-07-08
Add GPS plugin

PauloAguiar:
GPS is a turn-by-turn navigation. Draws the fastest route to a destination with live directions, progress and ETA, plus alternative routes using different teleports/transports, a searchable teleport-method catalog with availability modes, and destination search (places, dungeons, minigames, nearest-amenity). Built on Shortest Path (BSD-2, credited); adds the navigation layer and routes explorer on top.

GPS is different enough that after discussing it with the shortest path team, I decided to branch it off into its own plugin.

image


riktenx:

PauloAguiar:
@riktenx updated, removed the usage for both and set defaults instead.

PR #13441, opened 2026-07-08
Add Modern Chat Enhanced

GimmitheBong:
Modern Chat Enhanced is a maintained fork of the original Modern Chat plugin.

The original plugin appears to be inactive, so this fork continues development while preserving the original functionality.

This fork includes numerous improvements and bug fixes, including:

- Player icons
- Clan rank icons
- Emoji support
- Preserve Focus
- Bottom-anchored chat
- Start on All option
- Various UI improvements
- Numerous bug fixes and compatibility improvements

The goal is to continue maintaining and improving the plugin while remaining compatible with current RuneLite releases.

### DIFF
--- a/plugins/modern-chat-enhanced
+++ b/plugins/modern-chat-enhanced
@@ -1,0 +1,2 @@
+repository=https://github.com/GimmitheBong/Modern-Chat-Enhanced.git
+commit=93f1b9153db48481436704f639f148aa309f9d16


riktenx:
thanks for the submission but we don't allow new plugins (forks of existing ones or otherwise) that run client scripts which send server actions (which this plugin requires to operate)

PR #13450, opened 2026-07-08
regear

anonyser:
new plugin. a banking helper: you set the gear or inventory you want and it shows those items in fixed bank slots so you can withdraw them in the order you set. display only, no automation, you click everything yourself.


riktenx:
don't reimplement the bank tag layout system yourself, use https://static.runelite.net/runelite-client/apidocs/net/runelite/client/plugins/banktags/tabs/LayoutManager.html

anonyser:
updated. swapped the bank filtering/layout onto the layoutmanager api instead of moving the widgets myself, so it registers a hidden tag and builds a Layout for it. commit's bumped.

PR #13468, opened 2026-07-08
Add GE Companion plugin

xOwip:
Live GE prices, interactive price charts, watchlist, and bank value tracking — no account or API key required.

Source: https://github.com/xOwip/ge-companion


riktenx:
doesnt build

xOwip:
Thanks for the heads up! I'm aware of the org.json build failure and am currently migrating to Gson. Will push the fix shortly.

PR #13476, opened 2026-07-08
Add snakeskin bandana filter

silly-build:
Adds Snakeskin Bandana Filter, a plugin that hides players who are not wearing a snakeskin bandana.

The plugin does not automate gameplay, send input, or communicate with external servers.


riktenx:
why are there 2 files, there should only be one

riktenx:
we don't allow the unhiding specifically of attackers as this causes abuse potential. you will have to limit this to dynamic hiding purely based on snakeskin bandana or not if you want to submit this

PR #13487, opened 2026-07-09
Create alch-hero

B1ngu55l4y3r:
A plugin to make Alchemy more fun by incorporating Guitar Hero-like mechanics


riktenx:
your bs2d license needs to match the template here https://github.com/runelite/runelite/blob/master/LICENSE, just copy this and then add your own name/year

B1ngu55l4y3r:
> your bs2d license needs to match the template here https://github.com/runelite/runelite/blob/master/LICENSE, just copy this and then add your own name/year

Updated to match requested template - thank you.

PR #13498, opened 2026-07-09
Update splash-helper: networking and set up guide

PeppieLangWaus:
**New features:**

* Added **Server Sync** — an opt-in WebSocket connection to a Splash Helper server (`splasher.help` by default) that streams live session data 
* Added **session resume** feature that resumes a sessions if player continues splashing withing a configurable time frame
* Added configurable **Warning** (default 5 min) and **Critical** (default 2 min) timer thresholds.
* Added a **Sessions Kept On Disk** cap (default 200)
* Added a death counter to statistics
* Added an interactive sticky knight set up guide

**Minor changes and fixes:**

* Fixed session resume never actually engaging
* Fixed the notification sound never playing and notifications staying permanently muted after being muted once.
* Knight movements are now reported per hour instead of per minute.
* Safety mode now locks onto the first knight a session engages instead of potentially retargeting mid-session.
* Sticky-knight flag now persists across a session resume, and the timer chat-message dedup is keyed by game tick instead of message string.
* Potential-XP stat now falls back to the currently selected spell when no session is active.
* Removed some dead code
* Deduplicated the three near-identical tile right-click submenu builders into a single helper, and removed a couple of other redundant helpers.
* Tidied `Constants`: added `CASTS_PER_HOUR` for the "hours remaining" projection, dropped unused `VISUAL_NOTIFICATION_DURATION_MS` and `KNIGHT_OF_ARDOUGNE`.
* Updated the README with the new timer threshold colors, per-hour movements, Sessions Kept On Disk, session-resume countdown, and a Server Sync section describing what data is sent.


pajlada:
hello, you need to add the `build` key to your `runelite-plugin.properties` file - most likely `build=standard` if you're not using any special dependencies.
you'll also want to add a `version` key to the same file to ensure the correct version shows up in the plugin hub sidebar

riktenx:
it should definitely be `build=standard`, you have nothing in here that requires `build=gradle`. if you get build errors with that you just need to fix them.

PeppieLangWaus:
Sorry about that, had it set to "normal", not "standard". Should be good now

PR #13509, opened 2026-07-09
Update Hitsplat Customizer

nanopink:
Updates Hitsplat Customizer to keep unassigned hitmarks visible with My hits only.


nanopink:
Closing for now so the fix can be tested in-game before resubmitting the Plugin Hub update.

PR #13512, opened 2026-07-09
update route-planner v1.3

BlackDeath-Osrs:
Waypoint path customization, Undo/Redo functionality. Moved the toggle for dev/player mode to the side panel from the config.

### DIFF
--- a/plugins/route-planner
+++ b/plugins/route-planner
@@ -1,2 +1,2 @@
 repository=https://github.com/BlackDeath-Osrs/route-planner.git
-commit=ed5377f8fcf1de66d0829fe6ba1886b8b154dc9d
+commit=1cccc6767f6c522ed06f9e73a3cec9f4fdb1d6ef


riktenx:
your license is malformed and needs to match this template https://github.com/licensee/licensee/blob/main/spec/fixtures/bsd-2-author/LICENSE

don't include the extra junk from other plugins at the end, you follow attribution by following instructions of the license, not by copypasting theirs

BlackDeath-Osrs:
> your license is malformed and needs to match this template https://github.com/licensee/licensee/blob/main/spec/fixtures/bsd-2-author/LICENSE
> 
> don't include the extra junk from other plugins at the end, you follow attribution by following instructions of the license, not by copypasting theirs

Lisense now matches the linked bsd-2 author template. i diffed it against the fixture, only the copyright line differs. The shortest path attribution moved out to a separate notice. Notice is where i put that rather than appending it to the license. If you'd rather it be somewhere else. or if the readme credit alone is sufficient, happy to move it and make changes as needed.

riktenx:
sorry, that was the wrong template but it also looks like you didn't copy it correctly?

_this_ is the correct template https://github.com/runelite/runelite/blob/master/LICENSE

regardless it seems you didn't copy the exact text of the old one when you did this. make sure they exactly match (using the new one now) aside from your name and the year ofc

BlackDeath-Osrs:
> sorry, that was the wrong template but it also looks like you didn't copy it correctly?
> 
> _this_ is the correct template https://github.com/runelite/runelite/blob/master/LICENSE
> 
> regardless it seems you didn't copy the exact text of the old one when you did this. make sure they exactly match (using the new one now) aside from your name and the year ofc

Fixed, not sure what happened before guess that's what happens when i do stuff when i'm tired.  glad to change anything if required let me know.

PR #13513, opened 2026-07-09
DVD Bounce

vividflash:
DVD Bounce does exactly one thing: it bounces one picture around the client like the classic DVD-player screensaver. Every config option (image path, size, speed, colour shift, corner flash) parameterizes that single behavior.

Verified in-client 2026-07-09 (bounce, sliders, colour shift, corner flash, custom path + fallback, stretched mode, canvas resize, throttled-fps motion)


riktenx:
file i/o needs to happen within a plugin-specific subfolder under .runelite

PR #13525, opened 2026-07-09
Fix a noted items bug for the high alc plugin

bloogeyz:
Stops noted items being marked as untradeable


coopermor:
If you want isSellable could be
`return itemManager.canonicalize(itemId).isGeTradeable()`

bloogeyz:
@coopermor  Thank you :)

PR #13538, opened 2026-07-09
Roster: v1.1.3

Mylosis:
Updates Roster to v1.1.3.

What was wrong and what changed:
- Cloud sync: account data was persisted through ConfigManager without a @ConfigSerializer, so it was stored as toString() output and read back as a raw string. Every load fell back to the plugin's local file mirror, meaning data never actually synced between machines. Data is now stored as an explicit JSON string; existing installs recover automatically from the local mirror on first launch.
- Drag and drop: the Uncategorized section only accepted drops on its header, and a dropped card could land at an unpredictable position. The whole section is now a drop target with exact insert positions, and the insert indicator is computed from real card bounds. ESC now cancels an in-flight drag.
- Storage thread-safety: all storage access is confined to the EDT and the background file mirror writes a pre-serialized snapshot, removing a race between drag saves, login events, and the writer thread. The plugin no longer creates its own thread (uses the client's injected ScheduledExecutorService).

Testing done: fresh install from a clean .runelite; upgrade from v1.1.2 including force-quit immediately after launch; both list and grid views; drag reorder and cross-category moves; cloud-sync round-trip verified in client logs; 34 new unit tests; clean ./gradlew build.

Full changelog: https://github.com/Mylosis/roster/blob/main/README.md#changelog


riktenx:

Mylosis:
Removed, replaced with the component-local `Component.transferFocus()`. Manifest bumped to `65e5a92`.

PR #13561, opened 2026-07-10
Add quick-wiki plugin repository information

jake219:
A RuneLite plugin that lets you right-click any item, NPC, or object in-game and instantly view its wiki info — description, Grand Exchange price, high/low alch value, release date, members status, and more — all in a compact sidebar panel.

Features
Right-click "Wiki" option on items, NPCs, and objects
Item icon, name, and GE price with high/low alch
Release date, members status, and quest item info
Full item properties (tradeable, equipable, stackable, etc.)
Toggle between short and full wiki descriptions


pajlada:
move your code out of com.example and into your own package
remove your empty config
make your requests all have a plugin-specific user agent so the wiki knows the source of the requests

CC @coopermor is this plugin ok with the wiki api?

PR #13566, opened 2026-07-10
Added Inspect plugin

DevOldSchool:
Adds Inspect, an optional inspection and information plugin for OSRS.

Inspect provides:
- Item and NPC inspection using OSRS Wiki data
- Locally derived visible-player equipment inspection
- Recent inspection history
- Equipment recommendations and bank highlights
- Item prices, requirements, Slayer details and drop summaries

OSRS Wiki communication is disabled by default and must be explicitly enabled in the plugin configuration. Player equipment inspection uses locally visible RuneLite client data and does not upload, store or expose player information over HTTP.

--

Edit: Renamed from Enhanced to Inspect.

### DIFF
--- a/plugins/enhanced
+++ b/plugins/enhanced
@@ -1,0 +1,2 @@
+repository=https://github.com/DevOldSchool/enhanced.git
+commit=812c595461ec1e66740b5b619570bcec18490713
--- a/plugins/enhanced
+++ b/plugins/enhanced
@@ -1,2 +1,0 @@
-repository=https://github.com/DevOldSchool/enhanced.git
-commit=812c595461ec1e66740b5b619570bcec18490713
--- a/plugins/inspect
+++ b/plugins/inspect
@@ -1,0 +1,2 @@
+repository=https://github.com/DevOldSchool/inspect.git
+commit=812c595461ec1e66740b5b619570bcec18490713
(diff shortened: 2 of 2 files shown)


pajlada:
not official feedback, just my opinion:
Your plugin name is a generic word that doesn't really describe what your plugin does

official feedback:
I don't think the way you're making requests to the wiki is in line with their usage policy

DevOldSchool:
> not official feedback, just my opinion: Your plugin name is a generic word that doesn't really describe what your plugin does
> 
> official feedback: I don't think the way you're making requests to the wiki is in line with their usage policy

Thank you for the feedback.
I'll come up with a better name for the plugin once I can confirm the wiki usage is correct.

Was there a particular thing you think is not inline with the usage policy?

pajlada:
You're essentially scraping the wiki which I think they would not appreciate.

They have APIs you can use instead - see https://oldschool.runescape.wiki/api.php - it's most likely the parse or query module you're looking for.

pajlada:
Looking at your plugin a bit closer, I see you might be using their real api already - in which case ignore my feedback :) I just saw references to the wiki url and `/w/` which made me assume you were scraping

DevOldSchool:
Thanks for calling it out.

The bank equipment recommendations has been updated to use the API directly and I believe I'm following all the usage guidelines correctly now.

Additionally, the plugin has been renamed to Inspect to better match what it does.

PR #13570, opened 2026-07-10
Pvm-Score update; Add Maggot King

fatmownes:
Include Maggot King in the score UI and kill overlay.


ldavid432:
Please add `build=standard` to your runelite-plugin.properties

PR #13575, opened 2026-07-10
Add ranger-boots-dryness plugin

FeDoofy:
Plug-in to track dryness or ranger boots and med clues from fally guards! Let me know if any adjustments need to be made, this is my first time making a plug-in and I wanted to use it in my ironman as I'm grinding!

### DIFF
--- a/plugins/ranger-boots-dryness
+++ b/plugins/ranger-boots-dryness
@@ -1,0 +1,2 @@
+repository=https://github.com/FeDoofy/ranger-boot-dryness.git
+commit=064f4206fa40c0d1f64210157b646738d2314b5c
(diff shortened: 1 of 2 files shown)


riktenx:
the manifest goes in plugins/

FeDoofy:
What do you mean by the manifest goes in plugins?

riktenx:
https://github.com/runelite/plugin-hub#submitting-a-plugin

FeDoofy:
Hello Riktenx, 

I think I understood the issue - my file was in the wrong place and I didn't properly upload it with the correct commit number. Should be good now, lmk your thoughts.

FeDoofy:
So the reviews were successful? What happens now? When does it come to the game?

PR #13577, opened 2026-07-10
Update consumable-cooldowns to 1.3.3

CopyPastaOSRS:
- Migrate plugin to use gameval ItemID API instead of deprecated ItemID
- Added missing sailing consumables
- Added consumables from cow quest & blood moon rises


riktenx:
doesnt build, set `build=standard` in runelite-plugin.properties

PR #13581, opened 2026-07-10
pvp profit tracker 1.2.0

anonyser:
more side panel visual changes. opponent info is centered now, combat level shows from the same hiscore lookup, and you can keep notes and a win/loss count per opponent. all display only.


anonyser:
had to update in place, text was being cutoff so corrected that real quick. no changes other than what was mentioned.

PR #13582, opened 2026-07-10
fbclan-plugin: update to c6b017d

orgonag:
Updates Final Boss clan plugin from e57c800 to c6b017d, and extends the warning line to disclose boss personal-best times.

Changes since the previous pin:
- Clan PB leaderboard: a new Leaderboards panel tab shows the clan's top-3 personal best times per boss and a feed of recently broken clan records. Uploading is opt-in (config toggle, default off): new PBs are detected from the player's own kill-count/duration chat messages (detection logic mirrors the core Chat Commands plugin), and the player's existing PBs already stored locally by the core Chat Commands plugin are submitted once per session. Captures are skipped entirely on Leagues/Deadman/tournament/beta/speedrun worlds.
- Writes go through a server-side database function that only ever improves the member's own stored time; the anon key has no direct table access (reads come from two read-only views). Only the local player's own RSN, boss name, and time are sent, and only for verified clan members who opted in.
- Viewing the leaderboard requires no opt-in and sends no player data.
- README and in-client plugin description/tags updated to describe the current feature set.

Same already-disclosed third-party Supabase backend, no new domains. Warning line updated accordingly (adds "boss personal-best times"); the in-source privacy javadoc was updated in the same release.


plugins/fbclan-plugin
```diff
@@ -1,3 +1,3 @@
 repository=https://github.com/orgonag/fbclan-plugin.git
-commit=e57c800d87b74332f7ef00ede576bcef7e2eea54
-warning=This plugin submits your IP address, RSN, drop data, and Party to a 3rd-party server not controlled or verified by Runelite developers.
+commit=c6b017d4ad024c99e9acc2b06889c78c28febd0b
+warning=This plugin submits your IP address, RSN, drop data, boss personal-best times, and Party to a 3rd-party server not controlled or verified by Runelite developers.
```
riktenx: You do not need to update the warning going forward.

```suggestion
warning=This plugin submits your IP address, RSN, drop data, and Party to a 3rd-party server not controlled or verified by Runelite developers.
```

orgonag:
Reverted the warning back to the existing text — thanks!

PR #13589, opened 2026-07-10
Potion storage bars

Hannamber:


plugins/potion-storage-bars
```diff
@@ -1,2 +1,2 @@
-repository=https://github.com/mrchooch/potion-bars.git
-commit=a634eb39247e009ef0502130907f48309dd2ffd1
+repository=https://github.com/Hannamber/potion-bars.git
+commit=e5ba97d4b627a381e49d89eecd8152e568894b86
```
riktenx: I can see you renamed your GH account so this is fine. I am only accepting the repo URL change here. You can submit a followup PR with your commit change.

```suggestion
commit=a634eb39247e009ef0502130907f48309dd2ffd1
```

riktenx:
ok that is my mistake because i didn't realize it didn't build before that, sorry

you can just resubmit with your updated commit hash and i'll handle it

PR #13594, opened 2026-07-10
Update chat-sounds plugin

robmonte:
Updates the chat-sounds plugin.

Primarily fixes clan chat notifications not working as expected, adds lots of notification level customization, and then there was a good amount of code refactoring.

I am not the original plugin author however I am collaborator-level on the repo if that makes a difference for approval, as @ThePharros has been unresponsive recently.


riktenx:
@ThePharros do you approve this update and/or for @robmonte to be added to the authors= list?

PR #13602, opened 2026-07-10
Create bosshpbaradjustments

ClegDrop:
Allows you to adjust the vertical position of the boss HP bar when set to top-middle.


riktenx:

PR #13628, opened 2026-07-11
Update hitsplat-customizer

nanopink:
Renames the hitsplat-customizer repository to CustomizeALot. The plugin commit update will be submitted separately after this PR merges.


riktenx:
if you are going to rename your repo then please only submit that. you can submit a commit change in a followup

PR #13635, opened 2026-07-11
Update hitsplat-customizer

nanopink:
Updates hitsplat-customizer to the latest Customize a Lot commit after the repository rename merged in #13628.


riktenx:

nanopink:
Removed KeyboardFocusManager and now synchronize only this plugin's preset controls directly on the Swing event thread, without posting synthetic ProfileChanged events.

PR #13642, opened 2026-07-11
update osparty to v1.0.31

iodrareg:
- fixes cox layouts
- fixes reconnect of plugin after disabling it
- adds host transfer

### DIFF
--- a/plugins/osparty
+++ b/plugins/osparty
@@ -1,4 +1,4 @@
 repository=https://github.com/osparty/osparty.git
-commit=b772a83ae70f643130574a2a9546dfb18fcc645f
+commit=5ad85c0ba3df739e7c855280a9ec313400a37bd8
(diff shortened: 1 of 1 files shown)


riktenx:

iodrareg:
> * use of client.hopToWorld is no longer allowed

Where can we find updates surrounding these updates about what we can and can't use? I'm a bit confused as this was not a change within this PR and why the base client _can_ use this function, and this function is part of the client. Or will this function be deprecated in the next client release?

riktenx:
* notice i said "no longer allowed" -> it must be removed regardless of whether it's new or not
* this enforcement was not communicated publicly and not all retroactive things like this will be
* the core client can use it because core developers know how to use it. we don't trust random hub plugins to use it anymore

PR #13650, opened 2026-07-11
Update jumpscare to 1.1

vividflash:
Hotfix: stretched-mode rendering and an epilepsy-warning flow for flash mode. Verified in-client.


vividflash:
Had a viewer ask me about adding gif to the options, if this is any problem I will just remove that again and only do static.

PR #13658, opened 2026-07-11
Add OSRS FlipHub plugin

zFallan121:
Adds **OSRS FlipHub**, a Grand Exchange flip tracker.

### What it does
- Records GE offer history locally (buys/sells/completed/aborted)
- Per-flip and rolling profit/margins, GE buy-limit timers, live OSRS Wiki prices
- Item bookmarks and price context in a side panel

### Read-only / no automation
The plugin only observes `GrandExchangeOfferChanged` and related client events. It performs no automation and sends no input to the game. No RuneScape/Jagex credentials are requested, read, or transmitted.

### Network / data handling
- **Default (not linked): local-only.** All trade data stays on the user's machine. The only network calls are read-only price lookups to the public OSRS Wiki price API (`prices.runescape.wiki`).
- **Optional account link:** if the user pastes a FlipHub license key in the plugin settings, their GE offer events (item, quantity, price, offer state, timestamps) are uploaded over HTTPS to FlipHub's servers (`osrsfliphub.com`) to power an online dashboard. This is **opt-in** - nothing is uploaded until the user links - and is disclosed both in the plugin description and in a dedicated config section. Clearing the key or clicking **Unlink** stops uploads immediately.

Source: https://github.com/zFallan121/OSRS-FlipHub-Plugin-Public
Licensed BSD 2-Clause.

### DIFF
--- a/plugins/osrs-fliphub
+++ b/plugins/osrs-fliphub
@@ -1,2 +1,2 @@
 repository=https://github.com/zFallan121/OSRS-FlipHub-Plugin-Public.git
-commit=2a0380495ce2ac441a9db592dd1ea6d0290a7efb
+commit=4d938bbed6a0a0cde4c7c0bf2fe0e9d1fdded1ed
--- a/plugins/osrs-fliphub
+++ b/plugins/osrs-fliphub
@@ -1,2 +1,2 @@
 repository=https://github.com/zFallan121/OSRS-FlipHub-Plugin-Public.git
-commit=68a1bfc3d643d5979dd640105a448d5eec873893
+commit=2a0380495ce2ac441a9db592dd1ea6d0290a7efb
--- a/plugins/osrs-fliphub
+++ b/plugins/osrs-fliphub
@@ -1,0 +1,2 @@
+repository=https://github.com/zFallan121/OSRS-FlipHub-Plugin-Public.git
+commit=68a1bfc3d643d5979dd640105a448d5eec873893
--- a/plugins/osrs-fliphub
+++ b/plugins/osrs-fliphub
@@ -1,2 +1,3 @@
 repository=https://github.com/zFallan121/OSRS-FlipHub-Plugin-Public.git
 commit=f1222128e09149574b10a8453a1496553894c6ee
+warning=This plugin submits your IP address, and may submit various account data, to a 3rd-party server not controlled or verified by Runelite developers.
--- a/plugins/osrs-fliphub
+++ b/plugins/osrs-fliphub
@@ -1,2 +1,2 @@
 repository=https://github.com/zFallan121/OSRS-FlipHub-Plugin-Public.git
-commit=b0ded985fc86d6b4ef6af6cf1b4829429da8b38c
+commit=3e73dec3a423b59cccbf8f64b7055d345ead82ea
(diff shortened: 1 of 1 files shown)


riktenx:

riktenx:
your opt-in networking **MUST** be gated behind a clearly-labeled `@ConfigItem`, which is a boolean that defaults to false. `@ConfigItem::warning` on said config **MUST** be set to `This feature submits your IP address to a 3rd-party server not controlled or verified by Runelite developers.`

Call sites where OkHttp calls are made to this URL **MUST** check this boolean at the top of the method and either return early or throw a config error so we can verify you're respecting the opt-in.

ldavid432:
Why are you manually reading the profiles2 and settings.properties files instead of just using ConfigManager?

zFallan121:
> Why are you manually reading the profiles2 and settings.properties files instead of just using ConfigManager?

Older versions of the plugin stored local trade history in RuneLite config under our own group (osrsfliphub.localTrades.*). Since ConfigManager only exposes the active profile, the direct file read was a shortcut to migrate that legacy data regardless of which profile it was written under, it only extracts keys under our own config group prefix and ignores everything else. 
That said, I agree reading the raw files isn't needed.

I could replace it with ConfigManager.getConfigurationKeys() + getConfiguration() against the active profile and remove the profiles2/settings.properties reads.

coopermor:
"Older versions of the plugin"? This is a brand new plugin?

ldavid432:
> Older versions of the plugin stored local trade history in RuneLite config under our own group (osrsfliphub.localTrades.*). Since ConfigManager only exposes the active profile, the direct file read was a shortcut to migrate that legacy data regardless of which profile it was written under, it only extracts keys under our own config group prefix and ignores everything else. That said, I agree reading the raw files isn't needed.
> 
> I could replace it with ConfigManager.getConfigurationKeys() + getConfiguration() against the active profile and remove the profiles2/settings.properties reads.

Yes please remove the raw reading of those files.

Also your plugin isn't even published yet, why would you need handling for older versions of your config?

zFallan121:
there are no published older versions. The legacy handling existed for my own pre-submission development builds, which stored trade history in RuneLite config before the plugin had its own storage. That's not something any hub user will ever have, so rather than porting it to ConfigManager I've removed the entire legacy config path, including the profiles2/settings.properties reads. Pushing the change shortly.

Alexsuperfly:

zFallan121:
> Desktop::browse is a restricted api, requiring manual review, use LinkBrowser::browse instead

Thank you mate, appreciated

Alexsuperfly:
you cannot send [user.name](https://github.com/zFallan121/OSRS-FlipHub-Plugin-Public/blob/3e73dec3a423b59cccbf8f64b7055d345ead82ea/src/main/java/com/osrsfliphub/link/LinkAttemptService.java#L41),

plugins/osrs-fliphub
```diff
@@ -0,0 +1,2 @@
+repository=https://github.com/zFallan121/OSRS-FlipHub-Plugin-Public.git
+commit=f1222128e09149574b10a8453a1496553894c6ee
```
riktenx, line 2:
```suggestion
commit=f1222128e09149574b10a8453a1496553894c6ee
warning=This plugin submits your IP address, and may submit various account data, to a 3rd-party server not controlled or verified by Runelite developers.
```

PR #13659, opened 2026-07-12
Camera smoothing: Roe's middle mouse fix

ArtsicleOfficial:
Previously middle mosue would do excess movement compared to what the user inputted.
Also includes Roe's config tweaks.


riktenx:
merge conflicts

PR #13683, opened 2026-07-12
Update tictac7x-charges to v0.6.13

TicTac7x:


riktenx:
your plugin is too large, the reviewbot has a 200k llm token limit and you are at 220k, to continue to update this plugin you will need to reduce the size of your codebase. only source code under `src/main/java` is counted (no resources/tests) and comments are stripped

seems like you have a lot of generic "data class" stuff that could just be made into/loaded as json?

TicTac7x:
Which data classes are you referring to?

riktenx:
i didnt read into it that hard i may be wrong

TicTac7x:
It's logical that the plugin keeps growing in size, if new items keep being added in all the time, so I don't know how am i supposed to overcome this tbh. There is lots of boilerplate code and custom events wrappers around runelite apis to compensate for the lack of its features or logic, so dont know what to do here...

riktenx:
looks like https://github.com/TicTac7x/runelite-plugins/blob/plugin-charges/src/main/java/tictac7x/charges/store/ids/ItemId.java is mostly what i was referring to. like this file is not necessary at all. i dont think there's that much actual logic here. that couldnt necessarily be json but it could certainly just not exist

TicTac7x:
Ok, these are just nice to have converters, because the names in the runelite apis are from time to time very weird, these could be removed, i agree, but I would still hit the size limit sooner than later, if i keep adding in new items for this plugin.

riktenx:
most of these trigger definitions, like the C_ whatever stuff i think could also maybe be loaded from json. your event api stuff seems sensible and fine as is

you've also got just a lot of nesting so a lot of imports, wildcards will help there or flatten your package setup

you can definitely get this under 200k and still able to scale up with game updates

riktenx:
maybe consider it in terms of like how shortest path works; that plugin is within the limit despite the massive and ever growing map data because its stored just as that, data, rather than being hardcoded as you basically have with your various item implementations

TicTac7x:
Removed the ItemId as requested, hopefully its enough now.

PR #13691, opened 2026-07-12
add friend-groups

H-Mill:
Add Friend Groups
- Add friends to groups to easily organize/find them
- Display friend groups in-game and in runelite side panel
- Hop to friends from runelites side panel


riktenx:

PR #13703, opened 2026-07-12
Update deathdotter plugin

DapperMickie:
Adds a "hide 2d" config option (on by default to keep default behavior of the plugin the same).
No more (potential) issues with worldviews.


dmgear:
This is fine. I give permission to @DapperMickie to update the plugin.

PR #13706, opened 2026-07-12
update death-count

H-Mill:
add all-time death counter


riktenx:
stop spamming prs. just open one and update it if needed

PR #13730, opened 2026-07-13
Update lobster-pot

datlalo:
adds in-panel bounty submission and a world-map fallback to the surface when a clanmate can't be located


riktenx:

PR #13741, opened 2026-07-13
update pet-pathing-and-combat

MoukerJoe:
Smoother turning (look-ahead steering with damping), turn speed slider, and optional pet-watches-you facing.


MoukerJoe:
Bumped the commit — this update is bigger than the original description. Summary of what's new:

Smoother movement: look-ahead steering, turn-in-place pivots, rest beside the owner (collision-checked), turn speed slider
Keeps the pet visible + animating through render-range gaps instead of blinking out
Faux-run cadence for pets that have no distinct run animation
Transmog preview (WIP): a config field to render the follower as any pet by NPC id, built from the cache. Purely cosmetic and client-side, off by default (-1). It's a work in progress — most pets look right, a few effect-heavy ones don't sort perfectly.
All client-side/visual only; the real pet and thrall still exist and behave normally.

PR #13745, opened 2026-07-13
Add Merchant Flipper plugin

KevinNLPG:
Merchant Flipper is a read-only sensor plugin that watches your own Grand Exchange
offers and reports changes to a local companion desktop app over a 127.0.0.1-only
WebSocket. It never clicks, types, or automates anything in-game — it only reads
GrandExchangeOfferChanged events through the standard RuneLite API.

- No third-party server: the plugin only ever talks to 127.0.0.1 (your own machine).
- No reflection, no native code, one small dependency (Java-WebSocket).
- Full data-handling writeup: https://github.com/KevinNLPG/merchant-flipper-runelite-plugin/blob/main/PRIVACY.md

PR #13747, opened 2026-07-13
Add merchant flipper

KevinNLPG:
Merchant Flipper is a read-only sensor plugin that watches your own Grand Exchange
offers and reports changes to a local companion desktop app over a 127.0.0.1-only
WebSocket. It never clicks, types, or automates anything in-game — it only reads
GrandExchangeOfferChanged events through the standard RuneLite API.

Re-opening as a new PR after addressing feedback from the previous submission (#13745):
- Switched from Java-WebSocket to okhttp3's WebSocket support (already a transitive
  dependency of net.runelite:client, so no extra dependency is declared).
- runelite-plugin.properties uses build=standard.

- No third-party server: the plugin only ever talks to 127.0.0.1 (your own machine).
- Full data-handling writeup: https://github.com/KevinNLPG/merchant-flipper-runelite-plugin/blob/main/PRIVACY.md

### DIFF
--- a/plugins/merchant-flipper
+++ b/plugins/merchant-flipper
@@ -1,3 +1,3 @@
 repository=https://github.com/KevinNLPG/merchant-flipper-runelite-plugin.git
-commit=5e51a0d638947ef94e671d87b70eb87a5a2a7644
+commit=e8746344f847eca2f2f0cc7ee41669dd56575716
 authors=KevinNLPG
--- a/plugins/merchant-flipper
+++ b/plugins/merchant-flipper
@@ -1,0 +1,3 @@
+repository=https://github.com/KevinNLPG/merchant-flipper-runelite-plugin.git
+commit=5e51a0d638947ef94e671d87b70eb87a5a2a7644
+authors=KevinNLPG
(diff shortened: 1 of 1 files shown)


KevinNLPG:
Not 100% sure if everything is correct now. I guess not 100% because i still see some failed attempts

riktenx:
we dont allow plugins that rely on custom native applications

we also dont allow readmes to include instructions on manually building and sideloading the plugin

KevinNLPG:
Okay good to know! Then i think this is the end for now, because we wanted to sync our ge trades to a flipping application to automatic get the trade values instead of putting it in ourself. But i understand that it is not allowed thanks for the help though!

riktenx:
you can host a website or something that the plugin talks to, that's what the other 27 flipping plugins on the hub do. if you'd like to refactor your plugin to do so we can leave this pr up and you can push an update when that's done, otherwise go ahead and close it

KevinNLPG:
Yeah i think that is a little bit to far for me just wanted to have it local and not public as in a website! Thanks for the help once again! Maybe i will see you in the future with a better plugin.

PR #13751, opened 2026-07-13
Update maggot-king

randytkrx:
Fixes the boss rendering invisible after a respawn ([maggot-king#4](https://github.com/randytkrx/maggot-king/issues/4)): the object draw callback also receives actor-backed scene entries whose id is a scene index rather than an object id, so an npc index colliding with a hidden scenery id could hide the boss. Only entries with hash type 2 (real game objects) are filtered now. Thanks to JZomDev for the report and bopsec for the fix.

Also defaults the corpse preference to Open stomach.

### DIFF
--- a/plugins/maggot-king
+++ b/plugins/maggot-king
@@ -1,2 +1,2 @@
 repository=https://github.com/randytkrx/maggot-king.git
-commit=789b337e1881f8116b70e7ce27dba0e003438fb8
+commit=55e4d7f8e51be5f14c9c9fac1254eb03c1cf6c7c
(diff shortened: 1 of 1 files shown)


coopermor:
Writing eggs are not a guaranteed pet. This feels very misleading to add pet "drop rates" when you have no data to back it up.

randytkrx:
> Writing eggs are not a guaranteed pet. This feels very misleading to add pet "drop rates" when you have no data to back it up.

oh my bad, i misread the wiki pages, https://oldschool.runescape.wiki/w/Maggot_egg was reading the drop rate weightings will fix

PR #13756, opened 2026-07-14
Add Group Storage Tracker Plugin

MerryGusmas:
Group Storage Tracker is a RuneLite plugin for Group Ironman accounts that helps track valuable shared-storage items which have been withdrawn and not returned.

The plugin learns items from Group Storage, watches bank, inventory, and equipment state, and shows a bank-only item viewer so forgetful group members can quickly see what needs to go back into Group Storage.


riktenx:
* build failed

> groupstoragetracker: Missing LICENSE file

PR #13774, opened 2026-07-14
Add paint-overlays

vahnx:
Adds Paint Overlays, a RuneLite plugin for persistent painting on both the in-game scene and the world map.

Current features:
- Brush, shape, text, and eraser tools
- Scene and world map painting
- Persistent text notes, shapes, and strokes
- Undo support for recent actions
- Sidepanel-driven configuration

### DIFF
--- a/plugins/paint-overlays
+++ b/plugins/paint-overlays
@@ -1,2 +1,2 @@
 repository=https://github.com/vahnx/paint-overlays.git
-commit=ac71c08f2f55005d635311f77fc8177c9c3face2
+commit=295849eaeb5c6497ad4097b8c80cf055bb111068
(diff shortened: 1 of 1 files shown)


vahnx:
Converting this PR to draft while I investigate a persistence issue that can cause some drawings not to reload correctly after logout/restart.

I’m also looking into a separate performance concern where startup / first-use behavior can hitch on some computers.

I’ll update the plugin and resume review once persistence is reliable and the performance behavior is better understood.

riktenx:

vahnx:
Updated and the previous KeyboardFocusManager usage has been removed; the plugin now uses RuneLite's KeyManager with a KeyListener for the paint hotkeys. Also improved performance and the persistence issue and minor UI tweaks and icon change.

riktenx:

PR #13789, opened 2026-07-14
Update bank templates 1.6.4

TheSpryt:
- If no account token is set in Bank Templates but the Exchange Insights plugin on the same client has one configured, that token is now used automatically (read live from local config, never copied or transmitted anywhere new), so the account link and verified bank sync work with zero setup for users of both plugins. A token set in Bank Templates itself still takes precedence.
- The account link status is now a single button below the search box that toggles link/unlink by status. Unlinking asks for confirmation and is remembered locally and server-side, so automatic linking never silently relinks a character the user chose to unlink.
- The bank value line moved under that button and reads "Your current bank value: ...".


riktenx:
* merge conflicts

TheSpryt:
Fixed, sorry!

PR #13809, opened 2026-07-15
anvil: progress sidebar, Active now, and site-relay federation

AhmedFathy2001:
Updates the **anvil** plugin. Adds an always-on progress side panel for the Anvil clan-events platform, plus the plugin side of optional cross-clan federation.

**Sidebar**
- Per-event board summary + nearest-to-complete tiles
- **Active now** — tiles you and teammates are actively progressing (drops, kills, skill XP, boss KC), attributed per player
- Link to the event's standings on the site

**Federation (new — site-relay, opt-in, off by default)**
- The plugin's *entire* federation footprint is its own home site: it polls `GET /api/plugin/federation/state` and renders the per-clan boards the site already aggregated, and establishes federation via a one-time Discord device-login opened in the system browser (pinned to the Anvil broker host only). It never contacts a broker or another clan's site directly — all server-to-server. With federation off (default), the panel is byte-for-byte the single-home sidebar.

**Other**
- Set-aware progress for collection tiles in the in-game chat/toast (a full set reads "4/4", not "4/1").

Existing tracking/overlay behavior is unchanged. `./gradlew test` passes.

### DIFF
--- a/plugins/anvil
+++ b/plugins/anvil
@@ -1,2 +1,2 @@
 repository=https://github.com/AhmedFathy2001/anvil-plugin.git
-commit=d40f69128dfc855f9eae439a11364072bb9a5812
+commit=2f6836afc0627ee1d80ce8b4ee8cc31fa1cd8532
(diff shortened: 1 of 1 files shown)


riktenx:

AhmedFathy2001:
@riktenx Fixed that issue (Sorry i added another module that was just finished to the PR)

AhmedFathy2001:
Any updates needed from my side? @riktenx

Alexsuperfly:
thread interrupt and sleep are not allowed

AhmedFathy2001:
> thread interrupt and sleep are not allowed

Hi, sorry missed a few instances, should be fine now?

PR #13845, opened 2026-07-15
Add account snapshot export

trilliam25:
Adds Account Snapshot Export, a passive plugin that exports observed account progress and interface-backed game data to compact JSON files.

The plugin does not perform actions on behalf of the user. It only records data and relies on the player to open the relevant OSRS interfaces normally.


Alexsuperfly:

PR #13860, opened 2026-07-16
Add GE Offer Webhook plugin

John-AndreS:
Adds **GE Offer Webhook** — a plugin that POSTs a JSON webhook to a user-configured URL on every Grand Exchange offer change: placement, partial fill, completion, and cancellation, plus a full-slot snapshot on login. A listening service can use this to maintain real-time, gapless GE state without polling.

- **Observe-only:** it reads client-reported GE state and sends JSON; it generates no game input.
- **Outbound-data disclosure:** provided via the `warning=` field in this manifest and in the Webhook URL config option's description.
- Uses the client's injected `OkHttpClient` and `Gson` (no fresh instances); `build=standard`; BSD-2-Clause.

Plugin repository: https://github.com/John-AndreS/ge-offer-webhook

### DIFF
--- a/plugins/ge-offer-webhook
+++ b/plugins/ge-offer-webhook
@@ -1,4 +1,3 @@
 repository=https://github.com/John-AndreS/ge-offer-webhook.git
 commit=e2cc116d492d329c1aa678cacc5fca5f91293783
 authors=John-AndreS
-warning=This plugin submits your Grand Exchange offer data (item, price, quantity, and offer status) to a server or URL you configure. This URL is not controlled or verified by the RuneLite Developers.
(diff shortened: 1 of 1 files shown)


plugins/ge-offer-webhook
```diff
@@ -0,0 +1,4 @@
+repository=https://github.com/John-AndreS/ge-offer-webhook.git
+commit=e2cc116d492d329c1aa678cacc5fca5f91293783
+authors=John-AndreS
+warning=This plugin submits your Grand Exchange offer data (item, price, quantity, and offer status) to a server or URL you configure. This URL is not controlled or verified by the RuneLite Developers.
```
riktenx:
```suggestion
```

riktenx:
the warning is not necessary since the URL is user-input.

PR #13888, opened 2026-07-16
Update stack-value

RealSheeva:
Adds an optional Fill display option with an opacity slider. Off by default, so existing installs are unchanged.


RealSheeva:
Reopening — the build failure was a missing build=standard in runelite-plugin.properties, now fixed.

PR #13898, opened 2026-07-16
Update tictac7x-motherlode to v0.4.5

TicTac7x:


TicTac7x:
I feel weird about creating wrapper around getting varbits, but it's to start writing proper tests. Or are there some other options to achieve this? Are there any good examples, where plugins have tests, that use runelite apis?

Adam-:
Normally, people use mocks. Here is an example https://github.com/runelite/runelite/blob/9308cb4833432c9f74ebc1412a4daff7c6de6ac3/runelite-client/src/test/java/net/runelite/client/plugins/chatcommands/ChatCommandsPluginTest.java#L219-L220

TicTac7x:
Thanks, I will change it to follow that style then.

PR #13902, opened 2026-07-17
Add TCG Locked

randytkrx:
Adds TCG Locked, a challenge gamemode plugin.

You can only equip or use an item once you own its card in the OSRS TCG plugin. Equipment and teleport items stay locked until the matching card is pulled, a side panel tracks your unlocks, and lock icons show what is still locked. The plugin reads the OSRS TCG collection and stays inactive when no collection is present.

Repository: https://github.com/randytkrx/tcg-locked
Builds cleanly with gradlew build and includes a unit test for the item name matching.

### DIFF
--- a/plugins/tcg-locked
+++ b/plugins/tcg-locked
@@ -1,0 +1,2 @@
+repository=https://github.com/randytkrx/tcg-locked.git
+commit=e49b4c4d4cedaa5edb0d0b176b4b3f64857daafb


hildickethan:
is this not the same as https://github.com/runelite/plugin-hub/pull/13638

randytkrx:
Yeah, fair, this overlaps Bronzeman TCG (#13638) a lot. I've opened #13925 to remove it. Thanks for the catch.

PR #13922, opened 2026-07-17
Create disable-walk-here plugin

Shackin:
Plugin to disable "walk here" with sound feedback and drawing of disabled tiles.

Export/Import/Clear are added in a drop-down menu to avoid a side panel or in-game menu clutter.


cubeee:
Have you considered contributing to the existing plugin? https://runelite.net/plugin-hub/show/stopmisclickingtiles

PR #13925, opened 2026-07-17
Remove TCG Locked

randytkrx:
Removing tcg-locked; it substantially overlaps Bronzeman TCG (#13638), as raised in #13902.

### DIFF
--- a/plugins/tcg-locked
+++ b/plugins/tcg-locked
@@ -1,2 +1,0 @@
-repository=https://github.com/randytkrx/tcg-locked.git
-commit=e49b4c4d4cedaa5edb0d0b176b4b3f64857daafb


riktenx:
wrong

riktenx:
jfyi we do not care if you have a similar plugin to another or whatever

randytkrx:
> jfyi we do not care if you have a similar plugin to another or whatever

Yeah, I just prefer not to clutter the hub when there's basically the same thing with like a few minor differences, if anything I'll get into contact with the man that made bronzemode and see if he'll take a couple suggestions! Easier and cleaner that way, though i was working oon a version for this which was a group version, might reopen with those changes

PR #13930, opened 2026-07-17
Add Ranked Duels

martynasbutkus:
Ranked Duels lets players challenge each other to rated duels anywhere in Gielinor. Both clients independently report results to my backend at rankedduel.com (rating, world, gear snapshot, kill/death events); results only count when both reports agree. Players are auto-registered on first login using their account hash (or a SHA-256 of the login name for non-Jagex accounts) — no credentials are ever sent. If the player also runs PvP Performance Tracker, this plugin reads its public fight history via reflection to attach fight stats to the duel — read-only, wrapped in try/catch, degrades gracefully when PPT is absent. No staking, no wagering, no gameplay automation.

### DIFF
--- a/plugins/ranked-duels
+++ b/plugins/ranked-duels
@@ -1,0 +1,2 @@
+repository=https://github.com/martynasbutkus/ranked-duels.git
+commit=f3e5d8b912d34e27e9ed87ba7a2f118283e1e3f6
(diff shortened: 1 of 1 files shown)


riktenx:

martynasbutkus:
Hey! Both addressed: removed the PvP Performance Tracker reflection integration entirely, and replaced the Thread.sleep with a scheduled executor.

findarian:
The pvp leaderboard plugin already does this. Maybe it's worth working together for what you're trying to do?

plugins/ranked-duels
```diff
@@ -0,0 +1,2 @@
+repository=https://github.com/martynasbutkus/ranked-duels.git
+commit=9be47ac47de8b6b5814aed0df8714dd3de9c4222
```
riktenx, line 2:
```suggestion
commit=9be47ac47de8b6b5814aed0df8714dd3de9c4222
warning=This plugin submits your IP address to a 3rd-party server not controlled or verified by Runelite developers.
```

PR #13937, opened 2026-07-17
Update Trouble Brewing Highlighter

xboxnuker-rgb:
removal of uneeded groud item highlights


xboxnuker-rgb:
sorry if ive done anything wrong so far, new to github and this whole process but feel like im getting close to what i'm meant to do at least :)

PR #13942, opened 2026-07-17
Update TCG Locked

randytkrx:
Reworks TCG Locked into a group card collection challenge for the OSRS TCG plugin.

Since the last version it adds: monster locking, one click difficulty presets, an unlock reveal with item art and a chime, a collection lockbook, and RuneLite party play where the whole group pools its unlocks (a card owned by any member unlocks that item for everyone). Also fixes a party message threading issue and contains the lockbook grid in a scrolling box.

Repository: https://github.com/randytkrx/tcg-locked
Builds cleanly with gradlew build; includes a unit test for the item name matching.

### DIFF
--- a/plugins/tcg-locked
+++ b/plugins/tcg-locked
@@ -1,2 +1,2 @@
 repository=https://github.com/randytkrx/tcg-locked.git
-commit=e49b4c4d4cedaa5edb0d0b176b4b3f64857daafb
+commit=2f4755510007e5134c3b48d11641a7958186d53e


randytkrx:
Updated to 8fca024, which also fixes issues that left the plugin broken and unusable: OSRS TCG's recent profile update changed how it stores its collection, so this plugin could no longer read it and nothing would unlock. It now reads the collection through OSRS TCG's official PluginMessage API (query on startup/login, push-updated on collection changes), with the old config decode kept only as a fallback for OSRS TCG versions that predate the API. The OSRS TCG developer flagged the breakage and this is the integration path they support.

randytkrx:
Bumped to ba90f0a: also adds Bronzeman TCG compatibility. That plugin lists TCG Locked as a conflict (both enforce the same locks and it double-strips menus / double-draws padlocks when run together), so TCG Locked now defers all enforcement to Bronzeman TCG's restriction engine when it's enabled (on by default, toggleable), keeping only its own lockbook, unlock reveals and party features active. Running both is now complementary instead of conflicting.

randytkrx:
Bumped to 2f47555, which fixes the bugs that were leaving the plugin unusable for people:

- **Nothing unlocked at all.** OSRS TCG renamed the field its collection is serialized under, so the old read of its stored state decoded successfully but returned an empty collection. That path could not work any more and has been removed entirely — the collection now comes solely from OSRS TCG's own PluginMessage API, which is also the integration path its developer supports.
- **Items with no card were locked permanently.** Only around 5,100 items have a card, so things like a Salve amulet, a Max cape or any teleport tablet could never be unlocked by collecting anything — locking them was a dead end rather than a goal. A bundled catalog generated from OSRS TCG's own card list now decides what is part of the challenge, and items with no card are left alone. That is the rule monster gating already used.
- **Locking ran before the collection was known.** An unanswered query was indistinguishable from owning no cards, so anyone whose OSRS TCG plugin had not replied yet found every item locked and got a chat warning for each piece of gear they were wearing. Nothing is locked until the collection actually arrives, and the panel says so instead of claiming to enforce.
- **The panel's Refresh button did not re-query.** It repainted the same numbers, so a collection that had drifted could not be recovered by hand. It now re-asks, as does every login and world hop.

Monsters are also matched through their card names rather than their in-game names, which fixes 39 that have cards but were never gated. Unit tests cover the catalog's carded, uncarded and multi-card cases.

PR #13961, opened 2026-07-18
Update to Plugin

ZH-Systems:


ZH-Systems:
Feedback given in discord will re-review and update accordingly

PR #13965, opened 2026-07-18
osrscn 0.1.7

Aoldbald:
- "Use" menu wording now fits the context (picking an item up to use vs. using it on something)
- GE trade messages are composed from the item name table instead of one row per item
- lookup fixes: bare numbers matching random dialogue rows, skill guide pages sticking in English after a re-layout
- missing-text collection cleanup: stable anonymous file name, mid-relayout garbage no longer collected, local player name masked
- new **opt-in** auto-upload for the collected missing-text file — off by default, consent dialog on enable, sends only game English + an anonymous id. Happy to extend the warning line if you'd like it mentioned.

### DIFF
--- a/plugins/osrscn
+++ b/plugins/osrscn
@@ -1,3 +1,3 @@
 repository=https://github.com/Aoldbald/OSRS_CN_Runelite.git
-commit=b74187947a40b11c6814b08e33b945b8eb4429b1
+commit=721e44cb53e14b6fe83fe261d860546ad82fb369
 warning=This plugin submits your IP address to a 3rd-party server not controlled or verified by Runelite developers.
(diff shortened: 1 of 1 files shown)


Aoldbald:
Hi, checking in on this one.

If the holdup is the new opt-in upload:
- off by default, consent dialog on enable
- sends untranslated game English plus a random install id, no chat or account data
- goes to a Cloudflare Worker I run, then the data repo this plugin already downloads from

Happy to add it to the warning line or split it into its own PR if that's easier.

Aoldbald:
Updated this PR to 0.2.0 and took out the upload, as offered above.

There is no network change in it now. The diff against the last merged version is client side only: a cache for lookups that miss, which is most of the per frame cost, translation for the Grand Exchange and bank search prompts, and a fix for text that was meant to stay English being drawn with our own glyphs.

The opt-in upload will come back as its own PR later, so it can be looked at on its own.

PR #13983, opened 2026-07-18
Create last protection prayer

kp97-bit:
This plugin creates a small overlay showing which protection prayer was clicked last.

### DIFF
--- a/plugins/last protection prayer
+++ b/plugins/last protection prayer
@@ -1,0 +1,2 @@
+repository=https://github.com/kp97-bit/recent-prot.git
+commit=f56279c4833dc1c35a4443400f0cc7d116bd341d


coopermor:
We will not be hosting plugins that track your last prayer as they are to assist with boss mechanics and are not compliant with the 3rd party client guidelines.

kp97-bit:
There are countless plugins that "assist with boss mechanics." For example Nyloer plugin highlights aggressive nylocas during the waves, which breaks the rule of "no additional visual or audio indicators of a boss mechanic" in a vastly more egregious way than simply showing which prayer you clicked most recently. Per the letter of the law, the only rules this could fall afoul of is "No prayer switching indicators" - this is not a prayer switching indicator - it does not tell you which prayer to turn on. "No combat prayer recommendations" - this is not a prayer recommendation. It does not tell you which prayer to turn on. All it does is track which prayer you had clicked on last.

PR #13985, opened 2026-07-18
CARoadmap Release

itohti:
CARoadmap analyzes your current progression and recommends the best Combat Achievements to complete next. While you're fighting a boss, it automatically switches into a live combat view that helps you stay on track for encounter-specific tasks.


riktenx:
use of java.net HttpClient is not allowed, inject and use Runelite's OkHttpClient instead

Alexsuperfly:
file access needs to stay within your plugin directory, even [tmp files](https://github.com/itohti/CARoadmap/blob/4a8f35b3c9f2058477059d615461551c070e500c/src/main/java/com/caroadmap/data/CSVHandler.java#L66) in case they are unable to be cleaned up.

PR #13195, opened 2026-07-03
Add last overhead prayer plugin

justinizmlg12-boop:

PR #13990, opened 2026-07-18
Create last protection prayer - Please Review

kp97-bit:
Hi - Thank you again for reviewing my plugin. My last attempt to commit this was rejected along these lines: "We will not be hosting plugins that track your last prayer as they are to assist with boss mechanics and are not compliant with the 3rd party client guidelines."

I don't quite understand the meaning behind "to assist with boss mechanics and are not compliant." There are many plugins available on the plugin hub that exist to assist with boss mechanics. Regardless, I reviewed the rules again to make sure. 
I understand the concern regarding the rule against prayer-switching indicators, but I believe this plugin is materially different from the type of assistance that rule is intended to prohibit.

The plugin only records and displays a status after the player manually clicks it. It does not:

-inspect an NPC, boss, projectile, animation, attack style, or encounter state
-predict an upcoming attack
-recommend which prayer the player should use
-indicate when the player should switch
-warn that the current prayer is wrong
-highlight a prayer button or otherwise direct the player’s next input
-automate, remap, or modify any click

The displayed value is therefore not derived from game mechanics or combat conditions. It is simply confirmation of the player’s own most recent manual input, similar in principle to an input-history display. 

I would appreciate if you could reconsider the decision along those lines; or provide further context as to why this plugin would not be allowed.

Thank you.

### DIFF
--- a/plugins/last protection prayer
+++ b/plugins/last protection prayer
@@ -1,0 +1,2 @@
+repository=https://github.com/kp97-bit/recent-prot.git
+commit=f56279c4833dc1c35a4443400f0cc7d116bd341d


coopermor:
We feel that last prayer indicators are a violation of:
> Indicates what prayers to use in what order, for example in the Cerberus boss fight
Indicates which prayer to use in any combat situation

This sort of plugin only exists for the maggot king. As such we have decided not to allow them. Other plugins have been reject as well. See #13201 #13195 #13193 among others

kp97-bit:
I just don't really understand how this is materially different to something like the need to summon thrall indicator, or the mark of darkness indicator. The suggestion that you need to cast mark of darkness exists only to help you with a Yama mechanic (need mark of darkness to speed up attack rate at boulders), yet that is totally legal. Again I reiterate that this is an input/status-history display at its core - the fact that it can help at Maggot King is irrelevant because it doesn't prevent you from making mistakes at Maggot King. For example, if you misclick the wrong prayer to turn off a prayer, you can still mess up if you then refer to the plugin. There is no direct guidance, there is no counting, no tracking, no direction to the player.

raiyni:
Having to remember prayers for a boss and being reminded to summon a thrall aren't the same mechanic.

coopermor:
> no tracking

Except you are tracking the player's last prayer

kp97-bit:
> > no tracking
> 
> Except you are tracking the player's last prayer

Meant as in, no tracking of the boss, but fair enough.

PR #13999, opened 2026-07-18
Update dorgesh-kaan-lights to 2.0.3

tmgdejong:
Fixed a bug where the legacy overlay only showed lamps in the current area


coopermor:
By the way your logic is still wrong. onGameState change fires everytime you cross a loading line. The broken lamps never change for the player when they logout or move. So by clearing all the information when crossing a loading line or on the login screen you are losing information that is valid. https://github.com/andmcadams/dk-lights/blob/a507609a235604bd6f35dfc479f789ab94af1fc1/src/main/java/com/dklights/DKLightsStateManager.java#L223-L240

PR #14013, opened 2026-07-19
loadout-lab: move to the AKAddons org

ajkatz:
Ownership move only, mirroring goalplanner (#13641): the plugin repo now lives at AKAddons/runelite-loadout-lab (I am a public member; authors=ajkatz added per the org-repo convention). The pinned commit is unchanged - 9dd5e2d builds identically from the new location (full history was transferred).

A version bump to 0.3.1 will follow as a separate PR once this merges.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #14014, opened 2026-07-19
loadout-lab: 0.3.4

ajkatz:
Updates Loadout Lab to 0.3.4 (pin f0a95a0).

0.3.4 removes all network I/O: the "Wiki calc" chip, which POSTed the shown setup to the wiki's shortlink service, and the async monster thumbnails on the mob rows. The plugin now makes no network requests at all.

Highlights since 0.3.2: bank setup view (equipment cross + inventory laid out via core bank-tag layouts), trip supplies with per-mob overrides, thralls + Death Charge modeling with autocast-clash handling, prayer/boost assumption pickers with per-style defaults, divine potions preferred as BiS, a fully optional Display/Controls/Defaults settings surface, DWMS integration moved to its new PluginMessage storages API (2.11.5+), and an elemental-weakness accuracy fix verified against the official calculator.

### DIFF
--- a/plugins/loadout-lab
+++ b/plugins/loadout-lab
@@ -1,3 +1,3 @@
 repository=https://github.com/AKAddons/runelite-loadout-lab.git
-commit=9dd5e2d17b9905a38f16f9cd7f3e8b7cb2211fcc
+commit=dbead0ed083c695f01cc1392d5afbb127e4a9702
 authors=ajkatz
--- a/plugins/loadout-lab
+++ b/plugins/loadout-lab
@@ -1,3 +1,3 @@
 repository=https://github.com/AKAddons/runelite-loadout-lab.git
-commit=dbead0ed083c695f01cc1392d5afbb127e4a9702
+commit=f0a95a03a1ebc588f228de9607d4069b77ea4940
 authors=ajkatz


pajlada:
Please perform the repository change first @ajkatz, otherwise the diff viewing gets messed up

ajkatz:
Understood - I've reopened #14013 (the ownership move, manifest-only, commit unchanged) and it should go first. This PR is parked until that merges; I'll rebase it onto master afterwards so the diff here is a clean one-line commit= change. Sorry for the churn.

ajkatz:
> Please perform the repository change first @ajkatz, otherwise the diff viewing gets messed up

@pajlada https://github.com/runelite/plugin-hub/pull/14013

ajkatz:
Rebased onto master now that #14013 (the AKAddons ownership move) has merged. The diff here is now a single commit= line - repository= and authors= already match master. Ready for review, thanks @pajlada.

ajkatz:
Thanks for the catch @pajlada - the `Runtime.getRuntime()` use is gone. It was sizing the multi-mob roster thread pool via `availableProcessors()`; it's now a fixed pool of 4 min-priority daemon threads, so the source no longer references `java.lang.Runtime` at all. I also rolled the pin forward to 0.3.2 (current head) so the fix ships on the latest build rather than a superseded 0.3.1 - the diff here stays a single commit= line.

pajlada:
Please don't use `Thread.currentThread().interrupt();` - while it may be valid in your own thread, it's impractical for us to confirm you're interrupting your own thread.

ajkatz:
Addressed the `Thread.currentThread().interrupt()` note - it's gone. That call was never our cancellation mechanism: superseded searches are abandoned cooperatively via an `AtomicLong requestSeq` ticket (every worker checks `if (requestSeq.get() != ticket) return`). The two `interrupt()` calls were just the re-assert-the-flag idiom around a blocking `rosterPool.invokeAll()`; both catch blocks now simply abandon the computation, so nothing re-asserts the interrupt flag. Repinned to dbead0e. Thanks!

ajkatz:
new build has no networking features

ajkatz:
Closing in favour of #14564. This bundled 0.2.2 through 0.3.4 into one 205-file diff; releasing the backlog in sequence instead, smallest first, so each update stands on its own.

PR #14016, opened 2026-07-19
Bump mastering-mixology-extended to 1.9.1-ext1.0

PDBoegel:
## Summary

Bumps `mastering-mixology-extended` to commit `99e079b`, tagged as version `1.9.1-ext1.0` in build.gradle.

Adds an explicit version string (previously the manifest was pointing at a commit without a version, so users saw a commit-hash version in the client) and a README changelog entry documenting the current feature set.

No code changes.


PDBoegel:
Closing — this was only a version-string bump with no functional code change. Not worth churning a re-package/publish for. Will bundle any future version bump with actual code changes.

PR #14022, opened 2026-07-19
Add Plugin Doctor

FreeArcanes:
## Plugin

Plugin Doctor is a local performance diagnostic panel that measures whole-client FPS, client-thread CPU, client-cycle timing, GameTick intervals, JVM heap behavior, garbage-collection activity, and client stalls.

It provides controlled plugin ON/OFF comparisons so users can repeat matched trials and identify a likely performance contributor. Results are explicitly presented as correlations rather than direct per-plugin profiling.

Repository: https://github.com/FreeArcanes/Plugin-Doctor

## Main features

- Live client-health measurements.
- Timed general and plugin ON/OFF captures.
- Pause/resume with paused time and samples excluded.
- Rejection of contaminated pairs when another plugin changes.
- Basic user-readable and complete advanced reports.
- Opt-in assisted switching for only the user-selected plugin, with a warning, conflict refusal, and original-state restoration when possible.
- Detailed documentation for every metric, score, threshold, limitation, and privacy boundary.

## Compliance and privacy

- Standard build with no additional third-party dependencies.
- No reflection, native access, process execution, dynamic code, Java serialization, or input injection.
- No gameplay automation or game-state recommendations.
- No network requests, telemetry, packet inspection, or player-data collection.
- Captures and rankings remain in memory; reports are copied only after explicit user action.
- Assisted switching is disabled by default and acts only on the selected plugin after explicit consent.

## Validation

- Java 11 target.
- `gradlew.bat clean test shadowJar` passes.
- 14 tests, including paired-trial rejection, report formatting, pause accounting, and 225 px sidebar layout regression coverage.
- Root icon is 48x48.
- No `META-INF/services/net.runelite.client.plugins.Plugin` descriptor.

### DIFF
--- a/plugins/plugin-doctor
+++ b/plugins/plugin-doctor
@@ -1,0 +1,2 @@
+repository=https://github.com/FreeArcanes/Plugin-Doctor.git
+commit=65f8efd74320c9e34c729b607e1e656b0c4012b0


raiyni:
We already have the debug plugin and I don't think would allow new plugins to capture this kind of information.

FreeArcanes:
That’s a fair concern. I see Plugin Doctor as serving a different purpose from Debug Tools. Debug Tools provides general diagnostic information, while Plugin Doctor is intended to help users answer a specific question: “Does this plugin consistently affect my client’s performance?”
Its main feature is a controlled comparison workflow. It records the same conditions with one selected plugin enabled and disabled, rejects the comparison if other plugins change, and presents the ON/OFF results in a readable report. It only uses local client and JVM measurements such as FPS, client-thread CPU time, cycle and tick timing, heap usage, and JVM-reported GC time.
It does not inspect network traffic, collect account or gameplay data, upload telemetry, or claim to directly profile another plugin’s code. If a particular measurement or the optional assisted-switching feature is the concern, I’m happy to remove or limit it. I’d appreciate clarification on which capability would be unacceptable so I can adjust the plugin accordingly.

Alexsuperfly:
due to the sensitive nature of the Memory beans and such, we only allow use of them in our debug helper hub plugin managed by the core maintainers.

PR #14060, opened 2026-07-19
Update commit reference in character-export

DZWNK:


Alexsuperfly:

DZWNK:

Thanks. I removed the direct Thread.currentThread().interrupt() call from shutdown handling in d790a93e24c9d946949f47645b0447210b166d93.

The code still calls ExecutorService.shutdownNow() if the writer does not drain or shutdown waiting is interrupted. Is that acceptable, or should I also avoid shutdownNow() entirely?

Alexsuperfly:

PR #14065, opened 2026-07-20
update osrs-tcg

Azderi:


Azderi:
Changed PR from feature update to a Hotfix branch to fix plugins depending on osrs-tcg that broke with the update to limit config writes.

PR #14073, opened 2026-07-20
Fitzgerald.gg

Fitzgerald-gg:
**Fitzgerald.gg** — a companion plugin that captures OSRS gameplay (loot, levels, deaths, quests, slayer, collection log, combat achievements, and ~90 lifetime counters) and syncs it to the player's page on fitzgerald.gg, my own site. It's a single-plugin alternative to running a loot-webhook plus a stat-tracker.

A few things reviewers usually ask about, handled up front:

- **Opt-in.** Both `enabled` and `captureScreenshots` default to **false** and carry the standard IP-disclosure `warning`. Nothing is enrolled, captured or transmitted until the user turns it on. I've also set the marker `warning=` and will defer to your wording on it.
- **Fixed, author-owned endpoints.** Every call targets a hard-coded https://fitzgerald.gg/api/... path — no dynamic or remotely-fetched endpoints (same pattern as plugins/runeprofile). The base URL is editable only under Advanced, for self-hosters.
- **Only the local player's own data.** `LootRecordType.PLAYER` records are dropped, so a PK victim's name/loot is never sent; `onActorDeath` gates on the local actor. No other-player data leaves the client.
- **Screenshots** attach only to the user's own notable events and are off by default.
- **Self-service data rights.** From the side panel the user can lock their page behind a passphrase, list/unlist it from the public directory, export all their data as JSON, or schedule a cancellable deletion — all authenticated by the account's own token, no website login. The lock passphrase is never stored in RuneLite config.
- **Dependencies:** depends on the built-in Slayer plugin (`@PluginDependency`) for on-task tagging; degrades gracefully if it's disabled. No third-party deps, no bundled code — OkHttp/Gson come from RuneLite via injection.

### DIFF
--- a/plugins/fitzgerald
+++ b/plugins/fitzgerald
@@ -1,4 +1,4 @@
 repository=https://github.com/Fitzgerald-gg/fitzgerald-runelite.git
 commit=2f70ef0dd2b64b6cbefc2adb290549ea3a893376
 authors=Fitzgerald-gg
-warning=This plugin submits your gameplay data — loot, levels, deaths, quests, slayer kills, collection log, combat achievements, group-storage movements, lifetime counters, screenshots — and your IP address to fitzgerald.gg, a server not controlled or verified by the RuneLite developers.
+warning=This plugin submits your IP address, and may submit various account data, to a 3rd-party server not controlled or verified by Runelite developers.
--- a/plugins/fitzgerald
+++ b/plugins/fitzgerald
@@ -1,4 +1,4 @@
 repository=https://github.com/Fitzgerald-gg/fitzgerald-runelite.git
-commit=91f1ec2ad663769cc210dbc5a40db0400300c7e6
+commit=a828a7c3d98e845bfaaa15eb9babbc24c354e6f1
 authors=Fitzgerald-gg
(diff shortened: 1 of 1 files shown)


plugins/fitzgerald
```diff
@@ -0,0 +1,4 @@
+repository=https://github.com/Fitzgerald-gg/fitzgerald-runelite.git
+commit=2f70ef0dd2b64b6cbefc2adb290549ea3a893376
+authors=Fitzgerald-gg
+warning=This plugin submits your gameplay data — loot, levels, deaths, quests, slayer kills, collection log, combat achievements, group-storage movements, lifetime counters, screenshots — and your IP address to fitzgerald.gg, a server not controlled or verified by the RuneLite developers.
```
riktenx:
```suggestion
warning=This plugin submits your IP address, and may submit various account data, to a 3rd-party server not controlled or verified by Runelite developers.
```

riktenx:

Fitzgerald-gg:
Done — the export now writes to a plugin-specific subdirectory (.runelite/fitzgerald/export-<name>-<timestamp>.json) instead of the .runelite root.

PR #14096, opened 2026-07-20
Add PK Group Manager plugin

charlie1991cooper-ops:


coopermor:
This appears to violate multiple rules including
> Indicates players in an opposing clan in PVP
> Indicates whom your opponent's opponent is in PVP
> Removes or deprioritises attack, cast (or similar) options from the minimenu in PVP

PR #14099, opened 2026-07-20
Adds Better Notifications

david3200:
A RuneLite plugin that replaces abrupt warnings with a smooth,
shooter-style edge warning.

The configured color begins creeping inward when hitpoints or prayer reaches its
configured point threshold. The effect becomes stronger toward the critical amount,
where it can pulse and trigger RuneLite's configured notification sound.


david3200:
Going to update readme etc. first

PR #14116, opened 2026-07-21
Add sound-override plugin

jarredgoddard:
Adds Sound Override: replaces game sound effects and announces game events with user-supplied audio. Supports raw sound effect ID replacement plus 34 event presets. Detection logic ported (with credit) from c-engineer-completed and odablock-sounds, both BSD-2.

### DIFF
--- a/plugins/sound-override
+++ b/plugins/sound-override
@@ -1,0 +1,2 @@
+repository=https://github.com/jarredgoddard/sound-override.git
+commit=363cff2cede3a6cc98621cfc48a0e2c48aa9f3b3


ldavid432:
The generic ID replacement aspect is not allowed. From out rejected features list:
> ID based plugins: Plugins that use player provided IDs for the entirety of their functionality

The preset sound swaps should be fine

jarredgoddard:
Removed the generic ID replacement, including the configurable ectophial sound ID (converted to a hardcoded teleport preset). All sound swaps are now fixed presets.

PR #14120, opened 2026-07-21
better-dt2-boss-utilities: add plugin

TheSpryt:
Adds Better DT2 Boss Utilities, a quality-of-life plugin for the Desert Treasure 2 bosses, currently focused on Vardorvis. It runs alongside other DT2 plugins.

Features:
- Hide any of the 8 Vardorvis axe spawns (static, moving, or both, per direction), with optional muting of the hidden axes' sounds
- Highlight axes as hull, tile, true tile, or outline, skipping any that are hidden
- Highlight Vardorvis' heads by attack style (magic and ranged), each with its own colours
- Hover feedback on the spore captcha, with an optional outline around every spore
- Swap the Whisperer's and Vardorvis' projectiles to other boss styles

Repository: https://github.com/TheSpryt/better-dt2-boss-utilities

Tested in game against Vardorvis.

### DIFF
--- a/plugins/better-dt2-boss-utilities
+++ b/plugins/better-dt2-boss-utilities
@@ -1,0 +1,2 @@
+repository=https://github.com/TheSpryt/better-dt2-boss-utilities.git
+commit=7dcaaa6f9329518bfdd612b164fdf7d3def20045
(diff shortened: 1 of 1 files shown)


riktenx:
this fell out of the bot view, if you still care about this make a new pr (pr limits no longer apply)

PR #14134, opened 2026-07-21
update osrs-best-in-slot

osrsbis:


osrsbis:
Hi — this has been open since 21 July with a green build. Happy to answer anything that would speed up the review.

A short guide to the diff (0.7.0 → 0.9.0), since it's bigger than a typical bump:

- No new runtime dependencies. build.gradle changes only the version string; everything added is compileOnly or test-scope. The encoder dependency that blocked our earlier submission is gone — frames are uploaded raw and stitched server-side, so there is no muxer in the plugin.
- Store delivery capture is gated behind the existing uploadTradeScreenshots toggle, which still defaults to off.
- Off-book value events (drop / pickup / alch / death items-lost) — own account only.
- Two fixes: a chat-event sweep regression, and exact transacted gp via coin delta.
- Plugin warning text is unchanged from the approved version.

Happy to split this into smaller submissions if that's easier to review.

osrsbis:
Superseding this with smaller, individually reviewable changes, starting with the bugfix bump #14418

PR #14143, opened 2026-07-21
Update Flipping Masterminds Plugin

ca-gray:
Changes:
1. Better UI with customisation features of the listed item panel
2. Added filtering with GE volume 
3. Fixed visual bug with for some people the text box to enter min/max price was squished by default.
4. Updated GE send data requirement to send upon login to avoid missing player GE slot updates with quick logins. (still requires API token)

### DIFF
--- a/plugins/flipping-masterminds
+++ b/plugins/flipping-masterminds
@@ -1,3 +1,3 @@
 repository=https://github.com/ca-gray/Flipping-Masterminds.git
-commit=9f8bee61d569db7e2b3b32bec7202df4eb3badc3
+commit=bb932fda816e6720df96ecd7847685b09615b246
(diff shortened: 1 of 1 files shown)


Alexsuperfly:
use build=standard

ca-gray:
> use build=standard

Added and in a strike of luck, been able to update my wiki prices api calls to v2.

Alexsuperfly:
remove .class files

ca-gray:
Removed build folder, manually added my whole folder as i was lazy instead of doing commit via my regular way.

riktenx:
you can't shutdown() the injected okhttpclient

ca-gray:
Thanks for the review, updated the code so it no longer performs a shutdown()

PR #14186, opened 2026-07-22
Update party-defence-tracker version

Hyftar:
Updates the Tonalztics of ralos defence reduction value to match the new value from today's update


pajlada:
Hi, switch your `build=gradle` to `build=standard`, and include a `version` key in your `runelite-plugin.properties` file that matches the version you want to show up in the sidebar. That way, these PRs will be merged in quicker.
Thanks!

PR #14187, opened 2026-07-22
Add inventory-click-highlighter plugin

nick-michael:
A new plugin which will highlight clicks on items in the inventory. You can customise the highlight colour, outline colour, fade duration and exclude any specific actions.

https://github.com/user-attachments/assets/443c1f03-7350-45c1-ae94-cd8a82cd9c78


Alexsuperfly:

PR #14196, opened 2026-07-23
Update Guide Overlay 1.2.2

RunelitePlugin:
Fixed BRUHsailer guide display: long paragraph steps are now split into one action per step, so nothing gets cut off by the step length limit.
Existing saved progress migrates automatically.


RunelitePlugin:
I'm not sure if my previous pull-requests had issues, but I cancelled them both and published this one manually again.  Please let me know if there are issues.

PR #14201, opened 2026-07-23
Updated plugin to include real time charts and favorite items

AaronDemond:
- Updated the Bank Stats plugin to allow users to view real time price data and charts all within the client, accessible view buttons and hotkeys.
- Added item favorites for monitoring items easily


AaronDemond:
Here's a couple screen shots of it in action. Basically its made to track value of items in your bank over time. Handles new items bought / received and items sold or lost. Also has tooling for very quickly searching real-time market data for any item, all within the client so no hopping to ge-tracker. Theres a favorite section to monitor different item batches at one time, and finally export / import for different bank snapshots as starting points to track from. I've been using it myself for a good month now, and even though (as you can see) I'm down horrendously, its been a cool to see it work. The quick price lookup is really nice once you memorize the hot keys. Really hope you guys can give it a look soon!

image

image

image

image

Alexsuperfly:

AaronDemond:
> * use of Desktop::browse is not allowed, use LinkBrowser::browse instead

Thanks for the response. Was that the only issue? Anything else I should change before I update the pr?

riktenx:
we dont know until you fix that

AaronDemond:
> we dont know until you fix that

Oh okay no problem. Should be able to push that when I get back on the comp in a few hours. Thanks!

riktenx:

- you MUST have a `@ConfigItem` boolean, that defaults to false, that entirely disables the networking calls, its `warning` MUST be set to exactly `"This feature submits your IP address and various account data to a 3rd-party server not controlled or verified by Runelite developers."`
- all networking calls need to go into dedicated methods, and at the top of every method where an opt-in networking call is made you MUST check the boolean config and either return early or throw an exception. the body of the method MUST be only this check and the network call

AaronDemond:
> If you add networking to a 3rd-party server to an existing plugin it MUST be entirely opt-in:
> 
> * you MUST have a `@ConfigItem` boolean, that defaults to false, that entirely disables the networking calls, its `warning` MUST be set to exactly `"This feature submits your IP address and various account data to a 3rd-party server not controlled or verified by Runelite developers."`
> * all networking calls need to go into dedicated methods, and at the top of every method where an opt-in networking call is made you MUST check the boolean config and either return early or throw an exception. the body of the method MUST be only this check and the network call

So the user has to explicitly go into the config and approve the config? I may end up scrapping that feature entirely. We'll see. I will re submit either either tonight or tomorrow. Thanks

Alexsuperfly:
This pull request has been waiting for author action for >week. It is now being closed as stale. Feel free to open a new pull request once you've addressed the requested changes.

PR #14204, opened 2026-07-23
Update Flip Smart

mablack01:
## Summary
- Added an optional in-panel session stats display showing session profit, projected hourly rate, GP/hour, and logged-in time
- Improved GE trade history syncing to automatically backfill missing history in the background
- Fixed an edge case where round-trip tracking could misattribute trades sharing the same GE slot


mablack01:
Closing — opened ahead of our internal release schedule. Will re-open when we're ready to publish. Apologies for the noise.

PR #14206, opened 2026-07-23
smithingtracker

spook121:
updated smithingtracker to have better clarity for text, more sophisticated logic for math with live prices


coopermor:
Remove the file `src/main/java/com/smithingtracker/src - Shortcut.lnk` you added to your repo

coopermor:
You should use [getItemPriceWithSource](https://static.runelite.net/runelite-client/apidocs/net/runelite/client/game/ItemManager.html#getItemPriceWithSource(int,boolean)) rather than doing networking to the wiki prices API.

PR #14207, opened 2026-07-23
Add IRONSCAPE Optimal

GROBBS-hash:
Adds **IRONSCAPE Optimal** — the community Ironman Efficiency Guide (by Oziris & the ironman.guide community, v4 "Enhanced 2026" edition) as a step-by-step side panel.

- Ordered, tickable guide steps with automatic completion detection: skill levels, quest state, mid-quest varbit/varp checkpoints, item counts, teleport/arrival detection
- On-screen current-step overlay with live item/level counts
- Bank view showing each upcoming step's items with have/need counts (real, withdrawable bank widgets)
- NPC/quest-start/target-tile overlays and teleport click-path hints, Quest Helper-style
- Optional navigation handoff to the Shortest Path plugin via PluginMessage (degrades gracefully when not installed)

Repository: https://github.com/GROBBS-hash/ironscape-optimal (BSD-2-Clause)

The bundled guide content is used with the express permission of its authors (Oziris / ironman.guide), who are credited in the plugin descriptor, panel, and README.

### DIFF
--- a/plugins/ironscape-optimal
+++ b/plugins/ironscape-optimal
@@ -1,0 +1,2 @@
+repository=https://github.com/GROBBS-hash/ironscape-optimal.git
+commit=53d225c20c51a77f3cfdbbd4cb46e7f72f1366a4
--- a/plugins/ironscape-optimal
+++ b/plugins/ironscape-optimal
@@ -1,2 +1,2 @@
 repository=https://github.com/GROBBS-hash/ironscape-optimal.git
-commit=627cc6754e8bc6d3fed3129925962ba2e5e870ea
+commit=92f3c65c9b696901fd3f90ff5633126be0758ded
--- a/plugins/ironscape-optimal
+++ b/plugins/ironscape-optimal
@@ -1,2 +1,2 @@
 repository=https://github.com/GROBBS-hash/ironscape-optimal.git
-commit=324cdf572d9d7d36e567c9be1b2b8790cbd3536a
+commit=7e03f249847a816d969e5fb2395f0ede77fa69b2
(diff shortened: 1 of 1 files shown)


Alexsuperfly:
* LICENSE invalid, see the [plugin-hubs](https://github.com/runelite/plugin-hub/blob/master/LICENSE) for example

GROBBS-hash:
Thanks for the review! All three addressed in the updated commit:

- Reflection removed (the Quest Helper handoff is now just a chat message pointing players at QH)
- `hopToWorld` removed — world links open the world switcher in game; direct `changeWorld` only on the login screen
- LICENSE replaced with the verbatim BSD-2 text

GROBBS-hash:
Hi @Alexsuperfly — gentle bump on this one when you have a moment. All three review points (reflection, `hopToWorld`, LICENSE) were addressed, reply above from July 29, and the pin now points at the latest build. Happy to make any further changes if anything else comes up. Thanks!

riktenx:

Alexsuperfly:
whatever this [file access](https://github.com/GROBBS-hash/ironscape-optimal/blob/b094b21b6cfdb2b93a955b50ff33267b99b2fcba/src/main/java/com/ironscape/DataFiles.java#L119) is trying to do is not going to be allowed, i suggest you drop the whole "dev directory" thing and not do whatever this is trying to do.

GROBBS-hash:
Thank you for your response, I've changed it.

For context on what it was: most corrections to this plugin are data rather than code (a map pin a few tiles out, an item missing from a quest's kit, a note). All of that is JSON compiled into the jar, so fixing one character meant a rebuild and a client restart mid play-test. The config let me point at a checkout so those files were read from there instead, with an in-game command to re-read them. It was empty by default and unreachable on a normal install, but I take the point that it's file access outside the plugin's own directory and shouldn't ship.

`DataFiles` is now a one-line classpath read. I also removed a one-time copy out of `~/.runelite/bruhsailer` (this plugin's former name) since that reads a different directory under `.runelite`. The config-key migration it went with is `ConfigManager` only and stays.

Everything the plugin writes (captured map pins, local place overrides, the guide manifest, bank snapshots) was already confined to `~/.runelite/ironscape/` and is unchanged.

Pin updated to `1ea9094`.

PR #14226, opened 2026-07-23
Update Terain Toggle

milz-1:
Added hide lower plane functionality utilising existing Low Detail plugin like the Automatic Low Detail plugin as requested by https://github.com/milz-1/Terrain-Toggle/issues/4


Alexsuperfly:
remove .class files

PR #14231, opened 2026-07-23
Update stations-cozy-clutter to 0.3.0

StationEarthxo:
Updates Station's Cozy Clutter to version 0.3.0.

- Adds animated object and NPC decorations to the sidebar catalogue and placement workflow
- Persists animation metadata in Tilepacks while remaining compatible with existing Tilepacks
- Keeps catalogue/model caches and imported placements bounded
- Removes the `enabledByDefault` descriptor entry

Validated with a clean `gradlew clean test` build against the latest RuneLite release.

### DIFF
--- a/plugins/stations-cozy-clutter
+++ b/plugins/stations-cozy-clutter
@@ -1,2 +1,2 @@
 repository=https://github.com/StationEarthxo/Stations-Cozy-Clutter.git
-commit=f563e522fa1483f52f60b3daaf953d4882e623af
+commit=ff0217648f51e78c6b40f87866c96d48c17d344f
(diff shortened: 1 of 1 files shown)


StationEarthxo:
Updated the plugin commit to `1456dcd21b15ac87f4805a6472e9c8b3859086f1` after additional live testing. This revision also adds bounded sub-tile nudging and a move-in-place workflow for existing props. Nudge offsets are optional and backward-compatible with existing Tilepacks; moves and nudges participate in the existing undo system. The clean Gradle test suite and a normal GPU-enabled development client both passed before pushing.

PR #14238, opened 2026-07-24
Update stations-cozy-clutter to 0.3.1

StationEarthxo:
Updates Station's Cozy Clutter to version 0.3.1.

- Adds a sidebar Undo button and Ctrl+Z support while the panel is open
- Adds a placement-counted Delete All button with an explicit confirmation warning
- Makes Delete All recoverable through the existing bounded undo history
- Documents the new build-management workflow

Validated with the Gradle test suite and a normal GPU-enabled development client.


ldavid432:
use of KeyboardFocusManager is not allowed

PR #14259, opened 2026-07-24
3Dconnexion Space Mouse integration plugin

NeusAap:
This plugin adds 3Dconnexion Space Mouse integration.
You can change the pitch, yaw and zoom of your RuneScape camera using the 3dconnexion space mouse navigator!

Perfect for some RS grinds during CAD modeling ;)

### DIFF
--- a/plugins/spacemouse
+++ b/plugins/spacemouse
@@ -1,0 +1,2 @@
+repository=https://github.com/NeusAap/runelite_spacemouse.git
+commit=478bc7818d87c6868f76b15be132f87e38de4b12


NeusAap:
Ah, I see that it fails to build because of the external third party dependency on 'hid4java' (org.hid4java:hid4java:0.8.0).
I guess someone has to approve that external package and add it to the build.gradle file for the plugin hub?

I understand that this is probably un-preferred, but I wouldn't see a more platform independant way for HID control than this.

Let me know if I can do anything!

raiyni:
We generally do not allow native code for hub plugins.

NeusAap:
> We generally do not allow native code for hub plugins.

Alright, so no way to get it allowed/added?
It makes life way easier to access the HID devices to get the raw data from the hardware.

Or is there maybe a HID -> Java library already accepted/in the project I could use?

Would a plugin that communicates with a separate locally-running helper application over localhost be acceptable for the Plugin Hub? The RuneLite plugin itself would contain no native code.

riktenx:
1. no
2. no
3. no (we dont allow plugins that depend on custom native binaries either, even if separately installed)

PR #14275, opened 2026-07-25
Add OSRS Bingo plugin

austin-baker2:
**OSRS Bingo** — shows a clan's bingo board in-game and auto-logs on-board drops to the event for a human admin to approve. Backed by the OSRS Bingo service, which any clan can sign up for and run an event on.

**External connection (disclosed):** the plugin contacts exactly one host and nothing else, and is fully opt-in — with no board code entered it makes no network requests. Once a board code is set, it fetches the board and reports drops that match a board tile (item id + quantity, drop source, board code), optionally with a proof screenshot (toggleable). Every outbound request lives in a single class; details are in the repo's `SECURITY.md`.

**Jagex rules:** detection uses only RuneLite loot events (NPC kills, reward chests) — never the inventory or bank, so bank items can't create fake entries. PvP loot is never reported. The plugin never completes a tile itself; a human approves every submission. No automation.

**License:** BSD 2-Clause.

### DIFF
--- a/plugins/osrs-bingo
+++ b/plugins/osrs-bingo
@@ -1,3 +1,4 @@
 repository=https://github.com/OSRS-Bingo-Plugin/runelite-plugin.git
 commit=a211759cef2516929f8f33ac11b3b52a2a0a47ed
+warning=This plugin submits your IP address, and may submit various account data, to a 3rd-party server not controlled or verified by Runelite developers.
 authors=austin-baker2
--- a/plugins/osrs-bingo
+++ b/plugins/osrs-bingo
@@ -1,0 +1,2 @@
+repository=https://github.com/OSRS-Bingo-Plugin/runelite-plugin.git
+commit=bddd27c50dea195474f3a1467d58a0ac376743e8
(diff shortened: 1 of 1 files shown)


Alexsuperfly:
if youre submitting this from an org repo you have to set the authors field of the manifest to allow those github accounts authorization to submit hub updates
see: [117HD](https://github.com/runelite/plugin-hub/blob/7281906d623dd1ccd40663e67d63d6a700775e09/plugins/117hd#L3)

austin-baker2:
Done — added `authors=austin-baker2` to the manifest. I've also updated our publish workflow so future commits to the plugin repo are authored by my account rather than the publish bot. Thanks for the pointer!

plugins/osrs-bingo
```diff
@@ -0,0 +1,3 @@
+repository=https://github.com/OSRS-Bingo-Plugin/runelite-plugin.git
+commit=a211759cef2516929f8f33ac11b3b52a2a0a47ed
```
Alexsuperfly, line 2:
```suggestion
commit=a211759cef2516929f8f33ac11b3b52a2a0a47ed
warning=This plugin submits your IP address, and may submit various account data, to a 3rd-party server not controlled or verified by Runelite developers.
```

PR #14286, opened 2026-07-25
Added flati boss tracker plugin.

Flati:
The plugin tracks boss kills and posts the KC to an endpoint hosted at flati.is/osrs (configurable for custom backends).
Allows friend groups to keep boss KC data with each other and makes a visual graph to show boss trips with kill timers.
Also tracks agility laps and other counters kept by Grace.
Includes information to set up your own server but creating a new group on flati.is is also available by request.

image

image

image

image

image

image

### DIFF
--- a/plugins/flati-boss-tracker
+++ b/plugins/flati-boss-tracker
@@ -1,2 +1,3 @@
 repository=https://github.com/Flati/Flati-Boss-Tracker.git
 commit=89bc1a55bc601969a45ef98c075cc26fbde3310e
+warning=This plugin submits your IP address, and may submit various account data, to a 3rd-party server not controlled or verified by Runelite developers.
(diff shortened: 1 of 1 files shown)


plugins/flati-boss-tracker
```diff
@@ -0,0 +1,2 @@
+repository=https://github.com/Flati/Flati-Boss-Tracker.git
+commit=89bc1a55bc601969a45ef98c075cc26fbde3310e
```
riktenx:
```suggestion
commit=89bc1a55bc601969a45ef98c075cc26fbde3310e
warning=This plugin submits your IP address, and may submit various account data, to a 3rd-party server not controlled or verified by Runelite developers.
```

PR #14287, opened 2026-07-25
True tile animation 2line fix player owner stale

Posiedien76:
Fixed player owner going stale but not recording it


pajlada:
gotta fix your merge conflicts

PR #14317, opened 2026-07-26
Update dorgesh-kaan-lights to 2.0.3

tmgdejong:
### Update dorgesh-kaan-lights to 2.0.3

- Fixed a bug where the legacy overlay only showed lamps in the current area
- Fixed a bug where states were reset upon the `LOADING` state
- Removed unused imports


Alexsuperfly:
removing your name from authors means you wont be able to make any new plugin changes if this is merged, are you sure this is intended?

tmgdejong:
I had no clue it actually had to be your GitHub name in the authors field, this is my ign. I will change it back then, thanks for the headsup

PR #14327, opened 2026-07-26
Create session-movement-tracker

lwr27:
Records a session by session timeline of walking, teleports, and bank visits to a local file. Optionally syncs to a GitHub repo the user configures, opt-in and off by default.


Alexsuperfly:
use RUNELITE_DIR to get the clients .runelite directory instead of relying on the [systems user.home](https://github.com/lwr27/session-movement-tracker/blob/f5d3d687200f64c83d8fd858d1d6f105365252f4/src/main/java/com/rstracker/RouteTrackerPlugin.java#L350), then add your plugin specific subdirectory to that path

PR #14367, opened 2026-07-27
Update Gielinor travels

willocon:
Added ability to change URL endpoint to facilitate clan events for example. 
Added a message that is displayed in chat when a user completes a game they have already completed. 
Added the AudioManager class which allows the client to play a jingle upon completion of a game. Said jingle is added in the resources folder.


Alexsuperfly:
replace printStackTrace with logger statement

PR #14391, opened 2026-07-28
feat: add emojis-plus plugin

idyet:
# Emojis Plus

emojis-plus

### What it does

Emojis rock 🪨. [RuneLite's built-in **Emojis** plugin supports a bunch of them](https://github.com/runelite/runelite/wiki/Emojis). It actually supports a lot more than this list (over 1000 😱), accessible by their `:code:` (`:smirking_face:`, `:weary_cat:`, etc.). This plugin offers a way to discover 🔍  these renderable emoji codes. As you start typing a code `:sm`, a dropdown of matching codes is displayed in the chatbox. The dropdown can be scrolled with arrows. Escape, Tab, and Enter will dismiss the dropdown - this plugin does not autocomplete text.

### Notes for review

- **Depends on the built-in Emojis plugin** as a soft runtime data dependency - we read two cache files the built-in plugin produces, at a public path; no build-time or class-level linkage, no internal-API use.

### DIFF
--- a/plugins/emojis-plus
+++ b/plugins/emojis-plus
@@ -1,0 +1,2 @@
+repository=https://github.com/idyet/emojis-plus.git
+commit=2a9f9f3f4fa589c34eff15c11fe3bedd50acc779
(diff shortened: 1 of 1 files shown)


ldavid432:
Chat autocompletion is not allowed per our rejected features list:
> Plugins which programmatically insert text into the user's chatbox input for any reason (pasting messages, shorthand expansion, etc.). This is considered to be autotyping. 

PR #14392, opened 2026-07-28
add Player NPC Replacer to plugin hub

YoungSensei-TSU:
Player NPC Replacer

Shift-right-click a player (or an NPC) to replace their model with any other NPC's - purely cosmetic, client-side only. Supports overriding another player, overriding your own character (optionally keeping your own animations and equipment visible on top of the new model), and overriding an NPC's own model. Can be used for roleplaying, alleviating boredom, b roll for youtube content etc.


Alexsuperfly:
your LICENSE is invalid, just remove the statement at the bottom

PR #14413, opened 2026-07-28
add collection-log-progress

xboxnuker-rgb:
Adds colour-coded Collection Log progress, percentages, configurable colours, and page filters.


xboxnuker-rgb:
ready to go :)

xboxnuker-rgb:
Preserved Collection Log progress snapshots during normal loading transitions such as teleports, while still clearing data on logout, world hop, connection loss, or authentication.

PR #14420, opened 2026-07-28
card-exchange-highscores: bump to 0.3.5

CompilerOfDust:
Point at tcg-highscores dbb4848 (0.3.5), which defaults the upload setting to off and rewrites the config warning. Add a manifest warning naming the data sent and that osrscardexchange.com is a third-party server not controlled or verified by the RuneLite developers.


Alexsuperfly:
needs rebase, you have to have the IP address warning in the config and the manifest to be enabled again

PR #14423, opened 2026-07-29
Improved target true tile

Zotters:


Notloc:
You could've opened a PR against mine...

Zotters:
> You could've opened a PR against mine...

I'm sorry big dog. I can take it down if you want.

coopermor:
We would prefer you PR the fix to the original plugin so the 84k users with the plugin installed can have the fix.

Zotters:
> We would prefer you PR the fix to the original plugin so the 84k users with the plugin installed can have the fix.

Got it.

PR #14431, opened 2026-07-29
Update target-true-tile

Notloc:
Move the list clears to the top of render to prevent memory leakage when the render loop is failing.


coopermor:
In runelite-plugin.properties please add `build=standard`

PR #14437, opened 2026-07-29
osrs-tcg-exchange: new plugin

CompilerOfDust:
Adds **OSRS TCG Exchange** — a plugin for trading OSRS TCG cards with other players and opening the Card Exchange marketplace from in game.

- Right-click another player (beneath the game's *Trade with*) to open a trade page in the browser. Both players deterministically derive the **same** RuneLite Party passphrase from their alphanumerically-ordered RSN pair — no offer, no accept step and no polling.
- Right-click a Grand Exchange Clerk (beneath *Exchange*) to open the Card Exchange marketplace.

The plugin makes **no HTTP/API calls** — it only opens URLs on osrscardexchange.com. Those URLs include the two RuneScape display names, which is disclosed in the manifest `warning` and in the plugin's own UI. Trading is RSN-only: no account, no login, no token.

Repository: https://github.com/CompilerOfDust/OSRS-TCG-Exchange-TheCardExchange
Author: CompilerOfDust

### DIFF
--- a/plugins/osrs-tcg-exchange
+++ b/plugins/osrs-tcg-exchange
@@ -1,4 +1,4 @@
 repository=https://github.com/CompilerOfDust/OSRS-TCG-Exchange-TheCardExchange.git
-commit=008ebabdb904711cc48baf7ae73da1e1eab16bfe
+commit=a677e41426a93cd24e91f349cb3dd61dbcb31417
 authors=CompilerOfDust
 warning=This plugin opens pages on osrscardexchange.com, a third-party website not controlled or verified by the RuneLite developers. Those page addresses include your RuneScape display name and the display name of the player you choose to trade with.
--- a/plugins/osrs-tcg-exchange
+++ b/plugins/osrs-tcg-exchange
@@ -1,0 +1,4 @@
+repository=https://github.com/CompilerOfDust/OSRS-TCG-Exchange-TheCardExchange.git
+commit=5743d627ee73991f7aa83e5a065f6a2539fce67a
+authors=CompilerOfDust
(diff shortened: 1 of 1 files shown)


pajlada:
Hi, you can use the named `developerMode` variable and provide an extra setting for the base url override instead of your current `defaultApiBaseUrl` functionality.

Alexsuperfly:
your LICENSE is one we cannot host, we suggest BSD-2, see the [plugin-hubs](https://github.com/runelite/plugin-hub/blob/master/LICENSE) as an example

CompilerOfDust:
Amending both issues. moving to BSD 2 Licence and providing an extra setting for the base url

CompilerOfDust:
Thanks for the review — the fixes are pushed and the manifest is bumped to the new commit (`008ebab`).

- **License:** RuneLite can't host the previous license, so I relicensed the plugin source to **BSD-2-Clause** (and removed the AGPL file), following the existing hub plugins as the example. The BSD-2 covers all functionality — the *Trade Cards* and *Exchange Cards* features.
- **Base URL override:** I dropped the old `defaultApiBaseUrl` mechanism that read a system property / environment variable. The plugin now injects the named `developerMode` variable and only applies the `Web app URL` config setting when RuneLite is running in developer mode. Normal installs always use the live site, and the shipped plugin no longer reads any system properties or environment variables.

Ready for another look — thanks!

raiyni:
Your license file is not in a valid format.

CompilerOfDust:
Actioned, the LICENCE should now be in the valid format, closely matching the RuneLite exmaple provided

PR #14438, opened 2026-07-29
anvil: keep the home board visible when federated (+ sidebar/federation/recap update)

AhmedFathy2001:
Updates the Anvil plugin.

Changes since the pinned commit:
- Sidebar: keep the member's own clan board visible when they're federated with other clans (it could previously disappear, leaving only the relayed clans).
- Sidebar: score Leagues/points boards by point weight; exclude optional (bonus) tiles from the point and count totals.
- PvP kill tiles: honour the site-configured minimum loot value before crediting.
- Clearer in-game notice when the account token is rejected or the site is unreachable.


Alexsuperfly:

AhmedFathy2001:
> * use of thread sleep is not allowed; use a scheduled executor etc. instead

should be fixed

PR #14442, opened 2026-07-29
add live-on-clan

MilicoOSRS:
Adds the Live On Clan plugin.

Features clan announcements, rank requests, monthly MVP rankings, and Twitch live status for members of Wise Old Man group 1945.

Third-party integrations are opt-in and disabled by default. The plugin connects to the clan API only after the user accepts RuneLite's third-party warning.

Validated with Java 11 using `./gradlew clean test jar`.


plugins/live-on-clan
```diff
@@ -0,0 +1,2 @@
+repository=https://github.com/MilicoOSRS/live-on-clan.git
+commit=ce282d7f8ba1aa826a71b9d82b39eb588ae277cc
```
riktenx, line 2:
```suggestion
commit=ce282d7f8ba1aa826a71b9d82b39eb588ae277cc
warning=This plugin submits your IP address, and may submit various account data, to a 3rd-party server not controlled or verified by Runelite developers.
```

PR #14450, opened 2026-07-29
Add Private Group Manager

charlie1991cooper-ops:
Adds Private Group Manager to the Plugin Hub.

Private Group Manager allows players to create temporary private groups without requiring a Clan Chat or Friends Chat. It includes privacy reminders, group roles, membership controls, shared world calls, tile pings, and optional Entity Hider integration.

Designed for PvM, events, and other organised group activities.

### DIFF
--- a/plugins/private-group-manager
+++ b/plugins/private-group-manager
@@ -1,2 +1,2 @@
 repository=https://github.com/charlie1991cooper-ops/private-group-manager.git
-commit=4894b3bb454494e830480d74c9a32405db6131a5
+commit=8337e70d51c0813d21b39f49d4824c504fa09e9d
--- a/plugins/private-group-manager
+++ b/plugins/private-group-manager
@@ -1,0 +1,2 @@
+repository=https://github.com/charlie1991cooper-ops/private-group-manager.git
+commit=4894b3bb454494e830480d74c9a32405db6131a5


charlie1991cooper-ops:
The main reason I built this plugin was to simplify some of the coordination problems we regularly encounter during group events. We use voice chat for communication, but certain things can become unnecessarily slow or confusing when everything has to be communicated verbally.

The biggest and most useful feature for us is the world call system. Previously, someone would call a world over voice chat, and everyone else would then have to remember the number, open the world switcher, scroll through the list and find it.

With the plugin, a player can simply click to share their current world. That world is then clearly highlighted for everyone else in the group panel, and they can click the highlighted world to hop directly to it. It removes several unnecessary steps and makes coordinating a group considerably quicker and less error-prone.

Another issue we encountered was people unintentionally revealing the group's world or location to people outside the group. This can happen simply because someone has their private chat enabled or discusses the current world/location elsewhere without thinking about it.

To help with this, I added a polite reminder asking players to consider turning their private chat off. I deliberately made this a prompt rather than automatically changing anything, as I don't want the plugin taking control of a player's settings without their input.

I also included the remove from group option as part of the group-management functionality. From the discussion here, I now understand that there may be concerns with the way I've implemented that feature, and I'm completely open to changing or removing it if necessary for the plugin to meet Plugin Hub requirements. It isn't essential to the core purpose of the plugin.

Lastly, I included a convenient Entity Hider toggle. During group activities there are situations where we need to trade another group member quickly, particularly when time is a factor. Temporarily hiding other players makes finding and interacting with the intended person much easier.

Rather than implementing my own player-hiding functionality, I intentionally chose to use the existing Entity Hider plugin and simply provide a convenient way to toggle it. My reasoning was that this reused existing RuneLite functionality instead of duplicating it.

Overall, the intention of Private Group Manager isn't to automate gameplay or give players an advantage in combat. It's primarily a coordination and quality-of-life tool for groups that are already communicating and playing together.

I'm very happy to make changes where something doesn't fit Plugin Hub expectations. The world/group coordination functionality is the important part for me, so if some of the secondary convenience features need to be changed or removed, I'm completely open to doing that.

charlie1991cooper-ops:
The goal isn't to replace RuneLite Party or Party Hub. Private Group Manager builds on top of the existing Party system by adding lightweight group management features that parties don't currently provide.

The main features are:

Leader and Coordinator roles with permission-based actions.
Shared world calls with a wrong-world indicator and one-click hopping.
Shared tile pings for group coordination.
A single panel for managing temporary organised groups.

The private chat reminder and Entity Hider shortcut are simply convenience features included because they're commonly used alongside organised group activities. If those features are considered out of scope for Plugin Hub, I'm happy to remove them and keep the plugin focused on group coordination.

riktenx:
Use of client.hopToWorld will prevent your PR from being automatically reviewed by the bot. You are welcome to wait for manual review, but as a fair warning that is performed by volunteers on a "when someone has time and feels like it" timeline and I don't think there's much of that happening as of late. Up to you whether you care or not.

Nightfirecat:
[You are registering your communication service to the event bus when GroupManager is initialized](https://github.com/charlie1991cooper-ops/private-group-manager/blob/8337e70d51c0813d21b39f49d4824c504fa09e9d/src/main/java/com/cakepowder/privategroupmanager/GroupManager.java#L45-L47) but it is never unregistered; please ensure that turning off your plugin does this upon `shutDown()`.
LGTM otherwise.

Alexsuperfly:
This pull request has been waiting for author action for >week. It is now being closed as stale. Feel free to open a new pull request once you've addressed the requested changes.

PR #14478, opened 2026-07-30
Update Flip Smart

mablack01:
## Summary
- Active Flips is now derived from a single authoritative in-plugin offer store, so GE slot highlights, timers, and the Active Flips list survive world hops and other in-game activities instead of blanking or flickering.
- Fixed a bug where a cancelled-but-uncollected Grand Exchange slot could resurrect as a live offer.
- Session Stats: Projected Profit no longer double-counts partially-filled sell offers.
- Active Flips state is now pushed to the FlipSmart backend so the website mirrors what the plugin sees in-game (covered by the existing data-sharing warning below).


mablack01:
Added a few more last minute fixes that I was able to get done

mablack01:
Pushed some changes to remove dead code/dedupe some logic to get us under 200k tokens. I'll be focusing my next few updates on slimming down the plugin so this isn't a problem down the line

PR #14486, opened 2026-07-30
More status bars

Avexiis:
Adds status bar color options and warning flashes for status bar thresholds


Avexiis:
an old plugin I was working on is in my fork of the hub and is blocking this..gonna remove that and resubmit. totally forgot about that

PR #14554, opened 2026-07-31
Manual state tracker

shdshdw:
Manual State Tracker shows a single movable overlay image that you switch yourself with hotkeys. It never reads the game to decide what to display, which is the whole point: it stays clear of the plugin-hub rules around boss mechanic indicators and prediction, because you are the one tracking the state.

You build state sets in a side panel, each holding any number of states. Each state has a name, a hotkey, and an image, which can be:

- Text, drawn as large as it fits, with fill and outline colour pickers. For example: a single character makes a good phase counter.
- A game icon, searchable over every item in the cache plus all 32 prayers, 24 skills and 178 spells, pulled from the client's own sprites.
- A few built-in shape: four arrows, a looping arrow, a cross and a plus, also colourable.
- A custom PNG uploaded by the user
- A flat colour.

Typical uses: counting a phase yourself, remembering which prayer you meant to be on, tracking your place in a rotation, or marking which direction to go next.

### DIFF
--- a/plugins/manual-state-tracker
+++ b/plugins/manual-state-tracker
@@ -1,2 +1,2 @@
 repository=https://github.com/shdshdw/manual-state-tracker-runelite-plugin.git
-commit=047f45036f3da9ec90b21c2b906c55ab4f5a264d
+commit=d533ae804a7b6b99ae2f5f4a4dcc450abec7301d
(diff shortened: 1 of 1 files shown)


Alexsuperfly:

shdshdw:
> * LinkBrowser::open and Desktop::open are restricted APIs, remove your usages of them

Done. 
I'm surprised that the developer guide doesn't have a compiled list of restricted API endpoints https://github.com/runelite/runelite/wiki/Developer-Guide

PR #14558, opened 2026-08-01
Update Lending Tracker: fix leaving a group, multiple owners, Looking For duplicates

Guess34:
Fixes leaving a group (it didn't actually leave, and the departure never reached other members), adds support for multiple owners, and fixes a handful of smaller bugs - duplicate "Looking For" entries on edit, joining failing when the relay is cold, and loans from one group showing under another.

No new data is sent anywhere and no new permissions or APIs are used.


Guess34:
Added a second commit — data lifecycle fixes for items and requests reappearing after removal. No new APIs or permissions, same capability surface as before.

Alexsuperfly:

Guess34:
Removed LinkBrowser.open - the Screenshots button shows the folder path instead of opening it. LinkBrowser.browse is still used for the GitHub issues link.

PR #14563, opened 2026-08-01
Add superior-slayer-tracker

JJDeakins1:
Adds the Superior Slayer Tracker plugin.

Tracks superior slayer monster kills and unique drops (imbued heart, eternal gem, mist/dust battlestaff), with per-monster kill counts, dry-streak stats, and drop-rate expectations shown in a side panel.

Repository: https://github.com/JJDeakins1/superior-slayer-tracker

### DIFF
--- a/plugins/superior-slayer-tracker
+++ b/plugins/superior-slayer-tracker
@@ -1,0 +1,2 @@
+repository=https://github.com/JJDeakins1/superior-slayer-tracker.git
+commit=9fd572899373443b3075b8684c95ef6f17f0c72a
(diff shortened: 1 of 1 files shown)


coopermor:
Do you need FileReadWriter.java? If you can store the user's data using [ConfigManager](https://static.runelite.net/runelite-client/apidocs/net/runelite/client/config/ConfigManager.html) it would sync if the player is using RuneLite's cloud sync. Removing the file I/O would also make it quicker to merge.

JJDeakins1:
> Do you need FileReadWriter.java? If you can store the user's data using [ConfigManager](https://static.runelite.net/runelite-client/apidocs/net/runelite/client/config/ConfigManager.html) it would sync if the player is using RuneLite's cloud sync. Removing the file I/O would also make it quicker to merge.

Good idea I've removed all file I/O and moved storage to RS-profile ConfigManager keys as of 008c0e3

PR #14584, opened 2026-08-01
Update Kill Clog to 2.0.0

420kc:
Kill Clog 2.0.0

first-party collection log sync, opt-in and off by default:

* a new "Sync Collection Log to Killclog.com" setting publishes your locally tracked collection log and personal bests to your own killclog.com profile. nothing is sent until you turn it on, and opting out at killclog.com deletes everything stored
* player lookups now also read killclog.com alongside TempleOSRS and RuneProfile, so synced players' logs and personal bests show up in lookups and boss tooltips. looking up a never-synced name costs no extra requests
* the README and privacy copy name every endpoint the plugin talks to: TempleOSRS, RuneProfile, killclog.com, and the OSRS Wiki item mapping

also in this release:

* hiscore lookups now use the name-keyed json endpoint, so game updates that reshuffle the hiscores can no longer blank or misalign boss kcs
* personal bests now also come from the adventure log's counters scroll, so times you set before installing show up without having to re-kill anything. the log opens in either the original or the newer menu interface depending on your interface style setting, and both are read
* hovering a pet in the player summary shows its name and links its wiki page
* skill summaries follow the virtual levels plugin when you have it enabled, up to 126 and virtual totals
* new titan prestige for 200m xp in every skill
* mad angel collection log data maps correctly

note on the adventure log: runelite's own chat commands plugin parses personal bests from that same scroll, but it only watches the older menu interface, so for anyone using the newer interface style the parse is silently skipped and no adventure log personal bests are recorded at all. that is why this plugin reads it too rather than relying on the existing parse. i have a fix written for chat commands and will open it against runelite core separately.

### DIFF
--- a/plugins/kill-clog
+++ b/plugins/kill-clog
@@ -1,3 +1,3 @@
 repository=https://github.com/420kc/kill-clog-plugin.git
-commit=cb1efd9562d9051d38287f9e25f11355cb156f0e
+commit=a659b0ebbd87209a5e707b75b19d8fc03b4e3504
 warning=This plugin submits your IP address to a 3rd-party server not controlled or verified by Runelite developers. If you enable "Sync Collection Log to Killclog.com" (off by default), your collection log and personal bests are published to your killclog.com profile.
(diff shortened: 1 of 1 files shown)


420kc:
the chat commands fix mentioned above is now open: https://github.com/runelite/runelite/pull/20372

Alexsuperfly:
revert the warning text update

PR #14596, opened 2026-08-01
Add Bronzeman PVP TCG

0xG3r4rD:
Adds **Bronzeman PVP TCG**, a PvP-oriented bronzeman challenge mode built on a card collection.

- Equipping an item is blocked until you own its matching card, so gear is unlocked by pulling cards rather than by acquiring the item.
- Credits are earned only from PvP kills. A kill pays for one booster pack; an optional Hard mode pays 250 credits per 100k of loot value instead.
- A Defence level setting (1 / 20 / 30 / 40 / 50 / 60+) removes cards for gear you could not equip at that level, so the roll pool matches a pure or level-restricted build. Equip requirements are sourced from the OSRS Wiki.
- The card pool is curated down to ~1,190 equipable weapons, armour and jewellery.

**Attribution:** this is a fork of [OSRS TCG](https://github.com/Azderi/osrs-tcg) by Azderi (BSD 2-Clause; the copyright notice is retained in LICENSE), and the equip-restriction concept is ported from [Bronzeman TCG](https://github.com/Felmeme/bronzeman-tcg) by Felmeme. Both are credited in the repository README. The upstream OSRS TCG plugin is already on the hub; this fork runs alongside it — its package, config group, storage paths, party message types and chat commands are all namespaced separately so both can be enabled in the same client without conflicting.

Repository: https://github.com/0xG3r4rD/bronzeman-pvp-tcg

### DIFF
--- a/plugins/bronzeman-pvp-tcg
+++ b/plugins/bronzeman-pvp-tcg
@@ -1,2 +1,3 @@
 repository=https://github.com/0xG3r4rD/bronzeman-pvp-tcg.git
-commit=6e835041984c6ffd318cf34adee2fc07109e4cf4
+commit=146965399d11038b8bfef80f73db39e75753c25f
+warning=Card artwork is downloaded from the OSRS Wiki, which receives your IP address and the requested image URLs. If you configure a Discord webhook for pull alerts, card pull data is sent to Discord. No collection or account data is sent anywhere else.
--- a/plugins/bronzeman-pvp-tcg
+++ b/plugins/bronzeman-pvp-tcg
@@ -1,0 +1,2 @@
+repository=https://github.com/0xG3r4rD/bronzeman-pvp-tcg.git
+commit=d637d8c9e154fb36e16a5b0522c1b7993ed92b7b
--- a/plugins/bronzeman-pvp-tcg
+++ b/plugins/bronzeman-pvp-tcg
@@ -1,0 +1,2 @@
+repository=https://github.com/0xG3r4rD/bronzeman-pvp-tcg.git
+commit=6e835041984c6ffd318cf34adee2fc07109e4cf4
--- a/plugins/bronzeman-pvp-tcg
+++ b/plugins/bronzeman-pvp-tcg
@@ -1,3 +1,2 @@
 repository=https://github.com/0xG3r4rD/bronzeman-pvp-tcg.git
 commit=146965399d11038b8bfef80f73db39e75753c25f
-warning=Card artwork is downloaded from the OSRS Wiki, which receives your IP address and the requested image URLs. If you configure a Discord webhook for pull alerts, card pull data is sent to Discord. No collection or account data is sent anywhere else.


0xG3r4rD:
Adjusted certain config to be moved from the settings tab to the plugin menu itself.
image

plugins/bronzeman-pvp-tcg
```diff
@@ -0,0 +1,2 @@
+repository=https://github.com/0xG3r4rD/bronzeman-pvp-tcg.git
+commit=62a21975434e9e8f8a1126657b260ee32507596c
```
riktenx:
```suggestion
commit=62a21975434e9e8f8a1126657b260ee32507596c
warning=This plugin submits your IP address, and may submit various account data, to a 3rd-party server not controlled or verified by Runelite developers.
```

0xG3r4rD:
@riktenx thanks for the review. Rather than add the warning, I removed the feature it was about.

As of `1.17.0` the "Web album" collection sharing is gone entirely — the service, the snapshot builder, the stored share credentials, the config section and its API key field, and the sidebar live indicator (about 1200 lines). The plugin no longer contacts osrs-tcg.xyz at all, and no collection data, statistics, display name or account data is submitted anywhere.

I've updated the commit hash and added a warning covering what network activity actually remains:

- **OSRS Wiki** — card artwork is downloaded from the wiki, so it receives your IP address and the requested image URLs. This is not gated by a config option.
- **Discord webhook** — opt-in and empty by default; if a webhook URL is configured, card pull data is sent to Discord.

I based the wording on the existing `group-tcg` entry, which discloses the same wiki image fetching. Happy to adjust the text if you'd prefer something closer to your original, or to gate the wiki downloads behind a config option if that's preferable.

0xG3r4rD:
Gentle bump — no rush, just flagging this is still open.

Since the last review the third-party upload was removed outright in `1.17.0`, and the manifest now carries a narrowed `warning=` covering only what's left: OSRS Wiki card art, and the opt-in Discord webhook. `build` is green on the current commit.

@riktenx if the narrowed wording isn't what you'd prefer, I'm happy to take your original text as-is, or to gate the Wiki image downloads behind a config option — whichever makes this easiest to approve.

riktenx:
if youre only talking to the wiki and user urls now you dont need any warning

0xG3r4rD:
Warning line dropped — the manifest is back to just `repository` and `commit`. Thanks!

riktenx:
we would rather you not depend directly on the wiki for card art downloads

0xG3r4rD:
Updated to `1.18.0`, which removes the last external dependency: all 1,022 card images are now bundled in the jar instead of being downloaded from the OSRS Wiki at runtime.

- The only outbound request left in the plugin is the Discord pull webhook, which is opt-in and empty by default.
- Card art renders instantly on first run and works with no connection.
- The wiki no longer takes a request per card per player.

The jar goes from 2.9 MiB to 6.4 MiB as a result — under the 10 MiB limit and the 8 MiB warning threshold. Images are quantized to a 256-colour palette, which is visually identical on OSRS item renders and cuts the raw 15 MB of art down to 3.9 MB.

Happy to revert to the previous commit if you'd rather review the smaller jar first.

PR #14607, opened 2026-08-02
add bank-bridge

theshane0314:
Adds **Bank Bridge** — serves your bank, inventory, worn equipment and real levels to a gear-planning website over a **local-only WebSocket**. There is no server, no account, no token and no telemetry; nothing leaves the machine.

Source: https://github.com/theshane0314/bank-bridge-plugin

### What it does

Binds a WebSocket to `127.0.0.1`, first free port in `37767-37776` — the same range WikiSync uses, deliberately, so a page can scan one range and find whichever plugin is installed. A page connects and requests `GetBank`; it gets back item ids and quantities plus real skill levels. Browsers allow this from an `https://` page because loopback is a potentially trustworthy origin, which is the same mechanism the wiki's DPS calculator already relies on.

### Security

I've flagged the relevant points myself rather than leave them to be found:

- **Binds 127.0.0.1 explicitly**, not a wildcard bind plus a firewall rule.
- **A WebSocket is not subject to same-origin**, so every handshake's `Origin` is checked against an allowlist (`osrs.plaincandle.dev`, `localhost`, `127.0.0.1`, plus a user config field). Missing `Origin` is rejected; anything else is closed before a byte is sent.
- **Known limit, stated in the listing warning and the README:** a native program already running on the user's machine can forge `Origin`. It could equally read the RuneLite directory, so it isn't an escalation — but it is not claimed as a boundary against local software. The boundary is against websites.
- **It cannot observe a bank PIN.** A PIN is entered on a widget, and the plugin references no widget reads, no key or mouse listeners, no varbit/varclient reads, no chat, menu or script hooks. Its whole contact with the client is five calls: game state, local player, real skill levels, account hash (used only to name a local cache file), and the containers `ItemContainerChanged` hands it. A Gradle task `accountSafetyCheck` fails the build on any reference to those APIs and gates `build`, `check`, `run`, `runHarness` and `shadowJar`; it's tested by planting a deliberate violation and confirming the build breaks.
- No reflection, no native code. One runtime dependency, `Java-WebSocket`, with its transitive `slf4j-api` excluded so the client's own version is used.

### Threading

All game reads happen on the client thread inside `ItemContainerChanged` / `StatChanged` / `GameTick` and are cached as immutable lists behind volatile references. `WebSocketManager` holds no reference to `Client` or `ClientThread`, so a socket request is answered from cache and can never reach — let alone block — the client thread. Sends run on a daemon executor; the disk write is debounced and also off-thread.

### Testing

`./gradlew runHarness` boots the real `WebSocketManager` against a seeded store and drives it with a real WebSocket client — 22 checks covering origin accept/reject/missing, protocol, payload shape, error handling, config gating and port release. Nothing under test is mocked. Verified in-game against a live account, and port fallback confirmed live with WikiSync holding 37767.

### DIFF
--- a/plugins/bank-bridge
+++ b/plugins/bank-bridge
@@ -1,4 +1,4 @@
 repository=https://github.com/theshane0314/bank-bridge-plugin.git
-commit=b4bee8bec21df5988ee631cd0a3f902e747737fe
+commit=f3b078b27a40dbb11420a96ebcc0f8c1363f7fa0
 warning=This plugin opens a WebSocket server on 127.0.0.1 so a gear-planning website can read your bank, inventory, worn equipment and levels. Nothing is uploaded and there is no server or account involved, and only allowlisted sites can connect.<br>Be aware that another program already running on your own machine could forge the origin check and read the same data. It cannot see your bank PIN or anything you type.
 authors=theshane0314
--- a/plugins/bank-bridge
+++ b/plugins/bank-bridge
@@ -1,0 +1,4 @@
+repository=https://github.com/theshane0314/bank-bridge-plugin.git
+commit=f806042418cf1dd70ad1af0e93ab02e42e2fe38c
+warning=This plugin opens a WebSocket server on 127.0.0.1 so a gear-planning website can read your bank, inventory, worn equipment and levels. Nothing is uploaded and there is no server or account involved, and only allowlisted sites can connect.<br>Be aware that another program already running on your own machine could forge the origin check and read the same data. It cannot see your bank PIN or anything you type.
+authors=theshane0314
--- a/plugins/bank-bridge
+++ b/plugins/bank-bridge
@@ -1,4 +1,4 @@
 repository=https://github.com/theshane0314/bank-bridge-plugin.git
-commit=f806042418cf1dd70ad1af0e93ab02e42e2fe38c
+commit=b4bee8bec21df5988ee631cd0a3f902e747737fe
 warning=This plugin opens a WebSocket server on 127.0.0.1 so a gear-planning website can read your bank, inventory, worn equipment and levels. Nothing is uploaded and there is no server or account involved, and only allowlisted sites can connect.<br>Be aware that another program already running on your own machine could forge the origin check and read the same data. It cannot see your bank PIN or anything you type.
 authors=theshane0314


ldavid432:
Not necessarily a required change, but you're using the java websockets dependency which the auto reviewer doesn't have the ability to review. If you use OkHttp for your websocket needs instead, combined with setting your build key in runelite-plugin.properties to standard, it would mean your plugin will be reviewed at a **significantly** faster pace.

theshane0314:
Thanks for the pointer, that was useful. I've dropped the dependency and switched to `build=standard`.

One note in case it helps someone else: OkHttp couldn't stand in here, because Bank Bridge is a WebSocket *server* rather than a client. It binds a loopback listener and the browser connects in to it. OkHttp's WebSocket support is client-only (`OkHttpClient.newWebSocket`), with no accept path.

So instead of swapping libraries I removed the dependency entirely. The RFC 6455 handshake and frame coding are now written directly against `java.net.ServerSocket` in `ws/WSConnection.java` and `ws/WSWebsocketServer.java`. The plugin ships no third-party classes at all, so `build=standard` builds it cleanly.

Since that protocol code is now mine rather than a library's, I widened the test harness rather than assuming it works. It runs 32 checks, driven two ways: an independent client implementation (Java-WebSocket, test-only, never shipped) for the normal path including a payload past 64 KiB, and a raw byte-level client for the cases a library client won't produce: fragmented messages across continuation frames, unmasked client frames (refused with 1002, as the RFC requires of a server), an oversized declared length (refused with 1009), a plain HTTP GET (400 rather than an upgrade), and ping/pong. I mutation-tested the two riskiest checks to confirm they actually fail when the code is broken.

Also verified end to end: a real browser handshakes and completes a full request round trip, and the port fallback still works with a live conflicting listener (WikiSync held 37767, this took 37768).

One consequence worth flagging, since it's a reduction in what your CI sees. My `build.gradle` had an `accountSafetyCheck` task that fails the build on any reference to widgets, key or mouse listeners, varbits, chat, menu or script hooks, on the grounds that a bank PIN is entered on a widget and the plugin should have no code path that could observe one. `build=standard` replaces the build file, so that gate no longer runs on your side. It still runs on every local build, and I've added a `standardBuildCheck` task that compiles against your standard-build classpath so I catch dependency drift before it reaches you. Happy to move to whatever you'd prefer here.

Alexsuperfly:

theshane0314:

Removed — shutdown no longer uses any interrupt APIs. stop() closes the server socket instead of joining the accept thread, so the InterruptedException handling is gone entirely. Updated to f3b078b.

PR #14627, opened 2026-08-02
add pigeon

Brizz999:
Adds Pigeon, a profile-based Discord-compatible webhook notification plugin derived from Dink. Pigeon lets users create, clone, import, export, and enable multiple independently configured profiles, with per-profile primary webhook URLs and notifier-specific overrides.

The plugin uses the standard build type and has no additional runtime dependencies. Its test suite currently passes 447 tests across 39 suites.

### DIFF
--- a/plugins/pigeon
+++ b/plugins/pigeon
@@ -1,2 +1,2 @@
 repository=https://github.com/Brizz999/Pigeon.git
-commit=5ffe9cbb5834e218d6e7cba7dfb9c24ddd0ca6c7
+commit=a6f12fac3c1049a10e6b98a940bf7181c26f3d3e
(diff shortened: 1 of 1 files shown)


riktenx:
* use of runelite's DiscordUser stuff is not allowed

Brizz999:
Thanks for pointing this out. The Plugin Hub marker now references Pigeon commit `a6f12fa`, which removes all use of RuneLite's internal Discord user integration:

- Removed `DiscordService` and `DiscordUser` usage.
- Removed the `DiscordProfile` model and `discordUser` payload field.
- Removed the `sendDiscordUser` configuration option.
- Updated the associated tests and documentation.

Pigeon still delivers notifications through user-configured Discord-compatible webhook URLs, but it no longer accesses or transmits RuneLite Discord account/profile information.

Verification completed: all 447 tests pass, the RuneLite integration-snapshot build passes, and the Plugin Hub build is successful.

PR #14634, opened 2026-08-02
Add Brains Off Mixology Helper

giuqoob:
Adds Brains Off Mixology Helper, a visual guide for configurable,
station-ordered Mastering Mixology inventories.

The plugin displays complete manual mixing recipes, inventory batch
assignments, processing-station guidance, and a potion queue. Its
optional station guard only reorders the station's existing Check and
processing menu entries.

The plugin does not click, move the mouse, automate inputs, access
credentials, or communicate with external services. It uses the
standard Plugin Hub build and has no third-party runtime dependencies.


giuqoob:
Basically a plugin to help players do this spam tech method without paying too much attention https://www.youtube.com/watch?v=tfVtxUWhFJE
The player selects the inventory of pots to fill (or uses defaults) and the plugin guides the player one potion type at a time to fill in the inventory in the correct order so that the player can simply process 9 potions per station at once and then return them. The plugin keeps track of potions already created in previous runs.

PR #14639, opened 2026-08-02
Add The Anchor CC Plugin

LogicalSoIutions:


riktenx:
you can't get back random urls in api responses and call out to them, any urls in the plugin need to be put in by the user or hardcoded so we can see what they are

LogicalSoIutions:
Makes sense, resolved!

PR #14652, opened 2026-08-03
Create moonlight-antelope-hunting

Hampo:
Adds overlays to assist with moonlight antelope hunting


coopermor:
Your plugin should subscribe to ItemContainerChanged and count the logs in that event rather than in the render loop

Hampo:
> Your plugin should subscribe to ItemContainerChanged and count the logs in that event rather than in the render loop

Good shout, hadn't even though of that - somewhat new to plugin development and haven't quite got the nuances down. Sorted now.

PR #14665, opened 2026-08-03
Add Green Log

420kc:
Sidebar collection log with search, filters, and completion tracking.
Local-only and self-only: captures your own log from the in-game interface, tracks new unlocks from chat messages, and stores everything under .runelite on the player's machine. The plugin makes no network requests; drop-rate data is bundled at build time from the OSRS Wiki (CC BY-NC-SA, attribution kept with the dataset).
Repository: https://github.com/420kc/green-log

### DIFF
--- a/plugins/green-log
+++ b/plugins/green-log
@@ -1,0 +1,2 @@
+repository=https://github.com/420kc/green-log.git
+commit=66e7c971f8f475c0613720995a5423072cc79f91


420kc:
I originally wanted to get this out as a local only version of Kill Clog, but on further consideration, closing this to focus on Kill Clog's pipeline and keep the queue noise down. Cheers

PR #14668, opened 2026-08-03
Update object-marker-plus

Shackin:
Added ability to highlight an object's radius or clickbox by name or ID.


Alexsuperfly:
user supplied ID input is no longer being accepted for plugin usage

Shackin:
> user supplied ID input is no longer being accepted for plugin usage

Oh, sorry. Removed fully tagging by ID.

PR #14675, opened 2026-08-03
Updated inspect plugin to v0.2.3

DevOldSchool:
Fixes an issue with the player inspect menu option where it was treated as a standard menu option rather than a player menu option. This fix allows the player inspect menu option to be condensed with the Interface Styles plugin correctly.


DevOldSchool:
Requesting review please.

PR #14688, opened 2026-08-03
gustav-helper: bump commit (step number, in-plugin step reports)

dangraagu:
Bumps `plugins/gustav-helper` `commit=` to `858e76f` (from `d751143`).

- **Current step number** shown alongside the completed counter (`… — 116 / 960 done · Step 117`).
- **"Report wrong / missing info" button.** Opens a dialog asking what is wrong, shows the exact text that will be sent, and only sends when the user confirms. The report carries the guide, step id and number, step text, the tile the plugin pointed at and the NPC/object/item ids it highlighted — enough to fix a bad coordinate or highlight from the report alone.
  - **No account information** is included (no player name, account hash, or player position) — covered by a test.
  - Never automatic: the only caller is the confirmed dialog. Sent asynchronously off the client thread via RuneLite's injected `OkHttpClient` (no new dependency), rate-limited, and failures surface in a dialog rather than throwing.
  - Reports go to a forwarding endpoint the author controls, which holds the Discord webhook privately and rebuilds the payload (no mentions/embeds/attachments pass through). Users can point it at their own webhook via config.
- Ledger tab scrolls; item-overlay lookup is allocation-free on the per-frame path.

Tests: 79 Java + 28 + 27 Python, all green.

### DIFF
--- a/plugins/gustav-helper
+++ b/plugins/gustav-helper
@@ -1,2 +1,3 @@
 repository=https://github.com/dangraagu/Gustav-s-helper.git
-commit=1489fb47693e8eb4f10cbb4542899cd10ccaf08b
+commit=088a9a99d0d87666220f22b643e0ea3440a7aee4
+warning=This plugin submits your IP address, and may submit various account data, to a 3rd-party server not controlled or verified by Runelite developers.
--- a/plugins/gustav-helper
+++ b/plugins/gustav-helper
@@ -1,2 +1,3 @@
 repository=https://github.com/dangraagu/Gustav-s-helper.git
 commit=1489fb47693e8eb4f10cbb4542899cd10ccaf08b
+warning=This plugin submits your IP address, and may submit various account data, to a 3rd-party server not controlled or verified by Runelite developers.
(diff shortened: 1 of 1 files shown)


plugins/gustav-helper
```diff
@@ -1,2 +1,2 @@
 repository=https://github.com/dangraagu/Gustav-s-helper.git
-commit=d7511433cbd0efb92bf64d207ffd78ee5ca6415f
+commit=1489fb47693e8eb4f10cbb4542899cd10ccaf08b
```
riktenx:
```suggestion
commit=1489fb47693e8eb4f10cbb4542899cd10ccaf08b
warning=This plugin submits your IP address, and may submit various account data, to a 3rd-party server not controlled or verified by Runelite developers.
```

plugins/gustav-helper
```diff
@@ -1,2 +1,3 @@
...
+warning=This plugin submits your IP address, and may submit various account data, to a 3rd-party server not controlled or verified by Runelite developers.
```
riktenx:
```suggestion
```

riktenx:
* you MUST have a @ConfigItem boolean, that defaults to false, that entirely disables the networking calls, its warning MUST be set to exactly "This feature submits your IP address and various account data to a 3rd-party server not controlled or verified by Runelite developers."
* all networking calls need to go into dedicated methods, and at the top of every method where an opt-in networking call is made you MUST check the boolean config and either return early or throw an exception. the body of the method MUST be only this check and the network call

dangraagu:
Thanks for the review — applied. Step reports are now fully opt-in: a default-false `enableStepReports` ConfigItem with the specified warning text gates the single dedicated networking method (config check + call only), and the manifest warning line is added as suggested. Pin bumped to 088a9a9.

PR #14690, opened 2026-08-03
Update anvil

AhmedFathy2001:
Bumps anvil to a417480. Adds the clan's weekly (SotW/BotW) competitions and upcoming events to the side panel, and an in-game 'active missions' strip (with a live countdown and per-mission grow/decay values) for mid-event mission drops on a bingo board. Also fixes the side panel sitting empty for ~15s after signing in, reports a sign-in whose token RuneLite failed to persist instead of silently looking unsigned-in, and surfaces the reason when a cross-clan "share my RSN" is refused rather than rendering a refusal as success.


Alexsuperfly:

AhmedFathy2001:
> LinkBrowser::open is a restricted api, remove its use

Removed.

PR #14693, opened 2026-08-03
Update Gem Sack plugin to 1.3

serenibyss:
- Fix empty to bank option on Gem Bags, as the option name was renamed from "Empty" to "Empty-to-bank"
- Add compatibility with the bank's "Empty containers" button


ldavid432:
JFYI `Empty-to-bank` is from item charges improved. The action was not actually changed (Your code seems to handle it fine just wanted to mention that)

PR #14694, opened 2026-08-03
osparty to v1.0.51

iodrareg:
- cleanup V1 of the party system, now all dead code is no longer in the repo
- fixes bug with party creation and joining at the same moment creating a desync in the applicant's party frame
- fixes the rune pouch not being updated on filling it with different runes
- role layout fix
- fixes partyframe memberId
- Added account authentication first steps
- Add playerlookup
- Fix stale timer when some leaves the party
- Added Castlewars, GOTR and Wintertodt
- Disable ready check in raid
- Fixed ping related issues

### DIFF
--- a/plugins/osparty
+++ b/plugins/osparty
@@ -1,4 +1,4 @@
 repository=https://github.com/osparty/osparty.git
-commit=cc3a40411c1062f6ae23c3bf8b1edbc6669c3cca
+commit=2756a5751854bab678de974f3908cde33e9a9cac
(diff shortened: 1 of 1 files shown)


coopermor:
Reflection is not allowed

riktenx:
you can't send the hostnames of peoples' computers to your site thing

iodrareg:
> you can't send the hostnames of peoples' computers to your site thing

Updated

PR #14709, opened 2026-08-04
Update InventorySetups to v1.25.0

dillydill123:
- Improve GE Item Search
- Add Plugin Message API
- Opacity (Alpha) now respected in color picker for highlight color.

### DIFF
--- a/plugins/inventory-setups
+++ b/plugins/inventory-setups
@@ -1,2 +1,2 @@
 repository=https://github.com/dillydill123/inventory-setups.git
-commit=afd6e45adb24a90d3411c9760b84d2b43bfe7489
+commit=6f9678715ef3f4c5293480e3926a51c8418e4c44


coopermor:
Is invsetups_version.txt the only reason you need build=gradle? If so, can you switch to build=standard and use ExternalPluginManager.getDisplayData() to get your plugin's version at runtime? See an example [here](https://github.com/wise-old-man/wiseoldman-runelite-plugin/blob/af5fa37ee9369eed3536e0940d9239e31c23272a/src/main/java/net/wiseoldman/WomUtilsPlugin.java#L1377-L1387).

dillydill123:
@coopermor I think so, but I'm not sure if there is any other reason I need it. Would need to investigate and test more. Can I defer this change to the next update? Thanks for making me aware of it.

coopermor:
If you're using gradle build then you'll need to wait for a human review.

dillydill123:
Sounds good. I'm ok with that for now to get the features out without more delay. But will definitely check that out in the next update. Thanks!

Nightfirecat:
I'll echo what was said earlier about recommending `build=standard` if possible if you want more expedient merges in the future, but this all looks good for now.

PR #14717, opened 2026-08-04
Update repository URL for faux-bingo plugin

ThatOhio:
Was asked to update the repo url before I update the commit hash in a previous PR.

### DIFF
--- a/plugins/faux-bingo
+++ b/plugins/faux-bingo
@@ -1,2 +1,2 @@
-repository=https://github.com/ThatOhio/faux-bingo.git
+repository=https://github.com/faux-bingo/faux-bingo-runelite-plugin.git
 commit=012f37352906f81f4a5e5e3debf9dae3e2f73849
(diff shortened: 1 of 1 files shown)


ThatOhio:
I was asked to update the plugin url before changing any commit hash, however the build for this fails because I need to update the runelite-plugin.properties file to include the new build property. 

I am now uncertain how I should proceed, as to fix that I would need to also update the commit hash.

pajlada:
@ThatOhio Make an as-minimal-as-possible update to make your plugin buildable, and I will manually review the repo url change + code change.

In my mind, the "code change" should be a single line change to `runelite-plugin.properties` adding `build=gradle` or `build=standard`.

ThatOhio:
Understood! I thought that would be the case but wanted to confirm, thanks, I'll do that right now

pajlada:
@ThatOhio You should strongly consider switching this to `build=standard` going forward to make reviews go quicker. If you need the version in the network request, you can just specify the version in one of your java files too.

ThatOhio:
Yea I had that thought when doing the previous PR, I'm going to make some changes to strip the test dependencies and add a variable based version before opening the PR for the actual updates so I can use standard, thanks for the help!

PR #14728, opened 2026-08-04
waypoint-sync: multiple characters, sea charting, ironman detection

Xiodrade:
### What changed

  - **Multiple characters on one account.** The plugin now sends RuneLite's account hash
    alongside the display name, so a main, an ironman and a pure can be tracked separately
    under a single key. Previously a second character's sync overwrote the first.
  - **Sea charting.** Reads the Sailing charting completion varbits, a category nothing else
    covered.
  - **Ironman detection.** Sends the account-type varbit so money-making suggestions can drop
    Grand Exchange methods for ironman accounts.
  - **Failed syncs explain themselves.** The panel showed only an HTTP status code; it now
    shows the reason the server returned.
  - **Cheaper interface polling.** Collection-log and combat-achievement capture is now driven
    by `WidgetLoaded` instead of probing up to 80 widget children per group every other tick,
    with a slow poll retained as a safety net. No change to what is captured.

### DIFF
--- a/plugins/waypoint-sync
+++ b/plugins/waypoint-sync
@@ -1,3 +1,3 @@
 repository=https://github.com/Xiodrade/waypoint-sync.git
-commit=4f83e84811a9acbe74cbb2763628a6f931ebbc72
+commit=ccee06fd0dacf801262f0900f0a7aa5b059a73aa
 warning=This plugin submits your IP address, display name, quests, diaries, and combat achievements to a third party server not controlled or verified by Runelite developers.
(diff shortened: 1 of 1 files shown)


Xiodrade:
Pushed a follow-up: syncs triggered by in-game captures are now coalesced into one request rather than one per event, and a rejected sync is no longer treated as delivered. Caught while testing — a Sailing charting session generated enough requests to trip our own server-side rate limit during normal play.

Xiodrade:
3d10d6f - Reverted an edit to warning=;

a3f0995 -
  **Account hash** — removed. `client.getAccountHash()` is no longer read or sent. Characters on
  one account are told apart by display name alone. The only thing this cost is rename resilience,
  which is now handled on the site rather than by sending a permanent identifier.

  **Endpoint override** — removed. `baseUrl()` returned
  `System.getProperty("waypoint.url", DEFAULT_URL)`, a leftover from local development. The
  destination is now fixed in source and cannot be changed at runtime.

  Also included, found while testing: the plugin treated every `LOADING` game state as a logout and
  re-synced on the following `LOGGED_IN`. That fires on any region change — teleports, staircases,
  and continuously while Sailing — so a single session produced far more requests than intended.
  It now re-arms only on `LOGIN_SCREEN`, `HOPPING` or `CONNECTION_LOST`.

PR #14747, opened 2026-08-05
Update Varrock Bank Ruby Rings

andrew-friedman-phd:
Adds a random world-hop utility (with per-world cooldown tracking), session profit/hr and XP/hr estimates, Low Level Alchemy support, and a Crack the Clue III emote helper that highlights the next emote in the Emotes tab.

Repository: https://github.com/andrew-friedman-phd/varrock-bank-ruby-rings


coopermor:
I would recommend using the Random World Hopper plugin if you can. Using client.hopToWorld() will make your plugin no longer eligible for bot review and update could take several days or weeks to be approved depending on reviewer availability.

andrew-friedman-phd:
Appreciate the input; looking at the plugin, it looks like it uses the same function I do. I'd prefer not to require an additional dependency, so I am OK to wait for the full review if I need to.

PR #14810, opened 2026-08-07
update dvd-bounce

vividflash:
1.5 makes broken or oversized custom images fall back to the bundled logo instead of failing silently, and stops the overlay from interfering with overlay management. Update max gif to 30

Update Gradle to 8.1, still build=standard


vividflash:
Thank you

PR #14820, opened 2026-08-07
Add clan-turf

Zezi-cmd:
Clan Turf - claim Grand Exchange tiles for your clan by walking on them. Tiles
are colored per clan and ownership is tracked per world; walking onto a rival's
tile takes it over. A sync server makes rival clans' turf visible in real time,
and all turf resets daily.

Data disclosure: with the sync server enabled (default), the plugin sends the
player's clan name, the GE region coordinates of claimed tiles, and the current
world number to the server, so other clans' turf is visible to everyone. No
account details or personal information are sent. Users can turn the sync server
off in config to run fully local, in which case nothing is sent anywhere.

No automation: the plugin only reads the player's position and draws overlays.
The clan-chat commands only read clan messages to recognize !defend/!invade and
rewrite them locally. It never moves the character, sends chat, or performs any
automated game action.


riktenx:
use of client.hopToWorld will mean your plugin requires full manual review, which nobody is really doing right now so it will take an indefinite amount of time for your plugin to be reviewed

if you would prefer your plugin to be autoreviewed by the bot you will need to remove it

Zezi-cmd:
Thanks for the heads-up @riktenx - I've removed client.hopToWorld (the Invade/Defend world-hop) in the latest commit, so this should be eligible for auto-review now. Everything else stays local

Zezi-cmd:
Label's off now, thanks @riktenx . Does the reviewbot re check automatically from here, or is there something still needed from me to get it merged? Sorry, still figuring out this whole process.

plugins/clan-turf
```diff
@@ -0,0 +1,2 @@
+repository=https://github.com/Zezi-cmd/clan-turf.git
+commit=791d903fe1378f07ab0ae4eb6b5c49f8ea1da6da
```
riktenx, line 2:
```suggestion
commit=791d903fe1378f07ab0ae4eb6b5c49f8ea1da6da
warning=This plugin submits your IP address, and may submit various account data, to a 3rd-party server not controlled or verified by Runelite developers.
```

PR #14826, opened 2026-08-07
Update slayerdex

PokeyRS:
slayerdex: update to 1f8dd69


PokeyRS:
Added webhooks, they're off by default and locked behind a toggle box that issues a warning about the feature.

PR #14839, opened 2026-08-07
Update dopamine-simulator to 0.3.0

Zoinkwiz:
Dopamine 2 : The Search for more Dopamine.


coopermor:
This plugin now exceeds the token limit to be auto-reviewed. Is it possible to move the card data from the java files to a json file to reduce the token count?

PR #14847, opened 2026-08-08
Add ToA Wardens Tile Memory

dacluna:
Adds a ToA Wardens P3 memory aid that tracks only the player's own manual movement clicks.

The plugin watches three configured floor tiles and keeps the most recently clicked one highlighted until another of the three is clicked. It does not detect Warden attacks, animations, projectiles, timing, or determine a safe tile automatically.

### DIFF
--- a/plugins/toa-wardens-tile-memory
+++ b/plugins/toa-wardens-tile-memory
@@ -1,0 +1,2 @@
+repository=https://github.com/dacluna/toa-wardens-tile-memory.git
+commit=daa564a38065f8d57a58f57cdd6d0261738bd924


coopermor:
This is not allowed

dacluna:
Could you clarify which part of the guidelines this falls under? The plugin does not detect or infer any boss mechanic or determine where the player should stand. It only remembers which of three fixed, user-configured tiles the player manually clicked most recently. I’d like to understand the distinction for future reference.

coopermor:

riktenx:
and

dacluna:
Thanks. To clarify, the player is manually selecting which tile to highlight by left-clicking it. The plugin only applies that manual marking action to those three predefined tiles — effectively the same as manually marking a tile, just using left-click instead of right-click.

riktenx:
yes we understand. it's still not allowed, per the above

raiyni:
This is not the same as manually marking the tiles because the last tile is relevant for insanity p3.

PR #14851, opened 2026-08-08
Add chatbox escape blocker

ellatonin:
updated interfaces


coopermor:
Reflection is not allowed on the hub

PR #14870, opened 2026-08-08
Overhaul update to Faux Bingo plugin

ThatOhio:
We have an event coming up at the end of September and are overhauling how we run our bingo events.

In summary of the changes, we have drastically "dumbed down" the responsibility of this plugin in favor of streaming data to our API, where we can make similar determinations (if an item is bingo related, etc) without requiring updates to this plugin, or any form of client side logic at all.

There has been a lot of work done as well to better identify drops from NPCs, players, and activities with this update. Largely building off of changes other loot plugins have made to better support all the various ways loot is detected in this game.

The plugin still functions without the API enabled, just with severely limited functionality (primarily related to clan overlay and Wise old Man updating).

### DIFF
--- a/plugins/faux-bingo
+++ b/plugins/faux-bingo
@@ -1,3 +1,3 @@
 repository=https://github.com/faux-bingo/faux-bingo-runelite-plugin.git
-commit=66e2414b98fbe91118e6847d050490578199a9d6
+commit=88ddc978759290098e23d85b024d8d43ce2315ba
 authors=ThatOhio
--- a/plugins/faux-bingo
+++ b/plugins/faux-bingo
@@ -1,3 +1,3 @@
 repository=https://github.com/faux-bingo/faux-bingo-runelite-plugin.git
-commit=88ddc978759290098e23d85b024d8d43ce2315ba
+commit=027faa042942720d676c31421f90a3599642bfc9
 authors=ThatOhio


riktenx:
you can't get some url from an api response and call out to it. all urls the plugin talks to need to either come from the user or be hardcoded

ThatOhio:
> you can't get some url from an api response and call out to it. all urls the plugin talks to need to either come from the user or be hardcoded

Those URLs are always a discord webhook. We could have the players put them in, but we wanted to simplify things. 

If I added some checks to ensure the URL that was returned was on the discord domain, would that be sufficient for us to pull it from the api like that? I can make it match EXACTLY a discord webhook url even, because that's the only one we intend to work with.

ThatOhio:
> you can't get some url from an api response and call out to it. all urls the plugin talks to need to either come from the user or be hardcoded

Actually I retract what I just said, we were previously setting discord webhook urls on the client side but have since moved to simply having our website send those webhooks, so while the team model has a field for webhook, this plugin doesnt actually use it anymore. Can you be more specific on where you were seeing this? 

Only place I could think was the team icons, which are returned as an icon and displayed as an image (so that is a url coming back i suppose), we could potentially shift these to a hard shape like {apiBaseUrl}/v1/teams/{teamId}/icon but that seems a bit excessive in this case (the old version of this plugin already did the team icon stuff as well, thats not technically a new feature we just tweaked where it came from)

Alexsuperfly:
correct its the team icons

ThatOhio:
> correct its the team icons

Makes sense, I'll whip something up to serve the icons from the same url and update the pr soon.

ThatOhio:
@riktenx @Alexsuperfly Updated how we're handling the team icons so that everything is resolved to the same downstream url. 

Theres also a guard to not allow for redirects.

PR #14876, opened 2026-08-09
Update custom-emoji to 1.1.0

Fiffers:
New feature:
* Seasonal emojis for Halloween and Christmas, peep the readme for instructions

Some fixes:
* Fix bug causing incorrect emoji positions when emojis are proceeded by a `>`, `<`, or `@` character.
* Fix bug causing overhead emojis to not be rendered when the player who wrote the message is not on the users friends list
* Fix bug causing the suggestions to not work when the message starts with `/` or `//` which doesnt have a space between those slashes and the emoji [#60](https://github.com/LouisGameDev/runelite-custom-emoji/issues/60)


Alexsuperfly:

Fiffers:
All instances of LinkBrowser::open removed 👍

PR #14887, opened 2026-08-09
Add Tick Cursor plugin

Aschmies:
Cycles the  mouse cursor colour on each game tick. Supports 1-6 configurable tick slots, each with a user-selectable colour and opacity.

### DIFF
--- a/plugins/tickcursor
+++ b/plugins/tickcursor
@@ -1,2 +1,2 @@
 repository=https://github.com/Aschmies/AschysQOL.git
-commit=7ae072cdc214b33f5aaa1aa1415ff1b111888230
+commit=842d0143374789b3ac354f50aaa64568c578fb90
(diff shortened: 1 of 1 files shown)


riktenx:

Aschmies:
> * use of getenv is not allowed; use a jvm sysprop instead

I’ve removed the getenv usage and switched the cursor loading to a bundled PNG resource instead of reading from system files. That keeps the behavior the same on Windows while also making it work on macOS, I haven't tested on macOS just yet though, I will do that later in the week.

Aschmies:
Would someone please be able to add this description for me as now that its merged I'm unsure if this is something I can do. 

Cycles the mouse cursor colour on each game tick. Supports 1-6 configurable tick slots, each with a user-selectable colour and opacity.

ldavid432:
> Would someone please be able to add this description for me as now that its merged I'm unsure if this is something I can do.
> 
> Cycles the mouse cursor colour on each game tick. Supports 1-6 configurable tick slots, each with a user-selectable colour and opacity.

You need to add the description in your code and create a new PR with that commit hash

PR #14888, opened 2026-08-09
Add Zero XP Drops

Webben91:
Shows a movable MISS or 0 when an attack gives no combat XP.


Webben91:
I made this because I wanted misses to show like normal XP drops without having to set up weapon or animation IDs. It also works alongside the existing XP drop plugins.

PR #14892, opened 2026-08-09
create Station's Creature Collector

StationEarthxo:
## Summary

Adds Station's Creature Collector, a local-only cosmetic creature collection plugin.

- Earn capture balls through XP and credited boss kills
- Catch eligible non-humanoid creatures with a client-side animation
- Browse a regional collection catalogue with rendered portraits and silhouettes
- Call, resize, dismiss, and interact with caught cosmetic companions
- Protect official OSRS pets and their source bosses from catchability

All companion models, capture effects, rewards, and transformations are client-side only. The plugin does not automate actions or send interactions to the game server.

## Verification

- Clean Gradle test and jar build passed against latest RuneLite
- Plugin started in the development client without plugin-specific errors
- Plugin Hub icon is 48x48

### DIFF
--- a/plugins/stations-creature-collector
+++ b/plugins/stations-creature-collector
@@ -1,0 +1,2 @@
+repository=https://github.com/StationEarthxo/Stations-Creature-Collector.git
+commit=31d3c62c99ff81114c818bee393fe28cfa6c0548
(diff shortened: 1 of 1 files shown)


coopermor:
Reflection is not allowed

StationEarthxo:
Removed all production reflection in d49365a44443eadf1f59557fd4625a116c5858a8. The reflective access to the private boss-timer class has been replaced by an explicit boss-name registry with regression tests. A clean test/jar build passes, and src/main now contains no java.lang.reflect, Class.forName, getDeclaredMethod, setAccessible, or reflective invoke usage.

PR #14898, opened 2026-08-09
update tickcursor commit

Aschmies:
Updated.
Now contains a description + readme as well


ldavid432:
You need to add the description in runelite-plugin.properties

Also you'll want to remove the `disabled=` here to re-enable your plugin

Aschmies:
> You need to add the description in runelite-plugin.properties
> 
> Also you'll want to remove the `disabled=` here to re-enable your plugin

I think I finally sorted everything, hopefully. 
Thanks for your help :)

PR #14903, opened 2026-08-09
Add Gachaman Plugin

BraveH:
# Gachaman

**An RNG-governed challenge gamemode for Old School RuneScape, played entirely inside RuneLite.**

> Fate decides what you may train. Cards decide what you may wear. Chests decide everything else.

You don't choose your combat style — a wheel chooses it for you, and you train that until the
wheel changes its mind. You don't equip a sword because you found one — you equip it because a
card came out of a chest with its name on it, and you slotted that card into your loadout.
Everything in between is paid for in **Gacha Coins**, earned one kill at a time.

It's a gacha game bolted onto an MMO, and it turns "I got a drop" into "I got *permission*".

Inspired by OSRS TCG

### DIFF
--- a/plugins/gachaman
+++ b/plugins/gachaman
@@ -1,2 +1,2 @@
 repository=https://github.com/BraveH/Gachaman.git
-commit=432f16ca600b1d98f116b55a3939ae18ec3ca941
+commit=22e38fa0948f5e5679f23de49d76ef7c59c7eda8
--- a/plugins/gachaman
+++ b/plugins/gachaman
@@ -1,2 +1,2 @@
 repository=https://github.com/BraveH/Gachaman.git
-commit=5b886dbb75d89616a7b42207b532a2e7e2aa481b
+commit=761d83ed3ab2e966379edf43f02d283126fe0d46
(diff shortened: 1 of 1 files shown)


coopermor:
Remove the class files from your repo

coopermor:
Reflection is not allowed

BraveH:
Done, sorry forgot to remove them

BraveH:
I was using reflection to identify the actual state of the quest plugin quests so I can unlock specific monsters based on the step to unblock users during quests.

Removed the reflection calls/code

BraveH:
Added non-reflective code to handle specific quest progress so that npc’s are unlocked only for specific durations in quest progresses not all the time.

Alexsuperfly:
`too many tokens: 249632` you have to be under 200k for our review bot to run your plugin, otherwise it requires manual review, which at enough lines to be that many tokens no one will be wanting to do

BraveH:
Should now be less than 200k. Just a question though; I removed some features to get to this token count… if I end up creating an update after this plugin goes live… does the bot then work on the updated diff or on the entire plugin again? Ie… will I be able to add back these features or am I hard stuck by the code review bot?

coopermor:
The entire plugin is scanned.

BraveH:
Does this state mean that the automated bot check didn’t run? @coopermor

PR #14912, opened 2026-08-10
Create grimy-herb-stopwatch

ericgeoff:


ericgeoff:
Adds Grimy Herb Stopwatch, a Herblore training plugin that times manual grimy-herb cleaning inventories and tracks total herbs cleaned, average cleaning speed, fastest cleaning speed, and cleaning accuracy

PR #14917, opened 2026-08-10
Add Boss of the Week

marcushill13:
Adds **Boss of the Week**  a plugin for running clan competitions on a boss over a set period.

Someone creates a challenge (a boss, a start and end time, and what kills and drops are worth) and shares a code. Everyone else joins with it. The plugin counts kills and qualifying drops on its own and keeps a shared leaderboard, replacing the screenshots-into-Discord-and-a-spreadsheet routine clans currently use.

### External communication

The plugin talks to a small Cloudflare Worker, because a shared leaderboard cannot exist client-side. **The full source of that service is in `backend/` in the same repository**, and a clan can deploy their own and change the address in settings.

What is sent, and only for challenges the player has explicitly joined:

- their RuneScape name, for the leaderboard
- kills of that challenge's boss, and any drop the challenge counts
- a downscaled screenshot of each scoring drop, if the player leaves that setting on

Nothing is sent before a challenge is joined, and nothing about any other monster.

### Screenshots

Clans already verify drops by posting screenshots to Discord; this gathers the same evidence automatically. The picture is saved to the player's own screenshots folder, and a ~40KB copy is sent to the challenge's creator so they can verify it within the plugin. Only the creator can retrieve them, other participants cannot. The full size original never leaves the player's machine.

Taking screenshots and sharing them are two separate settings, both documented in the README along with exactly what is sent and who can see it.

### Notes

- Points are calculated server-side from the challenge's configuration, so the client cannot decide what a drop is worth.
- The README is explicit that this is trust-based rather than proof, including that RuneLite's loot detection cannot distinguish a boss drop from an item dropped at that moment.
- Boss drop tables are generated from the OSRS Wiki at build time and bundled (CC BY-NC-SA, credited); the plugin makes no wiki calls at runtime.

### DIFF
--- a/plugins/botw
+++ b/plugins/botw
@@ -1,3 +1,3 @@
 repository=https://github.com/marcushill13/botw.git
-commit=c09f2d501a4fdcd5edf88c74189c84314cc67640
+commit=2b79b8b8268c15959d6852081a02083bd087a0e7
 warning=This plugin submits your IP address, and may submit various account data, to a 3rd-party server not controlled or verified by Runelite developers.
--- a/plugins/botw
+++ b/plugins/botw
@@ -1,2 +1,3 @@
 repository=https://github.com/marcushill13/botw.git
 commit=c09f2d501a4fdcd5edf88c74189c84314cc67640
+warning=This plugin submits your IP address, and may submit various account data, to a 3rd-party server not controlled or verified by Runelite developers.
--- a/plugins/botw
+++ b/plugins/botw
@@ -1,0 +1,2 @@
+repository=https://github.com/marcushill13/botw.git
+commit=c09f2d501a4fdcd5edf88c74189c84314cc67640


Alexsuperfly:
Desktop::open is a restricted api, remove its use

marcushill13:
> Desktop::open is a restricted api, remove its use

Removed. It was used in one place: opening the folder after an evidence
export was written. Replaced with LinkBrowser.open, which is what the
client's own Screenshot plugin uses to open that same directory.
Fixed in 4e84a7b. :)
Hope thats ok

marcushill13:
Saw on someone elses PR that LinkBrowser was also restricted,

I've dropped opening the folder entirely rather than swapping in
LinkBrowser::open.
The export now shows the path and copies it to the clipboard to paste into a file browser.
Fixed in c09f2d5.

plugins/botw
```diff
@@ -0,0 +1,2 @@
+repository=https://github.com/marcushill13/botw.git
+commit=c09f2d501a4fdcd5edf88c74189c84314cc67640
```
riktenx:
```suggestion
commit=c09f2d501a4fdcd5edf88c74189c84314cc67640
warning=This plugin submits your IP address, and may submit various account data, to a 3rd-party server not controlled or verified by Runelite developers.
```

riktenx:
file io has to happen within a plugin-specific subdir under .runelite

marcushill13:
Moved, everything now writes under .runelite/botw (screenshots/ and exports/),
instead of the client's .runelite/screenshots. Paths are all in one class now.

Fixed in 2b79b8b.

Hope this is what you were after

PR #14922, opened 2026-08-10
Add ground-item-colour-groups

ellatonin:
Simple plugin that adds a side panel which groups items by custom colour from the default Ground Item plugin.

image
image


Alexsuperfly:
remove .class files
use of thread interrupt is not allowed

PR #14923, opened 2026-08-10
Add ground-marker-variables

Limy-Monkey:
Extends current Ground Markers plugin functionality with variable support.

Variables include `{metronome4}` as a 4 tick metronome, `{lvl_attack}` to display your current attack level, and more. Full information about which variables are supported are in the README of the repository.

This plugin does force-remove the Ground Marker overlays while the plugin is active. Re-adds Ground Marker overlays when the plugin is turned off. I would technically prefer if I could simply overwrite the text inside the existing Ground Marker overlay, but unfortunately this appears to be Java-private. For this reason, this plugin reads the data from Ground Markers' configManager so it can accurately recreate the tiles it is force-removing. All data is stored on Ground Markers side.

As a side note: I made this plugin in one afternoon. I have no major interest in maintaining this plugin long term -- I have another larger plugin project in active development. If anyone happens to be interested in what this plugin does and would like to take over this codebase, please reach out to me.

### DIFF
--- a/plugins/ground-marker-variables
+++ b/plugins/ground-marker-variables
@@ -1,0 +1,2 @@
+repository=https://github.com/Limy-Monkey/Ground-Marker-Variables.git
+commit=dea84e0b5d94e29ee79f3a85b367bfc78f0ec1e6


ldavid432:
The user configurable varbit based functionality is not allowed. From our rejected features list

> ID based plugins: Plugins that use player provided IDs for the entirety of their functionality can cause moderation issues and outcomes that break Jagex's plugin rules. Due to this fact we will not be accepting any new ID based plugins.

Limy-Monkey:
Thank you @ldavid432. I was not aware displaying varbits fell into that rule. I appreciate the feedback. I've removed the `{varbit<N>}` variable functionality.

Limy-Monkey:
Hi @raiyni - I see you added the Waiting for Author label. Can I ask why? I've removed the `<varbit>` functionality requested in the previous comment.

raiyni:
I am struggling to see a way in which this metronome configuration isn't just a use for side stepping current restrictions on automatic metronomes. 

eg, what is your example used for?

```
{metronome16 > 8 ? {metronome4} : }

    16 tick cycle, showing a 4-tick metronome for 8 ticks then hiding it for 8 ticks.
 ```

Limy-Monkey:
That example was a random simple example to explain the functionality.

Some examples I've thought of include:
- Showing a 7-tick metronome at Yama, but only show 3 -> 2 -> 1, instead of all 7 numbers
  - `{metronome7 <= 3 ? {metronome7} : }`
- Showing four separate 4-tick metronomes at Olm
  - `{metronome16 <= 4 ? {metronome4} : }`
  - `{metronome16 > 4 && metronome16 <= 8 ? {metronome4} : }`
  - two more separate ground-marker metronomes for ticks 9-12 and 13-16

Still, this plugin has no functionality currently to reset the current tick counter. It always simply uses
`(client.getTick() / M) % N`, with no hotkey or offset for the user to reset the tick counter.

I agree, there is added functionality here that current metronome plugins do not have, but I wouldn't think this constitutes a breach of Jagex's rules unless there are specific instances it could be abused that I'm not thinking of.

raiyni:
Yeah, counting boss mechanics like that wouldn't be allowed

Limy-Monkey:
My understanding of the rules is that any automatic tracking of boss attacks or when they will attack next are not allowed, but metronomes are allowed, just can't reset on boss attack automatically.

I am allowed to put a 7 tick metronome over my head at yama. Is there a reason that 7 tick metronome over my head couldn't just hide numbers 4-7?

I agree, I don't want this to be abused, I just don't believe there's anything truly different about this functionality.

Limy-Monkey:
When looking into this plugin, I found the Tile Marker Metronome plugin in the hub already which does the same thing, but with colors instead of numbers, and the ability to set a hotkey to reset the tick counter. That plugin is just a bit too buggy for me right now, and I prefer numbers over colors personally.

Limy-Monkey:
I apologize for pushing on this, but I honestly believe this one is within Jagex rules, unless there's something I'm missing. To demonstrate, I took a short screen capture. Top is Tile Marker Metronome (currently on plugin hub, with configuration shown in sidebar). Bottom is Ground Marker Variables plugin (this PR), set to `{metronome7 <= 3 ? {metronome7} : }` I'm happy to remove though if this is continued to be deemed as against the rules.


https://github.com/user-attachments/assets/7a04b8f2-80a3-4fdb-83b3-0417b328f368

ldavid432:
I think you already clearly covered how this is against the rules, it can be used to time boss mechanics. 

And I think if anything, that video is more detrimental to the tile marker metronome than it is helpful to your plugin.

Most metronome plugins just have a single not dynamic metronome which isn't really a problem because you still have to do some manual interpretation of the numbers, with yours you've showed you can essentially hide the metronome except for on the ticks on which you want to indicate a mechanic is happening.

Limy-Monkey:
That's a fair interpretation. Is the issue here centered around using the metronome inside the conditional? If so, I can remove this. Even having a metronome on a tile alone without conditionals is still an improvement for me over what exists in RuneLite today.

raiyni:
Yes

Limy-Monkey:
Thank you both for looking at my plugin so far. I appreciate y'all's reviews.

I've updated the plugin to explicitly remove the MetronomeLabelVariable class from being used inside conditionals.

Limy-Monkey:
I apologize for updating now mid-review. It was brought to my attention that my fix above was not fully complete, and while I removed `{metronomeN ? ...}`, I hadn't removed `{hasThralls ? {metronomeN} : }`.

I've updated my code to fix this. Any `{metronome\d+}` text inside of conditionals gets stripped of the squiggly brackets and set to `metronome` as a permanent fix. Thank you again for your patience with my first submitted plugin.

While updating the repo, I did also add an option to set a hotkey to reset the tick counter on the metronome, as is available on Visual Metronome and other major metronome plugins. Want to highlight this change as well for full transparency.

Limy-Monkey:
Going to close this PR for now. There are too many things I dislike about this plugin, it bugs out with sailing tiles, is very inefficient, would like to have party metronome sync functionality, etc. Going to focus on my other plugin for now, will try to come back to this later, will link this PR when I do re-open, but won't be for weeks/months.

PR #14928, opened 2026-08-10
Update Slayer task loot

Sintry1:
- Fixed an issue where death reclaimed items were counting to supply costs.
- Added the ability to Exclude items from the task loot list.
   - Exclusions can be done per-task or for all tasks.


Sintry1:
Converting to draft to fix another issue reported by a clan mate with icons

PR #14930, opened 2026-08-10
Update 07flip to 886ca0a

UserD40:
Update 07flip plugin to commit 886ca0a.

This update adds a small readout on your Grand Exchange offer slots so you can see at a glance how much of an offer has filled and how recently that item last traded, with colour and on/off options in the settings. It also fixes a bug where the plugin could fill a price into the quantity box when you set the quantity before the price.

The rest is reliability work on the buy/sell price lock: the target you were shown when you bought now survives a client restart, so the profit you were quoted is the profit you get when you come back to sell. Internally a number of small model files were merged into one, which cuts the plugin's source size noticeably.

### DIFF
--- a/plugins/07flip
+++ b/plugins/07flip
@@ -1,3 +1,3 @@
 repository=https://github.com/UserD40/Runelite07Flip.git
-commit=886ca0a4fd7cb4e9b60da3ce1a831b3672303671
+commit=0480bcda157819aa1046bf377c9e62064c717547
 warning=This plugin submits your IP address to a 3rd-party server not controlled or verified by RuneLite developers.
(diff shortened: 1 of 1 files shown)


UserD40:
Pushed one follow-up commit to this PR before review starts, tightening the same buy/sell price-lock work this update is about.

When you buy an item the plugin locks in the sell target you were shown, so the profit quoted at purchase is the profit you get when you sell. That lock is stored on our server, and the plugin had its own separate, shorter expiry timer running alongside it. The two could disagree, and an expired lock quietly turned into a live market price while still being labelled as locked. The plugin now follows the server's expiry only, and a lock that has expired simply disappears rather than silently changing value.

No new features and no scope added beyond the price-lock behaviour already in this PR.

riktenx:
you know that nobody reads your code or these ai-generated messages right? save your output tokens, we dont care and its just notification spam

UserD40:
Withdrawing this for now - we have further changes to finish on both the plugin and our server side, and we would rather submit one complete, reviewed-once update than have this sit in the queue while it changes underneath you. Will reopen with the finished build. Apologies for the noise.

PR #14938, opened 2026-08-10
Partychat

NonaNonia:
Switched from custom formatted messages to:
.type(ChatMessageType.FRIENDSCHAT)
for better formatting.
I also added requestSync() (as I've seen a similar party plug-in use it) because sometimes the member list got out of sync, resulting in the party list showing the sender of a message as logged out and <unknown> name.


tylerwgrass:
Resolve the branch conflicts

PR #14941, opened 2026-08-10
Update Rogue Chests FC to 3.0

tcourter1:
- Added encrypted password protected staff authorization with local unlock.
- Added encrypted, staff-only Party auto-join/leave for the Wilderness and Ferox Enclave. **THIS IS PASSWORD PROTECTED**
- Added an option to disable Party automation.


tylerwgrass:
Automatic party joining is not allowed

tcourter1:
> Automatic party joining is not allowed https://github.com/runelite/runelite/wiki/Rejected-or-Rolled-Back-Features

I'm aware of this policy. I didn't discuss it with him this time, but I have previously discussed this with Adam in regards to the BA Utilities plugin. We were able to get approval to add automated party features because of how locked down the feature was. 

For the BA plugin, joins/leaves are only possible if 1. the user opts in and, 2. they are within the bounds of the Barbarian Assault regions, and 3. The users are in a BA scrolled group.

In this case, the feature is even further locked down by using an encrypted, Staff-only password. At the maximum, there will be maybe 15 users that have this password, and therefore the ability to use the feature.

If necessary, I can pop into the RL Discord and discuss this particular usecase with him and try to get approval again. Just let me know! Thanks.

raiyni:
> Automatic party joining is not allowed [Wiki: Rejected or Rolled Back Features (runelite/runelite)](https://github.com/runelite/runelite/wiki/Rejected-or-Rolled-Back-Features)

I'd say this is considered still true as we don't actually know how large your group is. Also your hash is in the source code and people can just decrypt it.

tcourter1:
> > Automatic party joining is not allowed [Wiki: Rejected or Rolled Back Features (runelite/runelite)](https://github.com/runelite/runelite/wiki/Rejected-or-Rolled-Back-Features)
> 
> I'd say this is considered still true as we don't actually know how large your group is. Also your password is in the source code and people can just decrypt it.

Ok, I understand. I'm not trying to ruffle any feathers or anything. 

Would it still be considered auto-joining if I created a prompt asking if the user wants to join the Party? This feature would still be locked behind the password protection, but would require a manual button press to join the Party. In this case, all that's being done is removing the extra step of physically typing the party passphrase.

riktenx:
yes a button press is fine

tcourter1:
> yes a button press is fine

I removed all the automated features and put in a button on the side panel

PR #14946, opened 2026-08-11
Update ge-companion to v1.2.2

xOwip:
v1.2.2 — Performance & flipping stats update

- Fixed GE tax not applied to flipping stats (Margin, Profit at limit, ROI)
- Added Margin × Volume stat to flipping stats panel
- Added tooltips to all flipping stats explaining each metric
- Fixed variant item detail panel collapsing when removed from bank
- Added live "Last updated" timer in Bank tab (updates every 10 seconds)
- Fixed "Not yet scanned" truncation in bank context label
- Performance: idToName reverse map eliminates linear searches during bank scans
- Performance: cache variant display names to eliminate repeated getItemComposition() calls
- Performance: conditional bankValue/wealthValue writes — only saves when values change
- Bank history no longer records while logged out
- Updated README flipping stats section with GE tax explanation and new screenshot
- Replaced restricted Desktop.getDesktop().browse() with LinkBrowser.browse()

### DIFF
--- a/plugins/ge-companion
+++ b/plugins/ge-companion
@@ -1,2 +1,2 @@
 repository=https://github.com/xOwip/ge-companion.git
-commit=c3d472f860a0eaa61897eea86079bad230d570e2
+commit=db2fd987197aaa5fb921acfaefe46d4d42862869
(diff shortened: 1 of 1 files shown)


Alexsuperfly:

xOwip:
Fixed! Replaced all Desktop.getDesktop().browse() calls with net.runelite.client.util.LinkBrowser.browse() in commit db2fd98. Thanks for catching that!

PR #14949, opened 2026-08-11
Add OSRS Toolkit Sync

Wolklaw:
OSRS Toolkit Sync records actual Grand Exchange fill deltas to a durable local queue for the OSRS Toolkit desktop trade journal. It can optionally record completed player-to-player trades; that setting is disabled by default. The plugin uses local files only, makes no network requests, does not access credentials, and does not automate or alter game actions. The clean plugin build and its unit tests pass, and both GE-fill and opted-in player-trade paths were confirmed in game.

### DIFF
--- a/plugins/osrs-toolkit-sync
+++ b/plugins/osrs-toolkit-sync
@@ -1,2 +1,2 @@
 repository=https://github.com/Wolklaw/osrs-toolkit-runelite.git
-commit=35607926226ad825e3f779de7848abe54260300f
+commit=7eb9f2c8608cd4535aeccf285c8c9dcabc4d95f2
--- a/plugins/osrs-toolkit-sync
+++ b/plugins/osrs-toolkit-sync
@@ -1,2 +1,2 @@
 repository=https://github.com/Wolklaw/osrs-toolkit-runelite.git
-commit=7eb9f2c8608cd4535aeccf285c8c9dcabc4d95f2
+commit=85106b412f52fdf6510db150938c118efda07388
--- a/plugins/osrs-toolkit-sync
+++ b/plugins/osrs-toolkit-sync
@@ -1,2 +1,2 @@
 repository=https://github.com/Wolklaw/osrs-toolkit-runelite.git
-commit=85106b412f52fdf6510db150938c118efda07388
+commit=104c9ffe34b0faae7613ac918626d503da6862f3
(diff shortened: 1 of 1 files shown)


Alexsuperfly:

Wolklaw:
Thanks for the review, I fixed both. The retry now reschedules itself on the executor instead of blocking the thread with Thread.sleep/interrupt. Please take another look whenever you get a chance!

Wolklaw:
Updated to `104c9ff`. Rather than only fix the two items, I went back through AGENTS.md and swept the rest of the plugin for the same class of problem. Changes beyond your review:

- **No thread interrupt remains anywhere.** Both `ScheduledFuture`s are now cancelled with `cancel(false)`, and the executor is stopped with `shutdown()` rather than `shutdownNow()`.
- **`shutDown()` no longer does blocking file IO.** It was writing the offer-screen clear straight through; since `PluginManager` asserts the EDT in `stopPlugin`, that was a file replace on the Swing thread. It is queued on the executor now and drains during the orderly shutdown.
- **No client access off the client thread.** `startUp()` was reading `client.getLocalPlayer()` from the EDT, and `onConfigChanged` — which arrives on whichever thread changed the setting — was mutating state that is written from the client thread every tick. Both go through `clientThread.invokeLater` now, and the two fields shared across those threads are volatile.
- **No reference to the reflection package.** The offer-state read used a Gson `TypeToken`; it now walks the object slot by slot, which also means one damaged entry costs that slot instead of the whole file.

Tests pass (24) and the on-disk format is unchanged.

For context on the plugin as a whole, since it is a local file bridge: it makes no network requests of any kind — there is no HTTP client, no socket, and no dependency beyond runelite-client. It writes JSON under `.runelite/osrs-toolkit/` and nothing else reads or leaves the machine. Player-trade recording and the gear/bank snapshot are both opt-in and off by default. The desktop app that consumes the files is GPLv3 and open source at https://github.com/Wolklaw/OSRS-Toolkit. Happy to change anything here if the arrangement is not one you want to accept.

Wolklaw:
Updated to `334ebc0`. One follow-up commit, no behaviour change: the periodic housekeeping sweep sorted its file list with a comparator that went back to the filesystem for the sort key on every comparison. It now reads each file's modified time once when the directory is listed and carries it through both the sort and the delete loop. Tests still pass.

riktenx:
we dont allow plugins that rely on separate native applications that run on the users system. if you can instead interface with a website or something that would be ok

PR #14960, opened 2026-08-11
re-enable trade-tracker

botanicvelious:
Updated commit reference and re-enabled the plugin.


Alexsuperfly:

botanicvelious:
desktop has been replaced with linkbrowser, thank you

PR #13948, opened 2026-07-17
Add clan-bingo plugin

btvp97:
This is a plugin for clan bingo. Players must voluntarily put in a code that we provide them for their team, and it updates on certain task completion.

PR #14963, opened 2026-08-11
Misclickers Clan Bingo

btvp97:
This is a plugin for our clan to do a bingo together.


riktenx:
* you can't use this dependency
* give your plugin a more unique name than this

btvp97:
this plugin was previously submitted as #13948. A maintainer (riktenx) flagged the io.socket:socket.io-client dependency and asked for it to be removed. I've replaced Socket.IO-based live sync with simple HTTP polling and switched to build=standard, so this submission no longer has any third-party dependency or touches the shared verification confiig

riktenx:
well you've added a new dependency in the verification metadata so that's obviously just a lie

riktenx:
looks like you're just blindly driving this pr through an llm without any awareness of what's happening on it, hopefully by closing it here your llm will pick up on that and maybe check the comments on what needs to change

PR #14984, opened 2026-08-12
feat: add bank-settings plugin

idyet:
# Bank Settings

## What it does

Client-side options for the bank and deposit box. It lets you **disable** any of the three deposit buttons (Deposit inventory, Deposit worn items, Deposit loot) in both the bank and the deposit box. A disabled button stays visible but is dimmed and its click op is stripped, so it reads as greyed-out and does nothing — preventing accidental full-inventory / worn / loot deposits without hiding the button entirely.

It also, optionally:
- groups disabled buttons to the left of the strip so the active ones stay together, and
- adds small decorative icons beside the toggle rows on the bank settings page.

## Why "disable" instead of the game's own "hide"

The vanilla bank settings already let you *hide* these buttons (server side setting). This plugin intentionally does the opposite: it keeps the button on screen but inert, without having to switch to the bank settings page.

### Button disable option
disable-button-option

### Button disabled and shifted left
button-disabled

### All button toggle options
all-button-toggle-options

### In-game bank settings labels replaced with icons
bank-settings-icons

### DIFF
--- a/plugins/bank-settings
+++ b/plugins/bank-settings
@@ -1,0 +1,2 @@
+repository=https://github.com/idyet/bank-settings.git
+commit=08fa13df72cfb0553b3a54f586744f2b40fd6133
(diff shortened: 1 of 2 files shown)


ldavid432:
- You have also changed the pvp-arena file in this PR, remove that
- Fix merge conflicts
- Switch to `build=standard`

PR #14992, opened 2026-08-12
Create pot-stash plugin.

Zorga0:
Use default Bank Tags instead of the Potion Storage.

- Clicking the Potion Storage button opens a "potstash" Bank Tags tab instead of the Potion Storage interface.
- Right-click potions to set their withdraw dose (uses the same technique Bank Tags uses to withdraw potions from storage).

Almost everything relies on the existing Bank Tags plugin or mirrors it:
- The Potion Storage internals mirror `net.runelite.client.plugins.banktags.tabs.PotionStorage`:
  - same scripts (`POTIONSTORE_DOSES` / `POTIONSTORE_WITHDRAW_DOSES`)
  - same potion matching across dose variants (`matches()` / `findPotion()`)
  - same hidden widget preparation (`PotionStorage.prepareWidgets()`)
  - fires menu action 1007 at that widget (same mechanism Bank Tags uses to withdraw potions from storage)

Everything is 1:1 and goes through the same scripts and menu actions that Bank Tags already uses.

### DIFF
--- a/plugins/pot-stash
+++ b/plugins/pot-stash
@@ -1,0 +1,2 @@
+repository=https://github.com/Zorga0/pot-stash.git
+commit=37edc425f4c08f042ef912b56896475491360660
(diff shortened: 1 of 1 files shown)


ldavid432:
Use of client.menuAction is not allowed

Zorga0:
Hello! Thanx for helping me.
I updated and fully removed client.menuAction.

PR #15001, opened 2026-08-12
Xarpus Sounds Muter

MrJaspen:
Mutes selected Xarpus sound effects in ToB so players can isolate the audio cues they care about. Made this to help people learn how to do p2 xarpus with melee but I was hearing from my friends it's hard to hear the spit noise people mention as the audio cue. This plugin helps isolate that noise. No third party dependencies.


Alexsuperfly:
Player provided IDs is no longer a configuration allowed in feature sets of plugins, if you remove that and use the standard configs with IDs already defined that you have, that would be allowable.

PR #15006, opened 2026-08-12
Add Combat Insight

BornDied:
Adds Combat Insight, a RuneLite combat calculator that displays live max hit, hit chance, target-aware DPS, attack speed, and observed average hit.

The plugin uses only local RuneLite client data and bundled combat data. It does not communicate with external servers or automate gameplay.

Special attacks and defence-reduction tracking are intentionally outside the scope of this initial release.

### DIFF
--- a/plugins/combat-insight
+++ b/plugins/combat-insight
@@ -1,2 +1,2 @@
 repository=https://github.com/BornDied/combat-insight.git
-commit=9368f800573184d630667ffd33f2a2ae31ec376a
+commit=2c739351db706155b539400c1c28a618044e948f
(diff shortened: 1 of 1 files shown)


Alexsuperfly:
use `build=standard` in runelite-plugin.properties

BornDied:
Updated to use build=standard.

Alexsuperfly:
we are unable to host plugins under the GNU license, can you use BSD-2 like the [plugin-hub](https://github.com/runelite/plugin-hub/blob/master/LICENSE)?

BornDied:
Updated the project to use the BSD 2-Clause license as requested.

PR #15012, opened 2026-08-13
Add ParentsGuild plugin

kulcris:
Adds the ParentsGuild RuneLite companion plugin.

Plugin repository: https://github.com/kulcris/ParentsGuildPlugin
Commit: 20c997bb57709bcd82597ac55732fd960b199800


kulcris:
@Alexsuperfly Didn't realize this was in draft mode. It is now ready for review

riktenx:
you can't receive some random url from an http response and call out to it. every single url you make calls to needs to either be hardcoded or explicitly entered by the user

kulcris:
> you can't receive some random url from an http response and call out to it. every single url you make calls to needs to either be hardcoded or explicitly entered by the user

I think I got them all

PR #15018, opened 2026-08-13
Update clue-details

Zoinkwiz:


pajlada:
@Zoinkwiz reminder that you need to update the version in `runelite-plugin.properties` for the real version number is displayed in the plugin panel

PR #15021, opened 2026-08-13
Glamourer 1.7.6

jhughes:
* Fix glamours for animations that override the equipped items
* Multithreading fixes +hardening


Alexsuperfly:

PR #15031, opened 2026-08-14
Add OSRS Copilot

zerofata:
Copilot plugin for OSRS, uses integration from runelite and data available from various API's along with a basic regex & grammar routing system.

Requires a user to bring their own LLM endpoint. Primarily tested using an OAI endpoint from Llama.CPP (Qwen3.8 27B / Muse 30B) and an openrouter provider (GLM5.2 / DSv4 Flash 0731).

Effort is taken where possible to cache queries and reduce strain on API's, the largest queries are run once weekly under a Github CI and published on the plugin repo.

APIs used:
- oldschool.runescape.wiki/api.php
- prices.runescape.wiki/api
- secure.runescape.com
- raw.githubusercontent.com - osrs-copilot vocab-data branch, where larger weekly queries get stored and served.
- Whatever OAI endpoint a user configures

Users are required to opt-in and enable functionality before the plugin does anything involving an API.

Uses injected OkHttpClient/Gson throughout, no reflection, file IO confined to .runelite/osrs-copilot/.

Disclosure (although I'm sure it's pretty obvious): Codebase was written entirely by Claude.

### DIFF
--- a/plugins/osrs-copilot
+++ b/plugins/osrs-copilot
@@ -1,2 +1,3 @@
 repository=https://github.com/zerofata/osrs-copilot.git
-commit=defe9383d727e0739519a47fe4ada0e8f276d895
+commit=17b4debf4e13cdb828c5dce511199927829d59bc
+warning=This plugin submits your IP address and game state (skills, quests, inventory, equipment, bank) to a 3rd-party server not controlled or verified by the RuneLite Developers: the LLM endpoint you configure.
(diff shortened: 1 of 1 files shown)


Alexsuperfly:

zerofata:
Thanks for the quick look, I've removed it now.

Let me know if there's any other issues to change.

plugins/osrs-copilot
```diff
@@ -0,0 +1,3 @@
+repository=https://github.com/zerofata/osrs-copilot.git
+commit=d01a842e25d4b5e459a8ff1eacbe0bda93561159
+warning=This plugin submits your IP address and game state (skills, quests, inventory, equipment, bank) to a 3rd-party server not controlled or verified by the RuneLite Developers: the LLM endpoint you configure.
```
riktenx:
```suggestion
warning=This plugin submits your IP address, and may submit various account data, to a 3rd-party server not controlled or verified by Runelite developers.
```

PR #15038, opened 2026-08-14
tick-prayer-flick

spook121:
1 tick flick praying

### DIFF
--- a/plugins/tick-prayer-flick
+++ b/plugins/tick-prayer-flick
@@ -1,2 +1,2 @@
 repository=https://github.com/spook121/tick-prayer-flick.git
-commit=0c74a5886be081cd8c6956aaa7b1ffefa0420825
+commit=d62205336403b0c2fc7f2b9dcc93ddf359b18ad3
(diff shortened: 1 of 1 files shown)


Alexsuperfly:
use `build=standard` in runelite-plugin.properties

spook121:
why is a simple overlay taking forever to be approved? if anything else is needed please let me know

ldavid432:
> why is a simple overlay taking forever to be approved? if anything else is needed please let me know

Because there are 200 plugins in the queue in front of you and reviewer time is volunteer. You just need to wait

spook121:
will do! thanks.

tylerwgrass:
> Indicates what prayers to use in what order, for example in the Cerberus boss fight 

This is against the rules. Also your repo is structured completely wrong

tylerwgrass:
I've already linked you the rule. You can break the rule without doing automatic client interaction.

Read the guidelines

spook121:
again its not rule breaking. it doesnt just automatically tell you what to
do for whichever boss, thats not even in the code to be an option. its a
visual metronome that you set up with whatever prayer to whatever prayer
order you want. thats it. but its fine, i wont upload my plugins. if this
breaks the rules, then so does every other metronome and game tick counter.

tylerwgrass:
image

image

image

PR #15044, opened 2026-08-14
Add OSRS Toolkit plugin

Issamnagi:


ldavid432:
Use `build=standard` in your runelite-plugin.properties

PR #15047, opened 2026-08-14
Create bank-sorter

loutre29:
Allows you to organize your bank simply and efficiently while displaying item stats/GP. Works only in secondary tabs. The "All Items" tab remains manually organizable.

Available options:

* Group by categories
* Right click an item to change the default category
* Hide items with no filtered bonus
* Show values / stats on items
* Filter by equipement slots
* Filter by stats (you can combine multiple stats)
* Sorting by Name/GE Value/Quantity/weight/attack speed

### DIFF
--- a/plugins/bank-sorter
+++ b/plugins/bank-sorter
@@ -1,0 +1,2 @@
+repository=https://github.com/loutre29/osrs-bank-sorter.git
+commit=390a6fa3734d763df9eb53526f955d02b609587e


raiyni:
Please use the layout utility instead

loutre29:
The intent was a tag-free sorter easy to use for new players and that allow to sort your tabs easily to compare stuff, objects etc. I've seen no addon doing the same exact things.

loutre29:
What do I need to change precisely ? The goal of the addon is to compare and filter items in the bank. The categories are automaticaly set and the sort is made of different presets.
Please find attached a video of how the plugin is working.
Regards

https://github.com/user-attachments/assets/f087cdbe-8eaa-4748-a8a5-6d04bc343bf8

raiyni:
Use layouts like bank tag layouts would

PR #15051, opened 2026-08-14
Add Item Descriptions

tomaszantonik:
Adds Item Descriptions, a plugin that displays Wiki descriptions when hovering over items.

### DIFF
--- a/plugins/item-descriptions
+++ b/plugins/item-descriptions
@@ -1,2 +1,2 @@
 repository=https://github.com/tomaszantonik/item-descriptions.git
-commit=3f167c2fe1906386c440afc194ec8236dfd25f9e
+commit=758a3b30917987ec0bb4096fc5b50d374afe3a88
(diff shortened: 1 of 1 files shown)


Nightfirecat:
Requesting wiki pages for every item hovered by every user with this plugin installed is not a reasonable amount of traffic to be sending to the wiki. It would be better to instead scrape the wiki once, then have the plugin fetch that data on startup so you're not directing so much extra traffic to the wiki.

tomaszantonik:
> Requesting wiki pages for every item hovered by every user with this plugin installed is not a reasonable amount of traffic to be sending to the wiki. It would be better to instead scrape the wiki once, then have the plugin fetch that data on startup so you're not directing so much extra traffic to the wiki.

Thanks for your feedback! 
I made a few changes. Item descriptions are now store in a JSON file, and the plugin fetches this data once on startup through the loadDescriptions() method.
I added a GitHub workflow that runs once per day. Subsequent runs only request descriptions for new or missing items and update the existing JSON file.

Alexsuperfly:
LICENSE is invalid, use the [plugin-hubs](https://github.com/runelite/plugin-hub/blob/master/LICENSE) with your authorship

PR #15082, opened 2026-08-16
Add Facette Companion plugin

ethanmonlux:
Facette Companion is a read-only RuneLite plugin that exports a bounded view of the local player's live game state to one documented local JSON file.

The plugin performs no gameplay input, makes no network requests, and does not require Facette or any other external application. Facette is an optional reader of the exported file.


Alexsuperfly:

ethanmonlux:
Updated the submission to address the thread-interrupt issue.

The plugin no longer interrupts threads, and shutdown no longer waits for the publisher to finish. The final inactive snapshot is now best-effort during client exit, and the docs have been updated to match that behavior.

I also trimmed the public test set while keeping the core lifecycle, schema, writer, fixture, and dev-launcher coverage.

Build is passing on the updated commit. Thanks for taking a look.

PR #15088, opened 2026-08-16
Add Clan Piano MIDI

TheOffSeason:
Adds **Clan Piano MIDI**.

Lets you play the clan hall piano with a MIDI keyboard, and shares the performance with party members running the plugin.

The in-game piano is limited to 24 keys (C4-B5) and its Entertain mode quantises every note to the game tick, so a performance on it is heavily constrained. This renders the notes locally through the JDK synthesizer instead, giving the full MIDI range, polyphony and velocity, with no tick grid. Notes are shared over the existing party system so a clan hall event can hear the whole piece.

The plugin does not interact with the game. It reads a MIDI input device, plays audio, and exchanges note messages with the party. No input injection, no menu entries, no reflection, no HTTP, no file I/O.

Licensed BSD-2-Clause.

### DIFF
--- a/plugins/clan-piano-midi
+++ b/plugins/clan-piano-midi
@@ -1,2 +1,2 @@
 repository=https://github.com/TheOffSeason/clan-piano-midi.git
-commit=50b9e9cd88e0be623dc1dd5c1cbb9bc1f6929977
+commit=2e324d323c2b1aed4304c64390266cf7f05b3a20
--- a/plugins/clan-piano-midi
+++ b/plugins/clan-piano-midi
@@ -1,2 +1,2 @@
 repository=https://github.com/TheOffSeason/clan-piano-midi.git
-commit=10ba8616c557d599fb5836d113e9e631aece58e4
+commit=50b9e9cd88e0be623dc1dd5c1cbb9bc1f6929977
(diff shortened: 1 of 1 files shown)


TheOffSeason:
Thanks for the automated review. I'd like to check whether an exception is possible here, because I don't think `AudioPlayer` covers this case.

The `javax.sound` usage in this plugin is two separate things:

1. **Reading a hardware MIDI keyboard** (`MidiDevice`, `Transmitter`, `Receiver`, `ShortMessage`). This is the core of the plugin — it's how the user's piano keyboard is read. `AudioPlayer` plays sampled audio and has no facility for reading an input device, and as far as I can tell there's no MIDI support elsewhere in the client. `javax.sound.midi` is the only way to read a MIDI device in Java, so I don't have an alternative for this part.

2. **Rendering the notes** (`Synthesizer`, `MidiChannel`). This one I can convert to `AudioPlayer` with sampled piano notes, using note velocity as the gain. Happy to do that if it helps.

Is the restriction a hard rule, or is there room for MIDI input given there's no equivalent API? If input is acceptable I'll convert the rendering half to `AudioPlayer`. If it isn't, I'll withdraw the PR rather than waste reviewer time.

For context: the plugin reads the user's MIDI keyboard, plays a piano sound locally, and optionally shares notes with party members over the existing party system. It doesn't interact with the game at all.

PR #15095, opened 2026-08-16
Add RS-Bingo

Newrockku:
Side panel for following an rs-bingo.com bingo event: the team's board, tile checklists, point rates, per-player progress and standings, with optional in-client drop submission.

### DIFF
--- a/plugins/rs-bingo
+++ b/plugins/rs-bingo
@@ -1,4 +1,4 @@
 repository=https://github.com/Newrockku/RS-Bingo.git
-commit=ecc7c74dfb1ae63142907461dc7b2291892fe72f
+commit=52edd494a14e0225a623301384e5d50dd86868d7
 authors=Newrockku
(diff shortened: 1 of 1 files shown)


plugins/rs-bingo
```diff
@@ -0,0 +1,4 @@
+repository=https://github.com/Newrockku/RS-Bingo.git
+commit=ecc7c74dfb1ae63142907461dc7b2291892fe72f
+authors=Newrockku
+warning=This plugin sends your IP address and event code to rs-bingo.com, a third-party server not controlled or verified by the RuneLite developers, to load the event board. If you choose to submit a drop it also sends a screenshot of your client and your RSN for verification, and if you link your account it sends a token identifying you. Screenshots are only captured when you press the submit button.
```
riktenx:
```suggestion
warning=This plugin submits your IP address, and may submit various account data, to a 3rd-party server not controlled or verified by Runelite developers.
```

riktenx:
you can't get some url from an api response and call out to it. all urls the plugin talks to need to either come from the user or be hardcoded

PR #15112, opened 2026-08-17
Add RuneShopValue plugin

20pc-Nugget:
Adds a plugin that displays the sell value of runes at Ali Morrisane's shop.


Alexsuperfly:
only edit one file per PR

20pc-Nugget:
> only edit one file per PR

my mistake this is my second time doing this, won't happen again
am i required to change anything or is the script just awaiting approval

PR #15129, opened 2026-08-18
loadout-lab-ui 1.0.0

ajkatz:
The rich-interface companion for Loadout Lab: styled result cards, the mascot loading animations, and room for alternate layouts - the presentation layer, split out so the core plugin stays engine-focused under the source-size cap.

- Registers its renderer with Loadout Lab over the cross-plugin PluginMessage contract; without the core plugin installed it shows a small pointer panel only.
- No overlays, no game-state writes of its own; all game data arrives via the contract from the core plugin.
- Standalone-safe: with core absent it is a single passive panel.

### DIFF
--- a/plugins/loadout-lab-ui
+++ b/plugins/loadout-lab-ui
@@ -1,0 +1,2 @@
+repository=https://github.com/AKAddons/runelite-loadout-lab-ui.git
+commit=7b7173ac49b2342b7d273f679575b19ee2350b5c


abextm:
As you were told on Discord, your plugin needs to be a single plugin and cannot be split like this.

ajkatz:
> As you were told on Discord, your plugin needs to be a single plugin and cannot be split like this.

@abextm hi I was trying to have a conversation around how to best work with this restriction and didn't get any answers. Unfortunately I don't see any other path forward because my plugin size is over the limit of what reviewers accept. As I've said the main plugin will work without the UI plugin it just gives it a fuller UI.

abextm:
The only solution is to generate less code. The LLM cannot review things larger, and no human has time to look at vibe coded plugins like this.

ajkatz:
@abextm Before I rip this apart and merge everything back into one plugin I'd like you to actually consider the approach, because I don't think it got a real look.

RuneLite has a public cross-plugin API and the hub readme documents submitting multiple plugins. There's no written rule against a companion plugin. The split exists because of the size cap in the first place. The core plugin is the whole engine - all the game logic, works completely standalone, comfortably under the cap. The UI plugin doesn't touch the game at all, no overlays, no game state, it just draws what core hands it over the public API. Two smaller codebases with a narrow contract between them is less to review than one big one, not more.

And on vibe coded - this thing has a full test suite, golden regression outputs, wiki-verified data, and weeks of my own testing at bosses. I use AI to write code, I don't ship stuff I haven't tested.

coopermor:
The token limit exists to keep review costs down. By splitting large plugins into multiple it is just circumventing the token limit and increasing costs.

ajkatz:
> The token limit exists to keep review costs down. By splitting large plugins into multiple it is just circumventing the token limit and increasing costs.

That is a completely reasonable take and I understand. From my perspective, I'm creating one module which is slimmed down to be just the engine and is the piece that will need changes as the game changes. The other (this one) is a more permanent fixture which wouldn't need frequent changes. I get the perspective that you can't just trust that people wouldn't abuse this. 

I'll try to merge and condense my plugin but I just don't see how I can reduce this to a safe point beyond moving the engine to an endpoint.

abextm:
Basically, the token limit exists for a reason and trying to bypass it using any means is abuse of the service and disallowed. And while I don't think your use case of it is particularly horrible, if we allowed it here it would set a precedent for others to do the same to bypass this and other limits.

ajkatz:
> Basically, the token limit exists for a reason and trying to bypass it using any means is abuse of the service and disallowed. And while I don't think your use case of it is particularly horrible, if we allowed it here it would set a precedent for others to do the same to bypass this and other limits.

I get it but just think for the longterm health of the platform this restriction definitely helps ease the review process but actively limits the creativity of plugins actually using the full token allotment if there is a good purpose to do so and the code is benevolent/well-structured. By no means am i suggesting we take down 117HD or QuestHelper (or i'm even close to their user count) but this limit almost makes it impossible to produce something of that magnitude on the platform and reduces the ability to add artwork and robust UIs. 

Thanks for the quick feedback and keeping this platform running

PR #15136, opened 2026-08-18
update Better Slayer to 1.1

vividflash:
Mortimer task choice odds:

- fixed Panel to draw above his interface
- Setting to have best option colored in his list and/or as a panel/or neither
- Option to have odds quoted for a slayer unique roll, any unique item, a heart or gem, or an imbued heart
- Per superior odds corrected

Nieve instead of Steve:

- Gravestone rewritten for Glough
- Grave examine names Glough
- Player's closing line fixed


vividflash:
blocker is merged, undrafted. 

nieve verified fine ingame, 
mortimer i'm stuck on a long slayer task but want to push the fix out for the 162 already having it installed

PR #15145, opened 2026-08-18
Add Required Materials

SchauweM:
Adds Required Materials, which tracks the materials and skill levels you still need for Sailing and Construction builds, and can regroup your bank around them.

Repository: https://github.com/SchauweM/runelite-required-materials

- Clicking a build in-game starts tracking it. That works from the Sailing and Construction skill guides (both the old and new layouts), the Boat Customisation interface via Build or Check Materials, and the Furniture Creation menu.
- The side panel lists what you're tracking in a collapsible section per skill, showing each material as `have/need` against your bank and inventory, plus the skill levels the build requires, coloured green once you meet them. Bank contents are remembered between sessions, so the counts are there before you next visit a bank.
- A toggle button in the bank rebuilds the item grid so tracked materials come first, grouped per build with a `have / need` label and a tick or cross. It reuses the bank's own item slots rather than drawing an overlay, so withdrawing behaves normally, and toggling it off restores the bank exactly as it was.

Requirements are read from the OSRS Wiki rather than scraped from the game's interfaces. When you click a build, the plugin fetches that item's wiki page through RuneLite's injected OkHttpClient and parses its `{{Recipe}}` template for the materials and level requirements. Only the page name is sent, results are cached per page for the session, and no user or account data leaves the client. A handful of Construction activities have no `{{Recipe}}` on the wiki; for those it falls back to the material list the game already prints to chat, and the levels shown in the skill guide.

The plugin only reads game state and displays information. It does not automate anything, does not act on the player's behalf, and does not alter what the client sends to Jagex.

### DIFF
--- a/plugins/required-materials
+++ b/plugins/required-materials
@@ -1,2 +1,2 @@
 repository=https://github.com/SchauweM/runelite-required-materials.git
-commit=b65f1f33e7d35d5bfe56063692d90298fa2fb5eb
+commit=2ad57ad30647beca8cbd45f2a4416b0c84c2b1a0
--- a/plugins/required-materials
+++ b/plugins/required-materials
@@ -1,2 +1,2 @@
 repository=https://github.com/SchauweM/runelite-required-materials.git
-commit=dbddefd0fc9eb8c629532c39339ecda1f16ec844
+commit=a60b09e0b5b702c909e4168435b364db1f5cdaad
--- a/plugins/required-materials
+++ b/plugins/required-materials
@@ -1,2 +1,2 @@
 repository=https://github.com/SchauweM/runelite-required-materials.git
-commit=2ad57ad30647beca8cbd45f2a4416b0c84c2b1a0
+commit=dbddefd0fc9eb8c629532c39339ecda1f16ec844
--- a/plugins/required-materials
+++ b/plugins/required-materials
@@ -1,2 +1,2 @@
 repository=https://github.com/SchauweM/runelite-required-materials.git
-commit=3b8266d5271e238583426b4d6f28e8313850a96b
+commit=b65f1f33e7d35d5bfe56063692d90298fa2fb5eb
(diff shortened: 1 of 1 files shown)


SchauweM:
### Patch notes for the latest changes:

**Smarter about what's worth tracking**

Added toggles in the settings panel to stop tracking furniture or other things you're currently building. Clicking on furniture or boat facilities from the skill guides will still track them.

**Knows what's already built in your house**

Construction upgrades need the previous tier standing first — a Tool store 4 needs a Tool store 3. Those requirements are now shown, and confirmed against what's in your POH: visit it once in build mode and the plugin remembers what's there.

**Fixes**
- Some furniture tracked as nothing at all — anything whose in-game name differs from its wiki page, such as Oak chest of drawers.
- Upgrade requirements like "Tool store 2" were listed as an item to withdraw and sat at 0/1 forever. Now correctly labeled and tracked as a requirement
- The bank button is now a proper button with the plugin's icon and a tooltip, sitting next to the bank scrollbar.
- Tracked builds are grouped by skill, most recently used first, and clicking Build no longer shuffles an entry to the bottom of the list.

SchauweM:
### FINAL patch notes before release (if it get's approved)

- UX improvements:
  - The side panel now says when it's waiting on something — open a bank to get item counts, or visit your POH in building mode to confirm which upgrades are already built — instead of showing zeroes with no explanation. Both messages disappear once satisfied.
  - Fixed the panel showing that bank prompt on every login even after a bank had been opened.
- Side panel layout fixes: rows use the full panel width, and long build names wrap instead of being cut off.
- Part naming is now shared between the side panel and the lookup rather than implemented twice, with tests covering the boat-size and raft/skiff/sloop grouping rules. 43 tests total.

These patch notes mark the final version for submission. Excluding any changes required by the reviewers obv.

PR #15150, opened 2026-08-19
add tombs-of-arbitrage

exisrick:
## Summary
- adds Tombs of Arbitrage, a passive Grand Exchange flip journal and advisory companion
- provides local Trade, Ideas, Stats, and recovery views without automating game actions
- keeps optional paired Ideas read-only, user-initiated, and disabled by default

## Test plan
- [x] Public Java 11 CI: 697 tests, check, and build
- [x] Plugin Hub manifest points to the verified public commit


Alexsuperfly:
`too many tokens: 233503` you have to be under 200k for our review bot to run your plugin, otherwise it requires manual review, which at enough lines to be that many tokens no one will be wanting to do

PR #15161, opened 2026-08-19
Create runeshopvalue

20pc-Nugget:


Alexsuperfly:
LICENSE is invalid, use the [plugin-hubs](https://github.com/runelite/plugin-hub/blob/master/LICENSE) and just place your authorship in it

PR #15199, opened 2026-08-20
Add No GE Inventory Blink plugin

osrshool222:
Adds the 
o-ge-inventory-blink RuneLite plugin which stops the inventory from blinking/flashing while the Grand Exchange window is open. It does this by hiding the pulsing GE glow (GeOffersSide.GLOW) while the GE window is open.


osrshool222:
Closing: pushing back / removing plugin hub submission

PR #15208, opened 2026-08-20
Update Kill Clog to 2.0.1

420kc:
Kill Clog 2.0.1

* adds a list view for the boss grid, switched from the hamburger menu
* adds an off switch for chat emojis
* the chat message toggle now covers sync results and warnings too; setup guidance stays on, since a muted user can't know they muted it
* fixes pb lines for turkish locale clients
* fixes the show rank setting in compare mode
* aligns hub search tags with the client search

### DIFF
--- a/plugins/kill-clog
+++ b/plugins/kill-clog
@@ -1,3 +1,3 @@
 repository=https://github.com/420kc/kill-clog-plugin.git
-commit=2496ad195e9a639fa5be83d1bf5ff7d053351eb1
+commit=afc574ac78838fc28456dd99fe1c892c5649d66b
 warning=This plugin submits your IP address to a 3rd-party server not controlled or verified by Runelite developers.


coopermor:
The search is just local only, right? Is there a reason you need the RsnInputPolicy BLOCKED list?

420kc:
> The search is just local only, right? Is there a reason you need the RsnInputPolicy BLOCKED list?

@coopermor probably a little paranoid on my end. it's local only but the entered rsn string renders visibly in the plugin ui (search and player not found messages), so I wanted to add the rsn block list as a content safeguard that matched vanilla's hiscores, because currently the plugin could be used to render slurs or other abusive text and present screenshots that suggest it's Kill Clog's content.

Edit: It's pretty heinous having the hard list so I'll remove that and repin

coopermor:
What do you mean by vanilla's hiscores? I do not see any blocklist in core's hiscore search. https://github.com/runelite/runelite/blob/master/runelite-client/src/main/java/net/runelite/client/plugins/hiscore/NameAutocompleter.java

riktenx:
> because currently the plugin could be used to render slurs or other abusive text and present screenshots that suggest it's Kill Clog's content.

well sure but also any plugin that ever displays a user-input string anywhere could too?

420kc:
Thanks @coopermor and @riktenx  You were both right about the RSN handling. My initial reasoning was overcautious and I was conflating vanilla hiscores functionality with Jagex' actual content filter so I removed the list.

Separately, the updated pin includes a fix for a longstanding cross-plugin tooltip issue I reproduced during testing. Switching Kill Clog between click and hover modes was modifying Swing's shared tooltip settings, which could leave hover tooltips broken in other plugins until RuneLite was restarted. Tooltip listener management is now scoped exclusively to Kill Clog with regression coverage for switching from click to hover and back. Unrelated plugin tooltips are untouched.

PR #15212, opened 2026-08-21
add max-cape-qol

FreeArcanes:
Adds Max Cape QOL, which provides visible configurable shortcuts and compact paging for inventory and equipped Max cape menus. Source tests pass with gradlew.bat test.


ldavid432:

PR #15236, opened 2026-08-21
exchange-insights-capture: add plugin

TheSpryt:
Exchange Insights Capture, an instant replay plugin. It keeps the last few seconds of play in memory and writes an MP4 when you die, kill a boss, or fill a collection log slot. A hotkey saves one whenever you want. Clips land in folders named for your account and what triggered them, the way RuneLite files screenshots.

The plugin has no third-party dependencies. It encodes H.264 and writes the MP4 container itself, so the verification template needs nothing added. An earlier revision of this PR used jcodec, which has since been removed.

**Warning line.** The plugin can upload clips to Exchange Insights, a third-party service, so the entry warns about the upload and the IP address it implies. Linking an account turns that on. Nothing else in the plugin touches the network.

**What it records.** Capture reads the client canvas, falling back to a screen grab of the client window while that window is focused. Clips carry no audio.

**Since review.** Dropped `LinkBrowser::open`, `Desktop`, subprocess execution, `getenv`, `Runtime::availableProcessors` and thread interrupts. Clicking a clip copies its path to the clipboard, now that a plugin cannot open one.

### DIFF
--- a/plugins/exchange-insights-capture
+++ b/plugins/exchange-insights-capture
@@ -1,0 +1,3 @@
+repository=https://github.com/TheSpryt/Exchange-Insights-Capture.git
+commit=5e0fe48b27e5df5131763863c284af370c4f8d75
+warning=This plugin can upload your recorded clips and submits your IP address to a 3rd-party server not controlled or verified by Runelite developers.
--- a/plugins/exchange-insights-capture
+++ b/plugins/exchange-insights-capture
@@ -1,3 +1,3 @@
 repository=https://github.com/TheSpryt/Exchange-Insights-Capture.git
 commit=5e0fe48b27e5df5131763863c284af370c4f8d75
-warning=This plugin can upload your recorded clips and submits your IP address to a 3rd-party server not controlled or verified by Runelite developers.
+warning=This plugin submits your IP address, and may submit various account data, to a 3rd-party server not controlled or verified by Runelite developers.


riktenx:
use of jcodec is not allowed

riktenx:
> use of jcodec is not allowed

.

TheSpryt:
jcodec is gone. The plugin now encodes H.264 and writes the MP4 container itself, so this PR adds one file and touches nothing shared, and there are no third-party dependencies to verify.

The scan feedback above came from jcodec rather than plugin code, which logs through slf4j and has no system I/O of its own.

ldavid432:
Use `build=standard` in your runelite-plugin.properties

TheSpryt:
Changed to `build=standard` and repointed. It only needed `gradle` while there were third-party dependencies to resolve, and there are none now.

ldavid432:
* use of subprocess execution is not allowed

TheSpryt:
All six done, repointed.

- `LinkBrowser::open` and `Desktop` are gone. Opening a clip and revealing it in a file manager both went with them, along with the `explorer.exe` and `open -R` subprocesses behind the latter. Clicking a clip now copies its file path to the clipboard.
- `LinkBrowser::browse` is still used for the two https links, which I understand to be the permitted call.
- `getenv` is gone. It detected Wayland to avoid a screen capture that returns black; the existing blank-frame detection already catches that case by measuring it, so nothing replaced it.
- `Runtime::availableProcessors` is replaced by fixed pool sizes.
- No interrupts: executors stop with `shutdown`, and an interrupted encode no longer re-flags the interrupt.

One thing I would rather raise than have you find: capture falls back to `Robot::createScreenCapture` when the renderer returns nothing, bounded to the client window and only while that window is focused. Happy to drop it if that is not acceptable, though it is what makes capture work on the GPU renderers.

riktenx:
remove it

TheSpryt:
Removed, and repointed.

Capture now goes through `DrawManager::requestNextFrameListener` only. The canvas bounds, the pointer position measured against the grabbed rectangle, and the blank-frame detection all went with it, the last of those because Robot fails by returning a valid all-black image rather than an error and there is nothing left to detect.

TheSpryt:
@riktenx @ldavid432 Is there an approved way for a plugin to open its output folder?

riktenx:
not afaik

TheSpryt:
OK, I guess I remove that functionality then and repointed. Let me know if there's anything else

plugins/exchange-insights-capture
```diff
@@ -0,0 +1,3 @@
+repository=https://github.com/TheSpryt/Exchange-Insights-Capture.git
+commit=5e0fe48b27e5df5131763863c284af370c4f8d75
+warning=This plugin can upload your recorded clips and submits your IP address to a 3rd-party server not controlled or verified by Runelite developers.
```
riktenx:
```suggestion
warning=This plugin submits your IP address, and may submit various account data, to a 3rd-party server not controlled or verified by Runelite developers.
```

PR #15246, opened 2026-08-21
New plugin Lootshare (Beta)

pk-enjoyer:
## How it works

  It combines two existing RuneLite features, the Party system and LootManager, with additional logic to track loot (using runelites LootManager) and calculate fair splits.

  ## Usage

  Join a Party as you normally would, play OSRS, collect loot, and view the final split at the end.

  There are some configurable settings, but most are not client-side. Instead, the host controls these settings so splits remain fair and cannot be manipulated by individual members.
  image

  ## This is a beta release

  Several parts are still missing. The main reason is to keep the initial review manageable by reducing the number of lines reviewers need to go through. I’ll aim to stagger future changes into smaller reviewable chunks.


riktenx:
>  The main reason is to keep the initial review manageable by reducing the number of lines reviewers need to go through. I’ll aim to stagger future changes into smaller reviewable chunks.

we don't read your code whatsoever, a bot now does that, outside of extremely extenuating circumstances that i don't know whether they apply to you. so i am just gonna close this and when you are ready for it to be reviewed just submit it not draft

PR #15247, opened 2026-08-21
Update monsterstats to v1.3.0

Koitere:
Adding additional stats to be available in plugin (Combat Stats/Attack Speed). Added option to see defensive stats on monster examine. Added ability to customize keybind for tooltip hover display. Updated monster database to be consistent with current game version.


Koitere:
A large part of the change size is due to the database being updated with new monsters/stats. As always thanks for everything you guys do :)

PR #15286, opened 2026-08-22
Add Nex static calculator

sleepywaffle96-maker:


riktenx:
you cant add two plugins in one pr

PR #15289, opened 2026-08-22
Add groupscape plugin

LiamMorganDev:
Tracks a RuneScape group's players (skills, inventory, bank, position, etc.) and streams the data to a group-scoped web dashboard so other group members can view it live. Similar in shape to the existing group-ironmen-tracker plugin already in the hub.


plugins/groupscape
```diff
@@ -0,0 +1,2 @@
+repository=https://github.com/MayzrDev/groupscape-plugin.git
+commit=1fe0dfe71cc0f9a5c43405a5173703d3a0a5df89
```
riktenx, line 2:
```suggestion
commit=1fe0dfe71cc0f9a5c43405a5173703d3a0a5df89
warning=This plugin submits your IP address, and may submit various account information, to a 3rd-party server not controlled or verified by Runelite developers.
```

PR #15313, opened 2026-08-23
the-birdhouse: update commit

birdandworm:
Non-commit plugin change.

Changes:
- Fix raid loot (TOA/COX/TOB) not being reported to clan leaderboard
- Add 100k GP minimum threshold to match previous Dink config
- Allow clan leaderboard reporting without auth token via WOM membership validation
- Include playerName in all clan-loot request bodies


Alexsuperfly:
dont change the warning

PR #15316, opened 2026-08-23
Create bank-coffer-scanner

osrsrigid:
Adds Bank Coffer vs Alch — compares Death's Coffer sacrifice value (105% GE price) against High Alch value for every item in your bank, so you can see which items are worth sacrificing vs alching (particularly useful for Ironmen, who can't sell on the GE).

Items where the gap is large get a highlighted warning in the panel (configurable threshold), and each row has a button to permanently hide items you don't want in the list.

Read-only/informational only — no automation, no menu entry changes, no interface modifications.


Alexsuperfly:
LICENSE is invalid, can you use the [plugin-hubs](https://github.com/runelite/plugin-hub/blob/master/LICENSE) and just place your author attribution

PR #15320, opened 2026-08-23
Add OSRS Toolkit Sync

Wolklaw:
Resubmission of #14949, which was closed for depending on a native companion application running on the user's system.

The plugin no longer talks to anything local. It now POSTs the same event data over HTTPS to a self-hosted sync service ([osrs-toolkit-sync-server](https://github.com/Wolklaw/osrs-toolkit-sync-server), open source) instead of writing to `.runelite/`, so the data it sends is inspectable the way a local binary isn't. Cleared with @Pine on the RuneLite Discord: "info that is sent to a website can be verified a random native app can't so yeah."

What it records, all opt-in except Grand Exchange fills:
- Grand Exchange fills, offer opens/cancels, and where you're standing in the GE
- Player-to-player trades (off by default)
- PvM gear/bank/skill snapshots on bank open, for readiness checklists (off by default)
- NPC loot received (off by default)
- What was equipped/carried at the moment of death (off by default)

Repo: https://github.com/Wolklaw/osrs-toolkit-runelite

### DIFF
--- a/plugins/osrs-toolkit-sync
+++ b/plugins/osrs-toolkit-sync
@@ -1,2 +1,2 @@
 repository=https://github.com/Wolklaw/osrs-toolkit-runelite.git
-commit=35607926226ad825e3f779de7848abe54260300f
+commit=7eb9f2c8608cd4535aeccf285c8c9dcabc4d95f2
--- a/plugins/osrs-toolkit-sync
+++ b/plugins/osrs-toolkit-sync
@@ -1,2 +1,2 @@
 repository=https://github.com/Wolklaw/osrs-toolkit-runelite.git
-commit=7eb9f2c8608cd4535aeccf285c8c9dcabc4d95f2
+commit=85106b412f52fdf6510db150938c118efda07388
--- a/plugins/osrs-toolkit-sync
+++ b/plugins/osrs-toolkit-sync
@@ -1,2 +1,2 @@
 repository=https://github.com/Wolklaw/osrs-toolkit-runelite.git
-commit=85106b412f52fdf6510db150938c118efda07388
+commit=104c9ffe34b0faae7613ac918626d503da6862f3
--- a/plugins/osrs-toolkit-sync
+++ b/plugins/osrs-toolkit-sync
@@ -1,2 +1,2 @@
 repository=https://github.com/Wolklaw/osrs-toolkit-runelite.git
-commit=334ebc0c951149a689ccc5ffbce691a076569228
+commit=2a16b8561a95d62fad5311574640b79a89750dd6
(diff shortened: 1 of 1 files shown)


Wolklaw:
Bumped the pin from `2a16b856` to `a652df8`. Not a cosmetic update — the previously pinned build had a bug that made half the plugin silently do nothing, and I would rather it not be reviewed in that state:

- **Live Grand Exchange state never reached the service.** The client sent `POST` to two endpoints the service routes as `PUT`, so every offer-slot and offer-screen write came back `405` and was discarded. It failed quietly because a 4xx is treated as "this payload will never be accepted" and dropped rather than retried, while heartbeats and events use `POST` routes that were always correct — so the plugin kept reporting a healthy connection the whole time. Fixed, with a test that pins the verb per endpoint.
- **Removed the service address setting.** It is compiled in now. The server is open source, so anyone self-hosting is already building from source; as a text field it mostly offered a way to break your own sync with a typo.
- **Removed the connection overlay** and its config item. Pasting a token is now answered in the chatbox instead — accepted, rejected, unreachable, or "no token yet", which previously said nothing at all.
- **Updated the plugin description**, which still told people to enter the address that no longer exists.

The connection section is now just the opt-in switch (carrying the required warning string) and the pairing token. No new dependencies, no new permissions, and file IO is still confined to `RUNELITE_DIR`.

Wolklaw:
Bumped the pin from `a652df8` to `613cf30`.

No behavior change — this is documentation only, but it's directly review-relevant so I didn't want it left stale under review. A comment in `LocalSyncStore.atomicWrite()` said "the desktop app reads the offer-state file whenever it redraws the Grand Exchange slots." That was true when it was written and has been false since the desktop app I maintain stopped reading `.runelite/osrs-toolkit/` at all — but the comment stayed, and read on its own it describes exactly the arrangement this submission exists to avoid: a plugin feeding an application on the same machine.

The code under it never changed — `events/` is a network outbox (deleted once the sync service confirms it holds the event), `state/` is this plugin's own memory of the eight GE slots so a restart can still compute fill deltas correctly. Both are read by nothing but this plugin. Added a class-level doc saying that plainly, and corrected the one comment that said otherwise.

riktenx:
we don't really care, save some output tokens and have your llm not post these updates

plugins/osrs-toolkit-sync
```diff
@@ -0,0 +1,2 @@
+repository=https://github.com/Wolklaw/osrs-toolkit-runelite.git
+commit=613cf3094bedf301b8cc026e84dba96d0f61e239
```
riktenx, line 2:
```suggestion
commit=613cf3094bedf301b8cc026e84dba96d0f61e239
warning=This plugin submits your IP address, and may submit various account data, to a 3rd-party server not controlled or verified by Runelite developers.
```

PR #15361, opened 2026-08-24
goalplanner: update to 0.5.0

ajkatz:
Bumps Goal Planner from `840b7124` to `83966f32` (0.5.0).

This is a resubmission of #14871, which was auto-closed by the stale bot on
19 Aug after sitting labelled "waiting for author". I had two plugin PRs open at
the time (#14853 for loadout-lab, which has since merged); if this one was
parked for that reason, nothing here has changed and the tree is identical. If a
reviewer did want something specific, I could not find it on the PR or in the
linked threads, so please do say and I will turn it around promptly rather than
letting it go stale a second time.

The pinned tree is unchanged from #14871, so any earlier review effort still
applies.

## What's in 0.5.0

**Repeatable goals.** Goals can now repeat on a schedule (daily / weekly /
monthly) instead of being one-shot. A repeating goal re-bases its target each
period rather than storing a baseline, so it stays correct across missed periods
and level-ups. Optional per-period chunking ("300k XP each week") drives the
progress display.

**Wrong-account fix.** Goal progress could previously be attributed to whichever
character happened to be logged in. Storage is now keyed per character, so
progress follows the account that earned it.

**Per-character storage.** Follows from the above: each character keeps its own
goals, sections and tags rather than sharing one global set.

**Fallen From Grace.** Quest data for the new quest, plus the boss it
introduces, so it is trackable like the rest.

## Notes for review

- **Size.** The bot labels this `size-xl`. It is one feature arc (repeatable
  goals) plus a correctness fix that required the storage change, released
  together because the storage rework is what makes the fix possible. Happy to
  discuss slicing if that is preferred for future releases.
- **No new network access.** This release adds none. The plugin remains local
  only.
- **Source size** is 188,887 tokens against the 200,000 cap, gated in CI by
  `./gradlew checkTokens`.
- `./gradlew preSubmit` is green: the full test suite plus the token, glyph,
  docs and ADR gates.

### DIFF
--- a/plugins/goalplanner
+++ b/plugins/goalplanner
@@ -1,3 +1,3 @@
 repository=https://github.com/AKAddons/runelite-goal-planner.git
-commit=83966f32a445e5afb7bb75bfdcc236e0592cebc1
+commit=323d31a384df15b41913ea21847b45695d7b258f
 authors=ajkatz
--- a/plugins/goalplanner
+++ b/plugins/goalplanner
@@ -1,3 +1,3 @@
 repository=https://github.com/AKAddons/runelite-goal-planner.git
-commit=840b7124af3f65cf0ac9b89cfec2846a6d08b451
+commit=83966f32a445e5afb7bb75bfdcc236e0592cebc1
 authors=ajkatz
--- a/plugins/goalplanner
+++ b/plugins/goalplanner
@@ -1,3 +1,3 @@
 repository=https://github.com/AKAddons/runelite-goal-planner.git
-commit=323d31a384df15b41913ea21847b45695d7b258f
+commit=b4a04ce697a01fca1c72220433d9a2f2eec8153a
 authors=ajkatz


Alexsuperfly:

ajkatz:
Fixed in `323d31a` - `openDiscordInvite` now calls `LinkBrowser.browse(DISCORD_URL)`; the hand-rolled `Desktop.isDesktopSupported` guard and its fallback logging are gone with it. Repinned this PR to that commit.

Thanks for the review. I also swept the plugin for the other common hub restrictions (`System.exit`, `Runtime.getRuntime`, raw `new Thread`, `printStackTrace`, `System.out`) and found no further instances.

riktenx:

ajkatz:
@riktenx there's no second call site - apologies, the hit was my own doing. The fix landed in `323d31a`, but the replacement javadoc still contained the literal `Desktop::browse` while explaining why it wasn't used, so a grep of the tree kept matching it. Reworded in `b4a04ce` and repinned.

`git grep Desktop` over the pinned tree now returns nothing at all; `openDiscordInvite` is a one-liner calling `LinkBrowser.browse(DISCORD_URL)`.

PR #15379, opened 2026-08-24
add actual afk

devsiriso:


devsiriso:
image_overlay
image_panel
image_settings

PR #15387, opened 2026-08-24
Update runepouch loadouts

DapperMickie:
Adds compact layout


riktenx:
use of client.menuAction isn't allowed in this manner, the only thing hub plugins are allowed to use it for is the collection log sync thing (eg. wikisync)

PR #15389, opened 2026-08-24
Update Player Compass for zoom-scaled compass

BloodGateRS:
Updates Player Compass to scale letter size and spacing with the projected character/world scale, keeping N/E/S/W evenly spaced while zooming.


Alexsuperfly:
needs rebase

PR #15393, opened 2026-08-24
Update parentsguild

kulcris:
Removed need for bingo token
Added localization for timestamps
Optional bingo metric tracking for XP, boss KC, and clues.
Pending metric progress uploads before WOM confirms.
Metric progress display like 0(+20)/200 xp.
Reminder to log out/back in when a metric is met locally so WOM can update.
Optional WOM refresh on logout/world-hop, default off.
Multi-item proof item selection before taking a screenshot.
Multi-item tiles now show multiple item icons on the plugin board.
Multi-item tiles now show multiple item icons on the website board.
Better board progress text, including 0/1 for single drop tiles.


kulcris:
Sorry for the large changes so soon, realized i was missing some functionality.

PR #15400, opened 2026-08-24
Add Bankstand

CVE-078:
## Description

Bankstand ([bankstand.gg](https://bankstand.gg/)) tracks OSRS account progress. This plugin keeps your skills, quests,
achievement diaries, combat achievements and collection log synced to your Bankstand account
automatically, so you don't have to go to the site and look anything up by hand.

## Data sent, and to whom

Only to your own Bankstand account, at whatever Server URL you set (bankstand.gg by default).
Each capability is its own toggle in the config, and only skill XP is on by default. The
manifest's warning= line and the plugin's README both spell out exactly what each toggle sends.

Everything's private to your account by default. Sharing it with a group or making it public is
a separate decision you make later, on the site, not something pairing turns on for you.

## Review notes

- build=standard, no runtime dependency beyond net.runelite:client
- No reflection, no native code, no process execution
- No client.menuAction/client.runScript/client.invokeMenuAction anywhere, enforced by a test
  (NoAutomationApiTest) that scans the source for it
- The pairing token never goes through ConfigManager (another test, NoCredentialsInConfigTest,
  checks this) and is never logged
- docs/for-reviewers.md in the repo names the exact files that touch the network and gives you
  the grep commands to check nothing else does

Repository: https://github.com/CVE-078/bankstand-runelite


Alexsuperfly:

CVE-078:
Thanks for catching this. Fixed: the submit retry's backoff no longer blocks a thread at all, so there's nothing left to interrupt either. Each retry is now a task scheduled on the plugin's own `ScheduledExecutorService` after the backoff delay (CVE-078/bankstand-runelite#65). Bumped the pinned commit above to `cd9ed26`'s target.

PR #15406, opened 2026-08-24
Update nex-static-calculator

sleepywaffle96-maker:
Added a few features to get kc from Collection logs to update bar from previous kills and dynamic % calculated from each kill


riktenx:
why is this a draft?

sleepywaffle96-maker:
Just checking last few details before sending so I won't spam. I want to be sure everything works

PR #15416, opened 2026-08-25
Add Compact Mode

gabworknestio-beep:
Adds Compact Mode, a display-only plugin that scales the complete RuneLite game window to a compact size while preserving accurate mouse input.

It supports 35%, 25%, and 20% widths, restores the original window bounds, and can dock the window to the right.

Tested locally with the external plugin launcher on Windows.

### DIFF
--- a/plugins/compact-mode
+++ b/plugins/compact-mode
@@ -1,0 +1,2 @@
+repository=https://github.com/gabworknestio-beep/Compact-Mode.git
+commit=2f993b755410cb0519913c307d953d2e74c640fc


Alexsuperfly:
Manipulation of the client frame is something we're scrutinizing more in new plugins, after the click to minimize situation, so this plugin is going to be rejected.  
The scale argument that can be passed to the launcher configuration can be less than one and should perform similar to this.  There is also the Picture in Picture plugin already available, which does something sort of similar as well.

PR #15423, opened 2026-08-25
exchange-logger

clamzhi-dev:
Updates the exchange-logger manifest to https://github.com/istid/exchange-logger/pull/15 - fixes GE sale tax not accounting for the 2% tax, log files always being .log regardless of the selected format, added option to split logs into separate files for each account and adds item names alongside item IDs.


Alexsuperfly:
I see the original author is letting you take over their work, but you need to add your authorship to the plugin manifest like [117hd](https://github.com/runelite/plugin-hub/blob/2d655ba1b2b7d3bc356ebaa32129288efa392bca/plugins/117hd#L3) in order to be allowed to submit plugin updates. and fix the build

clamzhi-dev:
Thank you for your guidance. Is this sufficient?

PR #15429, opened 2026-08-25
Add RuneGlass

JeroenBreevoort:
Adds RuneGlass, an opt-in plugin that syncs your character's skills and XP to the RuneGlass companion app.


riktenx:
why is this a draft

PR #15448, opened 2026-08-25
create irons-pub-bingo

KaosRichie:
A plugin for running clan bingo events (mainly for the Iron's Pub clan). The host shares a board file; the plugin tracks drops, raid uniques, boss kill counts, NPC kills, pets, XP, agility laps, loot value and chat messages against the board's tiles, shown in a sidebar panel.

Team sync is optional and comes in two independent layers: live sync over RuneLite's party service, and a Google Apps Script team store the event host deploys, which carries progress between sessions and drives the admin sheet. Neither layer needs the other. The team store and the Discord webhook for completion posts are off by default, and their config toggles carry the third-party IP warning.

### DIFF
--- a/plugins/irons-pub-bingo
+++ b/plugins/irons-pub-bingo
@@ -1,0 +1,4 @@
+repository=https://github.com/KaosRichie/irons-pub-bingo.git
+commit=14082f8809acd8a4db252563366eec5989aabf1a
+authors=KaosRichie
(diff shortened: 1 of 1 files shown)


Alexsuperfly:
you cannot [auto join RL parties](https://github.com/KaosRichie/irons-pub-bingo/blob/14082f8809acd8a4db252563366eec5989aabf1a/src/main/java/com/ironspubbingo/IronsPubBingoPlugin.java#L2622), the party service is already overloaded enough without the potential of holding parties open that people might not even know they're in.

PR #15457, opened 2026-08-26
Add Ground Item Icons

Meowdyosrs:
Adds Ground Item Icons, a standalone plugin that displays item icons alongside Ground Items text.

The plugin works alongside the existing Ground Items plugin without modifying it.

Features:
- Displays item icons next to Ground Items
- Respects Ground Items visibility and filtering
- Toggleable icon display
- Configurable icon size

Untitled

### DIFF
--- a/plugins/ground-item-icons
+++ b/plugins/ground-item-icons
@@ -1,2 +1,2 @@
 repository=https://github.com/Meowdyosrs/GroundItemIcons.git
-commit=ee329c1d127245ecf4516b5c14f44f93b0189c26
+commit=1e02cc6e1ab91256398f249fe46fcdb91a3b755e
--- a/plugins/ground-item-icons
+++ b/plugins/ground-item-icons
@@ -1,2 +1,2 @@
 repository=https://github.com/Meowdyosrs/GroundItemIcons.git
-commit=022df1f
+commit=022df1f1cbdade1be19004f318fdfc1d5939edbc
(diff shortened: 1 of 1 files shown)


Fiffers:
Any chance you'd consider changing the name? It's a bit close to the ground loot icons plugin in both name and function, so it'd be much appreciated. 

If you don't want to, understandable. I just want to avoid confusion if at all possible. Earlier today someone wrote an issue for my plugin when the thing they wanted fixed was from an entirely different plugin 😂

Meowdyosrs:
> Any chance you'd consider changing the name? It's a bit close to the ground loot icons plugin in both name and function, so it'd be much appreciated.
> 
> If you don't want to, understandable. I just want to avoid confusion if at all possible. Earlier today someone wrote an issue for my plugin when the thing they wanted fixed was from an entirely different plugin 😂

Hi Fiffers! 
Yeah, that's completely understandable 😂 I can see how the names could cause some confusion, especially with both plugins dealing with ground item icons.

I'd prefer to keep the name though, as I think "Ground Item Icons" pretty directly describes what the plugin does. Sorry if that's a bit disappointing, but your plugin actually inspired me to make this one in the first place, so thank you!

Alexsuperfly:
use build=standard in runelite-plugin.properties

Meowdyosrs:
> use build=standard in runelite-plugin.properties

Completed, thanks

PR #15479, opened 2026-08-26
Update groupscape

LiamMorganDev:
Update GroupScape to the latest plugin commit (65bcd1b).


Alexsuperfly:
you cannot [programmatically fill in the players name](https://github.com/MayzrDev/groupscape-plugin/blob/f8ce9a9d25e5f6afff681a2caed819812fe87759/src/main/java/com/groupscape/GroupScapeTrackerPlugin.java#L433) to join a house to, copying to clipboard is fine though.

MayzrDev:
We can't due to runelite rules? And is this all inputs or just player name ones?

Alexsuperfly:
RL rules, any input, the only approved usage of that script so far is in grand exchange searches afaik.

MayzrDev:
Thanks, updated

PR #15510, opened 2026-08-26
Add Gielinor Safari

WoodyCode23:
This plugin is a way to collect and train NPCs that spawn randomly throughout the game to battle gyms and conquer cities. This can be played with friends via the party plugin for better rewards. The plugin learns new NPCS and all of their animations as they happen in game.

### DIFF
--- a/plugins/gielinor-safari
+++ b/plugins/gielinor-safari
@@ -1,2 +1,2 @@
 repository=https://github.com/WoodyCode23/gielinor-safari.git
-commit=296bf03e2470202bc48447098ecf7cc51158e7ec
+commit=4af21076aac2605dc53a4fc310087a8440e096d1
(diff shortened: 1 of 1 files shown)


Alexsuperfly:

WoodyCode23:
Fixed in 7b9a3d0, manifest bumped. Focus is now tracked with a listener on the spinner's own editor component instead.

Alexsuperfly:
the [Build and Deploy](https://github.com/WoodyCode23/gielinor-safari/blob/7b9a3d07054d5bbccab311437a560c30bc75beb2/README.md#build-and-deploy) section of your README is not even close to correct, remove it to not confuse users

WoodyCode23:
Removed in 4af2107, along with the internal design docs the README linked. Those described an earlier server-backed design that was dropped in favour of RuneLite's party system, so they were stale as well as confusing. Also went through the source and rewrote leftover development comments into explanations.

Alexsuperfly:
file io must be constrained to a subdirectory specific to your plugin within the RUNELITE_DIR

WoodyCode23:
Done in 526ef53, everything now lives under gielinor-safari/ in the RuneLite directory with the old files migrated across.

Also went through recent hub reviews and fixed two things preemptively in 2088159: the LICENSE is now the plugin-hub one with my attribution, and the wiki image lookup no longer trusts a file name out of an API response.

PR #15542, opened 2026-08-27
update ironpath

gabworknestio-beep:


coopermor:
The authors list in the plugin-hub file is a list of GitHub usernames authorized to submit updates. There's no need to change it. Only updating runelite-plugin.properties is sufficient.

PR #15545, opened 2026-08-27
Update homeassistant

db1996:
- Added live skill XP updates
- some minor bug fixes


Alexsuperfly:
use build=standard

db1996:
I changed it to standard now, thanks!

PR #15558, opened 2026-08-27
Add Coffer

Paul2357:
Coffer is a Grand Exchange flipping assistant: ranked flip suggestions, a per-slot plan, open-position coaching, a watchlist, high-alch and crash finders, and automatic P&L tracking.

It requires a free Coffer account and submits the player's Grand Exchange offers to the Coffer server for flip suggestions and cross-account fill data. This is disclosed in the plugin description, the README, and a `warning=` line in the manifest.

Source: https://github.com/Paul2357/coffer

### DIFF
--- a/plugins/coffer
+++ b/plugins/coffer
@@ -1,3 +1,3 @@
 repository=https://github.com/Paul2357/coffer.git
-commit=6ef656da6060f5833261b76210a10470a38365e7
+commit=a4e0c9f6c6976926a1d3ba10903233b61333193d
 warning=This plugin requires a (free) Coffer account and submits your Grand Exchange offers to a 3rd-party server not controlled or verified by RuneLite developers, for flip suggestions and cross-account fill data.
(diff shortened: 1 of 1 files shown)


Alexsuperfly:

Paul2357:
Thanks for the review! Both fixed in the latest commit (`a4e0c9f`) — replaced the session-restore retry loop's `Thread.sleep`/`Thread.interrupt` with a self-rescheduling `executor.schedule` chain. No blocking sleep or interrupt remain in the plugin.

Alexsuperfly:
[all file access](https://github.com/Paul2357/Coffer/blob/e7c242dcce5b647070300de355bda46f7261282e/src/main/java/com/gecopilot/client/auth/TokenStore.java#L17) must be within your plugin specific subdirectory of RUNELITE_DIR

Paul2357:
Fixed in the latest commit — the session token is now stored under `RUNELITE_DIR/coffer/` instead of a home-directory path. All plugin file access is within the plugin's RUNELITE_DIR subdirectory now.

PR #15580, opened 2026-08-27
Update Fitzgerald.gg

Fitzgerald-gg:
Adding local support and the inclusion of skill stat tracking.

### DIFF
--- a/plugins/fitzgerald
+++ b/plugins/fitzgerald
@@ -1,4 +1,4 @@
 repository=https://github.com/Fitzgerald-gg/fitzgerald-runelite.git
-commit=c7546de11eb0ae3cdb896af49183d76ee6786471
+commit=827fbe0ad01cf1623e4a474fbaeca7a179e59613
 authors=Fitzgerald-gg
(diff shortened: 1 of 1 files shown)


Alexsuperfly:

Fitzgerald-gg:
Sorry. I've removed the direct java.awt.Desktop use. Since LinkBrowser.browse only accepts http/https schemes and this button opens a locally-generated .html file, I've routed it through LinkBrowser.open(path) instead — is that okay? It seems to be the RuneLite helper for local paths, and it launches the file in the default browser via the wrapped OS opener. The cloud variant of the same button already uses LinkBrowser.browse. Updated the plugin commit accordingly.

Let me know if I need to change my approach.

Alexsuperfly:
you cannot use LinkBrowser.open

ldavid432:
LinkBrowser.open is not allowed either. If you want you can have a button copy the URL to the clipboard, that is what I have seen most plugins do instead.

Fitzgerald-gg:
Thanks for the pointer - I have amended to that approach.

In local mode the button now copies the page link to the clipboard as you suggested, rather than opening anything directly; cloud mode still opens the profile URL via LinkBrowser.browse.

PR #15583, opened 2026-08-27
Add World Examine

OSRS4LYFE:
Adds World Examine, a RuneLite plugin that displays normal OSRS examine responses above NPCs, objects, ground items, and interface items.

### DIFF
--- a/plugins/world-examine
+++ b/plugins/world-examine
@@ -1,0 +1,2 @@
+repository=https://github.com/OSRS4LYFE/World-Examine.git
+commit=0c08c8dedd291d48f93bf4dcf5f151b22cc83bde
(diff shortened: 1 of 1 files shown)


ldavid432:
Have you seen https://runelite.net/plugin-hub/show/examine-tooltip ? Seems like it might do what you want already

OSRS4LYFE:
> Have you seen https://runelite.net/plugin-hub/show/examine-tooltip ? Seems like it might do what you want already

Pretty close, yeah. The main difference is that World Examine renders the text directly above the NPC, object, or item and keeps it anchored to that target as it moves, rather than displaying the examine near the cursor. I was aiming for a more in-world, overhead-text style.

OSRS4LYFE:
Surprisingly more immersive!

ldavid432:
It can also do that but, that's not blocking this from being added or anything.

OSRS4LYFE:
I was playing with that plugins settings before making mine! Was just able to configure mine more to my liking and was hoping to share with friends :)

PR #15587, opened 2026-08-28
Update ClanHQ events tab compatibility

jewatson1994:
## Summary\n- update ClanHQ to the latest standalone plugin commit\n- preserve the configured clan name when the events tab encounters a temporary API/parsing error\n- accept legacy event-list payloads so current SOTW/BOTW entries continue to render\n\nThe manifest uses the full 40-character commit SHA required by Plugin Hub packaging. The standalone plugin branch contains the corresponding committed changes.\n\nValidation: git diff --check passes. Gradle validation was attempted, but this environment could not resolve the RuneLite dependency metadata because the wrapper download failed certificate validation.

### DIFF
--- a/plugins/clanhq
+++ b/plugins/clanhq
@@ -1,2 +1,2 @@
 repository=https://github.com/jewatson1994/clanhq-runelite.git
-commit=f27be32c686d918ddddd8d7a5e185ae8aa6fe3aa
+commit=d2ed8149261ba41d11d3495cc3c83d550c7fd901


jewatson1994:
The latest commit is now over 30 minutes old. Could the Plugin Hub checks be retried?

coopermor:
Your plugin was just updated 6 hours ago so it is on cooldown to be reviewed.

jewatson1994:
I did not realize there was a cooldown. I just though PR activity would trigger the bot and the bot would see its been > 30 minutes. Sorry about that. What is the cool down typically? Just so I know for the future

ldavid432:
We do not publicize that information and it shouldn't matter anyway. Just make a PR and wait, it'll get merged when it gets merged

PR #15593, opened 2026-08-28
Add Resize Window Smaller

Jordan-Hn:
Lowers RuneLite's 765x503 minimum window size so the client window can be dragged smaller. ClientPanel hardcodes its minimum to Constants.GAME_FIXED_SIZE, which propagates through ClientUI's layout manager into Window.setMinimumSize() and becomes the window manager's size hint. The plugin replaces it with a configurable value and calls the public ContainableFrame.revalidateMinimumSize().

No reflection, no native code, no third party dependencies. Does not move, focus, or minimize the window. Restores the stock minimum on shutdown.

### DIFF
--- a/plugins/resize-window-smaller
+++ b/plugins/resize-window-smaller
@@ -1,0 +1,4 @@
+repository=https://github.com/Jordan-Hn/RuneLite-Resize-Window-Smaller-Plugin.git
+commit=47da2a7cddf2ad8fa28b21f5a86a4f591291b065
+authors=Jordan-Hn
+warning=This plugin lets the client window be made smaller than RuneLite normally allows. Small enough sizes can make the game render incorrectly or become unstable.


coopermor:
Is there a reason you can't just change the scale to a number less than 1? https://github.com/runelite/runelite/wiki/RuneLite-Launcher-Configuration#Scale

Jordan-Hn:
> Is there a reason you can't just change the scale to a number less than 1? https://github.com/runelite/runelite/wiki/RuneLite-Launcher-Configuration#Scale

Scale lowers the floor but doesn't remove it. --scale=0.5 halves every Java 2D unit, so the 765x503 minimum lands at roughly 383x252 physical pixels, and you still can't drag below that. Changing it means editing launcher config and restarting, rather than just dragging the window.

The results also differ. Scale shrinks the whole client proportionally, sidebar and config panels included, and the game still lays out at 765 logical px, so you get the same view at reduced resolution. This leaves rendering at 1:1 and lowers the minimum instead, so the game reflows its interface for the narrower viewport and text stays at native size.

Different problems: scale is "make everything smaller", this is "let the window be smaller".

PR #15597, opened 2026-08-28
Create herb-harvest-guard

iwakeman:
Adds Herb Harvest Guard.

Prevents players from accidentally harvesting herb patches without their configured yield-boosting items.

Supports:
- Magic secateurs equipped or in inventory
- Optional Farming cape requirement
- Replaces the Pick option with "Missing yield boost" when requirements are not met
- Blocks the harvest action and displays a chat warning


Alexsuperfly:
remove the build folder

PR #15610, opened 2026-08-28
Add Expose Interface State

osrshool222:
## Plugin submission

This PR adds the `expose-interface-state` plugin.

The plugin exposes a read-only JSON snapshot of selected local RuneLite interface state on `127.0.0.1`:
- Bank open state
- Grand Exchange slot, buy, sell, quantity-input, and price-input windows
- Active interface tab
- Configurable custom prefix and last-update timestamp

The plugin does not automate gameplay and does not expose player names, coordinates, credentials, or other player data. The plugin repository includes a BSD-2-Clause license, Java 11 configuration, tests, and no runtime dependencies beyond RuneLite.

Validation completed locally with `gradlew clean build`.

### DIFF
--- a/plugins/expose-interface-state
+++ b/plugins/expose-interface-state
@@ -1,0 +1,2 @@
+repository=https://github.com/osrshool222/expose-interface-state.git
+commit=bd76f933a01623ea389efd97dfc9f8747b474579


raiyni:
I don't see this plugin having a legitimate use outside of botting

osrshool222:
That is a fair concern. I wanted to extend the existing [Expose Bank State plugin](https://runelite.net/plugin-hub/show/expose-bank-state), which already exposes read-only local interface state over localhost.
The additional states are intended for legitimate local integrations such as accessibility tools, companion overlays, stream/control interfaces, and development/testing tools.
The plugin only reads visible RuneLite widgets and exposes the latest snapshot on 127.0.0.1. It does not click, type, move the player, automate actions, or send data to an external server. It also does not expose the character name, coordinates, credentials, or other account data.
I understand that interface-state data could be misused. If preferred, I would be happy to contribute these additions directly to Expose Bank State instead of maintaining a separate plugin, to avoid fragmentation.

raiyni:
That plugin exists for a streamer to hide their bank through an OBS overlay and their usage can be verified. What is your usage from a human's perspective?

PR #15645, opened 2026-08-29
Update Rogue Chests FC to v3.2

tcourter1:
Added automatic Global Ban List syncing.
Expanded the nearby player counter to show both Enemies and FC members.


Alexsuperfly:
[this](https://github.com/tcourter1/rogue-chests-fc/blob/99a3b0e3a56300bb12c0ca9e5e5771f55ef8dc75/src/main/java/com/roguechestsfc/RogueChestsFcPlugin.java#L151) obviously doesn't actually do anything, nor will it be allowed. i have to tell what it is to review it and i don't want to compile the plugin to do so.

tcourter1:
> [this](https://github.com/tcourter1/rogue-chests-fc/blob/99a3b0e3a56300bb12c0ca9e5e5771f55ef8dc75/src/main/java/com/roguechestsfc/RogueChestsFcPlugin.java#L151) obviously doesn't actually do anything, nor will it be allowed. i have to tell what it is to review it and i don't want to compile the plugin to do so.

It’s the URL for the Google Apps Script endpoint that syncs the global ban list from our staff-maintained Google Sheet. It’s obfuscated only to avoid leaving the endpoint and token plainly readable in the source; the plugin deobfuscates them at runtime before making the request. I understand this isn’t secure storage, just light obfuscation.

edit: I removed the obfuscation.

PR #15651, opened 2026-08-29
Add Chat World Hop

sulfight:
Adds a menu entry to hop worlds directly from chat.

When a player's message in public, clan, friends, group ironman or
private chat mentions a world (w302, w 302, world 302), right-clicking
the message adds a "Hop to world 302" entry. A message mentioning
several worlds gets one entry per world.

This is handy for e.g. hopping to shooting star worlds that are scouted
in a friends chat, or for multi pking to call out a world.

Only worlds present in the live world list are offered, and the world
you are already on is skipped. The hop itself uses the same
openWorldHopper/hopToWorld path as the built-in World Hopper, so the
game's own warnings for PvP, high-risk and members worlds still apply.

Repository: https://github.com/sulfight/chat-world-hop

### DIFF
--- a/plugins/chat-world-hop
+++ b/plugins/chat-world-hop
@@ -1,0 +1,2 @@
+repository=https://github.com/sulfight/chat-world-hop.git
+commit=71fe61dcf5e326aa8bc625728b4077e46f0ccbc1
--- a/plugins/chat-world-hop
+++ b/plugins/chat-world-hop
@@ -1,2 +1,2 @@
 repository=https://github.com/sulfight/chat-world-hop.git
-commit=71fe61dcf5e326aa8bc625728b4077e46f0ccbc1
+commit=e2a850750eef8377727d7bea78fce72304ccae03


Alexsuperfly:
Core used to have this feature but it had to be removed since it ended up triggering anti-cheat protections, due to this we cannot accept it as a hub plugin.

sulfight:
Ah that's unfortunate. It's because of the context menu I assume? If I instead had a side panel like the other world hop plugin but that filters the clickable sidebar worlds to only those from recent chat messages would that be alright?

sulfight:
@Alexsuperfly hmm for the removal are you talking about this?: https://github.com/runelite/runelite/commit/c4a9a1165

I'm curious why this wouldn't be allowed, since we have "hop-to" functionality in other spots like how the world hopper plugin adds the ability to hop to users in clan chat, and it also adds a context menu option to hop to the user's current world: https://github.com/runelite/runelite/pull/4761

The only difference with my plugin is parsing the messages to get the world instead of having it be "the user who sent the message's world". Is there any explicit clarity somewhere on that not being allowed?

sulfight:
Ah ok I think I get it now, it's that the anti-cheat freaks out at the context menu for chat messages hop-to but the "hop-to" for friends list, friends chat, etc is fine. That's unfortunate hmm.

PR #15656, opened 2026-08-29
update osrs-fliphub

zFallan121:
Bumps `commit=` for `osrs-fliphub` to pick up an updated README.

No plugin code changed between the old and new commit — the diff is the README plus nine images added under `docs/`, so the built jar is identical. The README now documents the Grand Exchange offer-prompt suggestions accurately (previously it described them as firing during item search and showing both buy and sell prices at once, which was wrong) and shows the plugin panel in-client.

Images are committed to the plugin repo and referenced by relative path, so they resolve against the pinned commit.


zFallan121:
Closing this for now - reworking the plugin panel UI, so the README screenshots need reshooting against the new design. Will open a fresh PR once that's done. Apologies for the noise.

PR #15660, opened 2026-08-29
Add magic-imbue-necklace-reminder

Guge94:
Warns you if Magic Imbue is active without a binding necklace equipped while you're carrying one, so you stop losing 50% success rate on combination runes.


ldavid432:
Get rid of the .txt file extension on your manifest file, it should have no extension

PR #15667, opened 2026-08-29
update raids-last

richardverheyen:
Updates the raids-last plugin to the latest commit.

## What's new
Adds cross-player resolution for all commands (`!coxlast`, `!toblast`,
`!toalast`, `!raidslast`, `!raidslasttob`). Previously each command only
resolved for the player who typed it, since raid loot is only ever tracked
by that player's own client. Now, right as a player's own command is about
to send, their client publishes their current last item per raid to a
short-TTL (20s) Upstash Redis key, holding the message until that publish
completes - guaranteeing the data exists before any other client can
receive it. Anyone else resolving that command reads it back the same way.
Nothing persists past the TTL.

This requires the other player to also be running this version of the
plugin, and resolution now depends on that relay being reachable - both
documented in the README, along with the tradeoff of the relay's access
token necessarily being embedded in the plugin jar.

Repo: https://github.com/richardverheyen/raids-last
Commit: 82ab1afd4f95e6b1f7e4f286f8b280cc924f630d

### DIFF
--- a/plugins/raids-last
+++ b/plugins/raids-last
@@ -1,2 +1,2 @@
 repository=https://github.com/richardverheyen/raids-last.git
-commit=82ab1afd4f95e6b1f7e4f286f8b280cc924f630d
+commit=18f2ea6fbe6818ddce945bbdd42f3f95a24eab78
--- a/plugins/raids-last
+++ b/plugins/raids-last
@@ -1,2 +1,3 @@
 repository=https://github.com/richardverheyen/raids-last.git
-commit=18f2ea6fbe6818ddce945bbdd42f3f95a24eab78
+commit=2f4ea810aff1212081c7351104ad1c94d0d085c8
+warning=This plugin submits your IP address, and may submit various account data, to a 3rd-party server not controlled or verified by RuneLite developers.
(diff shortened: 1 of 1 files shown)


ldavid432:

richardverheyen:
Thanks for the review - fixed: removed the retry loop in RaidsLastRelay.fetch() that used Thread.sleep/Thread.currentThread().interrupt(). On reflection it wasn't needed anyway - the publish-before-send ordering on the write side already guarantees the data exists by the time it'd be read, so a failed fetch is a genuine failure rather than a race to retry around. Pushed as 18f2ea6.

Alexsuperfly:
`publishToRelay` must be gated behind a new config option thats default off with a `warning=This plugin submits your IP address, and may submit various account data, to a 3rd-party server not controlled or verified by Runelite developers.`

richardverheyen:
Thanks - addressed: publishToRelay (and the lookup path, which also reaches the relay) is now gated behind a new plugin setting, `crossPlayerLookups` ("Resolve other players' commands"), default off. With it off, no request to the relay is made in either direction and the plugin behaves exactly as it did before this feature existed (local-only resolution). The config item carries the standard warning text, and I've added the same warning to this manifest entry.

Checked a comparable plugin (Meteor Reporter, currently under review in #15668) for the expected pattern here - matched its approach: a default-false `@ConfigItem` boolean with `warning = "..."`, guard clauses at the top of every method that reaches the network, and the same warning duplicated in the hub manifest.

Pushed as 2f4ea81.

richardverheyen:
when can i expect an update on this one?

PR #15673, opened 2026-08-29
update impling-finder

Koopy98:
Adds a location filter system,193 per-location config toggles under a new collapsible "Filter Locations" section, letting users hide implings from specific named areas permanently across restarts. Also recovers the Drops reference panel


ldavid432:

PR #15699, opened 2026-08-30
Add Fullscreen toggle

windowwasher7:
adds an option to toggle fullscreen like a browser.


riktenx:

> Plugins that manipulate the client's window, including, but not limited to, minimizing or changing which window is currently focused.

PR #15706, opened 2026-08-30
update osrs-tcg

Azderi:
Updates OSRS TCG to v1.0.
Moving from a local-only beta to a server-authoritative model backed by osrs-tcg.net.


Adam-:
Replace your use of Desktop.browse with LinkBrowser.browse

Azderi:
> Replace your use of Desktop.browse with LinkBrowser.browse

removed, it was used as a fallback only

Adam-:
This exceeds the 200k token limit

Azderi:
@Adam-  does the review bot look at diff and related code or the entire thing in one? The plugin has had a big refactor and not much of the original is kept so I'm slightly worried I'd be stripping away too many features that users have had in the beta.

coopermor:
Every review is the entire source. Diffs are not considered.

Alexsuperfly:

PR #15715, opened 2026-08-30
bank-architect: 0.3.0

PKoka5:
Follow-up to #15681, which merged while this was still being worked on.

The row-filling option covered two different decisions under one switch: whether the four combat-style gear columns may be held straight with filler items, and whether a part-finished Herblore recipe row may be padded out. A player who wanted the first had no way to decline the second.

They are the same mechanism but not the same trade, so this splits them into two options, both defaulting to on. Nothing else changes.

- 932 tests, 0 failures
- Bank simulation 150/150 scenarios completed
- `gradlew build` green

### DIFF
--- a/plugins/bank-architect
+++ b/plugins/bank-architect
@@ -1,2 +1,2 @@
 repository=https://github.com/PKoka5/ironman-bank-architect.git
-commit=d5631c52bedfee228d95627d7f7e10b51a8648af
+commit=0e16d24293ad1ac0ff0d804bae17b8c5869de378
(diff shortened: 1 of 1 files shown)


PKoka5:
Updated this PR to 0.3.0 rather than opening a new one, since 0.2.1 had not been reviewed yet.

On top of the row-filling split that was already here, 0.3.0 adds:

- an off-by-default option to leave already-correct bank slots uncoloured, so the guide can stay on permanently;
- item classification from the OSRS Wiki's own category membership, fetched at development time into a pinned TSV. The plugin still makes no network calls and still classifies fully offline.

PR #15722, opened 2026-08-30
Update hiscore-notifications

malafel-dev:
Fixes:
- Players near the top of a boss leaderboard get the same notification after every kc .
- Boss leaderboard messages by default show the kc of the person who was passed on the leaderboards, which is unclear.

Adds:
- Notifications via in game chat message.
- Ability to get an up to date list of bosses without updating the plugin.
- Change from scraping the hiscore webpage to using the hiscore API provided by Jagex.


malafel-dev:
It looks like I'm using an old compiler locally, so I didn't get this error. Working on it.

PR #10707, opened 2026-02-23
Add PVP Entity Hider plugin

zuest:
## PVP Entity Hider

Combat-aware entity hider for PVP — hides other players while keeping combat targets visible.

### Why not Entity Hider?

This plugin automates what's **already achievable manually** with Entity Hider today:
a user can add their opponent as a friend, enable "Hide Others" with "Hide Friends" off,
and Entity Hider shows only that opponent. This plugin removes the manual friend-list
step by detecting combat interactions directly.

Key differences from Entity Hider:
- **Combat linger timer** — keeps opponents visible for configurable seconds after combat
  ends (default 7s), useful in multi-combat so targets don't vanish instantly
- **Health bar validation** — only registers actual combat (requires visible health bar),
  filtering out follows, trade attempts, and safe-zone clicks
- **Filtering options** — show/hide by friends, clan, FC, skull status, team cape,
  combat level range, and tile distance

### No menu modifications

This plugin **only affects visual rendering** (player model visibility via `shouldDraw`).
It does not modify right-click menus, attack options, or the minimenu in any way.

### Features
- Show your combat target
- Show players attacking you
- Show all players in PvP combat (opt-in, off by default)
- Combat linger timer (0–30s, default 7s)
- Filter by: friends, clan, FC, skull, team cape, combat level, distance

PR #15743, opened 2026-08-30
Add Combat Entity Hider

younghans:
Combat Entity Hider is a combat-aware replacement for Entity Hider.

Hide attackers overrides Hide others: when Hide attackers is disabled, active and recently detected player attackers remain visible even when Hide others is enabled.

Recent attackers remain visible for a configurable 0–60 second grace period, defaulting to 15 seconds. The plugin also declares conflicts with Entity Hider, Auto Entity Hider, and Dynamic Entity Hider so multiple hiders do not compete over rendering.

Tested successfully in-game and with the project test suite and RuneLite Plugin Hub standard packaging build.

This is inspired by Dead Man Mode breaches. So when you're hiding players, it still see the player that's attacking you.

### DIFF
--- a/plugins/combat-entity-hider
+++ b/plugins/combat-entity-hider
@@ -1,0 +1,2 @@
+repository=https://github.com/younghans/combat-entity-hider.git
+commit=5c739b3dfee09b5a305e021e1fffe86d5a22bd79


raiyni:
See https://github.com/runelite/plugin-hub/pull/10707

PR #15754, opened 2026-08-30
update chance-man

ChunkyAtlas:
feat: add configurable tool and quest roll systems, centralize locked item handling, optimize item dimming, and stabilize Show Drops/music tab restoration


Alexsuperfly:
the use of client.runscript in MusicWidgetController cannot be verified as an allowed script to run, either remove it or use a static script id

PR #15767, opened 2026-08-31
Create group-ironmen-tracker-ingame-companion

kingkj52:
Hey, this is an unofficial in game companion to the Group Ironmen Tracker plugin, which shows your group's banks, skills, quests, diaries and map positions in client. I love and use that plug-in constantly, and when away from the game, its great to check quick things; however, when actively playing, it can be slightly annoying to have to alt+tab and go to a browser to view the information. That is what this solves.

Reads the groupiron.men API (or a self-hosted instance) with credentials the user supplies, and optionally uploads the user's own data in the same format when the official tracker plugin isn't running (so it should in theory work without the original plugin, but I haven't tested this part thoroughly, so I'd highly suggest using the original as well). Both are disclosed in the config with an explicit warning listing every field sent.

No new dependencies. Adds no menu entries and modifies no game widgets, and the in game views are all plugin drawn overlays. A test in the repo greps the source on every build to enforce that, so it should be good to go.

Optionally drives the Shortest Path plugin via RuneLite's PluginMessage event; no dependency on it, and a no-op if it isn't installed. Clean and simple, used as a way to route to your group mates, nothing more.

Note: tools/mapdump/ has its own build.gradle declaring net.runelite:cache. It is not part of the plugin build, it's a build time generator for the checked in map-areas.json map data. It isn't a subproject, and build=standard replaces the root gradle files regardless.


Alexsuperfly:
LICENSE is invalid, dont explain where the other authors attributions comes from, just attribute them alongside yourself

kingkj52:
Thanks for the shout, done. :)

PR #15778, opened 2026-08-31
Add Trip Tracker plugin

OhansonB:
Loot tracker with trip scoping, GP/hour statistics, and trip comparison.

Features:
- Three view modes: individual drops (list), aggregated by NPC (grouped), and per-session trips
- Trip management with pause/resume, rename, and configurable inactivity timeout
- Side-by-side trip comparison with CSV, JSON, and pretty-print export
- Tracks NPC kills, PvP kills, pickpocketing, raids, minigames, farming harvests, bird houses, and chest loot
- Item sprite and text display modes
- Per-item and per-NPC exclusion via right-click context menu
- Real-time filtering across all views
- Persistent storage per account with automatic data retention limits


ldavid432:

PR #15800, opened 2026-08-31
Add Mouse Tab Switcher

JD-Michelena:
Adds Mouse Tab Switcher, a RuneLite plugin that allows users to assign Mouse 4 and Mouse 5 to supported OSRS interface tabs.

Features:
- Configurable Mouse 4 and Mouse 5 tab assignments.
- Option to disable either mouse button.
- Supports multiple OSRS interface tabs.
- Includes Logout.


ldavid432:

PR #15816, opened 2026-09-01
Overhaul Plugin Foundation

ZH-Systems:
Complete overhaul to add multiple features under one banner hence the size.


Alexsuperfly:
`too many tokens: 202899` you have to be under 200k for our review bot to run your plugin, otherwise it requires manual review.

coopermor:
Still too many tokens. 201786

riktenx:
* use of DiscordUser stuff is not allowed

ZH-Systems:
@riktenx I saw an email notice that you left a comment on the PR but when I check there is nothing from you, am I going crazy?

ldavid432:
You can ignore the comment/email, we discussed as a team and decided that was not valid.

riktenx:
use getResourceAsStream, not getResource

pajlada:
hello, not really a reviewer comment, but:
it seems like your repository contains 2 plugins, that makes it difficult to review.
additionally, it seems like you're reusing Dink's config key. you should probably not do that

pajlada:
your plugin still contains two plugins:
 - https://github.com/ZH-Systems/TangleDinkPlugin/blob/8f003606a44149c0ce4c168a965c1842313dc624/src/main/java/tccrewplugin/TangleCrewSyncPlugin.java#L26
 - https://github.com/ZH-Systems/TangleDinkPlugin/blob/8f003606a44149c0ce4c168a965c1842313dc624/src/main/java/tccrewplugin/TcCrewPlugin.java#L83

you shouldn't keep unused code around if you can avoid it, either move it to a separate branch or keep it in git history

PR #15819, opened 2026-09-01
Add Stations Party Pings

StationEarthxo:
## Summary
- Adds a League-inspired eight-direction radial ping wheel for RuneLite parties
- Shares user-selected tiles and ping types through RuneLite's built-in PartyService
- Includes configurable wheel size, duration, sounds, labels, icon size, halo, tile marking, and standard tile-ping fallback
- Uses bundled resources only and has no third-party runtime dependencies or servers

## Testing
- `./gradlew test`
- Verified two development clients in the same RuneLite Party can exchange custom pings in both directions
- Verified the sender renders locally without relying on websocket echo
- Targets Java 11 and uses the standard Plugin Hub build

### DIFF
--- a/plugins/stations-party-pings
+++ b/plugins/stations-party-pings
@@ -1,0 +1,2 @@
+repository=https://github.com/StationEarthxo/stations-party-pings.git
+commit=ce10ee1b40893127dfa413002c41f10ae43d6303


ldavid432:
I don't think we can allow the use of league images in your plugin

Per Riot's policy that you mentioned:

> We prohibit the use of our IP in games and apps

I don't know if legally there is a difference between a full on game and a plugin for a game but I don't think we want to test that

riktenx:
closing per above ip issue

PR #15825, opened 2026-09-01
Create doom-delve-advisor

James-K-Lawson:
Adds Doom Delve Advisor, a display-only claim-vs-descend advisor for the
Doom of Mokhaiotl delve. It reads the reward screen and chat to estimate
the expected value of pushing versus banking, and shows a recommendation
plus a per-level stats panel. It never interacts with the game.


coopermor:
Use `build=standard` to be eligible for bot review

PR #15833, opened 2026-09-01
update to 1.0.60

iodrareg:
- Type !p followed by a message to chat with your party.
- Type !p alone to toggle party chat mode on/off.
- RuneLite chat commands work inside party chat.
- Choose where party messages appear and customize the prefix.
- Messages from blocked players are hidden.
- Type !osparty to advertise your party in public, friends, or clan chat.
- Other OSParty users can click the chat line to apply for the party.
- Applications use the same checks and role picker as the Search tab.
- Clickable !osparty lines can be disabled.
- Parties created at the CoX board, ToB notice board, or ToA obelisk can be advertised automatically.
- Party details are filled from the game and your previous raid settings.
- ToB and CoX let you choose your role directly from the advertisement.
- Choose between Ask me, Always advertise, or Off.
- CoX and ToA ads can automatically follow in-game scale or invocation changes.
- In-game board syncing can be disabled.
- Use Find me a party to get offers for matching parties.
- Offers can appear as a sidebar banner, in-game card, or both.
- Creating a party now checks for existing parties doing the same activity and offers to join them first.
- Handle party invites, join requests, and party offers directly from chatbox cards.
- Cards show applicant stats, KC, and warnings.
- Accept or decline from either the card or side panel.
- The old pending-applicant corner overlay has been removed.
- Customize the defence tracker with current level, current/starting level, percentage, or combinations.
- Choose how drained defence is displayed.
- Customize magic-defence information.
- Choose levels or percentage drained for thresholds.
- Optionally show the full defence level for monsters with a defence floor.
- Optional sounds are available for ready checks, kicks, map pings, and friends-chat requests.
- All event sounds are off by default.
- Party members with Vengeance active show the spell icon on their model and in the Party tab.
- The Vengeance icon disappears once it is consumed.
- Ad reports now allow up to 500 characters explaining what is wrong.
- All ironman variants, including unranked group ironmen, are now correctly recognized for ironman-only parties.
- Other players' account identifiers no longer leave the server.
- Board cards and rosters now identify players using anonymous OSParty IDs.
- New settings cover raid party advertising, raid party prompts, following the in-game board, checking for similar parties, party-finding notifications, party chat, chat prefixes, event sounds, and defence tracker display options.

### DIFF
--- a/plugins/osparty
+++ b/plugins/osparty
@@ -1,4 +1,4 @@
 repository=https://github.com/osparty/osparty.git
-commit=2756a5751854bab678de974f3908cde33e9a9cac
+commit=8e853493f4a9d4de6588b5b589dcbd708dbb1a15
 authors=iodrareg
 warning=This plugin submits your IP address to a 3rd-party server not controlled or verified by Runelite developers.


Alexsuperfly:
`too many tokens: 208776` you have to be under 200k for our review bot to run your plugin, otherwise it requires manual review.

iodrareg:
> `too many tokens: 208776` you have to be under 200k for our review bot to run your plugin, otherwise it requires manual review.

Ok then put it to manual review, we have more features in the pipeline to be worked out and can hardly cut lines  and corners in our codebase without going in against the guidelines.

coopermor:
I don't think any human reviewer will have time to review your 10k line diff anytime soon.

GDreijer:
> I don't think any human reviewer will have time to review your 10k line diff anytime soon.

What size diff would be reasonable for you to review within, say, 1–2 weeks? We could start releasing smaller updates if that would make things easier.

Also, is there any way we could get access to a larger context window, either through an exception or by paying for it?

iodrareg:
@coopermor and/or @Alexsuperfly if this needs to be chopped up in bitesize pieces of 500 lines of diff per MR i'm fine with that. If that allows us to expand our plugin more, and you only have to look at the diff, that's fine by me. Let me know your thoughts.

Alexsuperfly:
with the amount of submissions we are receiving, there is no capacity for full manual review.

GDreijer:
I don't think "there is no capacity for full manual review" actually answers the problem we're trying to solve.

So far, we've been told:

* The automated review cannot process our update because it exceeds the automated reviewer's \~200k token context limit.
* Because of that, the update requires manual review.
* Manual review is currently not feasible because there is no capacity to review a diff of this size.

We've repeatedly tried to find a workable middle ground. We've asked what diff size would be considered reasonable, asked whether there are alternative review approaches, and explicitly offered to split the work into small, independently reviewable PRs, even suggesting something around 500 lines per PR if that would make the review process manageable.

We haven't really received a concrete answer on whether that approach would actually give us a path forward.

From our side, this is currently a dead end: once a plugin grows beyond the automated review limit, the apparent answer is neither "split it into smaller changes" nor "have it manually reviewed."

We understand that reviewing a 10k+ line diff isn't realistic, and we're not asking anyone to do that. **We're developers and we're perfectly willing to change how we structure and submit our work if that makes the review feasible.** What we're asking for is a defined process that allows us to do that.

OSParty is an actively developed plugin, and we have a substantial amount of functionality already implemented as well as further development planned. We don't want to start stripping functionality or deliberately making the codebase worse simply to get under a context limit if there is a better way to structure the review.

We're willing to:

* restructure the release;
* split the changes into smaller, logically scoped increments;
* rewrite/rebase the submission history if necessary;
* keep individual changes well below whatever size you consider reviewable; or
* follow another review process if one exists.

**What we need from the maintainers is simply a concrete answer as to which of these approaches is actually acceptable and will give us a path to getting OSParty reviewed and merged.**

If a sequence of smaller, under-limit PRs is acceptable, we're happy to do that. If it isn't, could you please explain what the expected route is for a plugin that has legitimately grown beyond what the automated reviewer can process?

We're trying to resolve this constructively, but at the moment the practical outcome is that **OSParty is blocked indefinitely: the automated reviewer cannot process the plugin, while the manual-review route is unavailable due to capacity.** If neither route is viable, we need to understand what mechanism the Plugin Hub expects developers in this situation to use. Otherwise, there is effectively no path forward for a plugin once it reaches this size.

PR #15841, opened 2026-09-01
re-enable object-name-markers

jtclowner:
Re-enables Object Name Markers after **removing configurable tile-radius expansion** beyond the tiles occupied by matched objects. This removes the ability to configure the plugin to show an object's surrounding tiles to "automatically indicate where to stand or not stand", bringing the plugin in line with Plugin Hub guidelines.

The updated plugin:

- Limits tile highlights to tiles occupied by the matched object.
- Accepts legacy `name:<number>` configurations for compatibility with existing user settings, but ignores the numeric suffix and applies no expansion.


jtclowner:
@Nightfirecat Does this radius removal change satisfy rules/requirements for re-enablement?

PR #15857, opened 2026-09-01
Add Bank Tab Storage

TimoklDev:
Bank Tab Storage is a plugin that allows the user to easily store, hide, show, import, and export bank tabs that utilize the bank tags plugin.

It stores the data in the active runelite profile, and only accesses the clipboard when following explicit user actions (import/export).
It does not make any network requests.

This plugin was made because the tag tabs were getting overwhelming for me, i had so many of them and after trying to manage a spreadsheet i figured making a plugin for runelite would be the better option.

### DIFF
--- a/plugins/bank-tab-storage
+++ b/plugins/bank-tab-storage
@@ -1,2 +1,2 @@
 repository=https://github.com/TimoklDev/bank-tab-storage.git
-commit=b5e89e7bfba81f43f4a45a8f813075f662feb177
+commit=c0c27185ffe9c4635c1708299ec9b9cf7c5160e6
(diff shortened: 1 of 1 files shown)


ldavid432:
Your license is invalid, it looks like you shoved the bottom part all onto one line. It should match the hubs: https://github.com/runelite/plugin-hub/blob/master/LICENSE

TimoklDev:
> Your license is invalid, it looks like you shoved the bottom part all onto one line. It should match the hubs: https://github.com/runelite/plugin-hub/blob/master/LICENSE

Fixed the license formatting and updated the pinned commit. Thanks!
Also updated some naming conflicts where some packages still had the older local name.

TimoklDev:
@ldavid432 Anything else i need to do here?
It's my first plugin on the plugin hub so not really sure what the normal procedure is here!

ldavid432:
Just need to wait for further review

PR #15864, opened 2026-09-01
bank-architect: point at the renamed repository

PKoka5:
Repository URL only; the pinned commit is unchanged at `4635b3d`, so this is the same plugin build that is live today.

The repo was renamed from `ironman-bank-architect` to `bank-architect`, matching the plugin id and the display name. GitHub redirects the old URL, so nothing is broken either way — this just makes the marker point at the real address.

The 0.4.0 version bump follows as its own PR once this lands, as requested.


coopermor:
Please make the PR change the repo URL but keep the commit hash the same. Once it is merged you can submit a 2nd PR updating the commit hash to your new version.

PKoka5:
Thanks — done. This PR now changes only the repository URL and keeps `4635b3d`; I will open the 0.4.0 commit bump separately once this is merged.

PR #15874, opened 2026-09-01
Update fitzgerald (Chronicle): local-first rewrite

Fitzgerald-gg:
Updates plugins/fitzgerald. The plugin has been rewritten and renamed from Fitzgerald.gg to Chronicle; same repository, same entry, so existing installs follow the update.

**What changed**
Plugin was originally developed as a client for fitzgerald.gg. The client tracked the counters and events, and the server would enrich item names and prices, derive the per-resource skilling rollups, compute drop rarity and dryness, and backfilled history from the hiscores. Following feedback from users, the plugin has been reworked to be local-first.

The journal is plain JSON under .runelite/chronicle/, the side panel computes every view on the client, and the reference tables it needs are bundled. Nothing is read back from a server; ChronicleApiClient is the only class in the plugin that touches the network, and there is no longer a URL of any kind baked into the source.

Removed outright: screenshot upload and account enrolment. The Cloud/Local mode dropdown is also gone - local journalling is no longer a mode you pick, it is what the plugin does. The network surface went from 12 endpoints to 4 optional upward POSTs.

**Desire to drop the warning= line**
This was added when the plugin's purpose was syncing to my server, and it was accurate then. It no longer is. Cloud sync is off by default with the server field blank, and every one of the four calls is gated on both being set, so a user who never configures a server makes no network calls at all. With it off the panel reads "Journaling locally: nothing leaves this computer. Cloud sync lives under Advanced in the plugin settings."

**The current line reads:** 
This plugin submits your IP address, and may submit various account data, to a 3rd-party server not controlled or verified by Runelite developers.

That is now unconditional about something conditional, and it is shown to every user including the ones who never enable sync.

The disclosure has not gone away, it moved to where it fires. The cloudSync config item carries its own warning attribute, which RuneLite raises as a modal "Are you sure?" dialog defaulting to No before the setting is written, so nobody can turn cloud sync on without reading it:

"This feature submits your IP address, and your own account's activity, to the 3rd-party server you configure below: a server not controlled or verified by the RuneLite developers"

Happy to keep a reworded one if you would rather; my objection is not to warning people but rather the negative implications of the warning when the default no longer operates in that fashion.


riktenx:
please first submit with only your commit change, and then submit a followup that removes just the warning and keeps the (new) commit

Fitzgerald-gg:
Done. This is now the commit bump only. I'll open the warning removal as a follow-up once this is merged if approved 👍

Alexsuperfly:

Fitzgerald-gg:
Removed. Nothing else in the plugin creates a thread or an executor.

PR #15890, opened 2026-09-02
Add Chat-XL

4E-DEVOPS:
Adds Chat XL, a chat readability plugin that lets players use alternative RuneScape fonts while preserving chat wrapping, alignment, and the native chatbox experience.


coopermor:
If there are issues you don't need to close the PR. Just push a new commit to your branch

coopermor:
In Java the image size becomes `LENGTH * WIDTH * 4` so you need to resize the actual dimensions.

4E-DEVOPS (commented):
Relocated /images outside of /src

PR #15925, opened 2026-09-02
Add Slayer Atlas

nifte:
Adds Slayer Atlas: a side panel with info, locations, gear/prayer recommendations, and Shortest Path routing for Slayer tasks.


ldavid432:
Your license is invalid, please make sure it exactly matches the hubs (aside from the author): https://github.com/runelite/plugin-hub/blob/master/LICENSE

ldavid432:
use client.menuAction is not allowed

PR #15941, opened 2026-09-02
Update Kill Clog to 2.1.0

420kc:
Kill Clog 2.1.0

* an optional Character model setting adds a one-click button that publishes your current player and follower to your killclog.com profile
* pinned hover-mode tooltips now suppress new summary popups until the pinned tooltip closes


ldavid432:

PR #15950, opened 2026-09-03
Add Chat-Over-Keybinds

Endrit-alt:
Allows bound keys to still be used in chat. This is a temporary fix until "Press Key to Chat" Feature gets added as a core runelite feature.

Currently you can not keybind a letter, and still use it in chat.

https://github.com/runelite/runelite/pull/20321

### DIFF
--- a/plugins/chat-over-keybinds
+++ b/plugins/chat-over-keybinds
@@ -1,0 +1,2 @@
+repository=https://github.com/Endrit-alt/Chat-Over-Keybinds.git
+commit=e63400c4a8dbf6259d8ae4af718db602f3afcd46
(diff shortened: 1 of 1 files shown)


coopermor:
Reflection isn't allowed on the plugin hub.

Endrit-alt:
It's a temporary solution. It cannot be fixed without reflection, and is supposed to be fixed within runelite core, which can possibly never happen.

I've made a small PR to core that would allow this without reflection  : https://github.com/runelite/runelite/pull/20480

Until then this fixes it (It allows keybinds to still be used in chat.)

Can an exception be made until "key to chat" or "Key event interceptor" is added to core instead of being part of the "Key Remapping" plugin? I can delete the plugin once that happens.

PR #15977, opened 2026-09-03
spoon-meter: new plugin

Dberko:
Adds **Spoon Meter**, which rates how lucky an account's collection log is.

It reads each collection log page as you open it, works out how many uniques that kill count
should have produced from a bundled rate table, and reports the mid-p percentile of a Poisson
model — an account-wide score plus a per-boss breakdown, with the driest missing item called out
for each source.

- Repository: https://github.com/dberko/spoon-meter
- Licence: BSD 2-Clause
- No network access of any kind. It only reads the collection log interface while it is on screen,
  and stores what it read under `~/.runelite/spoon-meter/` keyed by RS profile.

### How this differs from Collection Log Luck

[collection-log-luck](https://github.com/peanubnutter/collection-log-luck) is the closest existing
plugin, and the overlap is worth addressing up front. It uploads/pulls your log via
collectionlog.net and reports luck per item. Spoon Meter is entirely local — nothing leaves the
client — and its output is an aggregate: one percentile for the whole account, and one per boss,
rather than per item. It also models pages whose attempts are counted separately at different
rates: normal vs corrupted Gauntlet (1/120 vs 1/50), the three Dagannoth Kings on one page, and the
normal/hard/expert raid modes, summing expectations across them rather than applying one rate to
the largest counter.

### Notes for review

- Rates live in `src/main/resources/com/spoonmeter/drop_rates.json`, sourced from the wiki. Entries
  whose real rate varies with team size, points or invocation are flagged `approx` and shown with a
  `*`; the raid unique chances are config values rather than assumptions baked into the table.
- Items that drop in stacks are deliberately excluded from the table, because the collection log
  records items received rather than drops — one 132-axe drop would otherwise score as 132 uniques.
- The collection log parser matches on structure (the `Obtained: x/y` line, and the widget holding
  the most item children) rather than hardcoded child ids, so it should survive interface
  reshuffles.
- Tests cover the Poisson/incomplete-gamma maths against known values including large-lambda
  stability, drop table integrity (every item resolves to a real probability, every weight set sums
  to its declared total), and the rating pipeline end to end.

### DIFF
--- a/plugins/spoon-meter
+++ b/plugins/spoon-meter
@@ -1,0 +1,2 @@
+repository=https://github.com/Dberko/spoon-meter.git
+commit=b103698745866112954c26465658cef408b518f5
--- a/plugins/spoon-meter
+++ b/plugins/spoon-meter
@@ -1,2 +1,2 @@
 repository=https://github.com/Dberko/spoon-meter.git
-commit=28c7fe8b8c0fe1db7892366b28151299aed642b5
+commit=71f60e7c269c4dce8b7bcf044991eb937d3127ee
--- a/plugins/spoon-meter
+++ b/plugins/spoon-meter
@@ -1,2 +1,2 @@
 repository=https://github.com/Dberko/spoon-meter.git
-commit=b103698745866112954c26465658cef408b518f5
+commit=28c7fe8b8c0fe1db7892366b28151299aed642b5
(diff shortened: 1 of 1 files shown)


Alexsuperfly:
[auto typing](https://github.com/Dberko/spoon-meter/blob/59cc6c5064e608f1d295c667a74cc062b486c529/src/main/java/com/spoonmeter/SpoonMeterPlugin.java#L292) is not allowed

PR #15989, opened 2026-09-03
add playtime-logger

mreedon:
Logs real login/logout session times to a local CSV (`~/.runelite/playtime-logger/sessions.csv`) — start/end timestamps, duration, world hop count, which worlds were visited, and which account. Session boundaries are tracked off the client's `GameState` transitions directly rather than the built-in played-minutes counter, so a world hop doesn't split one play session into two. No network requests, no dependencies beyond `runelite-client`.

### DIFF
--- a/plugins/playtime-logger
+++ b/plugins/playtime-logger
@@ -1,2 +1,2 @@
 repository=https://github.com/mreedon/playtime-logger.git
-commit=c9016af4ff0ff9740ac70928e5238de519da7e10
+commit=d6a569bc2735f7fb265b3b3f4b4dedd5308ec797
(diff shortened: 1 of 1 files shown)


coopermor:
In startUp you can do something like
```java
clientThread.invokeLater(() ->
{
	if (client.getGameState().getState() < GameState.LOGIN_SCREEN.getState())
	{
		return false;
	}
	if (client.getLocalPlayer() == null)
	{
		return false;
	}
	if (client.getLocalPlayer().getName() != null)
	{
		playerName = client.getLocalPlayer().getName();
		return true;
	}
	return false;
}
```
so you don't need the onGameTick check. The invokeLater will run until it returns true.

mreedon:
Good call, thanks — switched to that pattern and dropped the onGameTick subscription entirely. Verified on-device against a real login + hops + logout session, name captures correctly.

Updated to `402e6c8b2b0685367ee3b6ba2721ba9a741d2682`.

PR #15992, opened 2026-09-03
Update Kill Clog to 2.1.1

420kc:
Kill Clog 2.1.1

* shows duplicate item counts in `!kclog` results
* answers RuneProfile-style `!log` commands when RuneProfile is not active
* uses verified killclog.com sync data for group ironman account badges
* restores Colosseum Glory to solo and comparison PvM summaries


420kc:
Apologies for the noise here. This has been superseded by #16025, which carries these fixes forward as part of Kill Clog 2.2.0.

PR #16009, opened 2026-09-04
Add prayer-alert

ilee2914:
Prayer Alert beeps while you attack an NPC with a combat style its overhead prayer protects against, so you notice you are hitting into protection instead of watching the icon.

It reads the NPC's overhead prayer sprites and your combat style, and plays a sound. No network use, no input, no menu actions, no game state changes.


ldavid432:
We cannot accept this plugin as it is essentially alerting the player about "boss mechanics" which goes against Jagex's 3rd party client guidelines.

PR #16010, opened 2026-09-04
Add Camera Scenes

vahnx:
## Description
Adds Camera Scenes, a RuneLite plugin for saving and loading named camera viewpoints from the side panel or configurable hotkeys.


Alexsuperfly:
setting exact yaw and pitch is disallowed outside of detached camera mode

TeodorLB:
> setting exact yaw and pitch is disallowed outside of detached camera mode

Is "camera pitch limiter" just an exception to that rule then? Since you can set exact pitch with it (set min and max to the same) (although not from a keybind)

Alexsuperfly:
that doesnt set the pitch, it just limits it, it could be configured to only allow one pitch but i think thats not quite the same as setting some exact value whenever you want.  also your commit is invalid

PR #16012, opened 2026-09-04
Lucky penny v0.1

Al-oza:
Tracks the charges, resources, and Grand Exchange value saved by Ghommal's lucky penny now that it can be used universally.

### DIFF
--- a/plugins/lucky-penny
+++ b/plugins/lucky-penny
@@ -1,0 +1,2 @@
+repository=https://github.com/Al-oza/LuckyPenny.git
+commit=0.1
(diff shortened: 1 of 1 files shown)


coopermor:
Is there a reason you're using file I/O rather than RuneLite's [ConfigManager](https://static.runelite.net/runelite-client/apidocs/net/runelite/client/config/ConfigManager.html). Using ConfigManager would sync user's data if they use cloud sync.

Al-oza:
Just familiarity, I’ll go read the API documentation and set it up using the Runelite configManager instead. Are there any other recommendations you have for changes to be made?

Al-oza:
Removed file I/O and replaced with RuneLite's [ConfigManager](https://static.runelite.net/runelite-client/apidocs/net/runelite/client/config/ConfigManager.html). I have also changed the commit hash to be representative of the most up to date version.

PR #16020, opened 2026-09-04
Create friendsmap

P2GR:


Nightfirecat:
Rather than [showing users a consent dialog upon launching the plugin for the first time](https://github.com/P2GR/friendsmap-plugin/blob/6ea4e20d8f6fd85e65f31505c147d79698608006/src/main/java/com/friendsmap/FriendsMapPlugin.java#L538) we would rather you add a `warning` to your plugin so they can accept these terms before installing and running your code:

```
warning=This plugin submits your IP address, RSN, player location, and friends/clan data to a 3rd-party server not controlled or verified by Runelite developers.
```

P2GR:
> Rather than [showing users a consent dialog upon launching the plugin for the first time](https://github.com/P2GR/friendsmap-plugin/blob/6ea4e20d8f6fd85e65f31505c147d79698608006/src/main/java/com/friendsmap/FriendsMapPlugin.java#L538) we would rather you add a `warning` to your plugin so they can accept these terms before installing and running your code:
> 
> ```
> warning=This plugin submits your IP address, RSN, player location, and friends/clan data to a 3rd-party server not controlled or verified by Runelite developers.
> ```

Will change, thank you.

Alexsuperfly:
`java.net.http.HttpClient.newBuilder`: Java's HttpClient is not allowed, @Inject and use RuneLite's OkHttpClient instead

PR #16022, opened 2026-09-04
Update WeCorpCC to v3

shimonos:
Updates WeCorpCC to v3.

Changes include:
- New in-game Corp overlay
- Corp HP display
- Corp target display in Boosting mode
- Client HP display
- DWH / Elder Maul attempt totals
- New voluntary Corp Mass Lobby
- FFA / Split mass advertisements
- Live member count while advertiser is at Corp
- Corp-only advertising checks
- Client AFK handling improvements

The Corp Mass Lobby only shares structured mass information:
world, member count, loot rule, and timing data.

No free-text chat is included.


ldavid432:
use of client.hopToWorld requires manual review, if you remove it the bot auto review could go much faster than that.

shimonos:
Thanks. I’d like to keep client.hopToWorld because the Lobby’s Hop button is an intentional core feature. The user manually clicks Hop to join the advertised Corp mass world; it does not hop automatically without user action. I’m happy to wait for manual review.

coopermor:
Just note this is your 12th version since the plugin was created 17 days ago. This means all future updates will require manual review and may take weeks due to limited reviewer time.

Nightfirecat:
It looks like one of the changes to the plugin is that it now sends requests to your server by default? This is not permitted without some kind of `warning` to the user before this happens, even if their IP address is the only thing which you would get from the request.

shimonos:
I’ve decided to withdraw this update. I’ll remove the external server/lobby functionality and fully test the remaining changes locally before submitting a future update. Thank you for the feedback.

PR #16024, opened 2026-09-04
Add runeassist-flipping

RuneAssist:
RuneAssist Flipping is a Grand Exchange flipping assistant: server-composed
flip suggestions, held-cost/portfolio tracking, and potion-decant support.

- Suggestions are composed server-side via the RuneAssist backend's
  `/v1/suggestion` endpoint, from a wiki-price-based scorer. There is no
  on-device scorer; the plugin sends account state and renders what the
  server returns.
- Flip history is server-owned once a device is linked to an OSRS account.
  Held-cost basis and portfolio valuation are computed server-side too.
- The `warning=` line covers everything sent, none of which is opt-in: coin
  stack size, held Grand Exchange stock with cost basis, GE offers and
  transactions, and IP address.
- Adapted from Flipping Copilot under BSD-2. Attribution is in `LICENSE` and
  `THIRD_PARTY_LICENSES.md` in the plugin repo. It has since diverged
  substantially: its own backend, scorer, decant handling and held-cost ledger.
- `build=standard`; no custom Gradle dependencies.
- Icon is 32x32, under the 48x72 limit.

### DIFF
--- a/plugins/runeassist-flipping
+++ b/plugins/runeassist-flipping
@@ -1,0 +1,4 @@
+repository=https://github.com/RuneAssist/runeassist.git
+commit=1d2adbdcaf397e32b668fe9b1aa5509976326dfa
+warning=This plugin submits your coin stack size, held Grand Exchange stock with cost basis, grand exchange offers, grand exchange transactions, and IP address to a 3rd party server not controlled or verified by the RuneLite Developers.
+authors=RuneAssist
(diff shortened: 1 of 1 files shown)


ldavid432:
`too many tokens: 255545` you have to be under 200k for our review bot to run your plugin, otherwise it requires manual review, which at enough lines to be that many tokens no one will be wanting to do

RuneAssist:
Thanks for flagging. Removed a large amount of unreachable code inherited from the fork's original backend, which brought it under 200k. commit= is repinned to the current tip and the build is green, should be ready for the review bot now.

PR #16026, opened 2026-09-04
bank-architect: 0.5.0

PKoka5:
0.4.0 -> 0.5.0. Commit bump only; the repository line is unchanged.

Three additions, all of which leave the existing behaviour as the default:

- **Frequently used is now a switch.** The quick-access gathering (best axe and pickaxe, hammer, graceful, diary rewards) is on by default as before, but a player who would rather keep tools with tools can turn it off instead of correcting each new tier by hand.
- **A rearranged tab weaves.** A tab holding tags from more than one category used to assemble one category block after another, so an order alternating between them collapsed. Once the player has rearranged that tab's tags, it is now assembled tag by tag in the order they wrote. Tabs in default order, single-category tabs and tabs with a shaped block on them stack exactly as before.
- **Block order within a tag is the player's.** A block is whatever the sorters never split — a catalogued set, a dose or charge family, or a single item. The contents of a block stay curated, so a dose family still reads 4 to 1 and a set still dresses in slot order. Nothing saved means the curated order stands byte for byte, and there is a per-tag reset back to it.

Full suite green on the pinned commit, and all three fixed-seed simulation reports are byte-identical to 0.4.0 — every one of these is a departure the player has to ask for.

### DIFF
--- a/plugins/bank-architect
+++ b/plugins/bank-architect
@@ -1,2 +1,2 @@
 repository=https://github.com/PKoka5/bank-architect.git
-commit=8ecb21ed4a76483a899a31066d9855c56f351750
+commit=2dcdfc36a97ddda622172c7b3913dd52bc185f75
--- a/plugins/bank-architect
+++ b/plugins/bank-architect
@@ -1,2 +1,2 @@
 repository=https://github.com/PKoka5/bank-architect.git
-commit=2dcdfc36a97ddda622172c7b3913dd52bc185f75
+commit=6c65638c71e197981ae2db488c7bd2b59b87e9b2


ldavid432:
`too many tokens: 205566` you have to be under 200k for our review bot to run your plugin, otherwise it requires manual review, which at enough lines to be that many tokens no one will be wanting to do

PKoka5:
Thanks for the number — that is genuinely useful, and I would rather fix it than ask anyone to read 200k tokens by hand.

I measured on our side to find the bulk. The diff under review here is only ~75 KB, so it is the plugin's source as a whole rather than this release that crossed the line: `src/main/java` is 1.01 MB, which is about the 205,566 you are seeing. Within that, one file is 141 KB on its own — `CanonicalItemClassificationOverrides.java`, 3,635 lines of exact item-ID overrides, 1,935 `case` labels mapping IDs to a classification. It is a data table that happens to be written in Java.

Before I spend the weekend on it, one question so I move the right thing: **does the count cover only Java sources, or the plugin's resources too?**

If it is Java sources, moving that table into a TSV file under `src/main/resources` — where this plugin already keeps its other lookup tables — takes roughly 14% off in one change, which should land us comfortably under 200k with headroom for the next few releases. If resources count as well, that move buys nothing and I will need a different plan; probably trimming the bundled registry, which is the real heavyweight at 1.7 MB.

Happy to close this PR and open a fresh one once the plugin is under the limit, if that is tidier for your queue than leaving it sitting here.

riktenx:
everything under src/main/java is counted, and comments are stripped beforehand. nothing else is counted

PKoka5:
Thanks riktenx — that answered it exactly, and the "comments are stripped" part is what made it measurable. Repinned to 6c65638, which should come in under the limit.

What changed: `CanonicalItemClassificationOverrides` was 3,635 lines and 1,935 case labels mapping item IDs to a classification — a lookup table wearing a switch statement. It is now a TSV beside the class, read once into a map. Reproducing your method locally (everything under `src/main/java`, comments stripped) the counted size goes from 807,942 characters to 766,175, which at the ratio implied by your 205,566 is about **194,900 tokens**.

The move is proved lossless rather than asserted: the old table was dumped for every item ID from 0 to 70,000, the new one dumped the same way, and the two files are identical. The plugin's three fixed-seed simulation reports are byte-identical across the change too.

Also picked up since the last push: a player-reported fix for graceful recolours ([#36](https://github.com/PKoka5/bank-architect/issues/36)) — swapping the base set for a course recolour moved it off the quick-access tab, because the same six item IDs were hardcoded in two places.

Two notes in case they are useful to you. The margin is thinner than I would like — about 5k tokens — so if this creeps back up I will move the remaining curated ID lists out of the semantic rule sets rather than come back asking for an exception. And if a repin on a PR you have already looked at is worse for your queue than a fresh one, say so and I will close this and open a clean PR instead.

riktenx:
nobody looked at it other than seeing that there were too many tokens in a summary view, we prefer you update prs in-place rather than close and open

PR #16027, opened 2026-09-04
anvil: one Anvil instead of one site per clan

AhmedFathy2001:
Updates `anvil` from 67d9a5a to e8b0bd3 (plugin 1.5.0 → 1.6.0).

The site this plugin talks to used to be one deployment per clan, so the
plugin held one URL and belonged to one clan. It is now a single site with
clans as rows, and this release follows it:

- The clan comes from the account token rather than the URL. Someone in
  several clans picks between them in the sidebar instead of retyping a
  site address.
- Federation is removed — the whole mechanism existed to let separate
  per-clan deployments talk to each other, and there are no longer any.
- An Anvil button in the collection log and clan windows, placed in the
  same title bar as WikiSync and RuneProfile and anchored so it does not
  overlap them.
- Fixes: raid kill counts report the mode actually killed (an Expert ToA
  drop was reporting the normal mode's count), a 99 survives a plugin
  reload instead of being swallowed by the stat baseline, clumped NPC
  kills are counted off ServerNpcLoot rather than one per event, and drop
  attribution stops crediting a guaranteed drop as a lucky one.

### DIFF
--- a/plugins/anvil
+++ b/plugins/anvil
@@ -1,2 +1,2 @@
 repository=https://github.com/AhmedFathy2001/anvil-plugin.git
-commit=67d9a5ac2440db8642d345f6ba35511decbac759
+commit=f4f686528faa9eeca1207475f7142b2524108f31
--- a/plugins/anvil
+++ b/plugins/anvil
@@ -1,2 +1,2 @@
 repository=https://github.com/AhmedFathy2001/anvil-plugin.git
-commit=914994a0e2f83b3fe88605c8383b2316bfb20ffa
+commit=0879570665308a95bf9ca78dbec17b56202cdcf2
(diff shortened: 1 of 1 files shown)


riktenx:
please migrate all of your file i/o to the new https://static.runelite.net/runelite-client/apidocs/net/runelite/client/util/Filepath.html utility. note that this will mean you will not be able to ship the log-reading feature

riktenx:
you appear to have multiple plugin-specific directories, be aware that the filepath stuff allows for migration of _one_ folder only through `PluginDescriptor.legacyDataDirectory`

AhmedFathy2001:
> please migrate all of your file i/o to the new https://static.runelite.net/runelite-client/apidocs/net/runelite/client/util/Filepath.html utility. note that this will mean you will not be able to ship the log-reading feature

Would I be able to use a chooser that is pointed at the client.log location that the user can simply click to choose the client.log and then we tail the logs from there? Or would that also not be allowed and I just have to try to reproduce/ask them personally for the client logs for debug? The debug log is quite useful since the plugin does handle all sorts of events and its very easy to have it written slightly differently in the chat messages by jagex for certain bosses/drops/collections as you may know the inconsistencies with Jagex's own messages in the chat.

riktenx:
thats probably fine

AhmedFathy2001:
> thats probably fine

Should be all done, let me know if anything else is required.

PR #16046, opened 2026-09-04
Transfer ownership of better-npc-highlight

SamuelDev:
I am requesting to regain control of the better npc highlight plugin after giving it over to the runelite maintainers a few months ago. That happened in https://github.com/runelite/plugin-hub/pull/13365


pajlada:
cc @riktenx

SamuelDev:
I will be added as a coauthor on rikten's repo/branch of this plugin, rather than transferring ownership back to myself.

PR #16094, opened 2026-09-05
 Add Olm Melee Hand DPS Tracker

Roleiz:
Tracks the damage done to the olm melee hand in an 8 tick cycle(cripple threshold) by all party members.

### DIFF
--- a/plugins/olm-damage-tracker
+++ b/plugins/olm-damage-tracker
@@ -1,2 +1,2 @@
 repository=https://github.com/Roleiz/olm-damage-tracker.git 
-commit=6c4c86a2b63ec490c27b7d2c9f4efa372adb46d1
+commit=2531cfd62ed1d89dbffd64cfb40ef7563215a262
--- a/plugins/olm-damage-tracker
+++ b/plugins/olm-damage-tracker
@@ -1,2 +1,2 @@
 repository=https://github.com/Roleiz/olm-damage-tracker.git
-commit=ac3e4e60a9763065874ff1f4b210b0c042569824
+commit=ebdbef60cd74746a8c59f691087a651dd373621b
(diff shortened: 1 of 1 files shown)


coopermor:
Remove the build/ directory from your repository

Roleiz:
Hi, still awaiting response here. I understand you guys have probably been swamped with other requests too. Any questions about the plugin I can answer to the best of my ability!

Nightfirecat:
I do not think we can add this to the hub; this sounds an awful lot like "Attack counters" and "Adds additional visual or audio indicators of a boss mechanic except in cases where this is a manually triggered external helper." from [Jagex's Third Party Client Guidelines](https://secure.runescape.com/m=news/third-party-client-guidelines?oldschool=1).

Roleiz:
Gotcha, I saw there was a "DPS Threshold Alerts" Plugin so assumed something like this but in a party scope would be fine. Would making it more generalized to instead be a party DPS plugin work better? For example 8 people in a party hitting their respective boss, accumulates their party damage into DPS so would show 10/50/100 DPS in the interface in a 6 tick cycle(since most dps weapons are 4-6 ticks) instead of the specific damage amount? Please let me know and thank you for reviewing my request.

Nightfirecat:
I'm hesitant to say that would be allowed--that basically is the same thing with a math calculation or two in between. The core dps plugin already supports parties, so it covers the "shared party dps" use case there...

Roleiz:
I took a look at the DPS Counter plugin and it looks exactly what i was looking for with the caveat that it doesn't accumulate party DPS but does take into account the generalization DPS plugin I was trying to explain in concept. It looks like the developer for it hasn't made any contributions changes since 2020 so I'll try reaching out to him to see if i can get an update in since he appears active on github still. Thank you so much for your time!

PR #16103, opened 2026-09-05
Update Chat Overlay

devrat0:
- improved pattern matching for CAs to remove CA_ID prefix from messages


coopermor:
You can use client.expandMacro if you'd like https://github.com/runelite/runelite/commit/1c056963afdca697b94b6d157302a474f5b741f9

PR #16123, opened 2026-09-06
ScapePath: add diary tasks + collection log

ClipToTik:
Updates the ScapePath plugin revision to `7b8c835`.

Incremental account-state data expansion only — no change to the plugin's purpose, permissions, dependencies, or networking:

- Achievement Diary: exact per-task completion for **Karamja** (the only region RuneLite exposes stable per-task task varbits for), added as an optional, additive `tasks` map keyed by the RuneLite `VarbitID` constant name. Tier-only regions are unchanged and omit the key.
- Collection Log foundation: overall and per-tab obtained/total slot counts read from stable `VarPlayerID`s; reported unavailable (never fake zero) until the game populates them.

The plugin remains a passive, read-only account-state companion. Reads use public RuneLite APIs on the client thread; the only network path is the existing **opt-in** HTTPS sync to `https://www.scapepath.com` behind a config gate (unchanged). No new permissions, dependencies, endpoints, telemetry, chat/PII reads, background polling, or UI. `schemaVersion` stays 1 (additive). `./gradlew clean build` passes: 134 tests, 0 failures. See `PAYLOAD.md` for the full data disclosure.


ClipToTik:
Updated `commit=` to `00b512a4ac4a5d519108ad38ba6fe0b03dd58c96` with the final pre-release polish:

- Canonical ScapePath graph/progression sidebar icon, loaded as a packaged classpath resource.
- Simplified, native side panel: a lightweight connection/status surface (connection state, RSN/account type, total/combat level) instead of a data dashboard.
- Removed debug/payload information from the user-facing panel (payload size, schema, per-section freshness, raw JSON view).
- Reduced UI churn — the panel re-renders only when a displayed value changes.

Account-state collection and the HTTPS sync architecture are unchanged, the opt-in third-party disclosure/warning remains, and the Combat Achievements implementation is unchanged. Clean build + checkstyle + tests pass locally, and the Plugin Hub build check is green.

PR #16170, opened 2026-09-06
Update Visibility Plugin

Endrit-alt:
- Added support for solid transparency
- Dfiferent QoL fixes

### DIFF
--- a/plugins/improved-tile-indicators
+++ b/plugins/improved-tile-indicators
@@ -1,3 +1,2 @@
 repository=https://github.com/LeikvollE/tileindicators.git
 commit=c367969cdaad43ff5c6047abe67a7d390e134caf
-authors=Endrit-alt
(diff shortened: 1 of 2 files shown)


ldavid432:
I don't think we will be allowing any lwjgl code atm. I see you already commented about this in the GPU API thread. If Adam confirms the API won't support this then we can take another look at this but I expect it should go through the GPU API.

Endrit-alt:
@ldavid432 When the API releases and it supports it I would just remove any lwjgl code and replace it with the proper API calls if that would be okay? Currently I don't have any other solution.

Do you want me to remove it? The benefit for my plugin is quite significant as it allows me to disable rear faces from showing through front faces, see gif of solid transparency. I've asked Adam, I'll update when if/when he replies.

[opacity.webm](https://github.com/user-attachments/assets/388ab5e6-3154-4123-922f-9101cdbccc62)

Alexsuperfly:
i dont see us reviewing gl code tbh, so thisl just not get merged to begin with

riktenx:
yeah just remove it for now

Endrit-alt:
> yeah just remove it for now

Removed it for now, I do hope the future GPU API will support it, but it is what it is!

PR #16185, opened 2026-09-07
Add Direct Messages

iIDezi:
Adds a local inbox for private messages with search, unread counts, timestamps, themes, and optional RuneProfile portraits.

Message history stays on the user's computer. RuneProfile portraits are optional and off by default.


Alexsuperfly:
use of client.hopToWorld, means this requires full manual review, i suggest that feature be removed in order to allow automated reviewing

iIDezi:
I have client.hopToWorld for right-clicking the friend/message and hopping to their shown world similar to how to works on your Friendslist

This is for seemless use for the user instead of switching between UIs

PR #16217, opened 2026-09-08
disable exchange-insights and exchange-insights-capture

TheSpryt:
Please disable both Exchange Insights plugins. The website and service behind them are being retired, so neither plugin will have anything to talk to. Both manifests keep their repository, commit and warning lines and gain a disabled flag with the reason.

- plugins/exchange-insights
- plugins/exchange-insights-capture

The open update for exchange-insights-capture (#16195) has been closed. Bank Templates is unaffected and continues; its update is #16218.


TheSpryt:
Switched from deleting the manifests to adding the disabled flag with a reason on both, as the bot asked.

PR #16240, opened 2026-09-08
Add Harrisun Mystic UI

opaf-osrs:
Puts the minimap and the orbs into one rectangular block above the inventory. Square minimap instead of the round one, the four orbs as flat colour filled blocks in a row, and it works in both resizable layouts.

No external calls and no bundled art. The frame and the orb icons are read off the client's sprite table at runtime, so the block follows whatever resource pack you happen to be running and falls back to the game's own art when there is none.

Closest existing thing is compact-orbs, which repositions the orbs. This does the map and the orbs as one panel and squares the minimap off, so they are not really doing the same job.

the block

Attached to the inventory so the two read as one panel:

attached

### DIFF
--- a/plugins/harrisun-mystic-ui
+++ b/plugins/harrisun-mystic-ui
@@ -1,0 +1,2 @@
+repository=https://github.com/opaf-osrs/harrisun-mystic-ui.git
+commit=f2867450dd04f70569a2fd836d3409313b7a3101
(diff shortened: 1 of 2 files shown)


ldavid432:

opaf-osrs:
Yep, that is gone. It was forwarding a click from an orb painted in the map's corner to the stock op, because the real widget could not sit there. The orb now lives in the strip beside the map where it is just the stock widget answering its own click, so there is nothing synthesised. Repinned.

PR #16258, opened 2026-09-08
Add Local Event Bridge

ashy0019:
Adds Local Event Bridge, a headless plugin that publishes a fixed allowlist of gameplay observations to local companion applications over IPv4 loopback (127.0.0.1:41713).

The protocol is one-way with respect to gameplay and provides no commands for clicks, typing, menu actions, movement, or other game control.


Nightfirecat:
Sorry, this is not allowed per our [Rejected or Rolled Back Features page](https://github.com/runelite/runelite/wiki/Rejected-or-Rolled-Back-Features), specifically:

PR #16269, opened 2026-09-08
the-birdhouse: 1.4.0

birdandworm:
Bumps The Birdhouse to 1.4.0.

Auto-submission runs off the loot tracker, so anything a player gathers is invisible to it and board tiles asking for a mined, fished, hunted or minigame-earned item could never complete on their own. This release adds a tracker that diffs the inventory and only credits a gain that is anchored to gathering experience in the same few ticks, which is what keeps it honest: withdrawing from a bank, buying on the Grand Exchange or from a shop, and trading all award no experience and so can never clear it. Items taken off the ground are refused, and anything the loot tracker already reported is skipped.

No new data leaves the client. Submissions carry the same item name, quantity and optional screenshot as an ordinary drop, so the existing warning is unchanged.

### DIFF
--- a/plugins/the-birdhouse
+++ b/plugins/the-birdhouse
@@ -1,3 +1,3 @@
 repository=https://github.com/birdandworm/TheBirdhouse.git
-commit=fa04695ade4e78405441733905daae5fadae7815
+commit=80edf9713493d20d2b46d0bd2ab7c84e30b78b5b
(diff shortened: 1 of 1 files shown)


birdandworm:
Updated to include kill-count tile support. Bosses that pay out through a crate or reward pool - Wintertodt, Tempoross - book their loot under a different name than the boss, so a kill-count tile naming the boss never matched. The game's own count message now credits those, suppressed when the loot tracker already named the same source so an ordinary boss is not counted twice.

PR #16273, opened 2026-09-08
New elywatch plugin

myadoran:
This is a niche plugin that reads gear equip events. If the event is an elysian being equipped or unequipped, an API call is made to a user-defined host through the HttpClient java library.

The use case is connecting ely equip status to a Home Assistant instance. Equipping it turns a lamp on through one webhook, unequipping it turns it off through another.

Though I don't expect this plugin to exactly go viral, I'm happy to broaden it a bit if other people need something similar.

Cheers for making Runelite so flexible, I'm no programmer but with your documentation it was still accessible!

### DIFF
--- a/plugins/elywatch
+++ b/plugins/elywatch
@@ -1,0 +1,2 @@
+repository=https://github.com/myadoran/ElyWatcher.git
+commit=3a975fc914aff9e5424edd8ed3c23f9a18f20fce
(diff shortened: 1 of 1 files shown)


Alexsuperfly:
`java.net.http.HttpClient.newHttpClient`: Java's HttpClient is not allowed, @Inject and use RuneLite's OkHttpClient instead

myadoran:
Thank you for checking, it now uses async requests under OkHttp.

PR #16282, opened 2026-09-09
add shell-songs

jimmyununu:
**Shell Songs** is a song book for the seven musical shells rewarded by Crab Quest (Shell C, D, E, F, G, A and A#).

Pick a song in the sidebar and the plugin shows which shell to play next, how many times in a row to play it, and the whole song as a row of notes. It advances when the player clicks "Sound" on the correct shell. It also draws a small overlay with the upcoming notes and outlines the next shell in the inventory with the click count on it. Users can add their own songs in the settings.

The plugin only reads the user's own `MenuOptionClicked` event for the "Sound" option on a shell item. It does not play shells, send input, edit menu entries, or use the network. Shell item IDs are hard-coded. No dependencies beyond the standard build.

Repository: https://github.com/jimmyununu/shell-songs

### DIFF
--- a/plugins/shell-songs
+++ b/plugins/shell-songs
@@ -1,2 +1,2 @@
 repository=https://github.com/jimmyununu/shell-songs.git
-commit=3e50027f929612e6b7c9444314890e6d237fd05e
+commit=b4e1bc018eb28dea484d2d8824e46e3e458f909a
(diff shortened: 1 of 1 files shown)


ldavid432:
I think you will need to remove any copyrighted songs. Most of the rest seems to be nursery rhymes which should be fine

jimmyununu:
Thanks. Removed the two songs that were still in copyright (Seven Nation Army and Baby Shark). The remaining 20 are all public domain: nursery rhymes, folk songs, traditional carols, and 19th-century pieces (Stephen Foster, Beethoven's Ode to Joy, Jingle Bells). Added a note in the source that only public-domain material may be added. Manifest updated to b4e1bc0.

PR #16294, opened 2026-09-09
chore: disable yama attack reminder plugin

jdockerty:
Removing this plugin as it is not being updated/unused.

There are many Yama plugins which are much better nowadays too.

### DIFF
--- a/plugins/yama-reminder
+++ b/plugins/yama-reminder
@@ -1,0 +1,3 @@
+repository=https://github.com/jdockerty/yama-reminder.git
+commit=682887c8b1bfc7441a5723f7137b12a8ecb67a9a
+disabled=unmaintained and superseded by many other plugins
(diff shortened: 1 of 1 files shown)


jdockerty:
> add the disabled flag with a corresponding reason.

Question for maintainers: does this disabled flag remove it from being shown on the list of plugin hub search or does it just show it as "unavailble", but still in the list.

My intention is to remove it entirely, since I've not touched this plugin in some time and no longer use it. There are also parts which are incorrect that I never got around to fixing, so I don't believe it is useful for anyone else either.

coopermor:
It will be removed from the hub search as well as any clients it is installed on.

jdockerty:
@coopermor Thanks for the info

I've gone ahead and dropped in the `disabled` line instead :bow:

PR #16297, opened 2026-09-09
add somnia

GucciSync:
Adds Somnia, a community plugin featuring [GucciSync's](https://www.twitch.tv/guccisync) Twitch emotes, automatic Somnia party joining, and soundbites for supported game events.

Emotes and audio are bundled with the plugin. Twitch synchronization is a maintainer-only build task and no media is downloaded at runtime.


coopermor:
* Auto-rejoin for parties: Would put undue strain on the party servers.
* Twitch/BTTV/FFZ/7TV emote plugins: Distributing the emotes for these plugins would likely require skirting over the legal agreements of these services, and thus won't be accepted.

PR #16306, opened 2026-09-09
Update chart-plotter

Dazuzi:
- Added an optional info overlay.
- Overhauled routing engine.
- Performance/wasted work improvements.


ldavid432:

Dazuzi:
I addressed the thread interrupt issue, and while I was at it, made some further improvements and adjustments.

PR #16332, opened 2026-09-09
update womclanplugin

ShakyPizza:
Updates `womclanplugin` from `ec67912` to `94a09b4`.

This follow-up reduces routine WOM API traffic and makes the existing refresh controls
respect the API server's live rate-limit state across RuneLite restarts. No new dependencies,
endpoints, permissions, or API key are added.

### API impact

API load is reduced substantially:

- A routine sidebar refresh now requests only `/v2/groups/{id}`: **one request instead
  of four**.
- Achievements, activity, and name changes are fetched only when Clan Details is opened.
- Those three optional history responses have an independent one-hour cache, so reopening
  Clan Details does not refetch them.
- Automatic overview refresh remains once per hour.
- The five-minute manual cooldown now survives plugin/client restarts.

The client records `RateLimit-Limit`, `RateLimit-Remaining`, `RateLimit-Reset`, and
`Retry-After` from WOM responses. A `429` persists the server deadline and blocks manual,
automatic, and detail requests until it expires. A successful response with zero remaining
also blocks proactively until its reset time. There is no automatic retry loop.

### Cache and persisted state

- Successful raw JSON responses are stored per group under `RuneLite.RUNELITE_DIR` at
  `wom-clan-stats/group-{id}.json`.
- Cache writes use a temporary file and atomic replacement where supported.
- Cache read/write failures are non-fatal; the plugin continues using the network/in-memory
  behavior.
- Cached data is restored immediately and labelled `Cached` until a network sync succeeds.
- Two timestamps are stored through `ConfigManager` in the separate `womclan.network` group:
  `manualAllowedAt` and `serverRetryAt`. Keeping them outside `womclan` avoids triggering the
  plugin refresh listener.

The status tooltip exposes the last successful-data timestamp and the current WOM request
budget while that server window is still active. Existing data remains visible after a failed
or rate-limited refresh.

### Testing

`./gradlew clean test build` passes. There are 63 tests, all passing, including focused tests
for one-request sidebar fetching, persistent cache restoration/history freshness, restart-safe
deadlines, server backoff longer than the local cooldown, cached-state formatting, and API
budget formatting.


Alexsuperfly:
use RuneLite.RUNELITE_DIR instead of RuneLite.CACHE_DIR

ShakyPizza:
> use RuneLite.RUNELITE_DIR instead of RuneLite.CACHE_DIR

Updated thank you @alexsuperfly

PR #16352, opened 2026-09-10
add pvp-hud

50percentelf:
Adds PvP HUD, a configurable PvP combat HUD with multiple layouts, local combat timers, resource information.


50percentelf:
image

Please let me know if there are any concerns. I've redacted anything that comes close to the line/crosses on what might not be allowed based on previous pvp feature rejections.

IE opponent tick cycle is stubbed, no player highlighting, no opponent freeze timer or opponent tb timer etc. 

I do think maybe I'll get some pushback on the self pj/log timer but I haven't found evidence of it being rejected before, willing to discuss.

PR #16368, opened 2026-09-10
Add Crab Rave plugin

Laporchmonkey:
Adds Crab Rave, a small RuneLite plugin that plays a bundled 30-second Crab Rave audio clip when the player performs the Crab Dance emote.

The plugin only looks for the local player's Crab Dance animation and does not perform any gameplay automation or external network communication.


ldavid432:
We cannot accept copyrighted content in plugins

PR #16378, opened 2026-09-10
Add slayer-prep-assistant plugin

palle00:
Added slayer-prep-assistant plugin with repository info and warning.


palle00:
I know there are other plugins that offer similar functionality, but some of them require manual database updates and, frankly, aren't very pleasant to use or look at.

palle00:
Bump

palle00:
Bump so it doenst go inactive

riktenx:
you can't get a url from some server response and make requests to it. we need to be able to see every single url your plugin talks, you will need to hardcode the wiki and just plug in the path or something

PR #16388, opened 2026-09-10
Add Account Almanac

P6-P6:
Tracks bank contents, Grand Exchange offers, skills and wealth across every account you play, with cross-account totals, snapshots over time and charts.

All data is stored as JSON under .runelite/accountalmanac/. The plugin makes no network requests of any kind.

### DIFF
--- a/plugins/account-almanac
+++ b/plugins/account-almanac
@@ -1,2 +1,2 @@
 repository=https://github.com/P6-P6/account-almanac.git
-commit=2b8fa030aafabc6ea877398a0e8dadb447ebe04c
+commit=791fa3b73d1c3511b221f6e9416c1367d84c09be
--- a/plugins/account-almanac
+++ b/plugins/account-almanac
@@ -1,2 +1,2 @@
 repository=https://github.com/P6-P6/account-almanac.git
-commit=ac8859af210206f1dc0b13cd66b59578ec8ff8ec
+commit=2b8fa030aafabc6ea877398a0e8dadb447ebe04c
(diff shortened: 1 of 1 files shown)


P6-P6:
added a few more features

1. Quest points per account (###/###), read from the game's own vars so the denominator tracks quest releases
2. Interesting tab split into Skill XP · Random event items · Burnt, each with per-account pie charts and item sprites
3. F2P/Members filtering across the All items tab and both exports
4. Bank screenshots in the game's real steel frame, four sorts, quantity/value thresholds, adaptive width, hover tooltips
5. Grand Exchange image export with buy/sell filtering and seven sort orders
6. Market-price comparison — a column on the GE tab and on the export, showing how far each listing sits from market
7. Banded progress bars, red through amber to green
8. Two bug fixes: the missing thousands separator on "Top account's XP", and the blurry login name (synthesized bold on a font with no bold face)

P6-P6:
Two more since that list, now at `4fcb44b`:

- CSV exports (items, accounts, banks, offers, history, skills) plus a single-file HTML report
- Fixed light themes: gp and quantity cells used the game's stack colours, which are invisible on a pale background

riktenx:

P6-P6:
> * use of Desktop::browse is not allowed, use LinkBrowser::browse instead

Fixing it now. Appreciate it 


EDIT: Fixed in 8a335bd

P6-P6:
A few more cleanups since that fix, now at `b22c666`:

- Replaced the deprecated `getAccountType()` with `VarbitID.IRONMAN`
- Skill-only changes save every 10 minutes instead of every 5 seconds
- "Back up now" runs in the background instead of on the Swing thread

ldavid432:
LinkBrowser.open is also restricted

P6-P6:
Fixed, now at `3592272`:

- Removed `LinkBrowser.open` - the HTML export now just shows where the file was saved
- Removed `toFront()` / `requestFocus()` on the viewer window
- LICENSE now matches the hub's template exactly

P6-P6:
> LinkBrowser.open is also restricted

Fixed thank you.

PR #16393, opened 2026-09-11
Update reword-manager to 1.6.0

ejaz-karim:
Add rewording to NPCs/Objects with IDs
Merge upstream changes from example-plugin


ldavid432:
We don't accept new, user configured, ID based functionality in plugins

ejaz-karim:
> We don't accept new, user configured, ID based functionality in plugins
> 
> https://github.com/runelite/runelite/wiki/Rejected-or-Rolled-Back-Features

Even for just renaming NPCs? the plugin already does this with renaming Items using Item IDs. 
e.g.
5076,Bird's egg (Red)
5077,Bird's egg (Blue)
5078,Bird's egg (Green)

I understand the rule is there to prevent unfair advantages, but could this please be considered as an exception as I believe renaming Npcs and objects (which the plugin does with text anyway) using IDs won't give the player any advantage and won't break any Jagex rules. The use case for this feature was presented to me here https://github.com/ejaz-karim/reword-manager/issues/19 Where the user would like to rename Verzik Vitur differently according to the phase, because before you could only rename Verzik once using text and that would apply for all phases. If it's decided that this is still not acceptable, then I will remove this feature. Thanks.

ldavid432:
Unfortunately, I don't think we would want to make any exceptions for this rule. We have a hard enough time handling the grandfathered in ID based plugins.

ejaz-karim:
Should be good now. The relevant IDs code in the method "remapMenuEntryText" has been commented out. I added a bug fix too "Fix Rewording player & clan names not working if the name contains a space".

PR #16398, opened 2026-09-11
Add Random World Hopper++

andrew-friedman-phd:
## What it does

Hops to a random world while keeping recently used worlds on a user-tunable cooldown, so a random hop never sends you back to one you just left. Cooldown duration is editable directly from the sidebar panel to suit whatever activity you're doing.

- Sidebar cooldown list with a live countdown, plus a "Hop to random world" button.
- Configurable ping threshold and world-type filters (Normal, Deadman, Seasonal/Tournament, Quest Speedrunning, Fresh Start, PVP, Skill Total, High Risk, Bounty Hunter, F2P).
- Skill Total worlds are filtered against the account's actual live total level, so you won't get hopped onto one you don't qualify for.
- A small "World Hops" counter shown under the minimap while the sidebar panel is open, tracking hops since the last login.

Repository: https://github.com/andrew-friedman-phd/random-world-hopper

### DIFF
--- a/plugins/random-world-hopper
+++ b/plugins/random-world-hopper
@@ -1,0 +1,2 @@
+repository=https://github.com/andrew-friedman-phd/random-world-hopper.git
+commit=aa1e23609b163d4a31bd512bc55fde017c6599a9


andrew-friedman-phd:
For context for reviewers: this plugin's core hop logic (random-hop-under-ping-threshold, world-type filtering, per-world cooldown tracking, and the quick-hop widget state machine) is almost identical in functionality to the random-hop feature already approved in my Varrock Bank Ruby Rings plugin (#14719, https://github.com/andrew-friedman-phd/varrock-bank-ruby-rings). This plugin just extracts that behavior into its own standalone tool with a user-configurable cooldown duration instead of a hardcoded one, since it's generally useful outside that plugin's specific activity.

Nightfirecat:
[Do not use the "worldhopper" config group](https://github.com/andrew-friedman-phd/random-world-hopper/blob/aa1e23609b163d4a31bd512bc55fde017c6599a9/src/main/java/com/worldhopper/WorldHopperConfig.java#L9) (also in [this location](https://github.com/andrew-friedman-phd/random-world-hopper/blob/aa1e23609b163d4a31bd512bc55fde017c6599a9/src/main/java/com/worldhopper/WorldHopperPlugin.java#L64), as it conflicts with the core plugin's config group.
This plugin looks fine otherwise.

PR #16409, opened 2026-09-11
Plugin submission: Chunkblazer

btwinnn:
Chunk randomizer plugin

Edit: GPU plugin removed per request

### DIFF
--- a/plugins/chunkblazer
+++ b/plugins/chunkblazer
@@ -1,2 +1,2 @@
 repository=https://github.com/btwinnn/Chunkblazer.git
-commit=0c7594a50af0d86e1b02181c71fde06500c6c3c4
+commit=c28152243f1c3cd584581b72e81b38d7efa587ea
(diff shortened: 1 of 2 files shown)


ldavid432:
1. You should only modify/add one plugin per PR

2. We aren't accepting new GPU plugins currently. There is some ongoing work to have the GPU plugin expose an API that other plugins, like this one, can hook into. If you want to provide feedback on this, join the `gpu api` thread on discord.

btwinnn:
Understood @ldavid432 . I've deleted the GPU from the PR. I think the API hook-in is a great solution. Thanks for the info and your work.

btwinnn:
Removed "Player Discovery" features since they are listed as not allowed.

riktenx:
you can't receive arbitrary filepaths from a server response and construct a `File` from it, that makes it impossible to verify that you're only doing file io in a plugin-specific subfolder under .runelite.

alternatively if you migrated all of your file i/o to the new https://static.runelite.net/runelite-client/apidocs/net/runelite/client/util/Filepath.html this would not be a concern and would probably even help your plugin get merged faster, but that's not required

btwinnn:
Got it. Will work on migrating file I/O to use the new Filepath.
Merging speed isn't an issue. We appreciate ya'll either way.

Thanks goat @riktenx

btwinnn:
- Migrated I/O catalog and asset cache to new Filepath
- Fixes for API keys and Chunkblazer server auth
- Log cleanup
- Size caps on server downloads for security

PR #16419, opened 2026-09-11
Update commit hash for catastrophic-events plugin

mikeyf06:
Adds a $name token to the death message config, substituted with the local player's name before posting to Discord, plus a one-time migration that prepends the token to any existing custom message.


ldavid432:
It looks like you accidentally reverted your commit hash update

mikeyf06:
@ldavid432 oops, thanks for catching that. clicked the wrong thing. should be resolved now

PR #16431, opened 2026-09-12
Add previous-prayer-tracker

MustacheAltun:


coopermor:
We aren't accepting plugins that track your last prayer.

MustacheAltun:
> We aren't accepting plugins that track your last prayer.

oh, i thought it wasn't against the rules but i guess not? sorry

PR #16442, opened 2026-09-12
Update escape-crystal-notify (v1.14)

Xylot:
- Added Maggot King, Shellbane Gryphon, Mad Angel, and Brutus support.
- Updated Hueycoatl, Bryophyta, and Priff Rabbit entrance handling.
- Added customizable 3D crystal, ground circle, player outline, and teleport NPC indicators.
- Added per-encounter thresholds, defaults, and nearby encounter sidebar settings.
- Expanded six-hour warnings and made logout safeguards optional, disabled by default.
- Improved safe-area detection, instance requirements, and closest-entrance matching.

### DIFF
--- a/plugins/escape-crystal-notify
+++ b/plugins/escape-crystal-notify
@@ -1,2 +1,2 @@
 repository=https://github.com/Xylot/escape-crystal-notify.git
-commit=7038395a1c95d0a9ea0b083fd2648f2b90fab510
+commit=7c7aa2429259487f3233dda8ee354f3f00564a0c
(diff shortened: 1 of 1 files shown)


Nightfirecat:
For your debug config settings, please also limit this to only be available when developer mode is enabled so we avoid any possibility of providing user-provided ID lists for highlighting, etc. (As a reference for how you might do this, search for uses of `@Named("developerMode")` in the client codebase)

Xylot:
> For your debug config settings, please also limit this to only be available when developer mode is enabled so we avoid any possibility of providing user-provided ID lists for highlighting, etc. (As a reference for how you might do this, search for uses of `@Named("developerMode")` in the client codebase)

@Nightfirecat Added all developer mode checks

PR #16447, opened 2026-09-12
Update ba-world-scouter to 1.1

rsfost:
- Add option to disable side panel
- Sync flag icons from world hopper plugin
- Remove thread interrupts and sleeps from SSE handling


Alexsuperfly:

rsfost:
I think all sleeps and interrupts already run inside an executor, or are they still not allowed?

Alexsuperfly:
auto review will always fail as long as you have sleeps/interrupts, if you don't remove them then you'll have to be manually reviewed which typically takes significantly longer

PR #16449, opened 2026-09-12
Update templeosrs

44Mikael:
Local DB improvements/fixes and WikiSync collection log interface button conflict fix from following PRs:
- SMaloney2017/Temple-OSRS-Plugin#31 
- SMaloney2017/Temple-OSRS-Plugin#32


44Mikael:
Added build=standard to the plugin properties file for the build check to pass

coopermor:
If you're using h2 database then you can't use standard build and would need to use gradle build.

pajlada (approved):

PR #16451, opened 2026-09-12
Add Rune-Tags

4E-DEVOPS:
RuneTags adds interactive quick player profiles to RuneScape chat. View player identity, records, stats, metrics, notes, tags, and actions directly from mentions, @ tags, and chat references without leaving the game.


4E-DEVOPS (commented):
Updated WidgetInfo -> InterfaceID

PR #16455, opened 2026-09-12
Disable RuneScape Coach

jamesondaines:
I am withdrawing RuneScape Coach from public distribution and have made its source repository private. I am the same GitHub account that submitted the plugin in #14146; the username changed from `joelbridger` to `jamesondaines`. This keeps the manifest and adds RuneLite’s required `disabled` reason.

### DIFF
--- a/plugins/runescape-coach
+++ b/plugins/runescape-coach
@@ -1,2 +1,0 @@
-repository=https://github.com/joelbridger/runescape-coach-plugin.git
-commit=736ab5f8d093b5f08e9515cdb371e5a225109632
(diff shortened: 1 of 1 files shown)


pajlada:
Hi!

1) As the comment above from the RuneLite bot mentioned, the manifest file should stick around but a `disabled` key should be added.

2) The author themselves should make this change, or sign off on it here

jamesondaines:
I am the plugin author and confirm that I want RuneScape Coach disabled because it is no longer intended for public distribution.

jamesondaines:
Done—the manifest now keeps its existing fields and adds the requested disabled reason. I have also signed off as the author above.

ldavid432:
You are not the author though?

jamesondaines:
I am the same GitHub account that formerly used the username `joelbridger`; I renamed it to `jamesondaines`. The original submission, #14146, is shown by GitHub as authored by this account. I made the plugin repository private today as part of withdrawing it, which is why its old public link is no longer viewable.

PR #16466, opened 2026-09-12
Add Discord Loot Keys plugin

slabdab7100-lgtm:
easy way to export loot keys to discord


Nightfirecat:
Use of `URL::openConnection` is not allowed, @Inject and use RuneLite's OkHttpClient instead.

PR #16489, opened 2026-09-13
Add Knight Finder

CoreyUK:
Knight Finder finds worlds where a Knight of Ardougne is already pinned down for pickpocketing, shared between plugin users.

**What it does**

- Notices a pinned knight on the player's own world: one that has stayed within a two tile box for thirty seconds while somebody picks its pocket, or while it is in combat with a player (the "splashing" method). A knight that merely pauses in the street does not count.
- Outlines that knight with how it is pinned, how many people are on it and how long it has been there. This works with sharing off.
- With sharing enabled, reports it, and lists the knights other users have found in a side panel. There is no hop button.

**Third-party server**

Reports go to `https://knights.shep.rip`, which I run (the same service as Meteor Reporter, #15668). Sharing is opt-in and off by default; the single **Enable shared knights** toggle carries the standard IP warning.

A report contains the world, the knight's coordinates, a short name for the spot, whether it is in combat or trapped, and a count of players who have pickpocketed it recently. No player names or other player information are sent; the count is a number, and the server forgets a knight ninety seconds after the last report. The server stamps its own timestamps, validates every field, and rate limits reads and writes per IP.

Repository: https://github.com/CoreyUK/KnightFinder (BSD 2-Clause, built from the example-plugin template).


plugins/knight-finder
```diff
@@ -0,0 +1,2 @@
+repository=https://github.com/CoreyUK/KnightFinder.git
+commit=f5e13482099fa9b198c27788d6e7c0c4ab64b5b7
```
Alexsuperfly, line 2:
```suggestion
commit=f5e13482099fa9b198c27788d6e7c0c4ab64b5b7
warning=This plugin submits your IP address, and may submit various account data, to a 3rd-party server not controlled or verified by Runelite developers.
```

PR #16492, opened 2026-09-13
Add Lookout

CoreyUK:
Lookout keeps a small always-on-top view of the game in a corner of the screen while RuneLite is not the active window, and gets the player's attention when something in the game needs it.

**What it does**

- Shows a live copy of the game when RuneLite loses focus (after a short delay) or minimises, or on a hotkey. It only appears when RuneLite is likely out of view: if the client window is still on screen it uses a corner beside it, and if there is no such corner it waits until something covers RuneLite where the mouse is.
- Flashes - thick border, colour wash, a banner saying why, and the taskbar button - or brings RuneLite to the front, on: low hitpoints, being attacked, a Grand Exchange offer completing, a private message, or any RuneLite notification. Each is a toggle.
- Click to bring RuneLite to the front; drag to move (remembered, per screen); drag the corner to resize; a hide button; opacity and frame rate settings.
- The window never takes keyboard focus and stays off the taskbar and out of alt-tab.

**How it works**

Frames come from `DrawManager.requestNextFrameListener` at the configured rate, only while the window is showing. Focus is watched with a `WindowFocusListener` on the client frame; bringing RuneLite forward uses `ClientUI.forceFocus()`. No input is sent to the game and nothing leaves the machine.

**Relation to `picture-in-picture`**

The existing `picture-in-picture` plugin covers the same idea but has had no commits since March 2023, with open issues reporting it broken on Windows 11 (#28), a crash in its drag listener (#24), drag not working on macOS (#26), and requests for a hotkey (#25). Lookout is a fresh implementation rather than a fork: it adds the event alerts, the visibility heuristic, the hotkey, resizing, opacity and a hide button on top of the mirror window. Happy to offer it upstream instead if you would rather see that plugin revived.

Repository: https://github.com/CoreyUK/Lookout (BSD 2-Clause, built from the example-plugin template).


ldavid432:
You cannot change the window focus / bring to front

riktenx:

CoreyUK:
Thanks both — fixed in 6bcc4d4.

- Focus: the click action and alert option that brought the client to the front are removed, along with the taskbar flash. The plugin never touches the RuneLite window now; alerts either flash the Lookout window or send a notification through `Notifier`, so anything after that follows the user's own notification settings.
- `KeyboardFocusManager` is gone; the check for whether focus moved to one of the client's own dialogs uses `clientWindow.getOwnedWindows()`.

PR #16494, opened 2026-09-13
update voiced-dialogue

grabartley:
Updates voiced-dialogue to [v0.9.1](https://github.com/grabartley/runelite-voiced-dialogue/releases/tag/v0.9.1). The pinned commit is the `v0.9.1` release tag commit.

Thread interruption is gone from the plugin. Every executor closes with `ExecutorService.shutdown()`, and each pool carries a generation counter or close flag so a task queued at close is a no-op before it can reach the network, the disk, or the audio line. The wiki rate limit waits on a delayed `CompletableFuture` rather than sleeping a pool thread, so there is no `Thread.sleep` either.

Highlights since v0.8.0:

- Nearby NPC overhead chatter can be voiced, behind an opt-in toggle.
- NPCs learned from the wiki resolve their voice the same way the bundled voice table does, so a newly seen NPC sounds the same as a bundled one.
- Race parsing handles piped wiki links, fixing accents for NPCs whose infobox race field is a piped link.


riktenx:

grabartley:
@riktenx thanks for the catch, thread interrupts are gone and the PR now points at v0.9.1.

PR #16526, opened 2026-09-14
update quest-helper to v4.17.1

pajlada:


pajlada:
Will resolve system i/o part next release

PR #16536, opened 2026-09-14
Add Bank Portfolio Tracker plugin

2hBuilds:
2h Bank Portfolio Tracker: your whole bank valued, with how every item moved over 1, 7, 30, 90 or 180 days.

**What it does**

- A "Bank value" card (whole-bank total, gp and % move for the chosen window) over one row per stack: icon, name, the stack's value with its move in gp and %, then quantity x unit price and one item's move.
- Sort by percent change, gp change, item price or stack price; a gp filter with three editable presets.
- Optionally counts inventory and worn gear (read when the bank opens or on Refresh), coins and platinum, and untradeables at their tradeable parts' value or High Alchemy value.
- Remembers the last bank per account, so it works with the bank closed. Paged rows, 250 at a time.

**Prices and network.** Default is the Jagex guide price (RuneLite's own price table for "now"; the OSRS wiki's copy of the daily table for "then", one batched request per day). With "Use live prices" on, actively traded items use the wiki's traded series (`/latest` on Refresh and every 30 minutes, one `/24h` bucket per window per day), gated by five liquidity checks; thin items keep the guide price. All requests go to the OSRS wiki's own hosts and carry nothing about the player or the account (test-pinned). Storage is local under `~/.runelite/plugin-data/bank-portfolio-tracker/`, and all file i/o goes through `Filepath`. `build=standard`, no third-party dependencies, BSD-2.

**Versus Bank Price Changes.** That plugin shows the top 5 or 10 movers from the real-time trade endpoints. This one values the whole bank and every stack on a series chosen for accuracy (guide by default, live only where the market is thick), and adds inventory and worn gear. Different question, so a separate plugin.


Psychemaster:

PR #16540, opened 2026-09-14
Add Ironman Path Sync

stuntedrabbit:
## Ironman Path Sync

Repository: https://github.com/stuntedrabbit/ironman-path-sync

Companion plugin for https://ironmanpath.app (an OSRS Ironman skilling / gear guide). It sends the player's bank, inventory, worn items and real skill levels to the site so its Set builder can pick the best gear for a boss from what the player actually owns.

### How it works
- The site shows a 6-letter code. The player pastes it in the plugin's side panel and presses "Link with this code".
- The plugin sends one snapshot when the bank is opened (or on "Sync now"). The server answers with a private key that the plugin stores; the code is single-use and expires after 15 minutes.
- From then on the bank is sent each time it is opened (rate-limited to one send every ~30 s). "Unlink" deletes the key.

### External service
POSTs JSON to `https://ironmanpath.app/api/sync` (HTTPS). Payload: display name, skill levels (`client.getRealSkillLevel`), and item ids with quantities from the bank / inventory / worn containers. Nothing is sent without a code or key. No other data is read or sent.

### Notes
- No game automation, no input, no packet hooks. Reads item containers and skill levels only.
- Uses the injected `OkHttpClient` and `Gson`.
- Side panel (JPanel) with code field, Link, Sync now and Unlink buttons. Chat messages report the result.


Nightfirecat:
A `warning` should be added to your plugin manifest or to the pluginToken config key so that users enabling the plugin or config will be adequately warned that their data will be sent to a 3rd-party server. I recommend using the following format:

```
warning=This plugin submits your IP address, RSN, skill levels, and bank/inventory contents to a 3rd-party server not controlled or verified by Runelite developers.
```

PR #16556, opened 2026-09-14
Add Vale Totems Declutter

HogChampTV:
New plugin: Vale Totems Declutter - a QoL plugin for the Vale Totems (Auburnvale) activity that lets you pick the tree you're chopping and hides or deprioritizes the rest of the vale's clutter.

Optional toggle to just hide clutter if you are bringing logs during the vale totems activity.

Plugin repository: https://github.com/HogChampTV/vale-totems-declutter


coopermor:
We don't allow new plugins that take user supplied IDs. Hardcode the vale totems region in your plugin and only allow the config input to be object names and not IDs.

HogChampTV:
> We don't allow new plugins that take user supplied IDs. Hardcode the vale totems region in your plugin and only allow the config input to be object names and not IDs.

Thank you for the review, I believe I have made these necessary changes and updated the manifest to the new commit hash.

PR #16456, opened 2026-09-12
coffer: bump to 3736c00 (0.1.13-beta)

Paul2357:
Bumps Coffer to 0.1.12-beta.

Changes since the last accepted version: an in-panel item price chart (a small sparkline on flip/position cards, expandable per item) for Standard/Premium accounts; Discord alerts now available on all tiers; and a "Forgot password?" flow. Restricted-API note: the LinkBrowser/Desktop browser-open usages flagged in review are being removed in the next pushed revision (replaced with in-panel handling). No change to the data sent to the server.

PR #16558, opened 2026-09-14
coffer: 0.1.15-beta

Paul2357:
Bumps Coffer to 0.1.15-beta (commit 9c218c7).

Change vs 0.1.14: adds a Terms of Service / Privacy Policy agreement notice at the Create-account step in the plugin's login panel (a short static line stating that creating an account constitutes agreement to the linked terms; the URLs are shown as plain text since the panel doesn't open external links). No functional or data-handling changes.

Previous version merged as #16456.


Paul2357:
Updated to 0.1.16-beta (commit 05e9280). Adds one more UI fix alongside the Terms/Privacy notice: several panel labels used symbol/dingbat characters the OSRS panel font can't render (they showed as placeholder boxes) — replaced with font-safe text, and the settings/logout/watchlist icons are now drawn rather than glyph characters. No functional or data-handling changes.

Paul2357:
Minor correction to the note above: settled on 0.1.17-beta (commit 3ffdac3). The settings and log-out buttons are kept as their original glyph icons (they render cleanly); only the genuinely-unrenderable characters elsewhere were swapped for font-safe text. No functional change.

PR #16560, opened 2026-09-15
CardExchange TCG Online Gamemode (Independent)

CompilerOfDust:


riktenx:
why do you keep closing and opening new prs after you were told not to do so

CompilerOfDust:
Hi — some context on where I'm at with this. I've now redone this work twice: first to address comments that were outside the scope of what I'd submitted, and again because the PR had been locked for two months and I couldn't edit it in place.
It's now been three months in total, which does seem to be an extreemly large wait time. I think that is worth flagging for my own records.

riktenx:
that is fine then, i didnt realize a locked pr couldnt be pushed to

re:delay, there was a while where we weren't sure about other tcg plugins given jagex's endorsement of the original one, so several including this were on hold. but atm we feel enough time has passed and so they are now in queue

riktenx:

CompilerOfDust:
Removed both System.getenv fallbacks from the plugin — the API and web URL overrides are now
  read only from the -Dthecardexchange.apiUrl / -Dthecardexchange.webUrl JVM system properties.

plugins/osrs-tcg-online
```diff
@@ -0,0 +1,2 @@
+repository=https://github.com/CompilerOfDust/OSRS-TCG-Online.git
+commit=fd5a157635b3b46bc826ac5c0d9c7d56f846729d
```
riktenx:
```suggestion
commit=fd5a157635b3b46bc826ac5c0d9c7d56f846729d
warning=This plugin submits your IP address, and may submit various account data, to a 3rd-party server not controlled or verified by Runelite developers.
```

riktenx:
you can't get a url out of one server response and call it, we have to be able to see the exact hosts you're talking to. you appear to be doing this w/ card art. you will need to hardcode the card art url somehow

CompilerOfDust:
A quick note on the cart art. Currently, as you have identified, this is pulling from the wiki images. I plan to move card art into a public AWS S3 bucket (with a proxy bind to the domain)  to allow custom art. This change won't be done for a while, as it is a larger piece. I am assuming that caching the assets locally may be something I would like to discuss further in the future

raiyni:
Please scrape the images you need and host them on whatever backed you need so we don't end up with another data fiasco

CompilerOfDust:
I'm going on holiday for a few days. I have prepared the infrastructure to deploy the assets, but the code changes won't be pushed for the next few days

PR #16579, opened 2026-09-15
Update Clan Turf - fix Grand Exchange wall rendering

Zezi-cmd:
Fixes only, no new features or data changes:
- Fills the triangular gaps and broken outlines along the diagonal
  Grand Exchange walls, so claimed turf reads as one clean region.
- Vertically centers the clan name and count on the scoreboard bars.

No change to what the plugin sends. Data disclosure is the same as the
previous version and is spelled out in the README.


Alexsuperfly:
can you remove your usage of `java.net.http.HttpClient.newBuilder`, @Inject and use RuneLite's OkHttpClient instead

Zezi-cmd:
All done @Alexsuperfly, switched to the injected OkHttpClient

PR #16585, opened 2026-09-15
Move Menaphite Remedy Reminder repo to jtclowner-plugins

jtclowner:
- Adds a preserve off warning after fresh divine reapplication when no regular boost remains.
- Renames the plugin and updates its README/metadata.


ldavid432:
Don't change both the repo and commit in the same PR. Remove the commit change for now and add that in a followup PR.

You'll also want to add yourself as an author in this PR since you are moving the plugin to an org. `authors=jtclowner`

ldavid432 (approved):

PR #16589, opened 2026-09-15
Chat-XL v1.1

4E-DEVOPS:
Chat XL v1.1.0 completes a full performance-hardening pass across the custom chat layout pipeline. 

This update adds cached font resolution, shared text normalization, emoji/alternate-character fixes, row-first and fallback correlation improvements, optimized rank correlation with shallow recovery, persistent row indexing, active font/profile state, idempotent widget mutation, conditional revalidation, native presentation restoration, and optional performance diagnostics retained for future profiling.


4E-DEVOPS (commented):
Added game messages

PR #16590, opened 2026-09-15
pull request for region vision plugin

PinkyOSRS:
the plugin basically allows you to define certain regions, where you want entities (like your own player character or other players) to be hidden, or where you want a grid to be displayed, instead of having to turn these functions on or off globally


raiyni:
I don't think this can exist as a blanket plugin

PinkyOSRS:
> I don't think this can exist as a blanket plugin

how so? none of the features it has are restricted as far as I'm aware. the reason why watchdog doesn't work in boss rooms and such (like, to toggle on a plugin) is because it does a lot more, like possibly having triggers that tell the player what to do

this basically the does the same as manually turning three standard plugins on and off would. and neither having a grid, nore having entity hider functions, are restricted if I am not mistaken. please let me know if I'm wrong. thank you

PinkyOSRS:
> I don't think this can exist as a blanket plugin

actually thinking of it, it might be restricted in pvp? hide local player?

Alexsuperfly:

you need to also attribute the authors of the plugins youre yoinking all this from too.

as for legality, i would say this is probably fine. we do have a restriction on bundling other plugins, but this at least adds functionality on top of what would be the other plugins

PR #16637, opened 2026-09-16
Add Uncharged Jewellery Reminder plugin

FjoesPlugins:
Adds the Uncharged Jewellery Reminder plugin, which warns players when they enter the Wilderness carrying an uncharged amulet of glory, ring of wealth, or combat bracelet.

This should be usefull for pvmer's and pker's alike.


Alexsuperfly:

FjoesPlugins:
Alright, should be fixed now, let me now if anything else is wrong/missing, etc.

PR #16640, opened 2026-09-16
Add EVI Live (Local)

therealLeEvi:
EVI Live (Local) is a passive Grand Exchange companion plugin. It reports your own GE offers to a local bridge on 127.0.0.1 only (no external network calls), and shows a text hint plus an optional hotkey for EVI's suggested buy/sell price and quantity in the offer prompt, a highlighted row in item search, and a clickable "EVI item: <name>" row to jump to the suggested item (using RuneLite's standard Widget.setOnOpListener/setOnKeyListener widget-click API, not simulated input). It never opens a menu, clicks a button, or confirms an offer itself.

Plugin repository: https://github.com/therealLeEvi/evi-live-plugin
`build=standard` in runelite-plugin.properties.

### The companion bridge, and why it is mentioned here

This plugin is one half of a personal tool: the other half is a small Node.js server the user runs on their own machine. Since the plugin is only useful with it, its full source is public so it can be read rather than taken on trust:

**https://github.com/therealLeEvi/evi-live-bridge**

- It binds to `127.0.0.1` only and rejects requests whose `Host` header is anything else. There is no account, no sign-up, and no server operated by anyone else.
- Its only outbound requests are to two public sources: the OSRS Wiki real-time price API and the official Old School RuneScape news feed, both through a fixed allowlist of paths. There is no general-purpose proxy.
- `LOCAL-API.md` in that repository documents every endpoint this plugin calls.

What the plugin sends there: the Grand Exchange offers you place yourself (item, price, quantity, fill progress), your coin count so a suggestion is never larger than you can afford, and — only if you enable that setting — your inventory contents so idle stock can be suggested for sale. Accounts appear only as a salted pseudonym generated locally. No account name, password, or chat is ever read. With the bridge not running, the plugin simply shows that it is unavailable and does nothing else.

Happy to answer anything about either half.


riktenx:
all file i/o should be performed with https://static.runelite.net/runelite-client/apidocs/net/runelite/client/util/Filepath.html rather than direct java APIs

therealLeEvi:
Done — file I/O now goes through net.runelite.client.util.Filepath (via the Unchecked legacy-directory entry point, since Filepath's own documented "sanctioned" accessor for a plugin's directory hasn't landed in RuneLite yet). This keeps the existing pairing key and identity salt working with no migration needed. One thing worth flagging: Unchecked is called out in Filepath's own docs as a path the Hub's automated review can't fully cover, so let me know if you'd like anything adjusted there.

riktenx:
> since Filepath's own documented "sanctioned" accessor for a plugin's directory hasn't landed in RuneLite yet

that is wrong, you shouldn't be using unchecked, ever. use `getPluginDirectory()`. you will have the old folder only locally since you've been testing this yourself so you don't need to set `legacyPluginDirectory`

therealLeEvi:
You're right, and my earlier comment was wrong — `getPluginDirectory()` is present in the client (checked against `net.runelite:client:1.12.40-SNAPSHOT`), so there was no reason to reach for `Unchecked`.

Fixed in 9fb12daee516580a42d863ea8b97c7f57066b74c:
- `startUp()` now uses `getPluginDirectory()`; `Filepath.Unchecked` no longer appears anywhere in the plugin.
- Added `internalName="evi-live"` to `@PluginDescriptor`, which `getPluginDirectory()` requires.
- `legacyDataDirectory` is deliberately not set, per your note — the old `.runelite/evi-live` folder only ever existed on my own machine, and I moved my key and salt across by hand.

Thanks for the review.

riktenx:
all networking should be done by `@Inject`ing and using runelite's `OkHttpClient`

therealLeEvi:
Updated to 3.6.0 (`9c704b0a2379fe947990f3634d4eb8320db5266e`), which brings the submitted plugin in line with the version actually in daily use: an active-offers list in the sidebar, loss-aware sell suggestions that show the break-even price rather than hiding a losing sale, a button to dismiss a stale "you're holding this" reminder for good, preset minimum-profit tiers, trade durations up to 24 hours, and a world-type check so members-only items are not suggested on a free-to-play world.

I have also made the companion bridge's source public and linked it from both the plugin README and the description above, since the plugin is only useful alongside it and it seemed better for you to be able to read it than to take it on trust: https://github.com/therealLeEvi/evi-live-bridge (loopback-only, no external service, with its endpoints documented in `LOCAL-API.md`).

therealLeEvi:
Sorry — my previous comment answered the version sync and missed your actual point. Fixed now in [`fd3c616`](https://github.com/therealLeEvi/evi-live-plugin/commit/fd3c616bf35a9224fb4ae69056b03af048e94d03), which the PR now points at.

All networking goes through RuneLite's injected `OkHttpClient`. `LocalTransport.Http` takes it and derives its own with `newBuilder()`, so it shares the client's connection pool, dispatcher and interceptors, and overrides only what talking to a service on the same machine needs:

- 1.5s connect / 2s read timeouts — a loopback service either answers immediately or is not running
- `Proxy.NO_PROXY` and redirects disabled, so a request to `127.0.0.1` cannot be redirected off-machine
- `CacheControl.FORCE_NETWORK` on the suggestion GET, since it is a live reading of the player's own journal and current prices, and a cached one would quietly show a stale trade

`HttpURLConnection` no longer appears anywhere in the plugin. It is built in `startUp()` rather than at field initialisation, since injection has not happened at that point.

Since it is the whole of the plugin's network behaviour, for completeness: it talks to exactly one destination, `http://127.0.0.1:51743`, and nowhere else. No account, no external service, nothing leaves the machine — which is why the manifest carries no `warning=` line. The companion bridge it talks to is public and linked from the plugin README (https://github.com/therealLeEvi/evi-live-bridge), with every endpoint documented in `LOCAL-API.md`; it is what fetches prices, from the OSRS Wiki's public real-time API.

The plugin itself observes Grand Exchange offers, writes them to a local ledger with exact tax-matched FIFO accounting, and shows one suggested quantity/price as text in the offer prompt, optionally filling that single field on a hotkey. It never clicks, selects items or confirms offers.

This commit also stops it suggesting a ninth offer when all eight Grand Exchange slots are in use, and brings the submitted snapshot in line with the version in daily use. Tested against a real client, and `gradlew build check` is green.

therealLeEvi:
Updated to 3.7.0 ([`e85544a`](https://github.com/therealLeEvi/evi-live-plugin/commit/e85544aa71cacc8271f7cb4da0605823815e4abb)). Built and tested against the 1.12.39 release with `gradlew build check`. The plugin changes are small — three source files plus the build file:

- **`EviLiveConfig`** — every setting's description is now one short sentence (several were far too long to read as a tooltip), one setting that no longer has any effect is `hidden`, and the risk level defaults to Low. No `keyName` changed, so stored settings are unaffected; a test now fails the build if one ever does.
- **`EviLivePanel`** — the sidebar's read-only text areas no longer move their caret on update, which was scrolling the panel every two seconds.
- **`EviLivePlugin`** — sends Medium risk explicitly now that Low is the default, and reports an inventory check even when it finds none of the item IDs the local bridge asked about. Only IDs the bridge itself named are ever echoed back.
- **`build.gradle`** — pinned to the RuneLite **1.12.39** release rather than a snapshot. It has `Plugin#getPluginDirectory()` and `Filepath`, and a snapshot build could never reach the Plugin Hub manifest.

Nothing you reviewed has changed in behaviour: networking still goes through the injected `OkHttpClient` to `127.0.0.1` only, file I/O still goes through `getPluginDirectory()`, and the plugin still never clicks, selects or confirms anything.

therealLeEvi:
Repointed to [`c678eaf`](https://github.com/therealLeEvi/evi-live-plugin/commit/c678eaf17afab34731b22e4c3717498c4e3b3b90). README only, no code changes: the Privacy section said only the offers you place are observed, which understated it. It now lists everything the plugin reads (coin count, world type, whether positions the bridge asked about are in the inventory, and the full inventory only with an off-by-default setting), all sent to `127.0.0.1` only.

PR #16659, opened 2026-09-17
Update mystix

Dmurph24:
Updates Mystix to the latest release.

The main change fixes owned-item roadmap goals ("get N more of an item") going backwards or stalling when the item moves into a container the plugin could not see. The plugin already uploads the bank, inventory and vaults through its bank endpoint; it now also reads the Sailing cargo holds and trawling net, the seed box, tackle box, forestry kit and huntsman's kit (all plain item containers, through the existing ItemContainerChanged subscription), the rune pouch and plank sack (from their varbits), and infers the contents of the fish barrel, herb sacks, log baskets, gem containers and coal bag from the same game messages, inventory diffs and "Check" listings the community Fish Barrel, Log Basket, Coal Bag and Charges Improved plugins use. Each is uploaded as its own source on the existing bank endpoint; no new endpoint or data type is introduced. When the player runs Charges Improved, its tracked contents for those items are read from its config group (read only) and take precedence over the plugin's own inference.

Uploads are also less frequent: every bank-memory source now goes through one debounce (a change is held and collapsed for three minutes, or thirty seconds when it touches an item a goal is counting, and only a real logout flushes). Inventory and worn gear changes sync on that cadence instead of only while the bank is open. Region loads and world hops no longer trigger uploads.

Smaller fixes: a crew member banking the cargo hold and items emptied into a deposit box are accounted for locally until the bank is next read, emptying a storage item at the bank no longer counts its contents twice, and cargo hold container ids that arrive with the interface-owned flag set are recognised.

Loot drops are still queued and uploaded every three minutes, but a drop containing a stack worth 100k gp or more, or a single untradeable item, now uploads the queue right away so a rare drop reaches the app in seconds. These early uploads are limited to one per ten seconds and use the existing loot endpoint and payload; no new data is sent.


riktenx:
you can't get urls from some server response and call out to them, it makes it impossible for us to see what hosts you're talking to. every host the plugin talks to needs to come either from user config or be hardcoded

Dmurph24:
@riktenx Fixed in 2254bdc9638816bcecff64d24c1ae7f2d7ad7083: goal icons are now only fetched from the hardcoded `https://oldschool.runescape.wiki/images/` path. Any other URL from the server is ignored. The only other host the plugin talks to is the hardcoded `https://api.mystix.app`.

PR #16665, opened 2026-09-17
Add 69 damage plugin

jayuwu97:
Takes screenshot every time you deal 69 damage. Toggleable overhead text and chatbox text saying "Nice."


riktenx:
the file shouldn't have `.properties`

riktenx:
url also needs to have `.git` at the end iirc

riktenx:
you will need to address the above

riktenx:
additionally
* use https://static.runelite.net/runelite-client/apidocs/net/runelite/client/ui/DrawManager.html#requestNextFrameListener(java.util.function.Consumer) instead of java.awt.Robot
* don't use Thread.sleep or Thread.interrupt

jayuwu97:
admittedly i am clueless but lmk if there's anything else

riktenx:
all file i/o should be performed with https://static.runelite.net/runelite-client/apidocs/net/runelite/client/util/Filepath.html rather than direct java APIs

PR #16680, opened 2026-09-17
update osrs-fliphub

zFallan121:
Bumps `commit=` for `osrs-fliphub`, now at `f0d6ff6`.

**Flip ranks.** Lifetime flip profit across all of a player's characters now earns one of FlipHub's ten ranks, with the same names and thresholds as osrsfliphub.com. It is worked out locally from the trades the plugin already stores; nothing new is fetched or sent.

**On the Profile tab.** FLIP RANK sits beside the time range in the rank's colour, next to the rank's picture, and hovering either names the next rank. It always reads lifetime profit across every character, whatever range or profile is selected.

**The rank-up.** When a Grand Exchange sale takes that total into a new rank, the chatbox shows a level-up style message with the rank's picture, the local player gets the level-up spot animation and the Dance emote, and a game message is added to chat. The spot animation and the emote are client-side only; nothing is sent to the server. Each rank is celebrated once. A new setting, **Celebrate rank-ups** (on by default), turns it off.

Notes for review:

- The message is a `ChatboxPanelManager` input. It waits while a prompt or a conversation has the chatbox rather than closing it under the player.
- The rank pictures are PNG resources loaded with `ImageUtil.loadImageResource`. The chatbox copies are registered as sprite overrides under negative ids on start-up and removed on shut-down.
- The check is called from `TradeDeltaRecorder`, which only live GE offer changes reach. Adding up the characters runs on the plugin's existing scheduler, never on the client thread.
- A rank-up tester used during development lives under `src/test` and is loaded only by the dev launcher.

**README.** A Flip ranks section with an in-game screenshot, a linked contents list, section dividers, and a retaken Profile screenshot.

No new dependencies. `build.gradle`, `settings.gradle` and `runelite-plugin.properties` are unchanged from the published commit. `./gradlew clean build` is green at this commit, with 404 tests passing.


zFallan121:
On the token count: measured the same way as last time (all of `src/main/java`, comments stripped), this commit is 192.3k on a local cl100k count, about 2.6k more than `fd6619f`. Scaled by the ratio from the 201k your scanner read on `f2c7dfb`, that should come out near 198.5k, so under 200k by our count.

That said, our numbers have never quite matched the bot's, so please let me know if it's in fact over and I'll cut it back down.

PR #16683, opened 2026-09-17
Update clogman-mode

mozjay:
- Added fix to collection log syncing


riktenx:
some of your use of client.menuAction is not allowed, the only valid use of that method is as wikisync uses it

mozjay:
> some of your use of client.menuAction is not allowed, the only valid use of that method is as wikisync uses it

Thanks Ritken. I have patched to remove what I infer is the wrong usage and replaced it with how Runeprofile does the auto sync on clog opening, based on stuff here: https://github.com/runelite/plugin-hub/pull/12380

Hope that's ok - if I've misunderstood something I can just revert and add a note for users to just click the search button, no big deal

PR #16697, opened 2026-09-17
clan-management: update (screenshot redaction, clog counts, Boss KC board, fixes)

WoodyCode23:
Bumps `plugins/clan-management` to e8c7714.

Since the last merged build (f707007):

- Screenshots: optional chat redaction before upload (all chat / PMs only / nothing, default nothing), widgets hidden before the frame renders instead of painted over, dead black margins trimmed, no downscaling on large displays, no upload timeout on big images, and a 413 falls back to resubmitting the event without the image.
- Collection log: send the live varp counts with an unlock, and count the slot being announced (the varp lags its own chat message).
- Leaderboards: Boss KC promoted to a top-level board opening on current kills, whole numbers, per-board period memory, and boards fetch on switch instead of rendering empty.
- Fixes: ToA boards use total time rather than challenge time; death posts name the killer; an event that starts mid-session is picked up instead of only at login; the ornate pool check requires build mode for accounts that can enter other players' houses.
- New-member leaf: members inside the clan's new-member window get a small green leaf beside their name in the Members tab, drawn rather than an emoji (the panel font has no emoji glyphs).


riktenx:
do not tell users how to sideload the plugin in your readme. you can put dev build instructions in CONTRIBUTING etc

WoodyCode23:
Fixed, thanks. The build/sideload steps are out of the README and moved to CONTRIBUTING.md, so the README now only points users at the Plugin Hub.

PR #16699, opened 2026-09-17
Update commit hash for catastrophic-events plugin

mikeyf06:
Bumps to version 1.2.1. Fixes a token field that could corrupt on re-paste, a "Not linked" panel that could get permanently stuck even with a valid token if the poll loop hit an uncaught exception, a death message default that Reset wiped instead of restoring, and false accomplishment alerts firing for milestones a veteran account already had before accomplishment sharing was turned on (quest completions, 99s, and total-level milestones now detect off live chat messages instead of polling game state).


riktenx:
remove build instructions from your readme, you can put that in agents.md, contributing etc

mikeyf06:
> remove build instructions from your readme, you can put that in agents.md, contributing etc

fixed

PR #16712, opened 2026-09-17
Add RuneFolio Sync

Brettgod1355:
Adds RuneFolio Sync — a plugin that connects RuneLite to a RuneFolio account and syncs skills, quests, diaries, Combat Achievements, Collection Log, loot, and more to a browsable profile. Optional opt-in categories (bank/inventory snapshots, PvP history, screenshots) are off by default and clearly disclosed in-panel before enabling. Read-only telemetry — no automation or game interaction beyond reading state.


riktenx:

ldavid432:

Brettgod1355:
Sorry about the use of reflection, I thought I removed all of that earlier when I found and read the Rejected or Rolled Back Features

riktenx:
your license is malformed in some way. just take the exact content of the one from the runelite repo and use your own name and date

plugins/runefolio-sync
```diff
@@ -0,0 +1,3 @@
+repository=https://github.com/Brettgod1355/RuneFolio-Plugin.git
+commit=7e176c1828faccfe24440491106586efef7351c3
+warning=This plugin submits your IP address, RSN, and account progression (skills, quests, diaries, combat achievements, collection log, loot and, if enabled, bank, PvP and screenshot data) to a 3rd-party server (runefolio.app) not controlled or verified by RuneLite developers.
```
riktenx:
```suggestion
warning=This plugin submits your IP address, and may submit various account data, to a 3rd-party server not controlled or verified by Runelite developers.
```

PR #16726, opened 2026-09-18
AwesomeClanplugin

AwesomeClan:
plugin for the AwesomeClan clan. Syncs the clan roster to the clan's website every 30 minutes vOptionally update live  xp / kc updates to a per-member dashboard page via a personal token pasted into settings (supports multiple accounts,).


Nightfirecat:
A `warning` should be added to your plugin manifest or to the `pluginToken` config key so that users enabling the plugin or config will be adequately warned that their data will be sent to a 3rd-party server. I recommend using the following format:

```
warning=This plugin submits your IP address, RSN, clan roster, XP gain, and boss kill data to a 3rd-party server not controlled or verified by Runelite developers.
```

AwesomeClan:
added

PR #16747, opened 2026-09-18
Remove three-choices and npc-morph from Plugin Hub

osrssniffin:
I’m voluntarily removing these plugins. With the recent large increase in AI-generated Plugin Hub submissions, I don’t want to contribute additional plugins that I don’t consider meaningful enough to justify maintaining in the hub.


ldavid432:
> instead of removing the plugin hub manifest file, add the disabled flag with a corresponding reason.

PR #16776, opened 2026-09-19
Add Spyglass

Burrky:
Adds Spyglass, a local RuneLite activity and loot tracker with Session, Loot, and History views.

Spyglass tracks current activity, XP, loot, recent finalized activities, and starting loadouts.

The plugin makes no custom network requests, performs no gameplay or input automation, stores its data locally, and uses no third-party runtime dependencies.


riktenx:

Alexsuperfly:

PR #16777, opened 2026-09-19
new plugin 

no0bi1:
replace other player models with npc of your choice.... used primarily to hide distracting players


Adam-:
You should pick a different internal plugin name than `submit`

PR #16784, opened 2026-09-19
Chronicle rename

Fitzgerald-gg:
The rename finishes a change of identity rather than of behaviour: the plugin is called Chronicle everywhere internally now, and the hub id was the last thing still reading fitzgerald.


coopermor:
You can't rename the hub manifest file or it will uninstall the plugin for all users.

PR #16785, opened 2026-09-19
fitzgerald: update repository URL

Fitzgerald-gg:
This only moves the two lines that can move: the repository has been renamed to `Fitzgerald-gg/chronicle-runelite`, and the commit bumps to pick up a README rewrite with screenshots, a new icon, and corrected author/description metadata.

No functional change to the plugin: the only file in the jar's path that differs between the two commits is `runelite-plugin.properties`.


coopermor:
You need to modify this PR to only update the repository URL. Once it is merged you can open a second PR to update the commit hash.

PR #16791, opened 2026-09-19
Update Versatile Logger

PeppieLangWaus:
- Changed warning formatting.
- Fixed an issue where private messages only showed local players' username
- Added clan messages (drop/level up)
- unescape tags in logs


plugins/versatile-logger
```diff
@@ -1,3 +1,3 @@
 repository=https://github.com/PeppieLangWaus/versatile-logger.git
-commit=116b1543691e7b15cbc5c918afbce1f82d1ae285
-warning=This plugin can be configured to send nearly all in-game messages, notifications, and your IP address to a 3rd party server not controlled or verified by the RuneLite Developers.
\ No newline at end of file
+commit=f805fbaa37d1893f6cf1223962c99fe621e2209a
+warning=This plugin can be configured to submit data to a 3rd-party server not controlled or verified by Runelite developers including:\n    • IP address\n    • Username\n    • Chat messages\n    • Game messages
```
riktenx:
```suggestion
warning=This plugin submits your IP address, and may submit various account data, to a 3rd-party server not controlled or verified by Runelite developers.
```

riktenx:
i've applied a more general warning, just leave it as-is going forward

PeppieLangWaus:
Thanks, will keep it that way. Also was changing the warning why the `non-commit plugin change` was added?

riktenx:
yeAh

PR #16793, opened 2026-09-19
Update bestiary to 1.0.1 (add monsters, fix description)

iTheFish:
Adds 5 monsters (Araxytes, Gryphons, Shellbane Gryphon, Mammoths, Ghouls) and fixes the plugin description encoding.


riktenx:
you appear to be dynamically getting wiki urls from a server response and calling them, that makes it impossible for us to statically verify what hosts your plugin is talking to. all hosts you talk to need to be hardcoded or input by the user

PR #16809, opened 2026-09-19
Update Zenyte: opt-in live logging and automatic uploads

Callum-Upton:
Adds opt-in public live logging and automatic completed-log uploads, plus recognised-boss capture updates. All networking is guarded by a separate default-off Allow Zenyte networking switch with the exact reviewer-requested warning. Live logging and automatic upload also default off and require an upload-scoped website key; live sharing additionally requires website consent. Local recording remains available without an account.

Every HTTP execution uses a dedicated method containing only a fresh allowNetworking check and the HTTP call. Requests use RuneLite's injected OkHttpClient, fixed HTTPS zenyte.gg endpoints, no redirects or implicit retries, timeouts and a bounded background queue. Only regular replay data is transmitted; research files remain local. The key is masked and excluded from recordings.

Regular capture starts only for recognised bosses and raids; generic combat requires explicit research opt-in. Includes research commands/hotkey, research reset to off at startup, Vardorvis capture and capability metadata. Grotesque Guardians now uses one profile for Dawn and Dusk in instanced template region 6727, plane zero, preserving recording across transformations, individual NPC deaths and quiet transitions. The shared kill-count message confirms completion; player death and arena exit still close recordings. Existing Phosani plane and spell fixes are retained.

Validation: 98 automated tests and Gradle test/build passed on the submitted source, including network consent/revocation and Guardians transformation-without-further-hits, phase continuity, completion timing, arena exit and player death during exit grace. Recorded research supports the Guardians arena and completion identifiers; Dusk's additional transition form is verified against RuneLite game-value constants. Full live-client Guardians kill/death/exit validation remains pending. Dedicated Vardorvis live validation also remains outstanding. Live logging and automatic upload were tested in game before the HTTP-client change.

Source: https://github.com/Callum-Upton/glass-combat-logger/commit/c0fb8b6bfd02627fea45a0abd28365b79ffe436b


riktenx:

* you MUST have a `@ConfigItem` boolean, that defaults to false, that entirely disables the networking calls, its warning MUST be set to exactly "This feature submits your IP address and various account data to a 3rd-party server not controlled or verified by Runelite developers."
* all networking calls need to go into dedicated methods, and at the top of every method where an opt-in networking call is made you MUST check the boolean config and either return early or throw an exception. the body of the method MUST be only this check and the network call

PR #16810, opened 2026-09-19
Add better-brush-markers plugin repository info

ChronicOSRS:
Exportable and importable Tile Markers made via Brush Marker plug-in.


coopermor:
```
better-brush-markers: "build" must be set
in file /tmp/pluginhub-package14755136377159347739/better-brush-markers/repo/runelite-plugin.properties
You must add a build=standard or build=gradle entry.
build=standard is recommended unless you have dependencies or other changes to your build.gradle.
```

PR #16842, opened 2026-09-20
Add ground-marker-variables

Limy-Monkey:
Extends current Ground Markers plugin functionality with variable support.

Variables include {metronome4} as a 4 tick metronome, {lvl_attack} to display your current attack level, and more. Full information about which variables are supported are in the README of the repository.

This plugin does force-remove the Ground Marker overlays while the plugin is active. Re-adds Ground Marker overlays when the plugin is turned off. I would technically prefer if I could simply overwrite the text inside the existing Ground Marker overlay, but unfortunately this appears to be Java-private. For this reason, this plugin reads the data from Ground Markers' configManager so it can accurately recreate the tiles it is force-removing. All data is stored on Ground Markers side.

Does support party-sync using the RuneLite core party interface. Modeled after how Visual Metronome plugin accomplishes the same, with per-tick requests and responses, logically gated by Sync Target checks.

Has been PR'd (closed by me) once before: #14923 
 - Per comments there, this plugin does not support metronomes inside of conditionals.


Psychemaster:
Upcoming changes to how `@PluginDependency` works are likely to break this plugin, so you will soon need to remove the dependency on Ground Markers. See the following Discord message (and its' attached thread) for more details: https://discord.com/channels/301497432909414422/419891709883973642/1552729163146727515

PR #16843, opened 2026-09-20
update veritas

Hegget34:
Clan board moved to its own domain; plugin default updated.


Psychemaster:
You appear to be dynamically getting event URLs from a server response and pushing data to them, that makes it impossible for us to statically verify what hosts your plugin is talking to. All hosts you talk to need to be hardcoded or input by the user.

Hegget34:
Fixed. The plugin no longer takes any address from a server response. Event drops go only to the URL the player types into the Event URL setting, which is blank by default. The three hosts it can reach are the clan board (a setting with a default in the source), api.wiseoldman.net (a hardcoded constant), and the player's own Event URL. Updated the commit in plugins/veritas.

PR #16854, opened 2026-09-20
create daeyalt-sound

KeithIsSleeping:
Adds Daeyalt Sound, which plays an in-game sound effect when the Daeyalt essence relocates.

Only one rock in the Daeyalt Essence mine holds essence at a time. When it is exhausted a different rock becomes the active one, and it is easy to miss while alt-tabbed, so ticks get spent mining an empty rock.

**Source:** https://github.com/KeithIsSleeping/daeyalt-sound

### Note on overlap

`daeyalt-essence-mine` already notifies on this event, so I want to be upfront about the overlap. The difference is the alert mechanism: that plugin uses the `Notification` type, whose sound options are a native chime or a user-supplied `.wav` on disk. This plugin plays an OSRS engine sound effect by ID via `playSoundEffect`, which `Notification` cannot do, and follows the in-game sound effect volume while still playing when game sound is muted.

Happy to close this if you consider it too close to the existing plugin.


Hampo:
I don't mind adding this as a feature to `daeyalt-essence-mine` if you want

KeithIsSleeping:
> I don't mind adding this as a feature to `daeyalt-essence-mine` if you want

That would be great! I'll close this! I really prefer simple in game audio sound effects over the OS notifications, which feel like overkill to me.

KeithIsSleeping:
Closing this - the existing daeyalt-essence-mine plugin is taking on this feature, so this would be a duplicate.

Thanks for the time spent looking at it.

PR #16860, opened 2026-09-21
Create bank-pin-order

whtupyall:
Unscrambles bank pin buttons


raiyni:
We aren't interested in bypassing jagex security measures like this.

PR #16864, opened 2026-09-21
Update Golems Don't Die to 2.0 (Sailing)

Varzeki:
Vastly expanded wander capabilities of golems.
-Navigation for entire map
-Can use obstacles, transports, and Sail

Added Nameplates.

Improved golem count tracking by using varbit rather than chat message.

Framework in for telemetry, no network code included yet.
Intended to capture obstacle data; accuracy for traversal across the game can be improved.

Updated readme and media.

Exploration2


Alexsuperfly:
- use of thread interrupt is not allowed

PR #16881, opened 2026-09-21
Add runemargin plugin with configuration details

Tilley661:
Rune Margin is an in-game companion for the OSRS Grand Exchange flipping site
runemargin.co.uk. It shows live buy/sell margins, best-flip suggestions, set
combine/break arbitrage and potion-decanting plays in a sidebar panel, plus a
read-only margin overlay on the GE. It's a thin client of the public Rune Margin
API — information only, no input automation or offer setting.

Data: it sends the item ids you view (and, if you set one, your access token) to
runemargin.co.uk to fetch prices/analysis. The RuneScape name, GE offers and
profit/loss stay local.


Alexsuperfly:
we cannot accept the GNU license, can you use BSD-2 like the [plugin-hubs](https://github.com/runelite/plugin-hub/blob/master/LICENSE)

Tilley661:
@Alexsuperfly - Updated to BSD-2. Thank you!

PR #16890, opened 2026-09-21
Create skiller-safeground

DiceRollDev:
Info only visual overlay plugin, designed to map static f2p "p2p also later maybe" NPC threat Boundaries, including rendered custom attack metronomes as well as a safe or unsafe logout tile visual overlay for skiller safety.


Psychemaster:
Your plugin hub submission is not correct; please review https://github.com/runelite/plugin-hub/blob/master/README.md#submitting-a-plugin and update your PR accordingly.

PR #16897, opened 2026-09-21
Update poh-cosmetic-transmogs repository URL

jtclowner:
Updates the repository URL after transferring poh-cosmetic-transmogs from jtclowner to jtclowner-plugins. The published commit remains unchanged.

The 1.2.0 update is kept separate in #16658, now a draft pending this change.


pajlada:
To ensure you can continue to push updates, you should include your own GitHub account under the `authors` key in this plugin hub manifest file. See https://github.com/runelite/plugin-hub/blob/master/plugins/dink#L3 as an example.

PR #16922, opened 2026-09-22
Update InventorySetups to v1.26.0

dillydill123:
- Move to `build=standard`
- Extend `PluginMessage` API
- Cleanup GE Item Search filter


ldavid432:
Disregard previous comment

PR #16929, opened 2026-09-22
update veritas

Hegget34:
So sorry,

Small fix on top of the last update. The plugin's own hardcoded links were
being run through the host check intended for links supplied by the clan
board, which dropped the clan website from the Home panel. Constants in the
source cannot change at runtime, so they no longer go through that check.


Hegget34:
Two more changes since I opened this, both in the same commit range:

- Item tooltips in the loot tracker showing name, quantity, GE and high
  alch value. Item compositions are read on the client thread only and
  handed to the panel as finished data, never read while the panel is
  drawing.
- Fixes the player's name not appearing in the panel header. It was read
  once at LOGGED_IN, where the local player is often still null, so it
  frequently never got set at all.

PR #16936, opened 2026-09-22
Add Prevent TP at Mortimer

cotecode5:
Adds the registration pointer file for the Prevent TP at Mortimer plugin. This plugin protects players from accidentally teleporting out of Wyrmscraig Cavern and losing their 65k boat placement fee.


cotecode5:
This is my first time making a plugin, all i want is to stop accidently tping out when im at mortimer because im a UIM with no tp to boat or anything like that, I gotta do the back and forth with my boat everytime, its kinda for personal use. I can't test it and not sure also if the format of files is right but im trying anyway.
Please tell my any info you can on how to improve it. I would gladly make it cleaner with some pointers.
Thank You :)

